PR #44, opened 2014-08-17
Allow the dev server to use an existing http.Server

ampedandwired:
If the first argument to Server.listen is a node http.Server then the dev
server will use this. This allows the dev server to be embedded as middleware
within an existing express application.

```javascript
var app = express();
var devServer = new WebpackDevServer(webpack(config), {});
app.use(devServer.app);
var server = express.listen(3000);
devServer.listen(server);
```

I've implemented this as an overload of the dev server `listen` function, however it could be split into its own method if you think this is too messy or backwards incompatible (perhaps `devServer.listenWith(server)`).

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -1,5 +1,6 @@
 var fs = require("fs");
 var path = require("path");
+var http = require("http");
 var webpackDevMiddleware = require("webpack-dev-middleware");
 var express = require("express");
 var socketio = require("socket.io");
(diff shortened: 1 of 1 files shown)


sokra:
I won't merge this, because webpack-dev-server and application server should be separate by design. Best in separate processes. There is a proxy option which makes it behave like in one server, but best is to design your server in a way in which allow to have scripts anywhere.

PR #49, opened 2014-08-25
Log proxy errors instead of throwing exception

arnihermann:
If the upstream server is for some reason unavailable (for example, if it's being restarted/reloaded due to code changes) then `webpack-dev-server` should log the error and not crash the running process.

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -43,6 +43,14 @@ io.on("errors", function(errors) {
 	reloadApp();
 });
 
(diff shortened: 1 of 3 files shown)


brandonbloom:
This has been pretty reliable for my needs, but there's at least one class of proxy errors that shouldn't cause a visible error: Client-canceled requests. This happens to me if I show/hide img tags quickly, or set a timeout on XHR requests.

PR #93, opened 2014-12-30
Enable output options with multi config

mjackson:
This commit passes the `--output-public-path`, `--output-path`, and
`--output-filename` webpack options through to the server when the
webpack config is an array.


tomchentw:
+1 on this

PR #102, opened 2015-02-01
use a monospace font for error output

clavery:
The error output displays errors in a variable width font. Since this output is effectively the same as what's output by loaders to the CLI it can look strange when using loaders that output error information expecting fixed width fonts (in this example 6to5/6to5). Screenshot showing before and after this change.

before

after


sokra:
Thank

Issue #73, opened 2014-11-15
Socket.io 0.9.7


dschissler:
Perhaps socket.io could be updated to 1.2.

abrkn:
:+1:

TeffenEllis:
:+1:

uipoet:
No wonder, I've been seeing so many issues with the connection dropping. This is a must fix before I will try depending on this server for development again.

jhnns:
Send a PR :)

PR #105, opened 2015-02-12
update socket.io to 1.3.3

nelix:
fixes: #73 
This will not work in cases where (somehow, not through normal use of webpack-dev-server; perhaps you built --inline into a bundle?) you have a different client socket.io version.


natew:
:+1:

sokra:
Thanks

sokra:
This is broken... You get 

``` text
GET http://localhost:8080/socket.io/1/?t=1425790928139 400 (Bad Request)
```

PR #108, opened 2015-02-13
support sslPaths option for custom certificates

### LINKED ISSUE #107: Add option for providing own SSL certificate
Right now it's hardcoded to use the self-signed one in the repo. Would be nice to easily swap that out.

pstoica:
Fixes #107

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -144,12 +144,14 @@ function Server(compiler, options) {
 		}
 	}
 
+	var sslPaths = options.sslPaths || {};
+
 	this.listeningApp = options.https
 		? https.createServer({
 				// using built-in self-signed certificate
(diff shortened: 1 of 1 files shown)


sokra:
Instead of providing a path to key/cert/ca, add options to pass key/cert/ca buffer directly. No need for an additional indirection. Keep in mind you can write JS in webpack config file. You can fs.readFileSync there.

``` js
...
devServer: {
  key: fs.readFileSync(...),
  cert: fs.readFileSync(...),
  ca: fs.readFileSync(...)
}
```

pstoica:
Gotcha, fixed unless you also needed Buffer/String typechecking.

pstoica:
Would you be able to do a version bump to include this soon? I've been using a git repo/branch that has the prepublished files.

ruudk:
I'd like a version bump too.. Really want to use this feature!

PR #110, opened 2015-02-17
add cors

Dakuan:
This fixes a bug when using a hot-loader. See https://github.com/gaearon/react-hot-loader/issues/56

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -4,6 +4,7 @@ var webpackDevMiddleware = require("webpack-dev-middleware");
 var express = require("express");
 var socketio = require("socket.io");
 var StreamCache = require("stream-cache");
+var cors = require("cors");
 var http = require("http");
 var https = require("https");
 var httpProxy = require("http-proxy");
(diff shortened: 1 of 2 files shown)


troutowicz:
This isn't really a bug. When doing local development, why would you need CORS enabled? If you want the ability to access the the webpack dev server's socket.io instance anywhere inside your local network, you can just use the `--host` CLI flag. This flag was not included in the documentation, I updated the docs.

gilesbradshaw:
When you are writing an isomorphic application you want to do a hot reload of the client AND restart node.  Otherwise the two get out of sync when you next refresh the page.

jedwards1211:
@troutowicz CORS is also necessary when integrating with an existing server like Meteor, where the HTML is served up on Meteor, but the webpack bundles are served up by webpack-dev-server on a different port.

yinsang:
[chrome plugin](https://chrome.google.com/webstore/detail/allow-control-allow-origi/nlfbmbojpeacfghkpbjhddihlkkiljbi/reviews)

PR #113, opened 2015-02-20
Fix Node API inline option

tvararu:
This PR fixes the Node API by making the `config.devServer.inline` option be equivalent to `--inline`.

Before, the logic for `--inline` was only being triggered on `argv["inline"]`.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -65,6 +65,9 @@ if(!options.filename)
 if(!options.hot)
 	options.hot = argv["hot"];
 
+if(!options.inline)
+	options.inline = argv["inline"];
+
 if(argv["content-base"]) {
 	options.contentBase = argv["content-base"];
 	if(/^[0-9]$/.test(options.contentBase))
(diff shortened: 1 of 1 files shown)


tvararu:
Actually, after further testing I've found that it does not actually solve the entire problem. I'll update the PR.

tvararu:
I found the problem and my PR does not address it. For the `inline` feature to work, the webpack compiler needs to receive a configuration object with an `.entry` object that respects this format:

```javascript
{
  app: [
    require.resolve('webpack-dev-server/client/') + '?' + 'http://' + config.hostname + ':' + config.webpackPort,
    'webpack/hot/dev-server',
    './app/app.js'
  ]
}
```

Where `config.hostname` and `config.port` are `localhost` and `8080`in the default usecase.

This needs to be better documented. If someone could point me in the right direction, I'll gladly do so when I get a free moment.

sokra:
You cannot implement the inline option within the webpack-dev-server node.js API.

`--inline` need access to the webpack options. We don't have access to it via the node.js API.

Issue #115, opened 2015-02-22
Set CLI options from the config file


hokkos:
Hi,
Currently it is not possible to set the CLI options (like host, port, ...) from the config file, it would be very handy to be able to do that, because I want to set the host to the current local ip without having to change my npm run command.
This is probably [where it happens](https://github.com/webpack/webpack-dev-server/blob/9dff89ac02d3f5a932c0668ffe2b231165795313/bin/webpack-dev-server.js#L48).
Thanks.

troutowicz:
I ran into the same problem. I ended up creating a new JS file to launch the server, providing a way to dynamically set IP and port.

https://gist.github.com/troutowicz/bfd0e4eb0633ed77834e

troutowicz:
After writing up that gist, I realized there was a simpler way \o/

SpaceK33z:
@hokkos, in your config file, you can specify the port or host like this:

```
devServer: {
     port: 9000
},
```

Closing this for now, but feel free to re-open.

PR #116, opened 2015-02-22
Add option to set IP and port inside webpack config

troutowicz:
Issue #115 

When launching webpack-dev-server as a script using npm, using a specific IP and/or port requires hardcoding the values. With this change, a different route is offered. The host/port can be passed inside the devServer object contained in a projects webpack config. CLI flags will take precedence over these options.


mathieumg:
:+1:

sokra:
The idea is good... but don't write to `argv`. Write to `options`.

troutowicz:
@sokra, updated

sokra:
Thanks

PR #124, opened 2015-03-06
added option to proxy a path to another server

sakabako:
Add this to the configs and it will proxy any request to /api/* to localhost:5000.

```
    devServer: {
        proxy: {
            '/api/*': 'http://localhost:5000/'
        }
    }
```

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -111,6 +111,15 @@ function Server(compiler, options) {
 		app.use(this.middleware);
 	}
 
+	if (options.proxy) {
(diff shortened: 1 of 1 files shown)


sokra:
There is already an option to proxy requests and this shouldn't be required.

sakabako:
I'm assuming you mean the contentBase option, @sokra? That doesn't work for my situation because I have local files that I need to use with a service that's running on a different port. Unless I misunderstand it, with contentBase it's impossible to access my tests within the repo.

PR #133, opened 2015-03-21
Use socket.io-sockets instead of the one in web_modules

### LINKED ISSUE #125: upgrading socket.io completely broke hot module replacement
This commit: https://github.com/webpack/webpack-dev-server/commit/cf126c021f61490294dc2be1b93755b984ba9240

It upgrades socket.io, but the client file [here](https://github.com/webpack/webpack-dev-server/tree/master/client/web_modules) is still a really, really old version. Trying to use this newer server with the older version completely breaks (I see "unknown transport", and when I manually fixed that, I saw other errors).

I looks like this was published to npm, so HMR should be broken for everyone right now... Downgrading to 1.6.4 fixes it.

MrOrz:
Use socket.io-client instead of the one in web_modules to keep the socket.io client up-to-date.

Fixes #125 .


sokra:
Thanks, sounds good.

PR #137, opened 2015-03-25
Get rid of ANSI escaping

### LINKED ISSUE #136: Encoding(?) problems in the error output in the browser
This
screen shot 2015-03-24 at 8 00 15 pm
ends up as this
screen shot 2015-03-24 at 8 00 22 pm
in the browser

oddic:
It's also possible to convert it to style or whatever the console uses. fixes #136


sokra:
Thanks

PR #138, opened 2015-03-25
set a viewport on the live.html, to make mobile development easier

Swatinem:


sokra:
Thasnk

aboutqx:
How about a custom size viewport,like `width=750`?
That's useful when developing with h5 app on mobile.
May we need alternative `live.html` or automately cope 'our html meta' to the 'live' one

PR #140, opened 2015-03-26
Fix http-history-fallback overriding proxy settings

seanich:
I'm not sure if others have encountered this issue, but I found that requests were always being sent to index.html even though I had set up a few proxies for a backend server. Switching the order in which these options are applied has made everything work smoothly.


sokra:
merged

PR #148, opened 2015-04-05
added support for regex in proxy options

jamon:
provide configurations where the path is a regular expression, eg:

```
        proxy: [
            {
                path: /\/google(.*)/,
                target: "http://www.google.com/"
            }
        ],
```

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -135,17 +135,14 @@ new Server(webpack(wpOpt), options).listen(options.port, options.host, function(
 	if(options.historyApiFallback)
 		console.log("404s will fallback to /index.html");
 
-	if (options.proxy) {
-		var paths = Object.keys(options.proxy);
-		paths.forEach(function (path) {
-			var target;
-			if (typeof options.proxy[path] === 'string') {
-				target = options.proxy[path];
-			} else {
-				target = options.proxy[path].target;
-			}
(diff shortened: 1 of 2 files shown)


bin/webpack-dev-server.js
```diff
@@ -135,17 +135,14 @@ new Server(webpack(wpOpt), options).listen(options.port, options.host, function(
...
-			} else {
-				target = options.proxy[path].target;
-			}
-			console.log('proxying '+path+' to '+target);
+	if (options.proxy && !Array.isArray(options.proxy)) {
+		options.proxy = Object.keys(options.proxy).map(function (path) {
+			return ({
+				path: path,
+				target: typeof options.proxy[path] === "string" ? options.proxy[path] : options.proxy[path].target
+			});
```
sokra: This part should be moved into `lib/Server.js` to allow this config style without the CLI too.

AprilArcus:
This would be super useful for us! We are working on a mock server that has to support a very heterogenous legacy route scheme, and we'd like to be able to white-list certain routes and proxy the rest.

sokra:
@AprilArcus take a look at the note, fix it and create another PR.

nthtran:
I fixed this in #187. Please a have look.

PavelGavlik:
Shouldn't we close this PR when #187 has superseded this PR and it has been merged?

PR #152, opened 2015-04-12
fix historyApiFallback+proxy combo

Lendar:
Both `contentBase: "http://url"` and `proxy` are incompatible with `historyApiFallback`. This PR fixes that.

### Steps to reproduce:

Setup webpack with config

```
var server = new WebpackDevServer(webpack(config), {
  historyApiFallback: true,
  proxy: {
    "/api/*": "http://localhost:3100/"
  }
})
```

Then, open `/api/foo` in browser.

__Expected result:__ result from proxy is returned.
__Actual result:__ index.html is returned.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -109,13 +109,6 @@ function Server(compiler, options) {
 		res.end('</body></html>');
 	}.bind(this));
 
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #154, opened 2015-04-15
Strip ANSI escaping from warnings

nkbt:
Errors were cleaned up recently: 6ea80ce82ef5020e112f9f7c69c37134329e19ca, but warnings were not.


sokra:
Thanks

PR #156, opened 2015-04-15
failed proxy requests now return a 502 response rather than hanging

mrsharpoblunto:
Prior to this patch, any requests that fail to be proxied simply hang until they timeout - this can cause the entire dev server to become unresponsive if the number of failed pending requests hits 5 (the max number of open connections to the same server permitted by the default node http agent).

This patch ensures that such failed requests return with an immediate http 502 response.


.gitignore
```diff
@@ -1,3 +1 @@
 node_modules
-/client/live.bundle.js
-/client/index.bundle.js
```
sokra: Why removing this?

mrsharpoblunto: This change wasn't in the original pull request. I included live.bundle.js and index.bundle.js in my branch so that I could npm install from my repository. The added files and the change to .gitignore can be reverted (the changes in f170088)

MrOrz:
:+1: for this patch.

I make webpack-dev-server proxy point to my backend server. It was a real pain to kill the remaining backend server processes every time I made a typo that crashes the proxy. After switching to @mrsharpoblunto 's branch, I never have to do that again!

.gitignore
mrsharpoblunto: I've reverted the added files and .gitignore changes now

sokra: I'll cherrypick the first commit, when merging this..

sokra:
merged

PR #171, opened 2015-05-07
Update connect-history-api-fallback dependency and allow passing params to it

princed:
To make connect-history-api-fallback [options](https://github.com/bripkens/connect-history-api-fallback#options) configurable


sokra:
Thanks

princed:
Thanks for merging!

PR #190, opened 2015-06-02
Adding setup option for pre-configuring app middleware.

gmac:
By passing a `setup` function as a server option, one can pre-configure app middleware before `webpack-dev-middleware` is installed. This is useful for installing pre-webpack middleware that inspects an incoming request and makes adjustments to the Webpack server if needed.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -49,6 +49,11 @@ function Server(compiler, options) {
 	// Init express server
 	var app = this.app = new express();
 
+	// Allow pre-setup of app middleware
+	if (typeof options.setup === 'function') {
+		options.setup(app);
+	}
+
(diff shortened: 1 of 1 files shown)


heebtob:
I would love to see this feature. In my case I would like to use the passport middleware for authentication.
:+1:

danielstefanovic:
+1

mathieumg:
+1 :+1: A way to do the same post-setup could come in handy as well, I imagine.

mehwww:
+1

jnv:
:+1: 

Or perhaps you could have a `middleware` option similar to [browserSync](http://www.browsersync.io/docs/options/#option-server), although this is feasible with this solution.

PR #196, opened 2015-06-15
Proof of concept: send hot updates via sockets

gaearon:
I know this is pretty ridiculous with reading from `window` but that's not the point. It's just a proof of concept.

This PR works in conjunction with [this PR to Webpack](https://github.com/webpack/webpack/pull/1173).

It embeds small hot updates right into the socket response so that the client doesn't need to do any further requests. This saves some roundtrip time and, most importantly, lets the client go on without asking the server *while the server is performing the second pass* in `multiPass` mode. That's where the real savings are.

This cuts hot reload time (from “save” to seeing the changes) from 1500ms to 700ms on my machine.

I don't ask this to be merged, but I wonder if something like this makes sense for a future version. Making hot reload faster is important as it saves the brain cycles.

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -27,8 +27,19 @@ io.on("still-ok", function() {
 	console.log("[WDS] Nothing changed.")
 });
 
-io.on("ok", function() {
-	if(initial) return initial = false;
+io.on("ok", function(assets) {
+	if (!initial && !assets) return;
+	if (initial) return initial = false;
+
+	if (assets) {
+		if (!window.PRELOADED_HOT_ASSETS)
+			window.PRELOADED_HOT_ASSETS = {};
+
+		for (var url in assets) {
+			window.PRELOADED_HOT_ASSETS[url] = assets[url];
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


SpaceK33z:
Closing for now because the PR in webpack was also closed, and this would need a lot of work to implement correctly. Also it's been stale for over a year. If anyone would like to pick it up, please make a ticket first and we can discuss.

gaearon:
I never really knew what happened to webpack PR either because it was just silently closed.
Presumably the underlying issue was fixed?

SpaceK33z:
@gaearon, I don't know what happened either. I think that this is a really good idea though, so I've brought this up in the webpack team. They will discuss this in their meeting for next week.

gaearon:
👍 Thanks

everthis:
I think it's useful for those who develop webapps on remote machines.

niieani:
Any news on this front?

SpaceK33z:
@niieani nope, but feel free to finish this PR and the corresponding webpack one.

PR #198, opened 2015-06-18
Allow contentBase to override proxy

wmertens:
Only use the proxy if a file can't be served from contentBase. This allows greater development flexibility.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -219,13 +219,13 @@ function Server(compiler, options) {
 	};
 
 	var defaultFeatures = ["setup", "headers", "middleware"];
+	if(options.contentBase !== false)
+		defaultFeatures.push("contentBase");
 	if(options.proxy)
 		defaultFeatures.push("proxy");
 	if(options.historyApiFallback)
 		defaultFeatures.push("historyApiFallback", "middleware");
 	defaultFeatures.push("magicHtml");
-	if(options.contentBase !== false)
-		defaultFeatures.push("contentBase");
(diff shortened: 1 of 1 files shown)


sokra:
Could you update this PR to the latest commit?
Could you also check for correct order of the `if (options.historyApiFallback)` block?

watsoncj:
Is it me or is the proxy feature no longer reachable? Similar comments made in #213.

wmertens:
@sokra I updated the PR, with the new code it's much simpler :)

IMHO, the `historyApiFallback` is in the correct location.

sokra:
Actually I implemented the `features` feature because of your PR. The point is: The order (of the features) depends on the use case. With the `features` feature you can configure the order with the `features` option. Just override the order with i. e. `features: ["setup", "headers", "middleware", "contentBase", "proxy"]`. The order "proxy" before "contentBase" is more common so it's the default.

If you want to do something, you can add the `"features"` option to the documentation. :smile: 

Closing the PR, because it's now obsolete.

pix0r:
While allowing explicit control over `features` seems very flexible, it seems that someone with a limited understanding of webpack-dev-server (like myself) would find it easier if there was a documented config setting to prioritize `contentBase` over `proxy` without changing the way the other features are prioritized.

I was able to solve my problem by copying the runtime `defaultFeatures` into my config and editing from there to move `contentBase` ahead of `proxy`, but I could see this causing problems down the line if I change my config and don't update `features` accordingly.

@sokra I'm happy to lend a hand but not sure how (or if) you'd like to address this use case.

PR #204, opened 2015-06-26
Don't emit "still-ok" when there are errors

Stuk:
Emitting "still-ok" and returning means that no "error" event is emitted , forcing the user to manually refresh the page to see compile errors.

This change causes compile errors to be shown with no manual refresh needed.

Before:
before

After:
after


sokra:
Thanks.

Stuk:
Thank *you* for the tool and the merge!

PR #231, opened 2015-07-31
Fix: support html base tag via public path option

valorkin:
Hello @chentsulin 

Here is PR solving one particular case,
I have a project consisting from several angular apps
with set html base tag
```html
<base href="/user/">
<base href="/admin/">
...
```

So basically I am serving several angular apps on different routes
and this peace of code solves this problem

**BUT** it starts to serve files from file system!
how it could be fixed?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
+.idea
 /client/live.bundle.js
 /client/index.bundle.js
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
+.idea
```
SpaceK33z: This needs to be in your global `.gitignore`, it is not something specific to this repo.

SpaceK33z:
@valorkin, what do you want to do with this PR?

SpaceK33z:
I'm closing this PR because of inactivity. Feel free to comment, and we can always re-open it again.

PR #233, opened 2015-08-06
Gzip compression option

vectart:
Using this feature developer can evaluate gzipped size of generated assets.
It's also useful to check and optimise page load speed using throttled network and gzip compression.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -29,6 +29,8 @@ var optimist = require("optimist")
 
 	.boolean("https").describe("https")
 
+	.boolean("compression").describe("compression", "Enable gzip compression")
+
 	.string("content-base").describe("content-base", "A directory or URL to serve HTML content from.")
 
 	.string("content-base-target").describe("content-base-target", "Proxy requests to this target.")
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -10,6 +10,7 @@ var httpProxy = require("http-proxy");
 var proxy = new httpProxy.createProxyServer();
 var serveIndex = require("serve-index");
 var historyApiFallback = require("connect-history-api-fallback");
+var compression = require('compression');
 
 function Server(compiler, options) {
 	// Default options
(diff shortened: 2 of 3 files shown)


nkbt:
Why would you ever need compression for dev server?

vectart:
@nkbt Yes, the rationale is in PR description. As a developer, I want to evaluate size of initial assets considering gzip compression, network throttling and other aspects without deployment to stage or production server.

This compression option could be as useful as HTTPS one.

nkbt:
I find it quite awkward that people use webpack-dev-server for this matter. Usually if you want to evaluate speed/compression/https/etc you run some setup that is not too different from your production one. Dev is meant to be super-fast and dead-simple... But okay, if you still want it, why not to use webpack middleware with a simple express server and do any configuration including gzip in the app without making webpack-dev-server itself more complex? 

Okay, I made my point here. Not saying I am right, though. Maybe this update actually makes sense.

vectart:
@nkbt I don't use middleware to avoid writing a custom tool since I just need an option that adds only 8 lines and doesn't affect server speed so much. In result, I can see that my ~0.2–1Mb source files won't be a big deal in production.

Basically, the `compression` option doesn't look strange along with existing `historyApiFallback`, `https`, `headers` and `proxy` ones.

syranide:
+1 because JS compresses really well and this makes remote testing significantly faster too.

sokra:
This initial reasons also sound invalid for me.

But @syranide reasons looks valid.

> +1 because JS compresses really well and this makes remote testing significantly faster too.

So we better should enable this by default without an option, shouldn't we?

syranide:
> So we better should enable this by default without an option, shouldn't we?

@sokra It could imaginably delay the transmission slightly, but I imagine it probably cancels out by being faster... however, a lot of non-image resources (like webfonts) benefit from being compressed as well, so I would think the best solution is to accept a list of file extensions (or suffixes) to serve compressed, with the default being `.js` + `.map`?

SpaceK33z:
The PR #274 is already merged, which implements gzip compression with the `--compress` flag.

Issue #205, opened 2015-06-26
Can't get hot-reload to work for remote server with private IP


pwwolf:
I have a remote dev server with a private IP which I use to run webpack.  The machine is accessible from a public IP/host which is different from the IP that the server binds to.

If I specify --host 0.0.0.0, I can hit the server initially, but the reload doesn't work because my browser is trying to make a connection to "http://0.0.0.0:8080/socket.io/...".  Alternatively I've tried setting --host to my public IP, but it fails to start because it can't bind to that IP.

Is there any way to make the socket.io relative to whatever URL is in the browser rather than copying the --host value directly?... or just looking at document.location.origin

sokra:
If you don't use inline mode `--inline` you can specify the remote path in the query parameter of the client reference `webpack-dev-server/client?http://url:8080`

cdaringe:
hi @sokra, thanks for the great project.  can you add a bit more clarification?  this would be helpful to me as well!  _which_ client reference?  and can you confirm the proper query kv pair?  i assume my client needs to be uri encoded as well?

aaronjensen:
I just struggled w/ this for a while, and this is what I came up with:

```
inline: false
```

```
entries: ['webpack/hot/only-dev-server', 'your-entry.js']
```

In your index.html:
```html
    <script src="bundle.js"></script>
    <script>
      var devServer = document.createElement('script');
      devServer.setAttribute('src', 'http://' + window.location.host + '/webpack-dev-server.js');
      document.body.appendChild(devServer);
    </script>
```

This seems to work regardless of where you access it from...

kentor:
@aaronjensen did you get hot reloading to work? Even with your set up the page always refreshes because it can't find the manifest file. See https://github.com/webpack/webpack-dev-server/issues/262

aaronjensen:
@kentor yes, it works for me. I had to do something like `window.location.host.replace(':4000', ':4001')` because I was proxying through browser-sync, but other than that it works great.

eteeselink:
Note: this was fixed in this PR: https://github.com/webpack/webpack-dev-server/pull/263/files. the latest release on NPM includes this fix. It does exactly what @pwwolf suggested:

> Is there any way to make the socket.io relative to whatever URL is in the browser rather than copying the --host value directly?... or just looking at document.location.origin

but only if the host passed was `"0.0.0.0"`.

eteeselink:
(cc @sokra I therefore believe this issue can be closed)

[...]

PR #236, opened 2015-08-09
Added a flag for public ip to seed the inline frame.

wolfeidau:
I am currently using `webpack-dev-server` inside a docker container, this requires more control of the bind address and external public facing address to ensure it works.

To manage this I have added a flag which enables overriding of the ip address which is used in the `--inline` mode.

An example of this can be seen below, the server binds to `0.0.0.0` and the inline client uses `123.123.23.14`.

```
webpack-dev-server \
-d --hot --inline --display-reasons --display-error-details --history-api-fallback --progress \
--colors --port 8080 --host 0.0.0.0 --public 123.123.23.14
```

This provides a fix for #205

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -46,6 +46,8 @@ var optimist = require("optimist")
 
 	.describe("port", "The port").default("port", 8080)
 
+	.describe("public", "The public hostname/ip address of the server")
+
 	.describe("host", "The hostname/ip address the server will bind to").default("host", "localhost");
 
 require("webpack/bin/config-optimist")(optimist);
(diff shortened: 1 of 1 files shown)


sokra:
The `public` option should default to the `host` option, elsewise this would be a breaking change.

maybe `"://" + (options.public || options.host) + ":"` and no default value.

wolfeidau:
As recommended I have removed the default and as suggested defaulted the host value inline.

I also rebased this to the latest version.

Cheers

chollier:
I'd love to see that merged

sokra:
To support the usecase of #293 `public` should not only set the host part but also the port part:

``` diff
- "://" + (options.public || options.host) + ":" + options.port];
+ "://" + (options.public || (options.host + ":" + options.port))];

wolfeidau:
Updated as specified.

Cheers

zerkms:
Guys, any plans to merge this or introduce the public `ip:port` with some other solution?

sokra:
> Guys, any plans to merge this or introduce the public ip:port parameter with some other solution?

I'll merge this.

jeffreybrowning:
+1

PR #237, opened 2015-08-09
Sync url hash change to top frame for convenience.

danielking:


danielking:
This will make problems for page navigating.

PR #238, opened 2015-08-09
Add --cert, --key and --cacert options to CLI.

### LINKED ISSUE #235: CLI option to add SSL key and certificate
At the moment I need to use the API only because the CLI doesn't allow me to add a custom SSL key and certificate.

Would you be interested if I create a PR for this?

SpaceK33z:
This adds a `--cert`, `--key` and `--cacert` option to the CLI. The option names are the same ones curl also uses, so I think that makes sense.

Fixes #235

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -2,6 +2,7 @@
 
 var path = require("path");
 var url = require("url");
+var fs = require("fs");
 
 // Local version replaces global one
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

patrickhousley:
When can we expect a build with this code in it?

sokra:
Just published...

Issue #239, opened 2015-08-11
Improve server URL deduction


jeremejevs:
Right now, to get the server URL, the [client code](https://github.com/webpack/webpack-dev-server/blob/master/client/index.js#L6) just takes a random `script` element and removes everything from its source URL after the last slash, including the slash itself.

1. If there's no slash in that URL - it takes the script's name = `http://bundle.js/`.
2. If there's a dot and a slash - it takes just the dot = `http://./`.

My HTML file is auto-generated, so I have to use `webpack-dev-server/client?/` as my entry.

Consider utilizing `window.location` instead? Or improving the heuristics?

timkrins:
For my own development workflow, as I am only ever going to be serving on the root of a local network IP or localhost, I changed my index.js to say
`io = io.connect(window.location.host);`
Works for me, but obviously not elegant

SpaceK33z:
The heuristics have improved a lot lately. Does it work for you now, with `1.15.0`?

jeremejevs:
I don't remember what it was supposed to be, never committed it, sorry. Just removed `?/` and it seems to be still working, so there's that.

SpaceK33z:
Okay cool, so closing this because the heuristics are good enough now. There's still room for improvement, but that should be done with specific issues.

Issue #220, opened 2015-07-22
Add ability to configure the socket.io path


mingfang:
I'm running the webpack-dev-server behind a proxy server and need to change the default socket.io path from 
```/socket.io``` to ```/<myapp>/socket.io```.

Is this possible?

**Comment by vectart** [comment]
+1

**Comment by michaelzoidl** [comment]
+1

**Comment by syranide** [comment]
+1, somewhat related #46

**Comment by syranide** [comment]
I've put up a PR #240 for this.

**Comment by frontengineer** [comment]
+1

**Comment by syranide** [comment]
#240 merged

**Comment by mingfang** [comment]
#240 is a huge setup forward but it's not enough.
Setting host to window.location.host works.
But we also need to set port to window.location.port.
```
port: urlParts.port ? urlParts.port : window.location.port
```

SpaceK33z:
Setting the port was fixed in #430, which will be part of a 1.x release.

PR #240, opened 2015-08-13
Allow changing socket.io-path via query string

syranide:
Related #46, #220, (kind of) #239

This PR allows you to use `'webpack-dev-server/client?http://mydomain/mysocket.io/'` (domain is optional, just as path is). I.e. being able to change the `socket.io` path too and not just the domain. If there is no path you get the current behavior. `url.parse` parses `http://example.com` as having path `/` (so we must consider it no path), but that's not really a problem and preferable to avoid breaking existing configs.

If anyone wants to try it:
`"webpack-dev-server": "git://github.com/syranide/webpack-dev-server#socketiopath_npm"`

cc @sokra

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -1,9 +1,21 @@
+var url = require('url');
 var io = require("socket.io-client");
 var stripAnsi = require('strip-ansi');
 var scriptElements = document.getElementsByTagName("script");
-io = io.connect(typeof __resourceQuery === "string" && __resourceQuery ?
+
+var urlParts = url.parse(typeof __resourceQuery === "string" && __resourceQuery ?
 	__resourceQuery.substr(1) :
 	scriptElements[scriptElements.length-1].getAttribute("src").replace(/\/[^\/]+$/, "")
+);
+
+io = io.connect(
+	url.format({
+		protocol: urlParts.protocol,
+		auth: urlParts.auth,
+		host: urlParts.host
+	}), {
+		path: urlParts.path === '/' ? null : urlParts.path
+	}
 );
 
 var hot = false;


kentor:
Does this allow changing the domain? It's hard to test in VM where I use `10.0.2.2` instead of `localhost` to access the dev server asset, but the script still tries to access `localhost` for the socket.

syranide:
@kentor That already works, this allows changing the path too.

sokra:
Thanks

SimenB:
This destroys HMR for us. Using `client/index.js` as it was before this PR works.

image

syranide:
Hmm that's an interesting edge-case, you want to set the port but not the host?

SimenB:
That's correct. We always know the port as that's the port of `WebpackDevServer`. But the host IP changes. It's basically `localhost`, but setting either `localhost` or `0.0.0.0` makes it not available in the LAN. We want to access the running server from other machines than just the one running the dev-server

syranide:
It seems pretty clear to me that `http://:::3030` was a hack and should not be valid. It also seems to me that there's a bug in `url.parse`, when parsed and formatted it ends up as `http://:3030/::`, that seems nonsensical (but perhaps that simply an acceptable artifact due to the malformed url?).

Anyway, I imagine the correct approach here would be to special-case `0.0.0.0` and have it too be substituted with `location.host`.

cc @sokra

SimenB:
`http://:::3030` is perfectly valid IPv6. https://en.wikipedia.org/wiki/IPv6_address#Presentation

syranide:
@SimenB True, in that case then it's really a bug in `url.parse` that should be fixed in node and `0.0.0.0` won't be necessary. Also... I got lost in my thoughts a bit, `http://:3030` actually parses and formats correctly, however it ends up going wrong somewhere and ends up as `http://http`, so it seems the problem really is elsewhere. So that needs to be investigated, but it seems to me right now that this is not a bug in my PR (it just closed a previous loophole).

syranide:
Hmm, it seems `0.0.0.0` already happened...
https://github.com/webpack/webpack-dev-server/blob/master/client/index.js#L15

SimenB:
Using `0.0.0.0` works now. No idea why it didn't before... It's been that way for 6 months I don't really remember why we went with `::`

kentor:
@SimenB where do you specify that? I'm guessing that would solve my issue https://github.com/webpack/webpack-dev-server/issues/262

SimenB:
Specify what? The entry?

```js
  webpackConfig.entry = [
    'webpack-dev-server/client?http://0.0.0.0:3030',
    'webpack/hot/only-dev-server',
    path.join(__dirname, 'js/'app.js'),
  ];
```

syranide:
@SimenB `0.0.0.0` was added at the same time as my PR it seems, so just this last version.

SimenB:
Ah, ok. Makes sense :smile:

syranide:
@SimenB Also, FYI https://github.com/nodejs/node/issues/2929#issuecomment-141123167

SimenB:
Hangs forever using `'webpack-dev-server/client?http://[::]:3030'` Doesn't throw the error though

image

kentor:
@SimenB That works for connecting to the socket. But what do you specify as the `output.publicPath`?

SimenB:
@kentor `output.publicPath = '/'`

SimenB:
Doesn't work on windows, so for now we're stuck on 1.10.1...

image

syranide:
@SimenB The final URL is really funky, including encoded chars (which seems to indicate an URL being encoded as a hostname inside an URL) ... what have you written in your config? I'm assuming the URL you are visiting is `http://localhost:3030`?

https://github.com/webpack/webpack-dev-server/blob/master/client/index.js#L13

SimenB:
Entry is still `'webpack-dev-server/client?http://0.0.0.0:3030'`, and yes, localhost

syranide:
@SimenB Hmm, I don't see how that ends up being translated to what you printscreened above... could you perhaps put some `console.log`s in https://github.com/webpack/webpack-dev-server/blob/master/client/index.js#L13 and see where it's going wrong?

```js
// on line 10
console.log(urlParts);
console.log(window.location.hostname);
console.log(
        url.format({
		protocol: urlParts.protocol,
		auth: urlParts.auth,
		hostname: (urlParts.hostname === '0.0.0.0') ? window.location.hostname : urlParts.hostname,
		port: urlParts.port
	})
);
```

syranide:
Also:

```js
console.log(__resourceQuery);
```

SimenB:
Using 
```js
console.log(__resourceQuery);
console.log(urlParts);
console.log(window.location.hostname);
console.log(
  url.format({
    protocol: urlParts.protocol,
    auth: urlParts.auth,
    hostname: (urlParts.hostname === '0.0.0.0') ? window.location.hostname : urlParts.hostname,
    port: urlParts.port
  })
);
```

Linux (working):
image

Windows (failing):
image

syranide:
`__resourceQuery` is obviously all wrong on windows for some reason. So this is related to some weird manipulation by webpack or quirk in your config.

SimenB:
Oh seems to be Lync messing up the copy paste... Using `0.0.0.0` works!

syranide:
> I still think url.parse messes up, but this works for us now :smile:

It parses valid URLs correctly, so as far as we should be concerned it behaves as intended.

mingfang:
can we get port to be like this?
```
port: urlParts.port ? urlParts.port : window.location.port
```

jedwards1211:
@SimenB would the approach you guys are discussing work when connecting to your dev machine from other devices?

To accomplish that when I need to change the port I've been doing the following:
```html
  <script type="text/javascript">
    var scriptElem = document.createElement('script');
    scriptElem.type = 'text/javascript';
    scriptElem.src = new RegExp('https?://[^:/]+').exec(window.location.href)[0] + ':9090/assets/client.bundle.js';
    document.head.appendChild(scriptElem);
  </script>
```

SimenB:
Yeah, connecting from other machines works fine, as long as they're on the same LAN.

jedwards1211:
@SimenB What do your script tags look like?  I've been curious myself if there's a better way

SimenB:
I use `html-webpack-plugin`, so no manual script tags

jedwards1211:
Oh, are you using Meteor?

SimenB:
No, it's Backbone app, but the templates are using Handlebars, so I kept going with that for html-webpack-plugin (as I really dislike blueimp). Next version of the plugin will support lodash templates though, so we'll probably switch to that.

jedwards1211:
I see.  But now you're just using `http://0.0.0.0:3030...` for the `src` links of the scripts?  I didn't realize this would work

SimenB:
The script tags are generated, and are the entry chunks (of which we have 1).

image

No specifyin the port in scripts, as webpack-dev-server handles that part

PR #250, opened 2015-09-01
Add option to bypass proxying requests based on a function call

whiteinge:
The primary use-case behind this addition is to allow HTML5 history API support for proxied paths.

The example below uses the same pattern as the `historyApiFallback` URL rewriting code to avoid the extra behavior (proxying) when the request comes from a non-ajax browser request.

```
proxy: {
    '/thepath*': {
        target: 'https://otherserver.example.com',
        bypass: function(req) {
            if (req.headers.accept.indexOf('html') !== -1) {
                console.log('Skipping proxy for browser request.');
                return '/index.html';
            }
        },
    },
},
```

I'm very option to suggestions for alternate implementations that make better reuse of the existing `historyApiFallback` code. I'm not very familiar with Express middleware yet.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -134,19 +134,25 @@ function Server(compiler, options) {
 				}
 				options.proxy.forEach(function (proxyOptions) {
 					proxyOptions.ws = proxyOptions.hasOwnProperty('ws') ? proxyOptions.ws : true;
-					app.all(proxyOptions.path, function (req, res) {
-						if(typeof proxyOptions.rewrite === 'function') proxyOptions.rewrite(req, proxyOptions);
-						if (proxyOptions.host) {
-							req.headers.host = proxyOptions.host;
-						}
-						proxy.web(req, res, proxyOptions, function(err){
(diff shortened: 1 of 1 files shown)


whiteinge:
Oh, and [diff is best viewed with whitespace changes ignored](https://github.com/webpack/webpack-dev-server/pull/250/files?w=1) since I just wrapped the existing code in an if-statement. :wink:

whiteinge:
Polite bump (with the hope said 'bump' isn't overlapping with deadlines, fires, vacations, etc). @sokra the contributor graph suggests you're the sole primary over this repo. What are you thoughts on the above? Right approach / wrong approach? (And thanks for your time.)

whiteinge:
Rebased and resolved conflicts on top of the 'features' changes in 931bbaf.

sokra:
Thanks

sokra:
Could you add it to the documentation?

whiteinge:
[Happy to!](https://github.com/webpack/docs/wiki/webpack-dev-server/5aa0adf6ca135cc1a3a2c90d47106ab56a3e7314)

PR #255, opened 2015-09-04
Added support for multiple static virtual paths

ctindel:
I looked through a lot of different node.js projects before writing this trying to find a way to do what I want to do.  Given a complicated source tree that has a number of different subdirectories all supplying some static files I wanted a way to be able to define multiple static virtual paths for the webpack-dev-server.  

So, I added support for a new parameter "staticRoutes" which expects an array of objects that have a route and dir fields.  route specifies the static route as it will be seen from the browser and dir specifies a the directory in your source tree that will host those files.  For example:

```
var config = {
  devServer: {
    contentBase: 'src/public',
    publicPath: '/__build__',
    staticRoutes: [ { 'route': '/npm_scripts', 'dir' : __dirname + '/node_modules'},
                    { 'route': '/bower_scripts', 'dir' : __dirname + '/bower_components'} ]
  }
}
```

Now you will be able to specify something like:

```
<link rel="stylesheet" type="text/css" href="/npm_scripts/bootstrap/dist/css/bootstrap.css">
<script src="/bower_scripts/jquery/dist/jquery.js"></script>
```

In your html and those files will be served appropriately.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -137,6 +137,14 @@ new Server(webpack(wpOpt), options).listen(options.port, options.host, function(
 	else
 		console.log(protocol + "://" + options.host + ":" + options.port + "/webpack-dev-server/");
 	console.log("webpack result is served from " + options.publicPath);
+	if(options.staticRoutes) {
+		// Must be array
+		if (Array.isArray(options.staticRoutes)) {
+			options.staticRoutes.forEach(function (staticRoute) {
+				console.log("static route " + staticRoute.route + " is served from " + staticRoute.dir);
+			});
+		}
+	}
 	if(typeof options.contentBase === "object")
 		console.log("requests are proxied to " + options.contentBase.target);
 	else
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -149,6 +149,19 @@ function Server(compiler, options) {
 		app.use(this.middleware);
 	}
 
+	if(options.staticRoutes !== false) {
+		// Must be array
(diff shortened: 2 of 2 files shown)


aronwoost:
I would find this handy.

Is there a reason why this PR is pending? Input needed?

ctindel:
Yeah I'm not sure why it's pending, I'd have to resolve the conflicts that arose since then but only if he's interested in merging the PR in.

aronwoost:
Who is he?

ctindel:
https://github.com/sokra

aronwoost:
@sokra Is this feature of any interest for the project?

I'd would solve some issues for be, where assets come from different places.

Or do you think this should better live in some kind of middleware?

SpaceK33z:
This should be done in some kind of middleware, dev-server is not meant for assets outside of webpack. You could also run any webserver to serve these assets.

ctindel:
@SpaceK33z can you elaborate a bit more on what you mean by "assets outside of webpack" ?  I mean, why provide any option for serving static assets since they can all be served by middleware or a webserver?  Thanks.

SpaceK33z:
@ctindel, for assets that are bundled with webpack, you don't have this problem. If you would import the css and jquery in your webpack bundle, you wouldn't need to define custom static routes. The dev-server is built to handle webpack assets, not assets outside of this ecosystem.

ImJoeHs:
@SpaceK33z What about proxy? I mean dev server already handle the proxy to outside api, wouldn`t it be better to serve static files also?

PR #261, opened 2015-09-13
Correct console message for historyApiFallback

at0g:
Tiny update to correct the console message when passing an `index` option to the historyApiFallback.

With the following config, the console message should say: *404s will fallback to /default.html*

```
// webpack.config.js
module.exports = {
  devServer: {
    historyApiFallback: {
      index: '/default.html'
    }
  }
}
```


sokra:
Thanks

Issue #262, opened 2015-09-14
Dynamic hostname and port of hot update manifest


kentor:
Say my app is served on port 3000, and I have webpack dev server running on port 45537. I'm only using the dev server for assets/hot reloading, not serving the html files, that is taken care of by something else (django, rails, whatever). The thing is, I may access this app using different hostnames or IPs (e.g. `localhost`, `10.0.2.2`, the internal network IP of the server) so I want the hot reloading stuff to work no matter how/where I am accessing the app. 

Command used to start webpack-dev-server:
```
webpack-dev-server --config webpack.hot.config.js --host 0.0.0.0 --port 45537 --hot --no-info
```

conifg:
```js
const path = require('path');

module.exports = {
  entry: [
    'webpack-dev-server/client?http://0.0.0.0:45537',
    'webpack/hot/only-dev-server',
    path.join(__dirname, 'src', 'js', 'app.js'),
  ],
  module: {
    loaders: [
      {
        exclude: /node_modules/,
        loaders: ['react-hot', 'babel-loader'],
        test: /\.js$/,
      },
    ],
  },
  output: {
    filename: 'app.js',
    publicPath: '/',
  },
};
```

In my html file, I have this at the bottom to load webpack-dev-server.js and my app bundle:
```html
    <script>
      !function() {
        var host = location.host.split(':')[0];
        document.write(
          '<script src="http://' + host + ':45537/app.js"></' + 'script>'
        );
      }()
    </script>
```

The problem is when I reload a file, it tries to grab the manifest from port 3000 instead of 45537:

```
[WDS] App updated. Recompiling...
[WDS] App hot update...
[HMR] Checking for updates on the server...
GET http://192.168.1.45:3000/0f047ad04fb3ea0a8b54.hot-update.json 404 (Not Found)
```

Of course, specifing the `publicPath` as `http://localhost:45537/` would only work if I am on the same computer as the dev server, so that's not what I am looking for.

sokra:
You can set the `publicPath` at runtime (in webpack compiled modules):

``` js
__webpack_public_path__ = 'http://' + host + ':45537/js/'
```

kentor:
I'm not entirely sure I understand. I followed http://webpack.github.io/docs/configuration.html#output-publicpath, (the warning about the Hot Module Replace Plugin needing to be disabled scares me, if I can't use HMR then what's the point of this?) and created a file called `webpack.init.js` and in it put

```js
[...]

Issue #160, opened 2015-04-22
Responding to localhost as well as dev IP address


subblue:
Is there a way to get the webpack-dev-server to serve content from both `localhost` but also requests from my local dev machine's IP address, or the `name.local` OSX address (for when testing from mobile devices)?

At the moment I have to specifically set the `host` option, which is a little cumbersome when my IP address changes and I don't want to hard code these details in the config script.

Thanks.

nkbt:
If you set server's listen address to 0.0.0.0 it will listen to whatever address is resolved on your machine. Not secure of course, but if you are only accessible from local network, then you should be ok.

talmobi:
Exactly what I was looking for, thanks you all.

SpaceK33z:
Closing because it has been answered a while ago.

PR #263, opened 2015-09-14
allow bind to all host (--host 0.0.0.0)

thenewguy:
'0.0.0.0' allows webpack-dev-server to bind to all addresses on the machine but cannot be navigated to so use window.location.hostname in this case


thenewguy:
#262 #205 #160

thenewguy:
I've raised an issue for the PR at #265

sokra:
Thanks

timkrins:
Yes, fantastic! Works nicely. Somewhat related to #239

Issue #191, opened 2015-06-04
Can not find /build/Release/bufferutil.node, /build/Release/validation in ws


chentsulin:
When I start my dev-server, got those warning:

```
WARNING in (webpack)-dev-server/~/socket.io-client/~/engine.io-client/~/ws/lib/BufferUtil.js
Module not found: Error: Cannot resolve 'file' or 'directory' ../build/Release/bufferutil in /Users/chentsulin/Projects/electron-react-boilerplate/node_modules/webpack-dev-server/node_modules/socket.io-client/node_modules/engine.io-client/node_modules/ws/lib
 @ (webpack)-dev-server/~/socket.io-client/~/engine.io-client/~/ws/lib/BufferUtil.js 8:19-57

WARNING in (webpack)-dev-server/~/socket.io-client/~/engine.io-client/~/ws/lib/BufferUtil.js
Module not found: Error: Cannot resolve 'file' or 'directory' ../build/default/bufferutil in /Users/chentsulin/Projects/electron-react-boilerplate/node_modules/webpack-dev-server/node_modules/socket.io-client/node_modules/engine.io-client/node_modules/ws/lib
 @ (webpack)-dev-server/~/socket.io-client/~/engine.io-client/~/ws/lib/BufferUtil.js 10:19-57

WARNING in (webpack)-dev-server/~/socket.io-client/~/engine.io-client/~/ws/lib/Validation.js
Module not found: Error: Cannot resolve 'file' or 'directory' ../build/Release/validation in /Users/chentsulin/Projects/electron-react-boilerplate/node_modules/webpack-dev-server/node_modules/socket.io-client/node_modules/engine.io-client/node_modules/ws/lib
 @ (webpack)-dev-server/~/socket.io-client/~/engine.io-client/~/ws/lib/Validation.js 8:19-57

WARNING in (webpack)-dev-server/~/socket.io-client/~/engine.io-client/~/ws/lib/Validation.js
Module not found: Error: Cannot resolve 'file' or 'directory' ../build/default/validation in /Users/chentsulin/Projects/electron-react-boilerplate/node_modules/webpack-dev-server/node_modules/socket.io-client/node_modules/engine.io-client/node_modules/ws/lib
 @ (webpack)-dev-server/~/socket.io-client/~/engine.io-client/~/ws/lib/Validation.js 10:19-57
```

I finally figure out this is a issue with `engine.io-client`

See https://github.com/Automattic/engine.io-client/issues/376

[...]

PR #272, opened 2015-09-19
Downgrade socket.io to work with node 4.x

noxan:
Solution was proposed in socketio/socket.io/issues/2228

(my local version lacked of `live.bundle.js` first, but guess this is related to the publishing procedure of this project - fixed with npm run-script prepublish)


noxan:
may fix this issue, but would reopen #191 - might be better to just wait for an fixed socket.io version.

PR #274, opened 2015-09-20
Add --compress option to enable gzip compression

everett1992:
The --compress option uses the `compression` express middleware to gzip
compress responses from the dev server.

A development bundle for a babel + react project can easily be 1MB. I work on a remove development machine where I would take up to three seconds just to fetch the bundle. Gzip here makes a huge difference.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -42,6 +42,8 @@ var optimist = require("optimist")
 
 	.boolean("history-api-fallback").describe("history-api-fallback", "Fallback to /index.html for Single Page Applications.")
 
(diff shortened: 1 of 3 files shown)


sokra:
Thanks

everett1992:
Is there any documentation that should be updated? I can take care of it if you point me to the repos.

What's the release schedule for this project like?

sokra:
http://webpack.github.io/docs/webpack-dev-server.html

in the top-right corner is an edit button

everett1992:
Cool, I'll update the docs once I this change is published to npm

PR #275, opened 2015-09-21
do not reload on error

Sinewyk:
This remove a useless page reload to print an error when it was already printed before page reload.

I dunno if that's going to break things for people, but it seems weird that nobody wanted this or proposed this.

NB: live.js actually dos NOT reload when there are errors, this would make the client match this behaviour.

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -58,7 +58,6 @@ io.on("errors", function(errors) {
 	for(var i = 0; i < errors.length; i++)
 		console.error(stripAnsi(errors[i]));
 	if(initial) return initial = false;
-	reloadApp();
 });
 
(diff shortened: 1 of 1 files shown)


sokra:
This could break hot reloading, as the app doesn't get notified on error.

cc @gaearon

Sinewyk:
The app does get notified of an error. It just doesn't reload once for nothing anymore. And after you fix your mistake the hot reloading picks up again and reload correctly.

I'm using `react-hot-loader` + `webpack/hot/dev-server` and so it reloads when it needs to reload (if I touch something not hot compatible) and hot-reload when it can (right now when editing React stuff only).

PR #279, opened 2015-09-23
Add setup option to prepend custom middlewares

jnv:
Take two after #190, this time using the new fancy `features` object.

Example usage:

```js
devServer: {
  //...
  setup: function (app) {
    app.use(function(req, res, next) {
      res.cookie('my_fancy_cookie', {lorem: 'ipsum'}, {httpOnly: false})
    })
  },
},
```


sokra:
Thanks

PR #284, opened 2015-10-02
Allow iframe to use browser fullscreen API

watsoncj:
This attribute allows the application in the iframe to use the fullscreen API.


sokra:
Thanks

Issue #249, opened 2015-09-01
Add API to close dev server


tricoder42:
I'm trying to close webpack dev server programatically, so I can run it it task with webdriverio.

The webpack API already provides *listen* method which delegates listen call and inits socket.io. I tried to add close method with similar functionality, but couldn't get it work (the process still hangs). 

```javascript
Server.prototype.close = function(callback) {
	this.io.close();
	this.io = null;

	this.listeningApp.close(callback);
	this.listeningApp = null;
}
```

Any ideas?

TheLarkInn:
Fixed

cbelhaffef:
Hi,
How i can stop webpack-dev-server programaticaly ? like webpack-dev-server close() for example ...

SpaceK33z:
Like you said, with `close()`?

cbelhaffef:
i used but nothing happen.. :(

SpaceK33z:
Which version of dev-server are you using? Can you share the relevant part of your code?

cbelhaffef:
version : 1.14.1

var WebpackDevServer = require("webpack-dev-server");
var webpack = require("webpack");

`var compiler = webpack({
......
.....
});
var server = new WebpackDevServer(compiler, {  
});
server.listen(3000, "localhost", function() {});

server.close();
`

the server started but not closed.

SpaceK33z:
Can you try changing `localhost` to `127.0.0.1`? This causes a lot of weird errors (see #183).

cbelhaffef:
Hi, 
nothing changed.

pavel06081991:
@SpaceK33z, have the same issue, "webpack-dev-server": "^2.4.2", the server is alive even after calling stop() method. Do you have any idea why it happens?

pkearney13:
Same here..

AlexLeung:
After I call the close it still seems to be alive.

PR #287, opened 2015-10-06
add close() api

jhchen:
Implements close as described in #249. I don't believe it's possible to support a callback since socket.io's close does not have a callback. Happy to add some tests though not sure how the are split up amongst the various webpack libraries.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -271,6 +271,11 @@ Server.prototype.listen = function() {
 	}.bind(this));
 }
 
+Server.prototype.close = function() {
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

sokra:
Could you add it to the documentation?

jhchen:
Of course. I just added a commented out call to https://github.com/webpack/docs/wiki/webpack-dev-server in case people copy/paste the code snippet. Let me know if there is a better way or place.

PR #289, opened 2015-10-06
Lookup files in webpack fs, fallback to real fs

DmitryKorlas:
This changes allow to use webpack-processed version of resources instead of read them 'as is' from `contentBase`. 

So, when client request some resource, for example `/index.html` it will lookup such resource in webpack middleware fs. If found - webpack-processed version will be returned, otherwise it will be requested from `contentBase` as is now.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -10,6 +10,7 @@ var https = require("https");
 var httpProxy = require("http-proxy");
 var proxy = new httpProxy.createProxyServer();
 var serveIndex = require("serve-index");
+var mime = require('mime-types');
 var historyApiFallback = require("connect-history-api-fallback");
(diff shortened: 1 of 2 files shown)


sokra:
I don't understand why this is neccesary. The middleware, which is applied before the contentBase, should serve the files from webpack fs.

Is there a incorrect `publicPath` configuration on your side?

DmitryKorlas:
Oh exactly, i see middleware reslove it. My mistake were not to include defined ```publicPath``` into url's to the certain templates (index.html etc) - is a reason to receive "raw" files via ```app.get("*", express.static(contentBase), serveIndex(contentBase))```. Thanks!

PR #298, opened 2015-10-17
Allow hot module replacement in "webworker" targets

elliottsj:
These changes are required to support hot module replacement when the bundle is run inside a web worker: https://github.com/webpack/webpack/pull/1521

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -1,12 +1,14 @@
 var url = require('url');
 var io = require("socket.io-client");
 var stripAnsi = require('strip-ansi');
-var scriptElements = document.getElementsByTagName("script");
 
-var urlParts = url.parse(typeof __resourceQuery === "string" && __resourceQuery ?
-	__resourceQuery.substr(1) :
-	scriptElements[scriptElements.length-1].getAttribute("src").replace(/\/[^\/]+$/, "")
-);
+var urlParts;
+if (typeof __resourceQuery === "string" && __resourceQuery) {
+	urlParts = url.parse(__resourceQuery.substr(1));
+} else {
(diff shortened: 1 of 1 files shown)


client/index.js
```diff
@@ -1,12 +1,14 @@
...
 
-var urlParts = url.parse(typeof __resourceQuery === "string" && __resourceQuery ?
-	__resourceQuery.substr(1) :
-	scriptElements[scriptElements.length-1].getAttribute("src").replace(/\/[^\/]+$/, "")
-);
+var urlParts;
+if (typeof __resourceQuery === "string" && __resourceQuery) {
+	urlParts = url.parse(__resourceQuery.substr(1));
+} else {
+	var scriptElements = document.getElementsByTagName("script");
```
elliottsj: Getting `<script>` elements is now behind a conditional to prevent web workers from accessing `document` (which would throw an error)

client/index.js
```diff
@@ -76,7 +78,19 @@ io.on("disconnect", function() {
...
+			hotEmitter.emit("webpackHotUpdate", currentHash);
+		} catch (err) {
+			console.warn(
+				"Failed to load \"webpack/hot/emitter.js\". Is your webpack dependency up to date?\n",
+				err
+			);
+		}
+		if (typeof WorkerGlobalScope === "undefined" || !(self instanceof WorkerGlobalScope)) {
+			// Not in a web worker; broadcast update to window
+			window.postMessage("webpackHotUpdate" + currentHash, "*");
```
elliottsj: Still doing `window.postMessage` here for compatibility with any 3rd-party script that still listens for `"webpackHotUpdate"` events on the window.

PR #300, opened 2015-10-20
[Bug Fix] error when a script on the HTML has no src

puppybits:
- when an element on the index.html page has a script tag with no source (let when using google analytics), webpack dev server will through an error.


jaylinski:
please merge this, I believe this would fix #367

jaylinski:
also see #396 #394

PR #303, opened 2015-10-27
Express middleware for History API fallback

cbas:
The Connect middleware has issues with dots in URLs due to how it
redirects the request. The Express middleware is much simpler and
leverages Request.sendFile (Express 4.8.0 or later) to serve a fixed
file (index.html) only if the express.static middleware failed to find
a file. No rewriting needed, no issues with special characters in paths.

BTW there are no tests for this project so I would appreciate help in reviewing the change.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -10,7 +10,7 @@ var https = require("https");
 var httpProxy = require("http-proxy");
 var proxy = new httpProxy.createProxyServer();
 var serveIndex = require("serve-index");
-var historyApiFallback = require("connect-history-api-fallback");
+var historyApiFallback = require("express-history-api-fallback");
 
 function Server(compiler, options) {
 	// Default options
@@ -161,8 +161,12 @@ function Server(compiler, options) {
 
 		historyApiFallback: function() {
 			if (options.historyApiFallback) {
-				// Fall back to /index.html if nothing else matches.
-				app.use(historyApiFallback(typeof options.historyApiFallback === 'object' ? options.historyApiFallback : null));
+				var contentBase = options.contentBase || process.cwd();
+				var args = Array.isArray(options.historyApiFallback)
+					? options.historyApiFallback
+					: ['index.html', { root: contentBase }]
+				var fallback = historyApiFallback.apply(undefined, args);
+				app.use(fallback);
 			}
 		}.bind(this),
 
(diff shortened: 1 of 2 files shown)


radotzki:
I'm having the same problem with connect-history-api-fallback.
This pull request solved the problem for me.

ChrisJamesC:
Any update on this PR. It solves a problem I also have. Can I do something to help speed up the adoption of a fix? 
At test URL that failed before and now works is `http://localhost:3000/hello.world`

ChrisJamesC:
An alternative fix is to replace the `historyApiFallback=true` line with: 
```
        historyApiFallback: {
          rewrites: [
            {
              from: /^\/app\/.*$/,
              to: function() {
                return 'index.html';
              }
            }
          ]
        },
```
This will make any page with a dot in the /app path to be redirected to index.html (as it should be if it were not considered as a file)
See https://github.com/bripkens/connect-history-api-fallback#rewrites for more info

radotzki:
@ChrisJamesC  thank you, that works for me!

domenkozar:
@sokra I'd consider this as the `.` limitation is really annoying. I mentioned this in https://github.com/bripkens/connect-history-api-fallback/issues/25#issuecomment-238037144, but since this has been reported before it doesn't seem it will get fixed upstream.

bripkens:
This is now fixed in `connect-history-api-fallback` and the `.` character rule can now be disabled.

SpaceK33z:
So, is this PR still relevant after the fix in `connect-history-api-fallback`?

SpaceK33z:
Closing this since it is now possible to do this as an option. This PR is a breaking change, and I don't think it's worth it.

davidhariri:
It would be helpful to mention that option we should specify @SpaceK33z

tmcgann:
Use the `disableDotRule` found here: https://github.com/bripkens/connect-history-api-fallback#disabledotrule. For example:

``js
historyApiFallback: {
    disableDotRule: true,
}
```

I personally encountered issues with the rewrite rule--for some reason it changed the bundle that was used, which was really confusing. The symptom I observed was the hashes in my output files were different for the rewrite URLs; all the other URLs had the same hash in the output. Not sure why, may need to use the `context` argument as part of the rewrite (see https://github.com/bripkens/connect-history-api-fallback#rewrites). But the `disableDotRule` did the trick for me!

roger-russel:
I tried with "disableDotRule: true"  but disabled the other files as .js or .css, which made my vuejs project broke.

The solution that I found was based on this [link](https://stackoverflow.com/questions/38576356/attempting-to-route-a-url-with-a-dot-leads-to-404-with-webpack-dev-server).

was change this:
app.use(require('connect-history-api-fallback')());

To this:
app.use(require('connect-history-api-fallback')({
  rewrites: [
    {
      from: /^\/ip\/allow\/.*$/,
      to: '/'
    }
  ]
}))

The trick is create a rewrite to every route that uses dot on it.

PR #304, opened 2015-10-28
(feat) Support all https server parameters

radotzki:
In order to enable passing all parameters to the https server.

CLI:
I did not break the CLI - just add passphrase option.

API:
Currently there is no documentation about how to use the https feature from the API - so you might want to add.
The old syntax still works: 

```
devServer: {
	https: true,
	key: fs.readFileSync('...'),
	cert: fs.readFileSync('...')
}
```

But you can set https to an object like this:

```
devServer: {
	https: {
		key: fs.readFileSync('...'),
		cert: fs.readFileSync('...'),
		passphrase: 'secret'
	}
}
```

<b>https:</b> object to be passed to https.createServer()


urish:
+1

sokra:
merged

PR #313, opened 2015-11-10
Forcing edge for IE document mode

kylebyerly-hp:
This forces IE to use the latest document mode. It overrides 'Display intranet sites in Compatibility View' which some companies enforce with group policy.


sokra:
Thanks

PR #320, opened 2015-11-17
Allow development on single page apps

stutrek:
**Use Case**

I am developing a single page app that changes the URL. When I hit reload I need the server to serve the same HTML file each time, however I also have static assets, such as fonts and images, that need to be served from the file system. You can also prefix paths with devserver: to have them served from the dev server's build output rather than the file system.

**Functionality**

This PR allows you to use an object or an array for contentBase, so you can serve files from more than one place. If the target is a file, that file will be served for all requests to that path instead of using a static server.

**Example usage**

This will serve files from `./src` for `bower_components` and `assets`, falling back to the built version of `index.html` if the files don't exist or the path is not one of the two.
```
	devServer: {
		contentBase: {
			'/bower_components*': './src',
			'/assets*': './src',
			'*': 'devserver:/index.html'
		}
	},
```

**Notes**

If the object used for contentBase contains only the property `target`, it will use the deprecated proxy. I assume you'll want to remove this at the next major release.

I'm very open to doing this differently, but without something of the sort there's no good way to do this. The community has been running a separate static server and using redirects or proxies, which are a pain to work with.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -198,7 +198,13 @@ function Server(compiler, options) {
 							});
 						}
 						contentBase.forEach(function (contentItem) {
-							if (contentItem.target.substr(-5) === '.html') {
+							var isFile = false;
+							try {
+								isFile = fs.lstatSync(contentItem.target).isFile();
+							} catch (e) {
+								console.log('Warning: Serving files from '+contentItem.target+', which does not exist!');
+							}
+							if (isFile) {
 								app.get(contentItem.path, function (req, res) {
 									res.end(fs.readFileSync(contentItem.target));
 								});
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -171,17 +171,42 @@ function Server(compiler, options) {
 				var contentBase = options.contentBase || process.cwd();
 
 				if(typeof contentBase === "object") {
-					console.log('Using contentBase as a proxy is deprecated and will be removed in the next major version. Please use the proxy option instead.\n\nTo update remove the contentBase option from webpack.config.js and add this:');
-					console.log('proxy: {\n\t"*": <your current contentBase configuration>\n}');
-					// Proxy every request to contentBase.target
-					app.all("*", function(req, res) {
-						proxy.web(req, res, contentBase, function(err) {
-							var msg = "cannot proxy to " + contentBase.target + " (" + err.message + ")";
-							this.sockWrite(this.sockets, "proxy-error", [msg]);
-							res.statusCode = 502;
-							res.end();
+					var isOldTargetRequest = contentBase.target && Object.keys(contentBase).length === 1;
+					if (isOldTargetRequest) {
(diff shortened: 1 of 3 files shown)


sokra:
Sounds like a usefull functionality.

> the 'magic' of checking for .html files

Could you instead check for file or directory?

sakabako:
Good call, @sokra, done.

matthewdenobrega:
+1 I'm running a proxy and this change would simplify my setup.

lib/Server.js
```diff
@@ -171,17 +171,48 @@ function Server(compiler, options) {
...
+						contentBase.forEach(function (contentItem) {
+							var isFile = false;
+							try {
+								isFile = fs.lstatSync(contentItem.target).isFile();
+							} catch (e) {
+								console.log('Warning: Serving files from '+contentItem.target+', which does not exist!');
+							}
+							if (isFile) {
+								app.get(contentItem.path, function (req, res) {
+									res.end(fs.readFileSync(contentItem.target));
```
sokra: Is there any standard middleware (similar to the static middleware) which allows to serve a file? fs read is t simple to serve a file (header, mimetype, partial content, content-size, error handling).

sakabako:
@sokra I changed it to use the same library as express-static to serve the static file. It gets mime-type, content-length, etc. now.

sakabako:
@sokra I just added an additional bit of functionality which I can remove from this and make a new PR for if you don't like or don't like how I did it.

Now, if you prefix a contentBase with `devserver:`, instead of serving from the local file system, it will serve from the dev server's file system. This is useful when generating files with hashes and referencing them in a generated index.html because the html file only exists in the dev server's file system.

sokra:
> I just added an additional bit of functionality which I can remove from this and make a new PR for if you don't like or don't like how I did it.

I don't like it. Serving webpack assets is the job of the webpack-dev-middleware.

Just serving them from the webpack fs is not good, i. e. it doesn't block on compiling etc.

Could you please remove the last commit? The remaining PR looks good.

sakabako:
Rebased on master. I need both changes and I imagine you'd rather only have one versioning task, so I changed the devserver: thing to what I _think_ you meant.

```
app.get(contentItem.path, function (req, res, next) {
	req.url = contentItem.target.replace('devserver:', '');
	this.middleware(req, res, next);
}.bind(this));
```

This is clearly better because there's less duplication. Now it's being blocked on compiling twice, once when the request is made and again when this.middleware is called. Initially the very first thing I tried was mutating the request then calling next, but the middleware executes before this does, causing it to 404.

sakabako:
@sokra do you dislike the concept of serving stuff from the dev server or the way it was implemented here? Is there an alternative for serving a single HTML file that's generated by webpack?

SpaceK33z:
Closing this, because it doesn't align with the goals of this dev-server (that is, serving & live-reloading webpack modules).

PR #302, opened 2015-10-24
Replace socket.io with SockJS as discussed in #229

SpaceK33z:
The socket.io dependency causes node-gyp to be used. This results in a lot of issues with Windows and Linux installs. SockJS is much, much lighter and is node-only. This also results in a 5mb package size decrease :).

This is a relatively big change, so it would be best if some people test this in their project.

Fixes #229, #276, fixes #195, fixes #267, fixes #258 and fixes #242

PR #328, opened 2015-11-27
Fix websocket connection for non-inlined scripts

### LINKED ISSUE #326: 1.13.0 [WS] Disconnected, net::ERR_CONNECTION_REFUSED
For versions greater than 1.12.1 The WebSocket connection seems to be broken.

In the console I get the errors:
```
GET http://localhost/info?t=1448470960512 net::ERR_CONNECTION_REFUSED
[WDS] Disconnected!
```

My Webpack Config looks like this:
```javascript
{
//....
plugins: [
    new webpack.HotModuleReplacementPlugin(),
],
devServer: {
    contentBase: "./app/static",
    port: 3000,

    inline: true,
    historyApiFallback: true,

    colors: true,
    stats: 'normal',
},
}
```

As you can see the port is set to 3000 and I also access the page via `http://localhost:3000/` but the WebSocket connections seems to ignore the port and tries to connect to `http://localhost/info`

I am new to webpack so I am not quite sure if I am doing something wrong but as it was working a few days ago and also works if I downgrade to 1.12.1 I suppose it's an error with webpack-dev-server.

SpaceK33z:
With my previous PR, #302, I broke the websocket connection for scripts that are not using the `--inline` option.

Fixes #326.

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -2,10 +2,13 @@ var url = require('url');
 var SockJS = require("sockjs-client");
 var stripAnsi = require('strip-ansi');
 var scriptElements = document.getElementsByTagName("script");
+var scriptHost = scriptElements[scriptElements.length-1].getAttribute("src").replace(/\/[^\/]+$/, "");
 
+// If this bundle is inlined, use the resource query to get the correct url.
+// Else, get the url from the <script> this file was called with.
 var urlParts = url.parse(typeof __resourceQuery === "string" && __resourceQuery ?
 	__resourceQuery.substr(1) :
-	scriptElements[scriptElements.length-1].getAttribute("src").replace(/\/[^\/]+$/, "")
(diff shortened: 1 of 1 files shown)


nicksrandall:
I can verify that this fixes #326. Please merge :smile:

sokra:
Thanks

nicksrandall:
Can we create a new release with this fix because the current release is broken?

laszlokorte:
:+1: for a new release

buunguyen:
Appreciate a release, the current version has been broken for 23 days!

nicksrandall:
@sokra thoughts on a new patch release?

SpaceK33z:
@sokra,  could you please release a new version with this fix? The current version is completely broken if you don't use `--inline`. Whoops.

PR #329, opened 2015-11-28
Open default browser if --open option is given

### LINKED ISSUE #311: Automatically open browser
Please add a flag to `webpack-dev-server` that will automatically open the browser to the configured url.

SpaceK33z:
This will open your default browser with the app url if you add the `--open` option to the CLI.

Fixes #311

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -2,6 +2,7 @@
 
 var path = require("path");
 var url = require("url");
+var open = require("open");
 var fs = require("fs");
 
 // Local version replaces global one
@@ -44,7 +45,10 @@ var optimist = require("optimist")
 
 	.boolean("compress").describe("compress", "enable gzip compression")
 
+	.boolean("open").describe("open", "Open default browser")
+
 	.describe("port", "The port").default("port", 8080)
+
 
 	.describe("host", "The hostname/ip address the server will bind to").default("host", "localhost");
 
@@ -137,6 +141,9 @@ if(argv["history-api-fallback"])
 if(argv["compress"])
 	options.compress = true;
 
+if(argv["open"])
+	options.open = true;
+
 var protocol = options.https ? "https" : "http";
 
 if(options.inline) {
@@ -156,11 +163,13 @@ if(options.inline) {
 }
 
 new Server(webpack(wpOpt), options).listen(options.port, options.host, function(err) {
+	var uri = protocol + "://" + options.host + ":" + options.port + "/";
+
 	if(err) throw err;
 	if(options.inline)
-		console.log(protocol + "://" + options.host + ":" + options.port + "/");
+		console.log(uri);
 	else
-		console.log(protocol + "://" + options.host + ":" + options.port + "/webpack-dev-server/");
+		console.log(uri + "webpack-dev-server/");
 	console.log("webpack result is served from " + options.publicPath);
 	if(typeof options.contentBase === "object")
 		console.log("requests are proxied to " + options.contentBase.target);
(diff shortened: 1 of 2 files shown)


claycephas:
Not sure if this is environmental for me, but I get this error:

    Error: ENOENT: no such file or directory, open '.....\node_modules\webpack-dev-server\client\live.bundle.js'

sokra:
@claycephas `npm run prepublish` if you want to run it from git checkout.

@SpaceK33z it opens the wrong uri when `inline: false`.

SpaceK33z:
@sokra, how did you test this? I'm using `cd example && node ../bin/webpack-dev-server.js --open`, and it opens `http://localhost:8080/` for me, which is correct.

Even when I add this it seems to work fine:

```js
devServer: {
    inline: false
}
```

sokra:
It should open the printed url.

`http://localhost:8080/webpack-dev-server/` in non-inline mode.

SpaceK33z:
Okay, I intentionally did it without the `/webpack-dev-server` part, but I see now that's quite confusing. It's fixed now :).

Thank you for merging all my PR's so far and the constructive feedback @sokra :).

claycephas:
ok, I updated my package.json file to have this entry:

    "webpack-dev-server": "CodeYellowBV/webpack-dev-server#98fd2f45b4a81"

then ran

    npm install
    cd node_modules/webpack-dev-server
    npm run prepublish

now this works as expected

    ./node_modules/.bin/webpack-dev-server --open

Please merge!

sokra:
Thanks...

Could you also update documentation page to include this CLI option?

SpaceK33z:
@sokra, I have edited the wiki to include the `--open` option.

sandwichsudo:
Is it possible to use this option when opening webpackDevServer from a node script?
eg

              const server = new WebpackDevServer(compiler, {
			hot : true,
			harmony : true,
			host : '0.0.0.0'
		});

SpaceK33z:
@sandwichsudo, nope, not at the moment.

Gopikrishna19:
This doesn't work for me.. am I missing something?
Here's my package:
```js
"scripts": {
  "start": "babel-node server/index.js | webpack-dev-server --open"
},
"devDependencies": {
  "webpack-dev-server": "^1.14.1"
}
```
And webpack.config:
```js
devServer: {
    contentBase: './dist',
    hot: true,
    noInfo: true
}
```

SpaceK33z:
@Gopikrishna19, apparently it was not part of the 1.14.1 release. In the next 2.0 version it should work.

GuyMograbi:
webpack is awesome!

just reached here. half a year later. is it possible to have a patch for this as 2.x is not out yet? I will gladly help if you guide me.

SpaceK33z:
@GuyMograbi, I happen to be working on a v1 release.

calledT:
It's better I could decide which url to open with the `--open` option is string.Because I want to add querystring to the open url or the default name of my html is not index.

LucienLee:
Before webpack 2.0 released, I think using `[opn](https://www.npmjs.com/package/opn)` module is the simplest to open any url in browser.

controversial:
Is it possible to specify this in `webpack.config.js`?

alflennik:
I'd love to see this in the config.

```
devServer: {
  open: true
}
```

jonkoops:
@alflennik Maybe open a new issue with that as a feature request?

pavel06081991:
@SpaceK33z, how to pass open option when webpack-dev-server is used as module? (https://github.com/webpack/docs/wiki/webpack-dev-server#api)

lucaritossa:
Is there a way to specify a different url path or the standard url `http://host:port` will be always used?
I need to open a url similar to `http://host:port/my-custom-path/` (angular project inside an ASP.NET MVC existing project)

TrySound:
@pavel06081991 No way. It's only CLI feature.

PR #330, opened 2015-11-28
Show horizontal scrollbar in errors text

### LINKED ISSUE #322: overflow error text rendered light-on-white
When using webpack-dev-server iframe view, the error text is rendered so that it overflows. The text is light gray on a red background, but because the html body is white it becomes hard to read when you scroll to the right.

Making the body background-color something like black or the same color red would solve this?

Ideally the pre element rendering the text would wrap the lines to stay inside the width of the viewport, but I don't know if that is possibe.

screenshot 2015-11-19 11 41 06

SpaceK33z:
If the error text is too long, it now shows a scrollbar instead of overflowing out of the errors element, making the text unreadable (see issue #322).

localhost_8080_webpack-dev-server_bundle_and_1__npm_run_prepublish____npm_run_inline__node_

Fixes #322

### DIFF
--- a/client/style.css
+++ b/client/style.css
@@ -50,6 +50,7 @@ html {
 	line-height: 1.4;
 	color: #eff1f5;
 	background: #bf616a;
+	overflow: auto;
 }
 
 #okness {


sokra:
Thanks

PR #343, opened 2015-12-11
Add '--manual-reload-on-fail' option to CLI

install:
- Added the flag '--manual-reload-on-fail' that will stop the automatic
  reloading of a page when HMR fails.
- Takes precedence over the '--hot' flag.
- Useful for debugging when using the CLI and HMR fails.


SpaceK33z:
PR #510 does about the same, and has no merge conflicts, so closing this for now.

PR #357, opened 2015-12-28
Closing listeningApp when server closes

### LINKED ISSUE #348: server.close() desn't work since 1.13
I see you have changed the close method [#L310](https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L310). This worked fine with 1.12 but is broken in 1.13 and 1.14. Port doesn't get released.

NicholasBoll:
I ran into the same issue as in #348. When multiple sockets were added in 2dd3058746ea8ce9247726df616c8fd72c09c461, the `listeningApp` was no longer being closed by closing the server. The `close` method now explicitly closes the listening app when the server closes, thus closing the port and freeing the node process to exit gracefully (assuming it was shut down by something other than `Ctrl+C`).

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -313,6 +313,7 @@ Server.prototype.close = function() {
 	});
 	this.sockets = [];
 	this.middleware.close();
+	this.listeningApp.close();
 }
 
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #359, opened 2016-01-03
Use http-proxy-middleware instead of http-proxy.

jamsesso:
I've found that the configuration of the node-http-proxy (NHP) library doesn't allow for some basic use cases (including my own). For example, it doesn't seem possible (through configuration only) to proxy a request from `host1.com/api/user/1` to `host2.com/user/1`.

This PR changes the implementation to use the [http-proxy-middleware (HPM)](https://www.npmjs.com/package/http-proxy-middleware) library which is built on top of the original NHP, so all current configurations should work. The HPM library has a lot more configuration options available.

I have ported over `proxyOptions.bypass(...)`, but without a `proxy` object, there is no need for `proxyOptions.configure(...)` anymore.

Comments in the code can be removed on request, I wanted to make my intent clear. Cheers

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -7,8 +7,7 @@ var sockjs = require("sockjs");
 var StreamCache = require("stream-cache");
 var http = require("http");
 var https = require("https");
-var httpProxy = require("http-proxy");
-var proxy = new httpProxy.createProxyServer({secure: false});
+var httpProxyMiddleware = require("http-proxy-middleware");
 var serveIndex = require("serve-index");
 var historyApiFallback = require("connect-history-api-fallback");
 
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

mjrussell:
:+1: for this PR. It adds support for native web sockets which is currently not working in the latest release. Would resolve #283. Also seems like the HTPM has much more mature documentation.

One issue I did have though is that the syntax for describing the url to match is a little different.

Before this PR:
```js
  ...
  proxy: {
    '/api/*': {
       ...
    }
  }
  ...
```

After:
```js
  ...
  proxy: {
    '/api': {
       ...
    }
  }
  ...
```

mjrussell:
@sokra Should there be some documentation regarding the proxy path tweak?

mjrussell:
Also I noticed a crash when the server being proxied to shuts down, opened an issue with the HTPM since the trace looks like its on them

jamsesso:
Thanks for the merge. If there is a documentation repository I can make the changes for the proxy path tweak. I tested the `/api/*` path and I am pretty sure it worked. Can you verify that it is not working before the documentation is updated?

mjrussell:
I did run it on my setup and it was definitely not obeying the `/api/*`. The docs on the npm registry seem to suggest you should just use `/api`. 

I don't have a great open source example for a demo on hand. @sokra added a rewrite in 33ebd1f. If you can prove that this is unnecessary feel free to and then that check can be removed.

unindented:
@sokra What's the schedule for `webpack-dev-server` releases? I'd love to get this on npm.

(I can't install it straight from git because npm doesn't seem to run the `prepublish` tasks.)

mik01aj:
+1, I wait for the npm release.

geddski:
@mik01aj the 1.14.1 release has it.

sinedied:
Any guidance on how to make this work behind a corporate http proxy?
In my old gulp-based setup I could setup a custom agent to do this (via http-proxy-middleware `agent` option), how can I do the same here?

PR #374, opened 2016-01-18
Extend contentBase to support array with multiple sources.

BerndWessels:
Hello

This is a small but important PR that will allow us to have multiple sources as content base.

It opens a raft of development build features and solves masses of messy workarounds to achieve the same thing.

One typical example on how it would be used in a `webpack.config` is:

    var WHITELABEL = 'visa';
    config.devServer = {
      contentBase: ['./public/common', './public/' + WHITELABEL]
    }

In this example we are now able to separate `common` and `whitelabel` specific assets, externals and  sources cleanly in different folders - rather than having copies of all common files in each whitelabel folder.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -197,7 +197,11 @@ function Server(compiler, options) {
 			if(options.contentBase !== false) {
 				var contentBase = options.contentBase || process.cwd();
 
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #376, opened 2016-01-19
Bump http proxy middleware for websocket crash

mjrussell:
See chimurai/http-proxy-middleware#44


mjrussell:
I know this would auto upgrade due the the `~` but in case somewhere were to do some weird shrinkwrap I think its good to be on the safe side...

sokra:
Thanks

PR #380, opened 2016-01-21
Prevent application from covering up webpack errors

BHSPitMonkey:
Currently, my web application is getting rendered on top of the build errors/warnings, preventing me from seeing them. This CSS change seems to fix it.

### DIFF
--- a/client/style.css
+++ b/client/style.css
@@ -39,6 +39,7 @@ html {
 	height: -o-calc(100% - 30px);
 	height: calc(100% - 30px);
 	border: 0;
+	z-index: -1;
 }
 
 #errors {


benzen:
Which is the context in which is affect youy, the inline mode or the iframe mode?

BHSPitMonkey:
@benzen What specifically do you mean? The problem in my case manifested by the <iframe> containing my application being drawn over the webpack error/stacktrace printout; The z-index fix in this patch solved it for me.

benzen:
I was wondering if you used this: https://webpack.github.io/docs/webpack-dev-server.html#iframe-mode

BHSPitMonkey:
Ah, I wasn't aware it went by that name. But yes, that's where this bug comes in (the fix I made only affects CSS for the `<iframe>` element).

sokra:
Thanks

SpaceK33z:
Sorry, had to reverse your PR. It caused issues with registering clicks in the iframe itself (#559, and probably #564). If you can find a way that doesn't affect the iframe, a new PR is welcome!

BHSPitMonkey:
We could make it so the webpack messages and the application aren't stacked/positioned in the same place, and just lay them out sequentially?

PR #383, opened 2016-01-22
Fix webpack peerDependency

Globegitter:
Just tried to install the new beta version and it broke because of a version mis-match

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "author": "Tobias Koppers @sokra",
   "description": "Serves a webpack app. Updates the browser on changes.",
   "peerDependencies": {
-    "webpack": ">=2.0.3-beta <3"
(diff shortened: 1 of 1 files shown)


acusti:
Would be nice if this was fixed. Can’t install webpack-dev-server 2.x. I’m not clear why `">=2.0.3-beta <3"` fails on `2.0.7-beta`, but it does. And either way, it seem silly that the lower bound of the currently listed peer dependency range (2.0.3-beta) doesn’t exist as a published version of webpack.

Hatlen:
Webpack has released some new versions so I think this https://github.com/webpack/webpack-dev-server/pull/411 pull request might replace this.

Also found this site http://semver.npmjs.com/ that does an excellent job in explaining how npm version strings works.

PR #384, opened 2016-01-24
fix hostname resolving on newConnection

svsool:
correct behaviour when:
1) window.location.hostname is empty or undefined
2) window.location.protocol not http or https

I'll get bug with ```ionic emulate ios``` because default protocol in WebView is :file and hostname is empty string, hence SockJs url resolve to "http:/sockjs-node" that caused a bug ```SyntaxError: The URL's scheme must be either 'http:' or 'https:'. 'file:' is not allowed.```

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -59,10 +59,18 @@ var onSocketMsg = {
 };
 
 var newConnection = function() {
+	var hostname = urlParts.hostname;
+
+	if (urlParts.hostname === '0.0.0.0') {
+		if (window.location.hostname && !!~window.location.protocol.indexOf('http')) {
+			hostname = window.location.hostname;
+		}
+	}
+
 	sock = new SockJS(url.format({
 		protocol: urlParts.protocol,
 		auth: urlParts.auth,
-		hostname: (urlParts.hostname === '0.0.0.0') ? window.location.hostname : urlParts.hostname,
+		hostname: hostname,
 		port: urlParts.port,
 		pathname: urlParts.path == null || urlParts.path === '/' ? "/sockjs-node" : urlParts.path
 	}));


dperetti:
The issue affects Electron as well !

client/index.js
```diff
@@ -59,10 +59,18 @@ var onSocketMsg = {
 };
 
 var newConnection = function() {
+	var hostname = urlParts.hostname;
+
+	if (urlParts.hostname === '0.0.0.0') {
+		if (window.location.hostname && !!~window.location.protocol.indexOf('http')) {
```
SpaceK33z: Why the check for `http` here (not saying it is wrong per se)? If it has a valid reason, could you add a comment above? And also please give an example for when the hostname does not exist in the comments.

We get so many PR's around the detection of this url, so comments will help other people here.

svsool: when protocol === 'file', for example in electron, hostname = undefined and hence SockJs url resolves to "http:/sockjs-node" that caused a bug SyntaxError: The URL's scheme must be either 'http:' or 'https:'. 'file:' is not allowed

SpaceK33z: Could you add that to the code as a comment though? This is not immediately clear, and since there are no tests yet, it's easy for this to get accidentally removed.

SpaceK33z:
Other then the comment above and the merge conflict, this looks good. Will merge if you finish this :).

SpaceK33z:
Pinging @svsool, are you still interested in finishing this? This would fix a couple of other problems. I could takeover if you don't have the time.

svsool:
Fix it and create another fresh pull https://github.com/webpack/webpack-dev-server/pull/594

client/index.js
adarrra: I have a problem with this part as I need host '0.0.0.0' but socket port from urlParts.port, not from self.location.port. Am I right that I can't setup socket port explicitly?

unflores:
This change actually breaks my configuration setup. We are using docker and this needs to connect on '0.0.0.0:3035' but due to the line change I end up with 'dev.server.com'.

unflores:
This change actually breaks my configuration. We have a docker setup and run this from the docker container. The setup is '0.0.0.0:3035' b/c we need to broadcast this for the serving box. However, the change at line 65 effectifely resets the host to 'dev.server.com' and the port to '', which no longer works. :/

Issue #327, opened 2015-11-27
Expose a way to pass params for `express.static`?


dickeylth:
In <https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L203>,

```js
// route content request
app.get("*", express.static(contentBase), serveIndex(contentBase));
```

Now I want to pass param `{index: false}` to `express.static` so that it looks like `express.static(contentBase, {index: false})` to disable directory indexing, because it makes me feel inconvenient to access other html files under the directory.

Thank you!

SpaceK33z:
Fixed with #385, which will be part of the next 1.x release.

PR #385, opened 2016-01-26
add `staticOptions` for expose express.static options pass in

dickeylth:
Issue #327 (Expose a way to pass params for `express.static`)


sokra:
Thanks

sokra:
Could you add it to the documentation too?

dickeylth:
@sokra OK, I'll add it

PR #387, opened 2016-01-27
Option registerAdditionalRoutes added,

fru:
Could this be added so that I can add more routes to the dev server?


SpaceK33z:
PR #537 fixed this in a more flexible way.

Issue #117, opened 2015-02-25
TypeError due to assuming all <script> elements have a src attribute


steveme:
Hi,

Am seeing a lot of : 
```
Uncaught TypeError: Cannot read property 'replace' of null webpack-dev-server.js:1
```
Which is from here:
https://github.com/webpack/webpack-dev-server/blob/master/client/index.js#L5

As i've got script elements in my markup that don't have a src attribute.

We're using fluxible, so our HTML has inline state in a <script/> tag like: https://github.com/yahoo/flux-examples/blob/master/fluxible-router/components/Html.jsx

Thanks.

Steve

jhnns:
Yep, that's an error... that's probably not a good fallback to get the base path. @sokra when is `__resourceQuery` not set?

sokra:
It only checks the current script tag, which should be the script tag for the webpack-dev-server client. `__resourcePath` is used when putting the `webpack-dev-server/client?...` in the entry. I don't think this is wrong. I think the error is somewhere else:

* `<script>` tag to the WDS client is included with `async`, `defer` or inlined
* `webpack-dev-server/client` is added to the entry point but no host argument is included.

voronianski:
@sokra @steveme @jhnns I have very similar issue:

```bash
webpack-dev-server --hot --inline --progress --colors

70% 1/1 build modules
TypeError: Object /Users/me/github/app-webpack/node_modules/webpack-dev-server/client/index.js?http://localhost:8080,webpack/hot/dev-server,webpack/hot/dev-server has no method 'replace'
at Tapable.<anonymous> (/Users/me/github/app-webpack/node_modules/webpack/lib/NormalModuleFactory.js:42:26)
```

webpack config:

```javascript
'use strict';

var webpack = require('webpack');

module.exports = {
    entry: [
        'webpack/hot/dev-server',
        './src/app.js'
    ],
    output: {
        path: __dirname + '/public/build/',
        filename: 'bundle.js',
        publicPath: '/build/'
    },
    resolve: {
        extensions: ['', '.js', '.jsx']
    },
    module: {
        loaders: [
            {
                test: /\.(js|jsx)$/,
                exclude: /node_modules/,
                loaders: ['babel-loader?experimental']
            }
        ]
    },
    plugins: [
        new webpack.HotModuleReplacementPlugin(),
        new webpack.NoErrorsPlugin()
    ]
};
```

[...]

PR #394, opened 2016-02-01
fix undefined src attribute in scriptHost

wrasdf:
Hi there,

This is a fix for V1.14.1.

Thanks,
Kerry


jhnns:
Hi, thanks for your contribution. We've merged #447 which was almost the same change.

However, now I think we should rather fix the underlying problem as discussed in #117 before publishing this.

PR #396, opened 2016-02-04
Fix the case when the last script has no src

Rudolph-Miller:


jaylinski:
duplicates #394

SpaceK33z:
As said above, a dupe of #394, and there are more reliable ways to fix this.

Issue #402, opened 2016-02-11
Prepublish script fails on webpack 2 because of removed --colors option


MatthiasWinkelmann:
The prepublish script currently uses the --colors option which was removed from webpack. This results in the error Error: ENOENT, open '[...]/node_modules/webpack-dev-server/client/live.bundle.js'.

sokra:
fixed

PR #404, opened 2016-02-11
fixes webpack-dev-server#403 and webpack-dev-server#402 and updates d…

### LINKED ISSUE #403: Prepublish script depends on jquery which is missing from dependencies
As the title says, the prepublish script currently fails because it requires jquery.js. The error message is 

```
ERROR in ./client/live.js
Module not found: Error: Can't resolve 'jquery' in '[...]webpack-dev-server/client'
 @ ./client/live.js 1:8-25
```

MatthiasWinkelmann:
This fixes #403 and #402 and updates the dependencies to webpack 2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,6 @@
   "version": "2.0.0-beta",
   "author": "Tobias Koppers @sokra",
   "description": "Serves a webpack app. Updates the browser on changes.",
-  "peerDependencies": {
-    "webpack": ">=2.0.3-beta <3"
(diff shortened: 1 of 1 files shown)


MatthiasWinkelmann:
The travis built is failing on the webpack version specifier, but I haven't found a way to successfully require the 2.0 betas in package.json because webpack apparently doesn't follow semver.

sokra:
partly merged

PR #410, opened 2016-02-22
Permit webpack@2 betas

pyrotechnick:
Ranges don't work on pre-releases.

Heads-up: `2.0.3-beta` never existed (or was unpublished).

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "author": "Tobias Koppers @sokra",
   "description": "Serves a webpack app. Updates the browser on changes.",
   "peerDependencies": {
-    "webpack": ">=2.0.3-beta <3"
+    "webpack": "^2 || 2.0.4-beta || 2.0.5-beta || 2.0.6-beta || 2.0.7-beta || 2.1.0-beta.0"
   },
   "dependencies": {
     "compression": "^1.5.2",


package.json
```diff
@@ -4,7 +4,7 @@
   "author": "Tobias Koppers @sokra",
   "description": "Serves a webpack app. Updates the browser on changes.",
   "peerDependencies": {
-    "webpack": ">=2.0.3-beta <3"
+    "webpack": "^2 || 2.0.4-beta || 2.0.5-beta || 2.0.6-beta || 2.0.7-beta || 2.1.0-beta.0"
   },
```
Hatlen: Same thing here as my comment on your pull request for fixing babel-loader's package.json

(changing `2.1.0-beta.0` to `^2.1.0-beta.0` would make it also match 2.1.0-beta.1 and 2.1.0-beta.2, which also has been released now)

Hatlen:
:+1:

package.json
pyrotechnick: Same thing here as my reply to your comment on my pull request.

`^2.1.0-beta.0` also matches `2.1.0` which nobody has tested since it doesn't exist yet.

Hatlen: :+1:

pyrotechnick: Superseded by #411.

Thank you @Hatlen ^w^

PR #415, opened 2016-02-26
Output build errors to console

denis-sokolov:
Before this, if postcss throws a syntax error, because user's css is invalid,
the dev server silently sits there, and the browser is stuck there waiting.
No feedback to the developer is available.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -29,6 +29,7 @@ function Server(compiler, options) {
 	}.bind(this);
 	compiler.plugin("compile", invalidPlugin);
 	compiler.plugin("invalid", invalidPlugin);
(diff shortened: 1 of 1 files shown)


SpaceK33z:
Just tested this with the webpack config in `example/`, by replacing the `less` loader with `postcss`. Then I made a syntax error in `style.less`, and I got this nice message in the console, and in my console in the browser:

1_____bin_webpack-dev-server_js__node_

So it already works?

SpaceK33z:
Closing this, because of the reason above. Feel free to explain it, we can always reopen this.

denis-sokolov:
A lot of time has passed since February. Perhaps it's been fixed.

Issue #417, opened 2016-02-29
webpack-dev-server not accepting promise


crohde7707:
Noticed that when using webpack, you are able to return a Promise instead of an object as a valid parameter. I would assume that webpack-dev-server would be the same case, but it looks like it was an oversight. The code in question is as follows

webpack-dev-server/bin/webpack-dev-server.js
```
var wpOpt = require("webpack/bin/convert-argv")(optimist, argv, { outputFilename: "/bundle.js" });
var firstWpOpt = Array.isArray(wpOpt) ? wpOpt[0] : wpOpt;
var options = wpOpt.devServer || firstWpOpt.devServer || {};
...
```

webpack/bin/webpack.js
```
var options = require("./convert-argv")(yargs, argv);
...
function processOptions(options) {
	// process Promise
	if(typeof options.then === "function") {
		options.then(processOptions).catch(function(err) {
			console.error(err.stack || err);
			process.exit(); // eslint-disable-line
		});
		return;
	}

	var firstOptions = Array.isArray(options) ? options[0] : options;

	var outputOptions = Object.create(options.stats || firstOptions.stats || {});
	if(typeof outputOptions.context === "undefined")
		outputOptions.context = firstOptions.context;
...
}
processOptions(options);
```

Tried looking through the issues log for something like this, but couldn't find it, and it doesn't look like its fixed in the latest version

PR #419, opened 2016-03-03
Update webpack-dev-server to accept a promise

crohde7707:
This pull request is associated with issue #417.

Webpack allows for a Promise instead of a config object, and this logic was missing from webpack-dev-server. This pull adds that same logic


crohde7707:
@sokra If i branch off the latest 1.x version, and make the change there, would it be possible to get a patch version released? Not sure of the timeline when 2.0 will be released, and would like make this fix available.

sokra:
Thanks

shanebdavis:
@crohde7707 & @sokra:
I'd also like to see this backported to the current 1.x release. What's the status? Can I help?

PR #423, opened 2016-03-08
Allow passing false to contentBase to disable static file serving

davidblurton:
The implementation support disabling contentBase by passing false, but the option isn't passed through correctly.


bin/webpack-dev-server.js
```diff
@@ -168,7 +168,7 @@ if(argv["content-base"]) {
 	options.contentBase = {
 		target: argv["content-base-target"]
 	};
-} else if(!options.contentBase) {
+} else if(options.contentBase !== false) {
```
sokra: this always overrides the current `options.contentBase`

PR #430, opened 2016-03-16
Guess server's port from current page.

pcorpet:
Allow to use port from current page, the same way we allow to use hostname.

This is useful in setups where the webpack server is for instance in a docker container and doesn't know statically on which hostname and port it will be accessed from. It's already possible to use "0.0.0.0" as an address to use current page's hostname. I'm adding the same trick for the port.

The config should then look like this
"webpack-dev-server/client?http://0.0.0.0:0"

---
This change is [Reviewable](https://reviewable.io/reviews/webpack/webpack-dev-server/430)


sokra:
Thanks

Issue #340, opened 2015-12-10
Bundled certificate expired today


starjumper30:
The bundled certificate used in --https mode expired today.

pfrankov:
+1

dickeylth:
+1

PR #436, opened 2016-03-24
Update self signed ssl certificates

Smotko:
This should resolve issue https://github.com/webpack/webpack-dev-server/issues/340.

The new certificate will expire in 5 years (Thursday 24 March 2021).

I believe most of the data is the same, but I have added localhost as a Common Name so that Chrome will accept it.

screen shot 2016-03-25 at 00 03 37


sokra:
Thanks

phorozan:
is this released?

wbobeirne:
I'm still getting the broken red lock of doom in Chrome Canary (Version 53.0.2766.0 canary) it's complaining that the cert hasn't been verified by a third party. Is this something that can be resolved?

screen shot 2016-06-13 at 6 33 50 pm

wwselleck:
I'm getting a red lock as well, but with a different error (Chrome 51.0.2704.84)

screen shot 2016-06-14 at 12 13 33 pm

wwselleck:
I just upgraded to these versions

```
"webpack": "^2.1.0-beta.0",
"webpack-dev-server": "^v2.1.0-beta.0"
```
and am now getting the same error as @wbobeirne 

screen shot 2016-06-14 at 12 24 38 pm

wbobeirne:
@wwselleck it looks like you're using the old certificate, likely meaning you're on a version of webpack-dev-server that hasn't been updated since March 24th. The current stable version on npm (1.14.1) has the old cert.

wwselleck:
@wbobeirne Refresh, I updated and got the same error as you 😄

deevus:
Is there a definitive solution for this? What is the earliest version that has the new certificate?

wwselleck:
I was able to get past the "This certificate has not been verified..." error by just proceeding anyway on Chrome, and adding a security exception on Firefox, once I got to the red lock screen. After that calls to the webpack dev server were going through.

deevus:
I have another project that enforces HSTS so proceeding only works as long as I don't visit the other site. That doesn't work well when I'm trying to implement SSO D:

ashier:
Got this working via pointing the key, cert and ca to the correct files. 

`
var fs = require('fs');
var path = require('path');
var webpack = require('webpack')
var WebpackDevServer = require('webpack-dev-server')
var config = require('./webpack.local.config')

var ops = {
    cert: '/path to webpack ssl/server.crt',
    ca: '/path to webpack ssl/ca.crt',
    key: '/path to webpack ssl/server.key'
};

new WebpackDevServer(webpack(config), {
  https: true,
  cert: fs.readFileSync(ops.cert),
  ca: fs.readFileSync(ops.cert),
  key: fs.readFileSync(ops.key),
  requestCert: true,
  rejectUnauthorized: true,
  publicPath: config.output.publicPath,
  hot: true,
  inline: true,
  historyApiFallback: true
}).listen(3000, '0.0.0.0', function(err, result) {
  if (err) {
    console.log(err)
  }

  console.log('Listening at https://localhost:3000')
})
`

Then adding the certificate to my Keychain Access and trusting the certificate.

Hope this helps.

- ashier

PR #439, opened 2016-03-29
Add hot-only option to allow setting `only-dev-server` from cli or config

SimenB:
I currently have to mess around with `entry` manually as I want `only-dev-server`. This allows me to just add the options, or cli flag.

current:
```js
entry: [
  'webpack-dev-server/client?http://0.0.0.0:3030',
  'webpack/hot/only-dev-server',
  origConfig.entry
],
devServer: {
  host: '0.0.0.0',
  port: 3030,
  hot: true,
},
```

wanted: 
```js
devServer: {
  host: '0.0.0.0',
  port: 3030,
  hotOnly: true,
  inline: true
},
```

Ideally I'd PR this into `webpack-dev-server@1` (as IE8 support is dropped, we'll probably never upgrade), but there doesn't seem like there's a branch for it.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -23,6 +23,7 @@ var yargs = require("yargs")
 
 require("webpack/bin/config-yargs")(yargs);
 
+var ADVANCED_GROUP = "Advanced options:";
 var DISPLAY_GROUP = "Stats options:";
 var SSL_GROUP = "SSL options:";
 var CONNECTION_GROUP = "Connection options:";
@@ -37,6 +38,11 @@ yargs.options({
 		type: "boolean",
 		default: true,
 		describe: "Inline mode"
+	},
+	"hot-only": {
+		type: "boolean",
+		describe: "Do not refresh page if HMR fails",
+		group: ADVANCED_GROUP
 	},
 	"stdin": {
 		type: "boolean",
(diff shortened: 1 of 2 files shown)


SimenB:
@sokra ping

bin/webpack-dev-server.js
```diff
@@ -38,6 +39,11 @@ yargs.options({
 		default: true,
 		describe: "Inline mode"
 	},
+	"hotOnly": {
```
sokra: `"hotOnly"` => `"hot-only"` for the CLI argument

SimenB: `yargs` normalizes it, so both works. I can change it though :smile:

SimenB:
@sokra Updated

SimenB:
@sokra ping.

Also, thoughts on a v1 branch?

SimenB:
@sokra ping.

sokra:
Thanks

harmony7:
Hi, I have a comment on this. It seems that with `--hot-only`, the `HotModuleReplacementPlugin` plugin does not get loaded as it would with `--hot`. At the current time I can work around this problem by supplying both `--hot` and `--hot-only` to get the behavior I want.

Digging deeper it seems that this processing is done in `webpack/bin/convert-argv`.
```
		ifBooleanArg("hot", function() {
			ensureArray(options, "plugins");
			var HotModuleReplacementPlugin = require("../lib/HotModuleReplacementPlugin");
			options.plugins.push(new HotModuleReplacementPlugin());
		});
```

How would we go about getting this fixed? Is it something that needs fixing in webpack? (Why is it webpack's job to handle ARGV for webpack-dev-server, anyway?)

SimenB:
That's by design. It's for the entrypoint `webpack/hot/only-dev-server`. You still need to activate hmr with `hot`

harmony7:
Thanks for your response.

Okay.  I saw this code change you made in Server.js:

```
	this.hot = options.hot || options.hotOnly;
```

So I thought you were trying to make it so you had to specify only one or the other to activate `hot`

Issue #367, opened 2016-01-11
Getting errors on client build. 


ddaza:
Updated to webpack 
```
├── webpack@1.12.11
├── webpack-core@0.6.8
├── webpack-dev-middleware@1.4.0
├── webpack-dev-server@1.14.1

```

I am seeing an error: 

```
Uncaught TypeError: Cannot read property 'replace' of null
```

in line 5 of `webpack-dev-server/client/index.js` : 

```javascript
var scriptHost = scriptElements[scriptElements.length-1].getAttribute("src").replace(/\/[^\/]+$/, "");
```

globexdesigns:
+1

zag2art:
+1

cloudsan:
+1, rollback to @1.14.0 is fine for me,

csaintc:
:+1:

le0pard:
The same error after update to 1.4.1

liujason:
+1 rolledback to @1.14.0 fixed the problem.

jaylinski:
also getting this error on client when adding an inline-script-tag

Stas404:
+1 hi all.

kwtangerine:
+1. thanks for the tip, rolling back to 1.14.0 works for me too

christophehurpeau:
:+1:

kenjim83:
+1 me to.  1.14.0 works.

Qrysto:
+1, same issue as this and #381. Why is this unresolved for so long?

goodseller:
+1

jaylinski:
@sokra can you have a quick look at this? there are already two issues and three PRs open for this problem.

EwanValentine:
+1

jakubzitny:
+1

denich:
+1

foxtree:
+1

incleaf:
+1

numical:
+1

saponifi3d:
+1

EvNaverniouk:
For those who want to make the leap to Webpack 2.0 beta, this issue appears to be fixed in webpack-dev-server `2.1.0-beta.0`

SpaceK33z:
Dupe of #117.

PR #447, opened 2016-04-06
fix bug when src attribute not defined on last script element

svsool:


ptz0n:
LGTM :+1:

jaylinski:
also see #367

SimenB:
#394

PR #450, opened 2016-04-08
Update to v2.0.0-beta.1

colinrymer:
The `package.json` has been updated to work with webpack 2.1.0-beta, but the beta published to NPM doesn't support that version of webpack.

This PR introduces a newer version of the beta that can be published to NPM allowing developers to more easily use this in conjunction with the latest webpack beta.


sokra:
done

PR #477, opened 2016-05-05
set protocol to https if window using same

tsongas:
Fixes webpack-dev-server hot reloading failure when using Cloud9 IDE due to mixed protocol error.


tsongas:
Oops I just realized this is a duplicate of PR #470.

Issue #446, opened 2016-04-06
Don't reload on proxy errors


nhjk:
I proxy /api/* requests to a backend server. There was an error proxying the request, but webpack-dev-server reloaded every time it got this proxy error, making it very difficult to even see as devtools would refresh before I could see the error. I had no idea if the error was in my frontend code, webpack, webpack-dev-server etc.. until I was finally able to screenshot the console before it refreshed.
![webpack error](https://cloud.githubusercontent.com/assets/3460415/14306003/766c5392-fb90-11e5-91ab-a82be31872fd.png)

MarkPieszak:
Just a tip, you can preserve console logs inside chrome with the checkbox.

phillipj:
@nhjk I stumbled upon this my self, so I opened #478. Fingers crossed it'll get merged :)

nhjk:
@MarkPieszak thanks for the tip, can't believe I've been missing that checkbox this whole time.
@phillipj looking forward to the merge as well! I'm closing this issue because that pull request will resolve it.

PR #478, opened 2016-05-06
Don't reload app upon proxy errors.

phillipj:
Previously the clientside application would reload whenever a proxy answered with an error response
or in other ways behaved badly. This change removes that reloading behavior, as reloading the clientside
application probably won't fix any issue the proxy has anyway.

/ping @SpaceK33z as he made these socket handlers in the first place

Refs https://github.com/webpack/webpack-dev-server/issues/446

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -54,7 +54,6 @@ var onSocketMsg = {
 		for(var i = 0; i < errors.length; i++)
 			console.error(stripAnsi(errors[i]));
 		if(initial) return initial = false;
-		reloadApp();
 	}
 };
 


SpaceK33z:
👍  seems fine to me.

sokra:
Thanks

rthewhite:
Any idea on when this might be released? It's a quite annoying.

SpaceK33z:
@rthewhite, this was released as part of `1.15.0`.

rthewhite:
@SpaceK33z thank you for the heads up 👍

PR #480, opened 2016-05-15
Fix error message box display when having not found module

vinhlh:
With many errors like
```
bundle.js:40842 Uncaught Error: Cannot find module "./../CandidateHeader"
```

The screen show no thing, only the error message in console log.

The root cause is webpack-dev-server's client shows the iframe, and it hides the below error message box.


vinhlh:
already fixed in https://github.com/webpack/webpack-dev-server/commit/2c9a461b3433bb9d58fa2cea576aab822b88d5fd

PR #496, opened 2016-05-31
Use 'document.scripts' instead of document.getElementsByTagName

perry-mitchell:
`document.scripts` preserves script-addition order, so we can add scripts in the HEAD as well as BODY.

Currently using `getElementsByTagName` fails if the script is placed in the HEAD.

EDIT: This is nicely discussed [here](http://www.2ality.com/2014/05/current-script.html), where the author mentions (discussing the `getElementsByTagName` alternative):
> There are two situations where this snippet will fail, though. When a script is loaded asynchronously using `<script async>`, the order of script execution is uncertain. Therefore the `document.getElementsByTagName()` trick will not work, as it relies on the currently executed script being the last of that collection. __The same holds true for `<script>`s that are not appended to the end of the document__. In the future this might become less of a problem, as `document.currentScript` __properly resolves async scripts__.

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -7,7 +7,7 @@ if (typeof __resourceQuery === "string" && __resourceQuery) {
 	urlParts = url.parse(__resourceQuery.substr(1));
 } else {
 	// Else, get the url from the <script> this file was called with.
-	var scriptElements = document.getElementsByTagName("script");
+	var scriptElements = document.scripts;
 	var scriptHost = scriptElements[scriptElements.length-1].getAttribute("src");
 	scriptHost = scriptHost && scriptHost.replace(/\/[^\/]+$/, "");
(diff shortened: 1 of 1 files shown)


SpaceK33z:
Hi, thanks for this PR. A lot of people have this issue, also see #117. Your PR is almost there, but could you make it to work like this:

1. Use `document.currentScript` if available. This is the most reliable way to get the current script, and works in [most browsers](http://caniuse.com/#feat=document-currentscript).
2. Fallback to `document.scripts`.
3. If that fails, throw an exception.

perry-mitchell:
Thanks @SpaceK33z - Will update this as soon as I can 🙂

perry-mitchell:
@SpaceK33z I've updated the PR - using `document.currentScript` as you said.

SpaceK33z:
Thank you! I have some tiny nitpicks, but I'll fix those in a separate commit.

PR #499, opened 2016-06-06
set headers on all methods to support e.g. CORS

davidreher:
Hey @sokra,

currently the headers are only set on GET responses. It would be great to also set them on at least OPTIONS as well to support e.g. CORS. My solution is to set headers on all methods supported by express.

Please also backport this to 1.x

cc: @mischi


sokra:
Thanks

PR #506, opened 2016-06-14
Return http.Server instance from listen

mjackson:
Return the `http.Server` instance from `listen` to [match the return value of the same function in express](https://github.com/expressjs/express/blob/31dd549f350accd7b4e3685c13f745e857557827/lib/application.js#L617).


SpaceK33z:
Thanks!

Issue #487, opened 2016-05-25
Support for multiple dev servers exported from a single webpack config


ryancole:
I'm using webpack CLI with a .js config file. The config file exports an array of configurations. Each configuration provides some dev server settings, using unique ports. When I start webpack dev server and point it to that webpack config, it launches a single server and chooses the last exported webpack config.

Would it be possible to make webpack dev server see these multiple exported configs and launch X number of needed servers?

SpaceK33z:
Launching x number of needed servers should not happen in webpack-dev-server. You can write a simple script that launches those. There is [npm-run-all](https://github.com/mysticatea/npm-run-all), which can run multiple scripts in parallel.

PR #509, opened 2016-06-15
Would something simple like this be a start to handling an array of webpack configs?

ryancole:
I work on a project that has several webpack configs. All of the configs are essentially the same, they just have different entry points and output locations. Because of the this, the configs are exported as an array from webpack.config.js. This works great for a simple webpack build, but webpack-dev-server doesn't handle an array of configs. Currently, just the first config is used if an array is encountered.

I'd think that as long as the config specifies a unique port for each config then webpack-dev-server should have no problem booting up more than 1 express server.

I just made this super basic, non-master-ready hack and it seems to do what I'm wanting. I get multiple servers. Would an approach like this be a viable option, and perhaps one day be included as part of webpack-dev-server? I opened issued #487 on this awhile back.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -278,4 +278,8 @@ function processOptions(wpOpt) {
 	});
 }
 
-processOptions(wpOpt);
+if (Array.isArray(wpOpt)) {
+	wpOpt.forEach(processOptions);
+} else {
+	processOptions(wpOpt);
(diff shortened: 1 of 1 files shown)


akhaku:
May want to fix the lint error? See the travis build...
+1 from me, this is something I would find useful, too.

SpaceK33z:
See #487.

PR #510, opened 2016-06-16
Add --skip-errors option to webpack-dev-server

### LINKED ISSUE #209: Client reloads app even when there is an error and the screen is just shown for a very short time.
When loading the application with the live client e.g. http://localhost:3000/webpack-dev-server/index.html the browser is reloaded on script/style changes. But when there is an syntax error (and using the NoErrorsPlugin) the browser is reloaded with the message: **Errors while compiling. App ready.** And an error screen is displayed just before the browser is reloaded. But if the browser is reloaded manually after this syntax error has occured (so it is the initial load) the error screen is displayed nicely with the title: **Errors while compiling. App updated with errors. No reload!**.

Is there a reason why the client reloads the browser on errors? Thinking of these lines in the client: https://github.com/webpack/webpack-dev-server/blob/8fddb2adbf71582d4728d85021a32bd3175571c4/client/index.js#L47-L48

Or are you open to change this somehow so the error message is shown until the syntax error is fixed and a new bundle is generated even when the error occurs during a regular reload (not just the initial load)?

Kovensky:
Intended to be used when NoErrorsPlugin is active in webpack, but can be
used without it. It should work without --hot too, but untested.

Prevents reloadApp() from being called when there are compilation errors
reported; ultimately stopping webpack/hot/{only-,}dev-server from trying
to fetch a nonexisting update (if NoErrorsPlugin is active) and/or
reloading the window.location.

Fixes #209.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -44,6 +44,11 @@ yargs.options({
 		describe: "Do not refresh page if HMR fails",
 		group: ADVANCED_GROUP
 	},
+	"skip-errors": {
+		type: "boolean",
+		describe: "Do not reload if there are compilation errors",
+		group: ADVANCED_GROUP
+	},
 	"stdin": {
 		type: "boolean",
 		describe: "close when stdin ends"
@@ -177,6 +182,9 @@ function processOptions(wpOpt) {
 
 	if(!options.hotOnly)
 		options.hotOnly = argv["hot-only"];
+
+	if (!options.skipErrors)
+		options.skipErrors = argv["skip-errors"];
 
 	if(argv["content-base"]) {
 		options.contentBase = argv["content-base"];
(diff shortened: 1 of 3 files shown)


SpaceK33z:
@Kovensky, could you resolve the merge conflicts? Will merge after that!

SpaceK33z:
@Kovensky ping

Kovensky:
@SpaceK33z I'm actually not sure this is necessary anymore, IIRC there were changes on the webpack side that might have fixed this, but I don't remember which.

SpaceK33z:
You're right, it seems that when using HMR, it doesn't refresh anymore on error. When not using HMR, it does still refresh, but I don't think that's really a problem for now.

At first I thought we could completely remove the `reloadApp()` thingy for errors and warnings, but the problem with that is that if you don't have the DevTools open, you'll see an old version of your app. That can be very confusing.

Issue #363, opened 2016-01-06
please open a new branch to support webpack 2


zhaxi:
Webpack 2.0 beta makes a lot of changes that current webpack-dev-server can't follow up, could you open a branch to support webpack 2.0 beta?

stevenjlho:
+1

ahlee2326:
+1

mindreframer:
that is true... could we keep up the 1.x releases for some time, until the rest of the ecosystem has caught up with changes? 

For example I really need this commit: 
- https://github.com/webpack/webpack-dev-server/commit/434651fc21ddaea8f4875c1f4dd0eaea754f17d0

I'm using a framework (Phoenix / Elixir) , that starts webpack in the background and currently it just continues to run after the main application stops. I was hoping for a 1.4.2 release, but it is only available in 2.0.0-beta, which I can't use, because there is a boatload of plugins not compatible with 2.0 webpack  (at least in the package.json). Updating them all just to be able to install webpack 2.0 is unfeasible. 

Would it be possible to release some more 1.4.x versions with changes that don't break backward-compatibility? 

That would be great!

SpaceK33z:
I'm currently working on making a 1.x release.

SpaceK33z:
Version `1.15.0` has been released. The [changelog](https://github.com/webpack/webpack-dev-server/blob/webpack-1/CHANGELOG.md) is in the `webpack-1` branch.

PR #515, opened 2016-06-29
Handle missing `src` attribute on unused last script

gigabo:
Hi,

This fixes a bug introduced in 3d3c000, in the case where:

- `__resourceQuery` is set, and
- The last script tag on the page has no `src` attribute

Previously the script tag's `src` attribute was accessed lazily only if `__resourceQuery` was not defined. The last script on the page may not be the script that loaded webpack-dev-server.  This can happen, for example, with the use of an asynchronous loader such as LABjs.

I've pointed this PR at master, but it is based on `v1.14.1`, which is the latest non-beta release.  We are not on webpack `>=2.0.3-beta <3`, so we would need this bug fix on the `^1.14.1` range.  How are you handling patches there?

Thanks,

Bo

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -2,7 +2,8 @@ var url = require('url');
 var SockJS = require("sockjs-client");
 var stripAnsi = require('strip-ansi');
 var scriptElements = document.getElementsByTagName("script");
-var scriptHost = scriptElements[scriptElements.length-1].getAttribute("src").replace(/\/[^\/]+$/, "");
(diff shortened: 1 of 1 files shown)


gigabo:
Oh, this is basically equivalent to #447.  Just need that applied in `^1.14.1`.

SpaceK33z:
I'll make sure to backport this for the next v1 release. Also see #363.

PR #518, opened 2016-06-30
Fix HistoryApiFallback when publicPath is set

### LINKED ISSUE #216: historyApiFallback doesn't work, for scripts, if you're nested multiple path sections deep
I'm using `historyApiFallback` so that my react application can use the newer client-side routing / HTML 5 history. I've set `historyApiFallback` to `true`. This works when the client-side route is something like `/users` or `/home`, etc. This does not seem to work if the client-side route is something like `/users/1`, though. It tries to look for scripts within my HTML file relative to `/users`. Even if my HTML script is like ...

```html
<script type="text/javascript" src="./static/app.js"></script>
```

it looks for that at `/users/static/app.js`.

meghprkh:
Currently HistoryApiFallback does not work properly when publicPath as it redirects request to `/index.html` which gives as it does not exist, instead we should redirect to publicPath (which redirects to `publicPath/index.html`).

Fixes #216

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -189,7 +189,10 @@ function Server(compiler, options) {
 		historyApiFallback: function() {
 			if(options.historyApiFallback) {
 				// Fall back to /index.html if nothing else matches.
-				app.use(historyApiFallback(typeof options.historyApiFallback === 'object' ? options.historyApiFallback : null));
+				var historyApiFallbackOptions = typeof options.historyApiFallback === 'object' ? options.historyApiFallback : {};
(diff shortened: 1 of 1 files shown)


SpaceK33z:
This is kinda tricky stuff, and I'm not sure if it's always the case that you want to use the `publicPath` for this.

Do other people have this problem?

SpaceK33z:
Closing for now. #216 was already fixed by another PR. In the docs it's also documented that you need to use a custom index when using `publicPath`.

We could theoretically do this automatically, but this is a breaking change that I only want to do if more people have this problem. Will revisit this PR when that happens.

PR #519, opened 2016-07-04
Added support for basic auth option.

slont:
Hi.
Added basic auth option in this branch.
For example:
```
devServer: {
  ...
  basicAuth: {
    id: "admin",
    passwd: "pass"
  }
}
```
We can use basic auth very easily.


SpaceK33z:
Same as #520?

PR #520, opened 2016-07-04
Add basic auth.

slont:


SpaceK33z:
This is a very specific use-case. AFAIK, the dev server provides enough hooks to do this in a Node script. Remember that the dev server is for development only.

PR #531, opened 2016-07-20
optionally provide middleware as value in options array

dcep93:
we would like to provide our own middleware to WebpackDevServer


graingert:
@dcep93 use options.setup

https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L255-L258
```js
const webpackConfig = {
    devServer: {
        setup(app) {
            app.use(new YourOwnMiddleware())
        }
    }
}
export default webpackConfig
```

dcep93:
aah, must have missed that

thanks so much

PR #532, opened 2016-07-21
Assign --cacert CLI argument to correct server options property

ryan-codingintrigue:
The `--cacert` CLI argument was being mapped to `options.cacert` which is never actually used in _Server.js_.

The correct property name is `options.ca`. In order to preserve documentation & backwards compatibility for any existing users currently passing the the `--cacert` flag, I've left the flag the same but mapped it to the `options.ca` property so that it gets passed to the `createServer` call.


SpaceK33z:
Thanks!

Issue #381, opened 2016-01-21
Uncaught TypeError: Cannot read property 'replace' of null


cdbkr:
Hi, with version **1.14.1** I'm receiving this error:
```
Uncaught TypeError: Cannot read property 'replace' of null
```
I'm using an inline script in the index.html to load the .JS file (which is in the folder declared  by publicPath) and I enabled the hot module replacement to receive updates.

Using **version 1.14.0**, it works fine (so at the moment I am not updating the webpack-dev-server module to 1.14.1).

matiss:
I have the same situation as the OP. I also have a inline script which loads the bundle and I am getting the same error when I upgraded Webpack to `1.14.1`. Downgrade fixed the error, though.

techird:
Same problem, seems a problem with `1.14.1`, `1.14.0` is alright.

reebalazs:
I can confirm that it breaks when you have a script with no src tag. This is a showstopper if you have injected js into the html.

It breaks on this line: https://github.com/webpack/webpack-dev-server/blob/5b94506f0c47d7fd0bde8d215b6195450e71fc7e/client/index.js#L11

1.14.0 works for me too.

#372 is a duplicate.

wedgex:
I'm getting the same error, downgrading to 1.14.0 temporarily as a work around.

Nebukam:
Same issue happened to me. Downgraded as well, works like a charm with 1.14.0

globexdesigns:
FYI: This is happening when you have a `<script />` tag in your code that doesn't have a `src` attribute.

adamhartford:
I was able to work around this with a hack-ish change:

```javascript
<script>
    document.currentScript.src =''; // So that the script executes, but later has a src.

    // ... Do rest of work.   
</script>
```

damonmcminn:
I also have this issue. The sole `script` tag on my page contains a `src` attribute:

`<script src="bundle.js"></script>`

Downgrading to *1.14.0* solved the issue.

marcelpanse:
Same problem here because of the Google Tag Manager script in the index.html

JustinMGaiam:
This issue affected us as well.  Fixing it required setting webpack-dev-server to `1.14.0` in package.json and then the running `npm install` again.

incleaf:
Same problem with 1.14.0

idoco:
Just to point out for anyone else having the same issue - 
[...]

PR #535, opened 2016-07-24
Fix issue with defer/async/empty script src

vutran:
This PR fixes issue #381 for `v1.14.1`.

When you have a page with a `script` tag with no `src` attribute or if the script is `async`, it's trying to call `.replace()` from a non-string.

This patch should also fix the loading issues for facebookincubator/create-react-app/issues/134


SpaceK33z:
Closing in favor of #496, which provides a cleaner solution.

PR #548, opened 2016-08-06
Added message for server starting address.

CrazyDeveloper:
With this issue where address is appended to rest of the text should be fixed eg. `70% 1/1 build moduleshttp://localhost:8080/` should be `70% 1/1 build modules http://localhost:8080/` .


SpaceK33z:
Fixed with PR #469.

PR #550, opened 2016-08-09
Updated connect-history-api-fallback to 1.3.0.

mik01aj:
This way we'll get the fix for https://github.com/bripkens/connect-history-api-fallback/issues/5.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "compression": "^1.5.2",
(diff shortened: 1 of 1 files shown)


SpaceK33z:
See the documentation for [caret ranges on npm](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004). If you reinstall your node_modules, 1.3.0 should automatically be installed.

PR #554, opened 2016-08-16
Add a flag to prevent websocket reconnect

bs85:
I managed to unify my "regular dev" and "hot" builds in a single one, that is served through my regular pipeline. The only problem is that if dev-server isn't running, the websocket client attempts to reconnect every 2 seconds and pollutes the console and request tabs a bunch.

This is a rather dirty fix, but there is no config mechanism for the websocket client, if you have a suggestion for something cleaner let me know and I'll make another PR.

Cheers

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -72,9 +72,12 @@ var newConnection = function() {
 
 		// Try to reconnect.
 		sock = null;
-		setTimeout(function () {
-			newConnection();
-		}, 2000);
+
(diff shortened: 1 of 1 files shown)


bs85:
PS: if you accept, it would probably be a good idea to mention it on the wiki
https://github.com/webpack/docs/wiki/webpack-dev-server

SpaceK33z:
If you want to use a build when the dev-server is not running, you should build the files to disk with webpack instead. The whole point of the dev-server is to provide a client with live-reload.

PR #556, opened 2016-08-20
feat: pass in-memory fileSystem

gdi2290:


SpaceK33z:
What exactly is your motivation behind this? Accessing the webpack filesystem can leads to various bugs and generally should not be done.

SpaceK33z:
ping @gdi2290 ;).

SpaceK33z:
Closing for now, as the in-memory filesystem is an implementation detail that should not be exposed to the user; also see the "Contributing" guidelines in the README.

gdi2290:
I need it to serve files that are tracked ad or transformed by webpack rather than reading the file again

gdi2290:
if that's the case there needs to be a way to grab the in-memory filesystem then this PR makes sense. If someone changed how the private in-memory filesystem is exposed then they have to make sure it's exposed here.

Issue #560, opened 2016-08-22
Proxy bypass not working in 1.15.0


cletusw:
A `webpack.config.js` like this (mostly pulled from [the docs](http://webpack.github.io/docs/webpack-dev-server.html))

```javascript
var path = require("path");
module.exports = {
  entry: {
    app: ["./app/main.js"]
  },
  output: {
    path: path.resolve(__dirname, "build"),
    publicPath: "/assets/",
    filename: "bundle.js"
  },
  devServer: {
    proxy: {
      '/': {
        bypass: function(){
          return '/test.html';
        }
      }
    }
  }
};
```

works fine with `webpack-dev-server@1.14.1`, but after updating to `1.15.0`, I get

```
$ node_modules/.bin/webpack-dev-server --content-base build/
/Users/claytonwatts/dev/test/node_modules/webpack-dev-server/node_modules/http-proxy-middleware/lib/config-factory.js:42
        throw new Error('[HPM] Missing "target" option. Example: {target: "http://www.example.org"}');
        ^

Error: [HPM] Missing "target" option. Example: {target: "http://www.example.org"}
    at Object.createConfig (/Users/claytonwatts/dev/test/node_modules/webpack-dev-server/node_modules/http-proxy-middleware/lib/config-factory.js:42:15)
    at new HttpProxyMiddleware (/Users/claytonwatts/dev/test/node_modules/webpack-dev-server/node_modules/http-proxy-middleware/lib/index.js:16:45)
    at module.exports (/Users/claytonwatts/dev/test/node_modules/webpack-dev-server/node_modules/http-proxy-middleware/index.js:4:12)
    at /Users/claytonwatts/dev/test/node_modules/webpack-dev-server/lib/Server.js:184:9
    at Array.forEach (native)
    at Object.Server.features.proxy (/Users/claytonwatts/dev/test/node_modules/webpack-dev-server/lib/Server.js:171:19)
    at Server.<anonymous> (/Users/claytonwatts/dev/test/node_modules/webpack-dev-server/lib/Server.js:274:20)
    at Array.forEach (native)
    at new Server (/Users/claytonwatts/dev/test/node_modules/webpack-dev-server/lib/Server.js:273:40)
    at Object.<anonymous> (/Users/claytonwatts/dev/test/node_modules/webpack-dev-server/bin/webpack-dev-server.js:183:1)
```

[...]

PR #563, opened 2016-08-23
Fix bypass proxy middleware

SpaceK33z:
The `bypass` feature was no longer working because the code added the bypass method as middleware AND `httpProxyMiddleware` as middleware. However, if a `bypass` method is given, it should only use that as middleware.

This was caused in #359, and effects `1.15.0` and the 2.x branch.

For now, this is based on `1.15.0`, because the people that reported issues with this are using that version. It should be merged to `master` too.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -171,22 +171,23 @@ function Server(compiler, options) {
 				options.proxy.forEach(function(proxyConfig) {
 					var bypass = typeof proxyConfig.bypass === 'function';
 					var context = proxyConfig.context || proxyConfig.path;
+					var proxyMiddleware;
+					// It is possible to use the `bypass` method without a `target`.
+					// However, the proxy middleware has no use in this case, and will fail to instantiate.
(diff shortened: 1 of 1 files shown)


SpaceK33z:
Verified to work in #560.

HZZformGD:
@SpaceK33z  excuse me , can i ask u a question about webpack-dev-server,  my English is not good, so hope you can know what i say,     my problem is when i  `npm run dev`  ,  i got this error 

```
D:\nodejs\vue+koa+MD\myblog\node_modules\http-proxy-middleware\lib\config-factory.js:42

        throw new Error('[HPM] Missing "target" option. Example: {target: "http://www.e
xample.org"}');

```
i google this, and found someone has this trouble , his trouble is similar to mine.  
And then i follow the suggestion——copy the `lib/server.js `  which is fixed ,and run again , unluckyly, i got the same error again , ps: i can run` node app.js` ,it works on chrome, it also works when i `npm run build`,

Issue #566, opened 2016-08-24
Proxy rewrite function no longer called in 1.15.0


jdkrebs:
Hey there,

in 1.14.1 the following rewrite function worked
```
  proxy: {
      '/api/**': {
        target: 'http://localhost:9001',
        rewrite: function(req) {
          req.url = req.url.replace(/^\/api(.+)$/, '$1');
          return req.url;
        },
        ignorePath: false,
        changeOrigin: true,
        secure: false
      }
    }
```

As of 1.15.0 The rewrite function doesn't get called anymore.

Was there a breaking change with how the rewrite function works? I understand there was a switch in how proxy works but I figured I would have to update to 2.x.x for those changes to manifest.

Thoughts?

SpaceK33z:
Confirmed that this is happening. Could you try if the fix in #567 works for you? The fastest way to check is to copy/paste the contents of the PR in `node_modules/webpack-dev-server/lib/Server.js`.

jdkrebs:
@SpaceK33z I gave this a quick shot and it didn't seem to fix the problem. 

I probably won't be able to fool around with it any further until tomorrow morning.

chimurai:
@SpaceK33z @jdkrebs 

This line in `http-proxy-middleware` kind of ignores/reverts the rewritten path in your fix. ([source](https://github.com/chimurai/http-proxy-middleware/blob/3b179bc43831e56a0117af3d8b874fafaed3cd4d/lib/index.js#L82
))
```javascript
// http-proxy-middleware
req.url = (req.originalUrl || req.url);
```

Not sure what the best approach would be to make it backwards compatible.

As alternative, you could use the `.pathRewrite` option offered by HPM.
The syntax is slightly different...
```javascript
    proxy: {
      '/api/**': {
        target: 'http://localhost:9001',
        pathRewrite: function (path, req) {
          return path.replace('/api/foo', '/api/bar');
        }
      }
    }
```

For more info on `pathRewrite`: [http-proxy-middleware/recipes/pathRewrite.md](https://github.com/chimurai/http-proxy-middleware/blob/master/recipes/pathRewrite.md#custom-rewrite-function)

---

@SpaceK33z 

I noticed a sudden surge in downloads since v1.15.0 got released and with it also a number of proxy related issues. Feel free to _cc_ me if you need any assistance.

SpaceK33z:
@chimurai, thanks for your help, really appreciate it!

[...]

PR #567, opened 2016-08-24
Fix for proxy rewrite not working anymore

SpaceK33z:
Apparently this whole config option was removed in PR #359.

I'm not 100% sure this fixes everything, so it should be tested by someone who uses this feature.


SpaceK33z:
I give up for now, it still does not seem to work. See last comment in  #566 for more info.

PR #569, opened 2016-08-26
Add optional delay to proxy responses (ms)

rvanmil:
This allows an optional `delay` parameter in the proxy config, which will then delay proxy responses with the specified amount of milliseconds.

``` javascript
proxy: {
  '/api': {
    target: 'https://other-server.example.com',
    delay: 2000 // 2 second delay
  }
}
```

I'm used to having this feature in https://github.com/typicode/json-server but couldn't find a way to do this with webpack-dev-server. I think it's a nice and easy way to simulate slow network responses and test your spinners etc.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -173,13 +173,20 @@ function Server(compiler, options) {
 					var bypass = typeof proxyConfig.bypass === 'function';
 					var context = proxyConfig.context || proxyConfig.path;
 					var proxyMiddleware;
+					var proxyDelayMiddleware;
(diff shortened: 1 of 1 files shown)


SpaceK33z:
Dupe of #570.

rvanmil:
Sorry, I thought I had to create a separate PR for both branches.

SpaceK33z:
Ah, didn't see that the branches were different. Anyway, it should only be `master`, from there we can see if we backport it.

PR #570, opened 2016-08-26
Add optional delay to proxy responses (ms)

rvanmil:
This allows an optional `delay` parameter in the proxy config, which will then delay proxy responses with the specified amount of milliseconds.

``` javascript
proxy: {
  '/api': {
    target: 'https://other-server.example.com',
    delay: 2000 // 2 second delay
  }
}
```

I'm used to having this feature in https://github.com/typicode/json-server but couldn't find a way to do this with webpack-dev-server. I think it's a nice and easy way to simulate slow network responses and test your spinners etc.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -178,7 +178,7 @@ function Server(compiler, options) {
 					// However, the proxy middleware has no use in this case, and will fail to instantiate.
 					if(proxyConfig.target) {
 						proxyMiddleware = httpProxyMiddleware(context, proxyConfig);
-						if (typeof proxyConfig.delay === 'number') {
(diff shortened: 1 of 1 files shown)


SpaceK33z:
Wouldn't this be easy to do on the backend server itself?

rvanmil:
Not if you're proxying to a backend you don't control.

SpaceK33z:
Sorry, I don't see enough use for this. There are two other ways to do it:

- Use the DevTools to slow down the response.
- Use a "delaying proxy" to the other proxy.

bebraw:
I think a documentation PR related to this could be welcome. You can document how to achieve delay so people know how to set it up.

PR #577, opened 2016-08-30
Avoid accessing `document` if `__resourceQuery` is defined

narkq:
Just as title says.
This is a cheery-pick from master branch.


narkq:
Should fix #117 for people who is using webpack-dev-server 1.x with inline mode.

SpaceK33z:
Thanks!

Issue #109, opened 2015-02-14
Option to console.log only important stuff


steida:
It is planned or should I send PR?

We don't care about these messages:
```
[HMR] Waiting for update signal from WDS...
client:14 [WDS] Hot Module Replacement enabled.
```

There should be an option (default) to log only warnings and errors.

maxdow:
+1

smussell:
I agree the following are not useful logs and they are cluttering up my console making it more difficult to sort through and find my own logs.

```` 
[WDS] App updated. Recompiling...
[WDS] Nothing changed.
````

There should be some way for me to disable these logs.  This same issue was filed on the react-hot-loader, but it is apparently a webpack issue.

elektronik2k5:
+1

fusepilot:
Would like this option as well. The logs are a bit verbose after everything is setup correctly.

luisrudge:
:+1:

FlorentD:
+1 as well

dalhundal:
As a very messy temporary work around, I have a script, `development.js` which i conditionally append to the `entry` in `webpack.config.js`.

````
"use strict";

// This is a workaround used alongside the webpack-dev-server hot-module-reload feature
//  - it's quite chatty on the console, and there's no currently no configuration option
//    to silence it. Only used in development.
// Prevent messages starting with [HMR] or [WDS] from being printed to the console
(function(global) {
    var console_log = global.console.log
    global.console.log = function() {
        if (!(
            arguments.length == 1 &&
            typeof arguments[0] === 'string' &&
            arguments[0].match(/^\[(HMR|WDS)\]/)
        )) {
            console_log.apply(global.console,arguments)
        }
    }
})(window)
````

It overloads `console.log` - skipping any messages starting with `[HMR]` or `[WDS]`

SpaceK33z:
Can you guys check if PR #579 works for you?

SpaceK33z:
PR #579 was merged, check that out for more info. Note that it hasn't been released yet.

SpaceK33z:
Released in `2.1.0-beta.3`.

ibrahima:
Quick question, I see that this was merged and is in v1.16, and seems to remove the [WDS] logs, but I still see `[HMR] Waiting for update signal from WDS...` in my console. Is there a different setting for HMR?

lgdexter:
Thanks for this @SpaceK33z

felipenmoura:
Hi
[...]

PR #579, opened 2016-08-31
Add option to control browser log level

SpaceK33z:
Previously, when you used `quiet: true` or `noInfo: true` via the Node API or CLI, this only had effect on the output in the terminal (issue #109).

With this PR, these settings will be respected in the browser console as well.

Maybe it would be more flexible to add a separate option for this, but this would also mean more configuration. Looking for feedback on this.

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -16,6 +16,7 @@ if (typeof __resourceQuery === "string" && __resourceQuery) {
 var sock = null;
 var hot = false;
 var initial = true;
+var connected = false;
 var currentHash = "";
 var logLevel = "info";
 
@@ -81,7 +82,10 @@ var newConnection = function() {
 	}));
 
 	sock.onclose = function() {
-		log("error", "[WDS] Disconnected!");
+		if(connected)
+			log("error", "[WDS] Disconnected!");
+
+		connected = false;
 
 		// Try to reconnect.
 		sock = null;
@@ -91,6 +95,7 @@ var newConnection = function() {
 	};
 
 	sock.onmessage = function(e) {
+		connected = true;
 		// This assumes that all data sent via the websocket is JSON.
 		var msg = JSON.parse(e.data);
 		onSocketMsg[msg.type](msg.data);
--- a/client/index.js
+++ b/client/index.js
@@ -17,16 +17,14 @@ var sock = null;
 var hot = false;
 var initial = true;
 var currentHash = "";
-var quiet = false;
-var noInfo = true;
-var logLevel = 3; // 3 = all, 2 = warnings and errors, 1 = only errors, 0 = quiet
+var logLevel = "info";
 
 function log(level, msg) {
-	if(logLevel >= 3 && level === "info")
+	if(logLevel === "info" && level === "info")
 		return console.log(msg);
-	if(logLevel >= 2 && level === "warning")
+	if(["info", "warning"].indexOf(logLevel) >= 0 && level === "warning")
 		return console.warn(msg);
-	if(logLevel >= 1 && level === "error")
+	if(["info", "warning", "error"].indexOf(logLevel) >= 0 && level === "error")
 		return console.error(msg);
 }
 
@@ -54,14 +52,14 @@ var onSocketMsg = {
 	warnings: function(warnings) {
 		log("info", "[WDS] Warnings while compiling.");
 		for(var i = 0; i < warnings.length; i++)
-			log("warn", stripAnsi(warnings[i]));
+			console.warn(stripAnsi(warnings[i]));
 		if(initial) return initial = false;
 		reloadApp();
 	},
 	errors: function(errors) {
 		log("info", "[WDS] Errors while compiling.");
 		for(var i = 0; i < errors.length; i++)
-			log("error", stripAnsi(errors[i]));
+			console.error(stripAnsi(errors[i]));
 		if(initial) return initial = false;
 		reloadApp();
 	},
--- a/client/index.js
+++ b/client/index.js
@@ -17,43 +17,58 @@ var sock = null;
 var hot = false;
 var initial = true;
 var currentHash = "";
+var quiet = false;
+var noInfo = true;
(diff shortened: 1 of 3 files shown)


SpaceK33z:
To test this:

Add this to your `package.json`:

```
"webpack-dev-server": "webpack/webpack-dev-server#browser-log-level",
```

After `npm install`, go to `cd node_modules/webpack-dev-server`, and run `npm run prepublish`.

client/index.js
```diff
@@ -17,43 +17,58 @@ var sock = null;
...
+			log("warn", stripAnsi(warnings[i]));
 		if(initial) return initial = false;
 		reloadApp();
 	},
 	errors: function(errors) {
-		console.log("[WDS] Errors while compiling.");
+		log("info", "[WDS] Errors while compiling.");
 		for(var i = 0; i < errors.length; i++)
-			console.error(stripAnsi(errors[i]));
+			log("error", stripAnsi(errors[i]));
```
gaearon: These are useful errors and I want them to be reported. They are related to user’s code.

client/index.js
```diff
@@ -17,43 +17,58 @@ var sock = null;
...
 	ok: function() {
 		if(initial) return initial = false;
 		reloadApp();
 	},
 	warnings: function(warnings) {
-		console.log("[WDS] Warnings while compiling.");
+		log("info", "[WDS] Warnings while compiling.");
 		for(var i = 0; i < warnings.length; i++)
-			console.warn(stripAnsi(warnings[i]));
+			log("warn", stripAnsi(warnings[i]));
```
gaearon: These are useful warnings and I want them to be reported. They are related to user’s code.

client/index.js
```diff
@@ -68,7 +83,7 @@ var newConnection = function() {
 	}));
 
 	sock.onclose = function() {
-		console.error("[WDS] Disconnected!");
+		log("error", "[WDS] Disconnected!");
```
gaearon: However this is generally not useful at all, especially because it enters a crazy loop where it prints this message every few seconds:

gaearon:
Hey, thanks for taking a look!

So my problem is there’s no way to separate useful output (compilation warnings and errors) from useless output (generally anything prefixed with `WDS`—our users have no idea what WDS even is, and they don’t need 20 messages about being disconnected either).

My guess is there’s some bug causing repeated disconnects. It just isn’t right that it keeps retrying and spamming the console IMO.

client/index.js
SpaceK33z: Agree. So would you say remove this log completely, or make sure it doesn't get repeated all the time?

gaearon: I think `error` and `warning` could be separate log levels from `server-warning` and `server-error`. And that it shouldn’t print disconnected more than once regardless of the log level. Ideally it wouldn’t print failed `GET *` either—not sure if there’s a way to prevent this.

SpaceK33z: So you mean instead of using the existing `quiet` option to determine the log level, introduce a separate option?

gaearon: I don’t know to be honest. It’s already a bit of a mess. Why do both `quiet` and `noInfo` exist?

SpaceK33z: `quiet` and `noInfo` are only used in webpack-dev-middleware. If `noInfo: true`, the console will only show user errors. If `quiet: true`, the console won't show anything.

Maybe it's too big of a hack to reuse this for client errors though. I can imagine most people use `noInfo` or `quiet` _because_ they can also see it in their browser.

If we'd make a separate option though, I would prefer using a "log level". Maybe `clientLogLevel`, which takes `null` (no logs), `"error"`, `"warn"`, `"info"` (default)?

gaearon: This sounds reasonable but I’d prefer if there was a way to show errors/warnings from compilation but without the server logs/errors/warnings.

SpaceK33z: Good point. We could exclude the compilation errors from the log level, so far I think no-one has complained about that.
I'll update this PR later today.

SpaceK33z:
@gaearon, I've updated the PR with your feedback.

- I no longer abuse `quiet` and `noInfo`. Instead, there is a separate option, `clientLogLevel`. You can set this to `error`, `warning`, `info` or `none`. It defaults to `info`.
  - Via the CLI you can use this with `--client-log-level`. Not sure about the name btw, so suggestions are welcome.
- The repeated message `[WDS] Disconnected!` will now only display once.
  - Unfortunately there [doesn't seem to be a way](https://stackoverflow.com/questions/14337351/can-i-prevent-the-chrome-developer-tools-console-from-logging-image-404-errors) to hide the `ERR_CONNECTION_REFUSED` errors. This is happening because the websocket tries to re-connect every two seconds.

gaearon:
Thanks for taking the time to work on this.

>This is happening because the websocket tries to re-connect every two seconds.

Is this useful? Can we make it reconnect three times and give up? Or reconnect with exponential backoff?

SpaceK33z:
> Is this useful? Can we make it reconnect three times and give up? Or reconnect with exponential backoff?

If your PC gets into sleep mode, and a tab with the dev-server is still open, the connection would not be restored when you get out of sleep mode. I've made #584 for this, because I think this is a bit outside the scope of this PR.

Do you think the rest of this PR is good?

gaearon:
Yep, thanks.

gaearon:
The only issue is we’re still on 1.x and can’t jump to 2.x until it’s stable.
Any chance you’d be interested in backporting this to 1.x?

farwayer:
@SpaceK33z It will be very nice if this option can control HMR output also.

ibrahima:
If I understand correctly, as @farwayer said the HMR output is still logged regardless, so we're still getting some unhelpful spam (which, in my case, is annoying because our front end tests get spammed with it constantly). Not sure whether it's even possible for that code to use the same logger since it's in a different project, but here's the source: https://github.com/webpack/webpack/blob/master/hot/dev-server.js

PR #590, opened 2016-09-04
Remove js-beautify and improve ESLint config

SpaceK33z:

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Before, there were two linters doing the same thing: linting and fixing lint errors. There were also two commands for this, `npm run lint` and `npm run beautify-lint`.
ESLint has gotten extremely popular (especially since JSCS was merged into ESLint) and is very actively developed, so it makes sense to move over to ESLint. Many people also have ESLint setup in their editor.

**What is the new behavior?**

Now, there is only one linter, ESLint. The js-beautify config is merged in the ESLint config
as best as I could do. There is almost no documentation about the rules of js-beautify,
so this was a bit hard.

ESLint was updated to v3, which provides many powerful rules, and auto-fixing of some rules (with `npm run beautify`).

`npm run beautify-lint` was removed, because `npm run lint` does the same now (which makes sense).

- [ ] Yes
- [ ] No

* Impact:

**Other information**:

I also checked the new ESLint config against the main webpack repo, and adjusted the rules accordingly. This config finds about ~ 50 new errors in the main repo, but these are all very valid (e.g. dead code, inconsistent code style). Optionally, I could update the main webpack repo after this.

Another interesting thing to do would be to make a new repo with this config, `eslint-config-webpack`. But that's outside the scope of this PR for now.

Pinging @TheLarkInn for review.


TheLarkInn:
LGTM!

PR #596, opened 2016-09-07
CLI: Describe that `--inline` toggles livereload

ajoslin:
- [x] Bugfix (CLI)
- [x] Feature (CLI)

`webpack-dev-server --help | grep live` doesn't help me, and `webpack-dev-server --help | grep reload` doesn't help me.

**What is the new behavior?**

Both `webpack-dev-server --help | grep live` and `webpack-dev-server --help | grep reload` help me

- [ ] Yes
- [x] No

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -37,7 +37,7 @@ yargs.options({
 	"inline": {
 		type: "boolean",
 		default: true,
-		describe: "Inline mode"
+		describe: "Inline mode (set to false to disable including client scripts like livereload)"
 	},
 	"hot-only": {
 		type: "boolean",


ajoslin:
cc @gdi2290

SpaceK33z:
I'm not sure about this. It's technically not true that setting `inline` to `false` disables live reload. If you use the iframe modus (by going to `http://localhost:8080/webpack-dev-server/`), you still have live reload. Also, if you manually include the client script, you also have live reload.

SpaceK33z:
Something like "Set to false to disable including client scripts" would be more correct I think.

ajoslin:
Alright. Could we explicitly say that "client scripts" includes livereload?

Perhaps:

`Inline mode (set to false to disable including client scripts like livereload)`

SpaceK33z:
Okay, fair enough. It's just that it's important to know that using `--inline false` does not mean that live-reload is disabled per definition.

ajoslin:
I understand that isn't exactly what it does. Maybe it would be better to have an option just for livereload? It's a very reasonable thing to want to disable IMO, and most tools like this have an option for it.

SpaceK33z:
That's already possible, using `--lazy` :). If you update the PR with the text you suggested above, I'll merge this.

ajoslin:
:+1: Done

SpaceK33z:
Thanks!

ajoslin:
👌 And thank you.

Issue #464, opened 2016-04-26
not possible to disable content base


pseidemann:
I'm managing all my static files with the file-loader so I don't want to serve any static files from the current directory.
I tried to do `webpack-dev-server --content-base /dev/null` which prevents at least loading any file from the filesystem but it is responding with http code 500 for urls which don't exist which is a little bit annoying.
is there something possible like `--content-base false` or `--no-content-base`?
from the code it looked like this argument is optional but it seems like it's always falling back to the working directory which I don't want.

pseidemann:
I worked around this issue with doing `--content-base nonexistent/` for now

SpaceK33z:
Your wish has been granted! Use with `--no-content-base`. Note that it's not been released yet.

pseidemann:
@SpaceK33z wow awesome, thanks!

PR #598, opened 2016-09-07
use contentBase from webpack config - fixes #597

### LINKED ISSUE #597: contentBase in webpack.config is ignored
webpack: 2.x 
webpack-dev-server: 2.1.0-beta.3
OSX 10.x

**Current behavior:**

When specifying a contentBase in your webpack.config file (and running webpack-dev-server from the command line, without any arguments), it is ignored & overwritten with the current path

**Expected/desired behavior:**

contentBase from the config file should be used

This bug was introduced by commit be9efdaa2e4539cd90eab27494a965c500c03320

wouterverweirder:
Bugfix for issue #597

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -184,7 +184,7 @@ function processOptions(wpOpt) {
 	if(!options.clientLogLevel)
 		options.clientLogLevel = argv["client-log-level"];
 
-	if(argv["content-base"]) {
+	if(!options.contentBase && argv["content-base"]) {
 		options.contentBase = argv["content-base"];
 		if(/^[0-9]$/.test(options.contentBase))
 			options.contentBase = +options.contentBase;


MoOx:
This fix looks good to me!

SpaceK33z:
Thanks. This has the consequence though that `--content-base` is ignored if you set it in your webpack config. I think that's fine for now, just need to make sure it will be mentioned in the changelog.

MoOx:
Maybe we could display a warning if a user specify both? Would help a lot :)

SpaceK33z:
Yeah, that would be very nice. PR welcome. I'll work on it in after today if nobody sends a PR.

MoOx:
The tricky part is that argv[content-base] is filled by default, so very tricky to catch when the user specify something from here...

SpaceK33z:
You could check if `arg["content-base"]` is the same as `process.cwd()`... bit hacky though. It would mean that there is an edge-case, if you specify exactly your PWD then it doesn't work. But if there is no other way, that's better than nothing.

MoOx:
> But if there is no other way, that's better than nothing.

Do not add default in yargs? (and define default later, like it was before)?

SpaceK33z:
Yeah that also would be possible. Just need to make sure that it is still possible to explicitly disable contentBase (see #464).

SpaceK33z:
@MoOx, are you planning to do a PR with this? Will wait with a new release if so.

MoOx:
Don't have the time atm. Sorry :)

PR #599, opened 2016-09-07
Don't use PWD as a default for --content-base in yargs

SpaceK33z:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

#598 fixed a bug where the `contentBase` from the webpack config was ignored in the CLI. However, there were more issues.

The default value according to `--help` is the PWD. This is not true. If `contentBase` is set in the webpack config, this default value is ignored.

**What is the new behavior?**

Yargs will not incorrectly display the default value of `--content-base` now.

Originally I intended to add an error message when `contentBase` was set from the config AND from the CLI, but this would be inconsistent with all other options. The config always gets preference.

- [ ] Yes
- [x] No

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -90,7 +90,6 @@ yargs.options({
 	},
 	"content-base": {
 		type: "string",
-		default: process.cwd(),
 		describe: "A directory or URL to serve HTML content from.",
 		group: RESPONSE_GROUP
 	},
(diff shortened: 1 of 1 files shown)


SpaceK33z:
cc @wouterverweirder @MoOx

MoOx:
lgtm! Thanks for the reactivity!

PR #602, opened 2016-09-08
Fix historyApiFallback for nested paths

aves84:
I don't quite understand how the files are serving, specifically why files on top level served fine while nested are not.

For example, i had follow sctructure:
```
/public
  /userpics/pic1.jpg
  index.html
  userpic.jpg
```
With config
```
devServer: {
  contentBase: 'public',
}
```
`GET /userpic.jpg` => `/public/userpic.jpg`
`GET /userpics/pic1.jpg` => `/public/userpics/pic1.jpg`

After adding `historyApiFallback: true` to config

`GET /userpic.jpg` => `/public/userpic.jpg`
`GET /userpics/pic1.jpg` => `/public/index.html`

But swaping `defaultFeatures.push("contentBase");`  and `defaultFeatures.push("historyApiFallback", "middleware");` fixes it.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -256,11 +256,11 @@ function Server(compiler, options) {
 	var defaultFeatures = ["setup", "headers", "middleware"];
 	if(options.proxy)
 		defaultFeatures.push("proxy");
-	if(options.historyApiFallback)
-		defaultFeatures.push("historyApiFallback", "middleware");
 	defaultFeatures.push("magicHtml");
 	if(options.contentBase !== false)
 		defaultFeatures.push("contentBase");
+	if(options.historyApiFallback)
+		defaultFeatures.push("historyApiFallback", "middleware");
 	// compress is placed last and uses unshift so that it will be the first middleware used
 	if(options.compress)
 		defaultFeatures.unshift("compress");


SpaceK33z:
Thank you!

SpaceK33z:
Sorry, I had to revert your fix. It broke historyApiFallback. If you run the example in `examples/history-api-fallback`, you'll see that it breaks after this PR. Feel free to create a new PR.

aves84:
It is my fault, I'm sorry. My example above is incorrect. All that in it will work properly. I have a real file is named differently, and now I found out that the problem is not in the sub-paths, but in the file extensions. 

`http://localhost:8080/userpics/dacd0b4488be967d911c99ae7bc1908444c26c3f ` redirected to `/index.html`, and `http://localhost:8080/userpics/dacd0b4488be967d911c99ae7bc1908444c26c3f .jpg` serving normal.

After moving `historyApiFallback` after `contentBase` I have 
it all worked, files was served and instead non-existent paths was served `/index.html`. It really was a error, but somehow everything worked.

I can assume that the problem is in some module ￼linked with obtain mime type of the file path by extension.

I tested in a folder `examples/history-api-fallback folder`
I did folowing file structure:
```
.
├── app.js
├── file
├── file.txt
├── index.html
├── README.md
└── webpack.config.js
```
Now:
```
/          => index.html
/test      => index.html
/test.html => 404        (error)
/file.txt  => file.txt 
/file      => index.html (error)
```

The correct solution would be to find out where there is url handled, but I could not do it. I suspect that is related to the module **send**.

As temporary solution I can offer all the same, to place `historyApiFallback` after `contentBase` and add the serving static files.
```
app.use(
  historyApiFallback(typeof options.historyApiFallback === "object" ? options.historyApiFallback : null),
  express.static(options.contentBase || process.cwd())
);
```
It fixes serving existing files without extensions, only a non-existent files with extensions are incorrectly processed.
```
/          => index.html
/test      => index.html
/test.html => 404        (error)
/file.txt  => file.txt
/file      => file
```

Sorry for my bad English.

aves84:
I found out reason of problems with paths containing dots, it's just dotRule in **connect-history-api-fallback**, so such paths not relevant.

Did create PR #617

PR #613, opened 2016-09-21
fix for #612

### LINKED ISSUE #612: in-memory resources are not loaded after proxy bypass
When using `proxy`'s `bypass` function, resources that are bypassed and are part of the in-memory webpack bundle are not retrieved. Instead, only files from the filesystem are retrieved.

Example:
Running the server using `--inline`

```javascript
proxy: [{
  context: ['**'],
  target: 'http://someOtherServer:1234/',
  bypass: function(req, res, proxyOptions) {
    if (req.url.indexOf('clientlib-all.js') != -1) {
      return '/bundle.js'; // this file is part of the webpack bundle and should be loaded from memory, but it's not. Instead it returns a 404 _or_ if the file exists on the file system, that one is loaded instead.
    }
    return false;
  }
}];
```

I've fixed this locally by manually editing `webpack-dev-server`'s `Server.js` file:
```javascript
// ... line 269
	var defaultFeatures = ["setup", "headers", "middleware"];
	if(options.proxy)
		defaultFeatures.push("proxy"); // add a second parameter with value "middleware" to this .push call and it works
// ...
```

webpack: tested with 1.13.2, although bug is in server, not in webpack itself
webpack-dev-server: confirmed on **1.16.1** but might also be present in 2.x
OS: OSX 10.11.6, local installation

skwakman:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

See #612 

**What is the new behavior?**

- [ ] Yes
- [X] No (as far as I know)

* Github Issue(s) this is regarding: #612

**Other information**:

added additional `middleware` feature to express' use-chain after proxy to enable proxy-bypass retrievals to in-memory resources. Fixes #612.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -292,7 +292,7 @@ function Server(compiler, options) {
 
 	var defaultFeatures = ["setup", "headers", "middleware"];
 	if(options.proxy)
-		defaultFeatures.push("proxy");
+		defaultFeatures.push("proxy", "middleware");
 	if(options.historyApiFallback)
 		defaultFeatures.push("historyApiFallback", "middleware");
 	defaultFeatures.push("magicHtml");


SpaceK33z:
Thanks.

SpaceK33z:
Whoops, accidentally closed.

PR #617, opened 2016-09-21
historyApiFallback fix

aves84:
fixing historyApiFallback rewrites existed paths without dots
change `example/history-api-fallback`


SpaceK33z:
Is it not possible to fix this using `disableDotRule`?

aves84:
As far I figured out, dotRule now helps serving paths like `/file.ext`, becouse don't allow rewrite them. With `disableDotRule` I suppose such paths will rewrites to `/index.html`

This option will help server unexisted paths like `/test.html` without this option it not will be rewrited, and response will be 404. But in my opinion this is pretty reasonable by default, I don't think specific file is common case.

lib/Server.js
```diff
@@ -235,7 +235,10 @@ function Server(compiler, options) {
 		historyApiFallback: function() {
 			if(options.historyApiFallback) {
 				// Fall back to /index.html if nothing else matches.
-				app.use(historyApiFallback(typeof options.historyApiFallback === "object" ? options.historyApiFallback : null));
+				app.use(historyApiFallback(
```
SpaceK33z: Can you move `historyApiFallback` to the line under it? That would make clear that `express.static` is passed as second param to `app.use`

lib/Server.js
```diff
@@ -235,7 +235,10 @@ function Server(compiler, options) {
...
+					typeof options.historyApiFallback === "object" ? options.historyApiFallback : null),
+					express.static(options.contentBase || process.cwd())
```
SpaceK33z: `options.contentBase || process.cwd()` is also used in the `contentBase` method. Can you put this in a variable, so both use the same variable to get the content base?

examples/history-api-fallback/file
```diff
@@ -0,0 +1 @@
+file
```
SpaceK33z: Could you make this "file.txt", so the browser doesn't want to download the file if you click it?

examples/history-api-fallback/app.js
```diff
@@ -1,2 +1,34 @@
 var path = document.location.pathname;
 document.write("It's working from path <b>" + path + "</b>");
+
+document.addEventListener("DOMContentLoaded", function() {
+	var tests = [
+		{ url: "/", name: "index", re: /^<!DOCTYPE html>/ },
+		{ url: "/test", name: "unexisted path", re: /^<!DOCTYPE html>/ },
```
SpaceK33z: `unexisted` -> `unexisting`

examples/history-api-fallback/app.js
```diff
@@ -1,2 +1,34 @@
...
+		{ url: "/file", name: "existed path", re: /^file/ },
```
SpaceK33z: `existed` -> `existing`

examples/history-api-fallback/file
aves84: I have done it, but now this path is not illustrative, it will work even without fix because active dotRule

lib/Server.js
aves84: it's a typo, sorry

examples/history-api-fallback/app.js
aves84: done

examples/history-api-fallback/app.js
aves84: done

lib/Server.js
aves84: I have made common variable `contentBase` in `function Server` scope, I hope this is ok.

examples/history-api-fallback/file
aves84: I have made common variable `contentBase` in `function Server` scope, I hope this is ok.

aves84: I have done it, but now this path is not illustrative, it will work even without fix because active dotRule

examples/history-api-fallback/app.js
aves84: done

examples/history-api-fallback/app.js
aves84: done

lib/Server.js
aves84: done

lib/Server.js
aves84: I have made common variable `contentBase` in `function Server` scope, I hope this is ok.

examples/history-api-fallback/file
SpaceK33z: Ohh sorry, you can leave it like that then.

aves84: May be better to add
```
devServer: {
		historyApiFallback: {
			disableDotRule: true
		}
	}
```
 to `webpack.config.js`? In this case it will be shown, that rewrite to index.html occurs before trying to serve file.txt.

SpaceK33z: Yeah good idea.

aves84: It is done.

Issue #625, opened 2016-09-27
historyApiFallback breaks with 2.1.0-beta.7 and 2.1.0-beta.6


iam4x:
Before with this configuration the index file was served with `200.html` since I've updated not anymore. Went through the PRs and documentation again I can't see what's wrong.

I'll just stick to beta.5 for now 👍 

```
historyApiFallback: {
  index: '200.html'
}
```

SpaceK33z:
This probably has to do with #617. Where does the `200.html` file come from? Is it served through the `contentBase` option or from a webpack loader / plugin?

SpaceK33z:
Never mind, found the issue and added a failing test.

@aves84, would you have time to look into this? #617 broke this (I don't blame you, it's very difficult without tests).

iam4x:
Thank's you for the fast answer and great investigation spreading some love 💖

aves84:
I think '200.html' is not valid request url, it shoud be like '/200.html'. In this case all works fine.
https://github.com/bripkens/connect-history-api-fallback/issues/9

mac2000:
Got into similar issue, in our case we have few entry points (e.g. multiple apps in same repository)

Our `webpack.config.babel.js` is something like this one:

```js
import path from 'path'
import webpack from 'webpack'

export default (app) => {
	console.log(`>>>>>>>> ${app} <<<<<<<<<<<`)
	return {
		entry: {
			app1: './entry/app1',
			app2: './entry/app2'
		},
		output: {
			path: path.join(__dirname, 'dist'),
			filename: '[name].js',
			publicPath: '/dist/'
		},
		devServer: {
			hot: true,
			//historyApiFallback: true
			historyApiFallback: {
				rewrites: [
					{ from: /.*/, to: `/entry/${app}.html` }
				]
			}
		}
	}
}
```

package scripts is:

```
{
	"start:app1": "webpack-dev-server --env app1",
	"start:app2": "webpack-dev-server --env app2"
}
```

Starting from beta.6 rewrites stop working, the simplest way to reproduce it is even not to use entry points and/or outputs, just create few html files in entry folder and try to run webpack dev server to see what happend - directory index handler will be called

SpaceK33z:
@aves84, you're right in that there should be a prepended slash, but even with it it doesn't work. Also see the tests in `test/HistoryApiFallback.test.js`.

aves84:
[...]

PR #627, opened 2016-09-28
fix for #625 broken historyApiFallback

mac2000:
fixing broken historyApiFallback, with this one - tests are passed and we can use custom index or rewrites property of historyApiFallback

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -279,11 +279,11 @@ function Server(compiler, options) {
 	var defaultFeatures = ["setup", "headers", "middleware"];
 	if(options.proxy)
 		defaultFeatures.push("proxy", "middleware");
+	if(options.historyApiFallback)
+		defaultFeatures.push("historyApiFallback", "middleware");
 	defaultFeatures.push("magicHtml");
 	if(options.contentBase !== false)
 		defaultFeatures.push("contentBase");
-	if(options.historyApiFallback)
-		defaultFeatures.push("historyApiFallback", "middleware");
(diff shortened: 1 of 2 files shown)


SpaceK33z:
Thanks!

aves84:
@SpaceK33z And serving the files without extensions or with `disableDotRule` is broken again. Great job :/

SpaceK33z:
@aves84, this solution doesn't break previous behavior. If you find a solution that doesn't break the old behavior (that is, showing the directory index instead of the page set with historyApiFallback), I'll happily merge a PR.

aves84:
@SpaceK33z in #625 I already wrote this solution - you need to extract `serveIndex` from `contentBase` to independent meddleware.

I don't interested anymore making PR, everyone of which are reverted :/

I will prefer just to switch back to browserSync with normal behaviour.

SpaceK33z:
@aves84, I can understand your frustration. I didn't understand your solution in #625, but now I get it.

The commit cb1b32fa6c57ec3ae9d16be129279df9448cb29f should fix the behavior for both of you. @aves84 and @mac2000, could you verify this? Sorry for the inconvenience.

aves84:
@SpaceK33z, for me it works fine, thank you.

mac2000:
@SpaceK33z @aves84 confirming things are working with this commit, glad that everyone will be happy again!

SpaceK33z:
Great. Thanks for your help!

SpaceK33z:
Fix is released in `2.1.0-beta.8`.

PR #631, opened 2016-09-30
Add support for pfx files as ssl connection options

meltuhamy:
- [x] An example has been added or updated in `examples/` (for features)
- [x] Docs have been added / updated (for bug fixes / features)

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

It is not possible to provide a pfx file as supported by [https.createSearver](https://nodejs.org/api/https.html#https_https_createserver_options_requestlistener) 

**What is the new behavior?**
It will be possible to provide a pfx file, either by passing a commandline argument --pfx and --pfx-passphrase that will pass it on to https.createServer.

- [ ] Yes
- [x] No

* Impact:

**Other information**:
Another PR has been created for webpack 1

### DIFF
--- a/examples/https/README.md
+++ b/examples/https/README.md
@@ -6,6 +6,11 @@ node ../../bin/webpack-dev-server.js --open --https
 
 A fake certificate is used to enable https.
 
+You can provide the following SSL options to override the fake certificate:
+
+* Certificate options e.g. `node ../../bin/webpack-dev-server.js --open --https --cacert=pathToCacert --cert=pathToCert --key=pathToKey`
+* PFX and Passphrase e.g. `node ../../bin/webpack-dev-server.js --open --https --pfx=pathToPfx --pfx-passphrase=PASSPHRASE`
+
 ## What should happen
 
 The script should open `https://localhost:8080/`. Your browser will probably give you a warning about using an invalid certificate. After ignoring this warning, you should see "It's working."
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -302,19 +302,28 @@ function Server(compiler, options) {
 	}, this);
 
 	if(options.https) {
-		// for keep supporting CLI parameters
-		if(typeof options.https === "boolean") {
+		// pfx support
+		if(options.pfx) {
 			options.https = {
-				key: options.key,
-				cert: options.cert,
-				ca: options.ca
+				pfx: options.pfx,
+				passphrase: options.pfxPassphrase
 			};
+		} else {
+			// for keep supporting CLI parameters
+			if(typeof options.https === "boolean") {
+				options.https = {
+					key: options.key,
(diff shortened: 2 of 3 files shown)


lib/Server.js
```diff
@@ -302,19 +302,28 @@ function Server(compiler, options) {
 	}, this);
 
 	if(options.https) {
-		// for keep supporting CLI parameters
-		if(typeof options.https === "boolean") {
+		// pfx support
+		if(options.pfx) {
```
SpaceK33z: Why would this need a new option, instead of it being in the `https` option? It's not used that much, so I would prefer to keep it in the `https` object.

examples/https/README.md
```diff
@@ -6,6 +6,11 @@ node ../../bin/webpack-dev-server.js --open --https
 
 A fake certificate is used to enable https.
 
+You can provide the following SSL options to override the fake certificate:
+
+* Certificate options e.g. `node ../../bin/webpack-dev-server.js --open --https --cacert=pathToCacert --cert=pathToCert --key=pathToKey`
```
SpaceK33z: Could you provide a working example here? The examples are also meant to quickly test if something still works.

lib/Server.js
meltuhamy: What do you mean? I think users might want to provide a --pfx cli option, - this will end up being on the `options`.

examples/https/README.md
meltuhamy: That's now added 👍

lib/Server.js
SpaceK33z: I explained it a bit weird, but I meant that `options.pfx` should work the same and `options.cert`. If you'd do `--cert=mycert.crt --pfx=mypfxthingy`, the `cert` option will be ignored.

meltuhamy: Ah, I now understand what you mean. I wasn't sure that was going to work but I tried it and it worked. I've updated the PR with the change. Thanks

examples/https/README.md
```diff
@@ -6,6 +6,11 @@ node ../../bin/webpack-dev-server.js --open --https
 
 A fake certificate is used to enable https.
 
+You can provide the following SSL options to override the fake certificate:
```
SpaceK33z: Please provide this advanced documentation in [the wiki instead](https://github.com/webpack/docs/wiki/webpack-dev-server/_edit). The README only covers the most simple use case.

SpaceK33z:
Thanks! I'll fix the README change myself, because I want to do a new beta release today ;).

SpaceK33z:
Released in `2.1.0-beta.8`.

meltuhamy:
Thanks @SpaceK33z ! :)

PR #637, opened 2016-10-04
sanity check that conn exists on connection

jskz:
Low-impact bug fix that prevents an edge case where null is passed (and consequently aborts the server, which is the current behaviour) on a new connection if all handles are occupied.

I am able to reliably trigger this by rapidly refreshing on a page in development.


jskz:
The check and return now happen immediately, and the spacing issue is resolved.

SpaceK33z:
Thanks!

Issue #651, opened 2016-10-12
Https without a defined ca file throws error


eldh:
**I'm submitting a bug report**

**webpack and webpack-dev-server version:**
"webpack": "2.1.0-beta.25",
"webpack-dev-server": "2.1.0-beta.8",

**Please tell us about your environment:**
OSX 10.11 Sierra
Running server via CLI

**Current behavior:**

**Expected/desired behavior:**

Hey! I'm having some trouble with the latest webpack-dev-server beta and https. Previously I enabled https through cli `—https —key [key] —cert [cert]`. This now causes `net::ERR_SPDY_PROTOCOL_ERROR`. So I changed to use the config file like this: 
```
https: {
  key: '[key],
  cert: '[cert]',
},
```
This caused the following error: 
```
Error: error:0906D06C:PEM routines:PEM_read_bio:no start line
    at Error (native)
    at Object.createSecureContext (_tls_common.js:68:17)
    at Server (_tls_wrap.js:749:25)
    at Server (https.js:26:14)
    ...
```
From reading the source I would guess I am expected to supply a ca file. Is that necessary? Is there somehting else could be the problem here?

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem along with a gist/jsbin of your webpack configuration. Be sure to specify how you are running the server.** 
Here's my config (slightly simplified): 
https://gist.github.com/eldh/d2dcea7088590afa38f80950a073208d

* **What is the expected behavior?**
I'd expect it to either work like it did in previous versions or provide an error explaining how I need to change my config.

* **What is the motivation / use case for changing the behavior?**
To make it easier to correctly set up https. Https is awesome.

SpaceK33z:
Okay, that should be fixed. A PR is welcome.

eldh:
I'll see what I can do. :)

SpaceK33z:
I tried to reproduce this, but it appears to work. I cloned the webpack-dev-server repo and executed these commands:

```bash
cd examples/https
openssl req -newkey rsa:2048 -x509 -nodes -keyout server.pem -new -out server.pem -subj /CN=localhost -sha256
node ../../bin/webpack-dev-server.js --open --https --key server.pem --cert server.pem
```

This opened the browser, and the page loaded successfully.

Can you provide more information?

SpaceK33z:
friendly ping @eldh

eldh:
[...]

PR #656, opened 2016-10-13
Read files supplied through https config for key, cert and ca

eldh:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Fix for https://github.com/webpack/webpack-dev-server/issues/651

Currently, the options (key, cert, ca) supplied through `https` are treated as if they were the output of `fs.ReadFileSync`.

**What is the new behavior?**
Now the options are assumed to be paths to these files instead, and we read them before sending to spdy.

- [x] Yes
- [ ] No

* Impact:

**Other information**:

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -334,16 +334,15 @@ function Server(compiler, options) {
 		}
 
 		// using built-in self-signed certificate if no certificate was configured
-		options.https.key = options.https.key || fs.readFileSync(path.join(__dirname, "../ssl/server.key"));
-		options.https.cert = options.https.cert || fs.readFileSync(path.join(__dirname, "../ssl/server.crt"));
-		options.https.ca = options.https.ca || fs.readFileSync(path.join(__dirname, "../ssl/ca.pem"));
+		options.https.key = fs.readFileSync(options.https.key || path.join(__dirname, "../ssl/server.key"));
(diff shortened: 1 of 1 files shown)


eldh:
Ok, this is my attempt to fix #651. Don't know if we should suport taking in both a file buffer and string in the config, i don't see why we'd want to do that though. 

Please let me know what I can do to make this ready to merge.

SpaceK33z:
I've marked this as breaking, since it *is breaking* for users that use `fs.readFileSync` themselves.

I thought you were having a new bug, but in previous versions it also didn't work like this, right?

eldh:
That makes sense. I haven't tried this config before, but switched to it since getting the `ERR_SPDY_PROTOCOL_ERROR` with my previous setup (using cli flags). I still get that error after this fix, but adding `{spdy: {protocols: ['http/1.1']}}` works around that issue.

So after thinking some more about this, I don't know if this PR really adds brings much value. I do think it makes more sense to supply a file path rather than a buffer though.

SpaceK33z:
I fully agree with you that it makes more sense to supply a file path instead of buffer, but I don't think the breaking change is worth it. webpack v2 and dev-server v2 already have a huge amount of breaking changes. We could do this in a later major version though.

The `ERR_SPDY_PROTOCOL_ERROR ` you're talking about is pretty serious and is something we should fix in this version. I'm going to test it myself this weekend. Let's continue that discussion in #651.

Issue #659, opened 2016-10-17
Listen to socket


resure:
**I'm submitting a feature request**

**webpack and webpack-dev-server version:**
"webpack": "^2.1.0-beta.25"
"webpack-dev-server": "^2.1.0-beta.9"

It would be really nice to have ability to run webpack-dev-server listening to a socket. There is a StackOverflow answer that suggesting [passing socket as port param](http://stackoverflow.com/questions/33398936/can-webpack-dev-server-run-on-a-unix-domain-sockets) but seems like it doesn't work in the current version.

SpaceK33z:
A PR is welcome.

SpaceK33z:
Fixed by PR #661 :).

PR #661, opened 2016-10-18
Option for listening to unix socket

resure:
- [x] An example has been added or updated in `examples/` (for features)
- [x] Docs have been added / updated (for bug fixes / features)

- [x] Feature

#659

It's possible to pass socket path as port param but it won't correctly handle access rights (chmod) to socket and `EADDRINUSE` errors.

**What is the new behavior?**

New option `--socket` added. It overrides port if present and adds additional handling to listen:

- If connection is successful, then it works like before
- If connection is failing (with `EADDRINUSE` error) then socket file is already present. In this case, we're trying to connect to that socket file.
  - If connection is successful, then socket is already used by some other server and current instance of webpack-dev-server stops
  - If connection failed, then socket is not used by other processes and can be safely unlinked. After that, it starts listen() like usually.
- After listen() successfully started, it sets socket chmod to 666 (read-write for everybody, it necessary since servers like nginx are usually launched from other users like `www-data`).

- [x] No

### DIFF
--- a/examples/listen-socket/README.md
+++ b/examples/listen-socket/README.md
@@ -1,0 +1,43 @@
+# Listen to socket
+
+Start dev server in socket mode:
+
+```shell
+node ../../bin/webpack-dev-server.js --socket ./webpack.sock
+```
+
+And then start node server that will use that socket:
+
+```shell
+node check-socket.js
+```
+
+You should see this responses to this command:
+
+```shell
+Successfully connected to socket, exiting
+```
+
+Common use-case for this feature is configuring upstream in nginx that points to webpack socket.
+Listening to socket instead of port is much more convenient because if you have many applications you don't have to resolve port conflicts since each of them has own unique socket.
+
+Example of configuring upstream with Websockets support in nginx:
+
+```
+location /webpack/ {
+	proxy_pass http://unix:/home/happywebpackuser/apps/todo/webpack.sock;
+	proxy_set_header Host $host;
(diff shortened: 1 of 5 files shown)


examples/listen-socket/README.md
```diff
@@ -0,0 +1,43 @@
+# Listen to socket
+
+Start dev server in socket mode:
+
+```shell
+node ../../bin/webpack-dev-server.js --open --socket ./webpack.sock
```
SpaceK33z: Can you remove `--open` here since it opens the wrong URL in this case?

examples/listen-socket/README.md
```diff
@@ -0,0 +1,43 @@
...
+Common use-case for this feature is configuring upstream in nginx that points to webpack socket.
+Listening to socket instead of port is much more convenient because if you have many applications you don't have to resolve port conflicts since each of them has own unique socket.
+
+Example of configuring upstream with Websockets support in nginx:
+
+```
+location /webpack/ {
+	proxy_pass http://unix:/home/happywebpackuser/apps/todo/webpack.sock;
+	proxy_set_header Host $host;
+    proxy_set_header X-Real-IP $remote_addr;
```
SpaceK33z: Using tabs and spaces here.

resure:
Done, I've appended that changes to the last commit.

SpaceK33z:
Thanks!

PR #669, opened 2016-10-25
customize hot reload, level 1 and level 2

jyz19880823:
**Split the hot reload to two level**
- Level 2: when you change anything, it will recompile and reload the page
- Level 1: just recompile
- If set the options.hot to true, the level will be 2

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

**Other information**:

Sometimes there is no need to reload the page on every change, especially when you are debuging in the developer console.
Set the options.hot to 1, it will just recompile.
If you want to reload, just get http://localhost:9000/refresh, this can be done by emacs shortcut.
If you want to toggle the level, request http://localhost:9000/toggleHot .

```
(defun save-and-refresh-webpack ()
     (interactive)
     (save-buffer)
     (request "http://localhost:9000/refresh"))
(global-set-key (kbd "the_key_you_want") 'save-and-refresh-webpack)
```

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -25,6 +25,9 @@ function Server(compiler, options) {
 	this.sockets = [];
 	this.contentBaseWatchers = [];
 
+    if(this.hot === true){
+        compiler._hot = 2;
+    }
 	// Listening for events
 	var invalidPlugin = function() {
 		this.sockWrite(this.sockets, "invalid");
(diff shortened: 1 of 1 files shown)


SpaceK33z:
Hmm, this is all feels like a hack for a very specific use-case. Is it possible to do this as a separate package? You could use the `setup()` option to access Express. In the current state I can't allow this, as it modifies `compiler`.

SpaceK33z:
I'm going to close this since it doesn't align with our [core ideas](https://github.com/webpack/webpack-dev-server/blob/master/CONTRIBUTING.md#core-ideas):

- "There are hooks to add your own features, so we should not add less-common features."

Sorry for the trouble. Creating an issue before making a PR would prevent this in the future.

If you want to make a package out of this and need more hooks, we can discuss adding those.

Issue #683, opened 2016-11-01
CLI should search for and use available port if none is specified


benwiley4000:
**I'm submitting a feature request**

**Current behavior:**
Two `webpack-dev-server` CLI instances without a specified port will both try to use port 8080, and the second will crash.

**Expected/desired behavior:**
If **no port** has been explicitly specified via config options or command args, then [`portfinder`](https://www.npmjs.com/package/portfinder) should be used to find the first available port after 8080. This behavior is identical to the [`http-server`](https://github.com/indexzero/http-server/blob/master/bin/http-server#L83) CLI. If I specify a port for each dev server, then they should still only try the specified ports.

**What is the motivation / use case for changing the behavior?**
If I haven't specified a port, chances are I just want the server to work, and I'll either pass `--open` or check stdout for the port I need to open in my browser. If I'm testing many webpack dev servers for several projects, I don't want to have to manually pass ports I've determined to be available.

It's already possible to do this via the Node API, but with this feature in the CLI I would have no need for the Node API ([almost 50 lines of JavaScript](https://github.com/benwiley4000/react-gif-player/blob/a00dbb1d8b7f028659b0f75d2ec3e8550deea46b/devServer.js) seems like a bit much for this).

*It's my opinion this behavior should be default, but at the least it should be an option.*

**Question**
I'm unsure whether this behavior should extend to the Node API as well (though I was initially only thinking of targeting the CLI).

*I'm already looking into this, and I'd like to submit a pull request.*

SpaceK33z:
I really like this idea. Do you have time to do a PR soon? If not, I could also fix it (seems like a fun feature to implement).

SpaceK33z:
To answer your question (didn't see that at first), it should only target the CLI.

benwiley4000:
@SpaceK33z thanks. I'll give this a stab in the next day or so.

PR #685, opened 2016-11-03
CLI finds and uses 1st available port when none specified.

benwiley4000:
Addresses #683.

I'm not sure about a new example - is there an easy way to verify that two webpack dev servers are running simultaneously at expected ports? *ETA: I imagine this could be tested via a SharedWorker or the localStorage polyfill that does the same thing - allowing two browser tabs to talk to one another. But a system-level test maybe would be better?*

As for docs, if it's important to document the use of `portfinder` on the wiki I can do so as soon as the PR is merged.

- [ ] Bugfix
- [X] Feature
- [X] Refactoring (no functional changes, no api changes)

Because `portfinder` is async, `startDevServer` had to be spun out into a new function.

- [ ] Build related changes
- [ ] CI related changes

### For details see #683

- [X] Yes
- [X] No

Only in the sense that if two dev servers are run with no port config, one of them will now use a different port (and work). If someone *wanted* their server not to work before, now they might be out of luck! Probably not serious enough for a major version number, you know?

**Note:** There was one situation where a real breaking change could take place but it was [mitigated](https://github.com/webpack/webpack-dev-server/pull/685/commits/33dad8f8ecfd3655c96e76e004030284e2232e60#diff-83c4482ab4d62154c877147880243858R299).

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -5,6 +5,7 @@ var open = require("opn");
 var fs = require("fs");
 var net = require("net");
 var url = require("url");
+var portfinder = require("portfinder");
 
 // Local version replaces global one
 try {
@@ -45,6 +46,11 @@ var SSL_GROUP = "SSL options:";
 var CONNECTION_GROUP = "Connection options:";
 var RESPONSE_GROUP = "Response options:";
 var BASIC_GROUP = "Basic options:";
+
+// Taken out of yargs because we must know if
+// it wasn't given by the user, in which case
+// we should use portfinder.
+var DEFAULT_PORT = 8080;
 
 yargs.options({
 	"lazy": {
@@ -143,7 +149,6 @@ yargs.options({
 	},
 	"port": {
 		describe: "The port",
-		default: 8080,
 		group: CONNECTION_GROUP
 	},
 	"socket": {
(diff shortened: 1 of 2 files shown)


bin/webpack-dev-server.js
```diff
@@ -294,6 +296,25 @@ function processOptions(wpOpt) {
 	if(argv["open"])
 		options.open = true;
 
+	// Kind of weird, but ensures prior behavior isn't broken in cases
+	// that wouldn't throw errors. E.g. both argv port and options port
+	// were specified, but since argv port is 8080, options port will be
+	// used instead.
+	options.port = argv.port === DEFAULT_PORT ? options.port || argv.port : argv.port;
```
SpaceK33z: There seems to be an error here. If I set the port via `options`, after this line it results in `options.port` being `undefined`.

benwiley4000: Aha, thanks for catching that. I didn't account for the case where `options.port` is defined but `argv.port` is not (I forgot since `argv` used to have a default).

What do you think of just changing it to:
```javascript
options.port = argv.port === DEFAULT ? (options.port || argv.port) : (argv.port || options.port)
```
? This should account for all cases I believe.

benwiley4000: It's a bit strange semantically when `argv.port` is *not* `DEFAULT` and therefore *favored* **because** it's `undefined`, but we'll still land on `options.port` in the end.

SpaceK33z: Okay, tested that and it appears to work in all cases :).

benwiley4000: My branch has been updated.

benwiley4000:
I'm sorry in my last update I briefly made a horrible typo. :sob: It is fixed now!

SpaceK33z:
Thanks!

PR #632, opened 2016-09-30
Fix for web worker support

oskarer:
**Please check if the PR fulfills these requirements**
- [ ] An example has been added or updated in `examples/` (for features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?**
- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

**What is the current behavior?** (You can also link to an open issue here)
https://github.com/webpack/webpack-dev-server/issues/628

**What is the new behavior?**
Use `self` instead of `window`.

**Does this PR introduce a breaking change?**
- [ ] Yes
- [X] No

**Other information**:
Should be no difference in functionality, `window` and `self` is essentially the same thing in UI thread. https://developer.mozilla.org/en-US/docs/Web/API/Window/self.

I have setup a repo which replicates the error here https://github.com/oskarer/webpack-dev-server-webworker-test

PR #716, opened 2016-12-18
Make sure we have self.postMessage()

ratson:

bugfix

No

**Summary**

Fix `TypeError` when `self.postMessage` is not a function.

No

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -44,7 +44,7 @@ function log(level, msg) {
 
 // Send messages to the outside, so plugins can consume it.
 function sendMsg(type, data) {
-	if(typeof self !== "undefined") {
+	if(typeof self !== "undefined" && typeof self.postMessage === "function") {
 		self.postMessage({
 			type: "webpack" + type,
 			data: data
@@ -143,7 +143,7 @@ function reloadApp() {
 		log("info", "[WDS] App hot update...");
 		var hotEmitter = require("webpack/hot/emitter");
 		hotEmitter.emit("webpackHotUpdate", currentHash);
-		if(typeof self !== "undefined") {
+		if(typeof self !== "undefined" && typeof self.postMessage === "function") {
 			// broadcast update to window
 			self.postMessage("webpackHotUpdate" + currentHash, "*");
 		}


SpaceK33z:
On what browser or platform is postMessage not available?

ratson:
I am not certain what cause the error, I suspect it is because I serve the script in service worker.
FYI, my webpack config have multiple entries and use react-hot-loader v3.

jamez14:
Also seeing this on Chrome v54.0.2840.98 on Mac OS X 10.11.6. I do not have any hot reloading though. I am only reloading whenever there is a change.

Using webpack-dev-server@^2.2.0-rc.0 and webpack@^2.2.0-rc.1

SpaceK33z:
@jamez14 and does this occur in a service worker?

jamez14:
@SpaceK33z I do not have any service workers with my application. Pretty straightforward application, using Angular & Redux.

jamez14:
Also, after adding this code in this PR in manually to my webpack-dev-server node_module, I see this:

screen shot 2016-12-21 at 10 36 58 am

screen shot 2016-12-21 at 10 40 22 am

even with setting `devServer.hot` to false. Previously I did not even set this, assuming the default was false.

SpaceK33z:
Aha, interesting. Perhaps `self` cannot be used as a replacement for `window` in all cases. From the MDN documentation it seemed that was the case. Could you check if replacing all usages of `self` -> `window` fixes your issue? Also make sure to **not** include the code for this PR when testing (so we know that it fixes this issue for sure).

jamez14:
@SpaceK33z yep, that fixed it. I reverted the code in this PR and just changed all `self`'s to `window`. It reloads as expected with no errors. Thanks for the tip. I can PR this if you'd like, but it might require some more refactoring than just changing self -> window

SpaceK33z:
Thanks for your help. Unfortunately we can't just change `self` -> `window` since that breaks support inside web workers, see #632 for that. We need to find a way to support both... If you're willing to research that and do a PR, that would be very welcome.

SpaceK33z:
I looked a bit more into this, but I can't reproduce any behavior where `self` doesn't have `self.postMessage` or `self.location.reload`. service workers [also have `self.postMessage`](https://developer.mozilla.org/en/docs/Web/API/Worker/postMessage).

@jamez14, if you add `console.log('self', self)` right before `self.location.reload();`, what do you get? It reports a `Window` object for me (which is correct). I have the feeling `self` is overwritten to something else in your config.

jamez14:
@SpaceK33z of course now it is working on my machine (as well as a coworkers). We were both seeing it previously for the past week or so. If I see this issue again, I'll be sure to test this scenario. Thanks for taking the time to investigate!

bebraw:
Time to close?

jamez14:
@bebraw hasn't been an issue since my last comment. So all good from my perspective.

bebraw:
Ok, let's close.

Marksion08:
`index.js?bed3:67 Uncaught TypeError: self.postMessage is not a function`

It was found TILL today~~~

add type checking  **'&& typeof self === 'function''**,  it works!

`
function sendMsg(type, data) {
  if (typeof self !== 'undefined' && (typeof WorkerGlobalScope === 'undefined' || !(self instanceof WorkerGlobalScope)) && typeof self === 'function') {
    self.postMessage({
      type: 'webpack' + type,
      data: data
    }, '*');
  }
}
`

PR #721, opened 2016-12-25
(fix) proxy config with bypass and without target

radotzki:
Bugfix.

No.

**Summary**
When there is a proxy configuration object which includes the bypass property but does not include the target property, the response is never sent back to the browser.
This is a regression from version 1.16.2.

No.


SpaceK33z:
Thanks!

PR #725, opened 2016-12-28
Replace 0.0.0.0 with localhost on open

mstijak:
**Summary**
It's common to set `host: '0.0.0.0'` to enable access to the development server from the network, but that doesn't play well if `open: true` is set. 
Chrome displays an error instead of the application.

The fix is simple. Replace `0.0.0.0` with `localhost` before passing the URL to the browser. 

No

**Other information**
See http://superuser.com/questions/846721/chrome-wont-access-http-0-0-0-04200.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -434,7 +434,7 @@ function reportReadiness(uri, options) {
 	if(options.historyApiFallback)
 		console.log("404s will fallback to " + colorInfo(useColor, options.historyApiFallback.index || "/index.html"));
 	if(options.open)
-		open(uri);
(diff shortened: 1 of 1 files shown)


Korotkin:
npm start -- --host 0.0.0.0

mstijak:
@Korotkin 
This PR fixes the probem if you use `--open` and `--host 0.0.0.0`. `0.0.0.0` is not a valid IP address and Chrome refuses to open it.

`npm start --host 0.0.0.0 --open`

SpaceK33z:
I merged #746, so it's now possible to add `--public localhost` and it will use the correct URL to open the site. It's a bit more config, but you can't always assume that it works correctly for `localhost`.

Issue #736, opened 2017-01-10
[enhancement] Trigger webpack devserver from console using 'rs' like nodemon


ezzaouia:
I guess it would be better to be able to trigger webpack devserver from console for instance using 'rs' like for nodemon.

SpaceK33z:
That could be interesting. Would you be willing to make a PR for this?

ezzaouia:
@SpaceK33z I'll check the code later on, the idea I guess is quite simple below are the need steps in pseudocode..

`
function restartDevServer() {
  stopDevServer();
  startDevServer(the_needed_options);
}

// an http wrapper is needed to be able to close express server
function stopDevServer() {
  httpServerWrapper.close(function () {
    process.exit(0);
  })
}

// need to call this function to attach a listener 
// on typing 'rs' string in the console and restart the server if so
function attachRSEventlistener() {
  process.stdin.setEncoding('utf8');
  process.stdin.on('data', function (data) {
    data = (data + '').trim().toLowerCase();
    if (data === 'rs') return restartDevServer();
  });
}
`

shellscape:
Closing due to inactivity. We'd still welcome a PR for this, however.

PR #740, opened 2017-01-13
handling [enhancement] Trigger webpack devserver from console using 'rs' like nodemon #736

ezzaouia:
Handling [enhancement] Trigger webpack devServer from console using 'rs' like nodemon #736

Some logs:
[at-loader] Ok, 0.616 sec.
rs

	⟳ Restarting devServer..
	------------------------
Project is running at http://localhost:3000/
Content not from webpack is served from /Users/deep/Code/my-starter/src

[at-loader] Checking started in a separate process...

[at-loader] Ok, 0.001 sec.
r
s
rs

	⟳ Restarting devServer..
	------------------------
Project is running at http://localhost:3000/
Content not from webpack is served from /Users/deep/Code/my-starter/src

[at-loader] Checking started in a separate process...

[at-loader] Ok, 0 sec.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -8,8 +8,6 @@ const net = require("net");
 const url = require("url");
 const portfinder = require("portfinder");
 
-const _ = require("lodash");
-
 // Local version replaces global one
 try {
 	const localWebpackDevServer = require.resolve(path.join(process.cwd(), "node_modules", "webpack-dev-server", "bin", "webpack-dev-server.js"));
@@ -386,7 +384,7 @@ function startDevServer(wpOpt, options) {
 function liftDevServer(compiler, options, uri) {
 	let server;
 	try {
-		server = new Server(compiler, _.cloneDeep(options));
+		server = new Server(compiler, Object.assign({}, options));
 	} catch(e) {
 		const OptionsValidationError = require("../lib/OptionsValidationError");
 		if(e instanceof OptionsValidationError) {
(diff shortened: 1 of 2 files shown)


bin/webpack-dev-server.js
```diff
@@ -372,11 +378,15 @@ function startDevServer(wpOpt, options) {
...
+	uri = domain + (options.inline !== false || options.lazy === true ? "/" : "/webpack-dev-server/");
+
+	liftDevServer(compiler, options, uri);
+}
 
+function liftDevServer(compiler, options, uri) {
 	let server;
 	try {
-		server = new Server(compiler, options);
+		server = new Server(compiler, _.cloneDeep(options));
```
SpaceK33z: Perhaps we could use `Object.assign` instead of pulling in the whole lodash dependency? I don't want to use lodash for this one use.

ezzaouia: I see, well I'll change..

ezzaouia: okay done !

SpaceK33z:
Note that I haven't forgotten this, but this costs a lot of time to properly check if it works and review.

Have you checked that changing something in the webpack config actually changes something? Because (looking purely at the code), there might be some issues with the `require` cache.

SpaceK33z:
I tested it out, but it seems like it doesn't work properly when you change something in the webpack config. I tested it by starting the example in `examples/modus-inline`. After starting the server, I removed all loaders in the webpack config. Then I pressed `rs` followed by an enter. This restarted the server, but it still compiled everything properly. I expected it to fail since I removed the loaders.

ezzaouia:
@SpaceK33z thank you so much for your review and test; humm there maybe a way to reinitialize the cache, I'm so sorry I can't check now, I I'll check this weekend and let you know once done.

Many thanks and please have a nice day/evening... ;)

shellscape:
@ezzaouia ping

shellscape:
closing because stale. please do ping if you have the chance to circle back to this

Issue #739, opened 2017-01-13
Enhancement request: Custom url for --open flag


bdwain:
Would it be possible to add the ability to specify a custom url to open in the browser with the --open flag? 

webpack-dev-server --host 0.0.0.0 --open somealias.com

i sometimes use an alias for localhost to communicate what environment i'd like my app to run in.

SpaceK33z:
Have you tried using `--public somealias.com`? Not sure if it fits your use-case.

bdwain:
i think that just affects the code that webpack injects into the page to talk to the dev server. I was looking for a way to open somealias.come in the browser when the page opens, but i'd still like to have the host be 0.0.0.0 so other computers on my network can access my dev server.

SpaceK33z:
Okay, fair enough. The PR #725 attempts to fix this, but only for `localhost`. Could you perhaps file a PR to do it with the `--public` flag? I'll close #725 then.

bdwain:
sure i can do that

shellscape:
It appears that #749 fixed this issue, so closing. please ping me if that's not the case and we'll reopen the discussion.

PR #747, opened 2017-01-18
have --open use the options.public path if it exists

bdwain:
This changes the behavior of the --open flag to open the browser to the --public url if it is specified. (the default will be the same).

This is also addressed for version 1 with #746


bdwain:
had an issue with the CLA so i closed this and opened a new one

PR #748, opened 2017-01-18
have --open use the options.public path if it exists (version 1 fix)

bdwain:
This changes the behavior of the --open flag to open the browser to the --public url if it is specified. (the default will be the host like before).

This fixes the behavior for version 1. Master is fixed in #749

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -18,48 +18,48 @@ var webpack = require("webpack");
 
 var optimist = require("optimist")
 
-.usage("webpack-dev-server " + require("../package.json").version + "\n" +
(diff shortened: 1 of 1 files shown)


SpaceK33z:
webpack 2 will be released next week, so we will not release new features for 1.x

bdwain:
i know that version 1 is not under active development, but since v2 is not technically stable yet and a bunch of people asked for this feature and it's small and already done, would you reconsider putting it in v1? It's not always easy to upgrade to v2 right away and this would make v1 better.

bdwain:
@SpaceK33z any chance you could reconsider based on what I said above?

PR #749, opened 2017-01-18
have --open use the options.public path if it exists

bdwain:
This changes the behavior of the --open flag to open the browser to the --public url if it is specified. (the default will be the host like before).

This is also addressed for version 1 with #748

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -330,14 +330,14 @@ function startDevServer(wpOpt, options) {
 	const protocol = options.https ? "https" : "http";
 
 	// the formatted domain (url without path) of the webpack server
-	const domain = url.format({
+	const domain = options.public ? `${protocol}://${options.public}` : url.format({
 		protocol: protocol,
 		hostname: options.host,
 		port: options.socket ? 0 : options.port.toString()
 	});
 
 	if(options.inline !== false) {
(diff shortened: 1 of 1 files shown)


bin/webpack-dev-server.js
```diff
@@ -330,14 +330,14 @@ function startDevServer(wpOpt, options) {
...
 	// the formatted domain (url without path) of the webpack server
-	const domain = url.format({
+	const domain = options.public ? `${protocol}://${options.public}` : url.format({
 		protocol: protocol,
 		hostname: options.host,
 		port: options.socket ? 0 : options.port.toString()
 	});
 
 	if(options.inline !== false) {
-		const devClient = [`${require.resolve("../client/")}?${options.public ? `${protocol}://${options.public}` : domain}`];
```
bdwain: this line behaves the same, but domain is also used on line 375 to calculate the url that gets opened, and that's where the behavior changes

bdwain:
@SpaceK33z Is there anything else you wanted done for this?

SpaceK33z:
@bdwain, I am planning to merge this after the 2.2 release (2.2 has been published but not yet with the `latest` tag). So it's fine!

bdwain:
cool!

SpaceK33z:
Thanks!

PR #764, opened 2017-01-28
Implement error overlay

SpaceK33z:
Feature

Yes

**Summary**
This adds an error overlay in the browser, like create-react-app has. It works very much like create-react-app's error overlay. In fact, I copied much code from it 😅 .

I asked on Twitter if people would [like this feature](https://twitter.com/keeskluskens/status/825008478536228865), and ~~everyone forgot Trump for a while~~ people were happy.

The error overlay is only enabled when you add the option `overlay: true` in your `devServer` config. It could be enabled by default, but that would kinda be a breaking change so *maybe* in v3.

Furthermore, it will only show errors, not warnings. I'm not sure yet if we should display warnings the same way. If we would do that, there also needs to be a way to disable it.

Nope

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -2,6 +2,7 @@
 var url = require("url");
 var stripAnsi = require("strip-ansi");
 var socket = require("./socket");
+var overlay = require("./overlay");
 
 function getCurrentScriptSource() {
 	// `document.currentScript` is the most accurate way to find the current script,
@@ -32,6 +33,7 @@ var hot = false;
 var initial = true;
 var currentHash = "";
 var logLevel = "info";
+var useOverlay = false;
 
 function log(level, msg) {
 	if(logLevel === "info" && level === "info")
@@ -71,8 +73,14 @@ var onSocketMsg = {
 	"log-level": function(level) {
 		logLevel = level;
 	},
+	"overlay": function(overlay) {
+		if(typeof document !== "undefined") {
+			useOverlay = overlay;
+		}
+	},
 	ok: function() {
 		sendMsg("Ok");
+		if(useOverlay) overlay.clear();
 		if(initial) return initial = false;
 		reloadApp();
 	},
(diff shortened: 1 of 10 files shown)


SpaceK33z:
If anybody would want to test this, that would be appreciated. [Info on how to test a PR](https://github.com/webpack/webpack-dev-server/blob/master/CONTRIBUTING.md#testing-a-pull-request).

STRML:
A few comments:

1. `overlay` option works fine. :+1:
1. I'm not able to get the actual syntax error location from something like `<div></span>` - I just get the stack with `Module build failed`. Using HMR and `devtool: "eval"`. Is there further setup required?

Restuta:
I am not using "webpack-dev-server" directly, I am using "webpack-dev-middleware", is there a way to have it as part of the middleware itself? @SpaceK33z

karl:
Just tested this in our codebase at work and it behaved correctly as far as I could see. We would find this really useful for making it clear when compilation had failed.

We'd definitely use this if it was merged in!

SpaceK33z:
@STRML, I just tested this in a project of mine with React and JSX. It looks like this when I change a `</div>` to `</span>`:

screen shot 2017-02-03 at 11 38 26

It didn't need any extra configuration. This was with `devtool: "eval"`. I haven't setup HMR though (don't have the time right now to test that out).

@Restuta, no, webpack-dev-middleware can't communicate to the webpage since it's only middleware. [webpack-hot-middleware](https://github.com/glenjamin/webpack-hot-middleware) does offer this functionality btw.

@karl, thanks for testing!

Restuta:
@SpaceK33z I meant hot-middleware, sorry

KutnerUri:
@SpaceK33z - is this still live? does it handle runtime errors?

I'm getting this less "sexy" overlay
Screen Shot 2022-02-22 at 17 51 46

PR #780, opened 2017-02-05
Handle `opn` promise rejection

alistairjcbrown:

bugfix

No

**Summary**

The dev server uses `opn` to open the browser, which returns a promise. In headless environments, this call will fail and the promise will be rejected. Since node v6.6.0, a `UnhandledPromiseRejectionWarning` will be output when a rejected promise has no handler.

This PR adds a `.catch` handler and outputs a user friendly message that the browser could not be opened, and not to use the open flag in headless environments.

No

**Other information**

None


SpaceK33z:
Thanks!

PR #782, opened 2017-02-07
Export logic of adding entry points to webpack config

danez:

This is a refactoring and adds a new feature.

No

**Summary**

It adds a new static method `addDevServerEntrypoints()` to the exports. This method could be reused in alternative implementations like the grunt plugin instead of[ copying the code](https://github.com/webpack-contrib/grunt-webpack/blob/master/tasks/webpack-dev-server.js#L43-L60).

no

**Other information**

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -5,8 +5,9 @@ const path = require("path");
 const open = require("opn");
 const fs = require("fs");
 const net = require("net");
-const url = require("url");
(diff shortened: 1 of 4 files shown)


bebraw:
The refactoring looks like a good idea to me so 👍 .

SpaceK33z:
Thanks, I like the separating into more modules, that'll make it easier to write tests for it. Sorry for the delay, it was a very busy week at work.

PR #787, opened 2017-02-12
Explicitely but gracefully handle SIGINT and SIGTERM signals.

### LINKED ISSUE #557: Graceful way to stop server.js through CLI/Docker?
I'm using webpack-dev-server under a Docker container, and it's default command is simply `node server.js`, which is sampled below. However, when I ask Docker to stop the container, it timeouts in 10s, meaning it's probably not being able to terminate the node process.

What would be the best way to implement some sort of `npm stop` command to gracefully shutdown webpack-dev-server?

-------

*server.js*
```js
'use strict'
require('core-js/fn/object/assign')
const webpack = require('webpack')
const WebpackDevServer = require('webpack-dev-server')
const config = require('./webpack.config')

new WebpackDevServer(webpack(config), config.devServer)
.listen(config.port, null, (err) => {
  if (err) {
    console.log(err)
  }
  console.log('Listening at localhost:' + config.port)
})
```

lucsky:
Fixes #557.

As specified in the Docker documentation, `a process running as PID 1 inside a container is treated specially by Linux: it ignores any signal with the default action. So, the process will not terminate on SIGINT or SIGTERM unless it is coded to do so.` which is the reason why a `webpack-dev-server` run from a Docker container cannot be stopped which forces Docker to kill it after the standard 10s timeout. Explicitely handling the SIGTERM and SIGINT signals fixes this and allow much faster graceful shutdowns.


bebraw:
Good idea. 👍

DawidJanczak:
Is there anything blocking this?

bebraw:
No blockers. Just waiting a merge by @SpaceK33z.

jaridmargolin:
Is the `server.close()` here necessary?

PR #790, opened 2017-02-16
feat - Allow overlay to be enabled for warnings (#789)

### LINKED ISSUE #789: Show warnings in client overlay
Feature

Compiler warnings will not be shown in the overlay and there is no option to turn that on.

Compiler warnings should be shown in the overlay, but not by default. As discussed with @bebraw, it could look like this:

```js
overlay: {
  errors: true,
  warnings: true
}
```

`overlay: true` would be the same as `overlay: { errors: true, warnings: false }`.

**If this is a feature request, what is motivation or use case for changing the behavior?**

Some people like to fix the warnings first before seeing the result, now you need to look into the DevTools or the terminal to see those warnings. Also, CRA shows them too.

bebraw:

Feature.

Yes.

**Summary**

Now you can use

```javascript
{
  overlay: {
    errors: true,
    warnings: true
  }
}
```

Closes #789.

No.

### DIFF
--- a/examples/overlay-warnings/README.md
+++ b/examples/overlay-warnings/README.md
@@ -1,0 +1,11 @@
+# Overlay
+
+```shell
+node ../../bin/webpack-dev-server.js --open
+```
+
+## What should happen
+
+The script should open the browser and show a heading with "Example: overlay with warnings".
+
+In `app.js`, make a syntax error. The page should now refresh and show a full screen error overlay that shows the syntax error.
--- a/examples/overlay-warnings/README.md
+++ b/examples/overlay-warnings/README.md
@@ -8,4 +8,4 @@ node ../../bin/webpack-dev-server.js --open
 
 The script should open the browser and show a heading with "Example: overlay with warnings".
 
-In `app.js`, make a syntax error. The page should now refresh and show a full screen error overlay that shows the syntax error.
+In `app.js`, uncomment the lines that should cause a warning to appear. The page should now refresh and show a full screen overlay that shows the warning.
--- a/client/index.js
+++ b/client/index.js
@@ -33,7 +33,8 @@ var hot = false;
 var initial = true;
 var currentHash = "";
 var logLevel = "info";
-var useOverlay = false;
+var useWarningOverlay = false;
+var useErrorOverlay = false;
 
 function log(level, msg) {
 	if(logLevel === "info" && level === "info")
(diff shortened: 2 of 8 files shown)


examples/overlay-warnings/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+```shell
+node ../../bin/webpack-dev-server.js --open
+```
+
+## What should happen
+
+The script should open the browser and show a heading with "Example: overlay with warnings".
+
+In `app.js`, make a syntax error. The page should now refresh and show a full screen error overlay that shows the syntax error.
```
SpaceK33z: Could you document how to trigger a warning here? This text only tells me how to trigger an error. Would be handy for testing, since I can't come up with something that triggers a warning out the top of my head right now :).

bebraw: Yeah, the problem is, how to trigger a warning.

I tried copying

```javascript
// This results in a warning:
// if(!window) require("./" + window + "parseable.js");
```

to the example from **modus-inline**, but that doesn't trigger a warning at all (not there either). Maybe something changed between webpack 1 and 2?

SpaceK33z: Jup, in webpack 1 that triggered a warning but in webpack 2 not anymore :'). Looking at webpack's source code on `warnings.push`, I found that it should trigger a warning when incorrectly cased modules are used. So perhaps add a file `caseSensitive.js` and import it with `casesensitive.js`?

bebraw: The problem is that those two files resolve fine on macOS. It would fail only on Windows and so on.

SpaceK33z: Yes, but isn't the purpose of the `CaseSensitiveModulesWarning` plugin to give a warning for that, exactly because it works on macOS but not on Linux/Windows? (I seriously don't know btw, I only looked at the code very quickly)

bebraw: The only obvious way I can see would be to go through eslint and eslint-loader. That would work regardless of the platform.

bebraw: Ah, yeah. I see now. I think I have to add the plugin to the configuration as it's not enabled by default.

bebraw: Got it. You have to do

```javascript
require("./casesensitive.js");
require("./caseSensitive.js");
```

It requires multiple, different imports pointing to the same place in order to work. This is enough for me. 👍

bebraw: Done.

PR #802, opened 2017-02-21
devClient does not work with dynamic entry.

### LINKED ISSUE #801: Can't apply devClient to  `configuration.entry` options when entry was a function
Hi, I use dynamic entry for my webpackConfig:

```js
// webpack.config.js
module.exports = {
  entry: () => {
    return { app: 'index.js' }
  }
  //...
}
```

when I use webpack-dev-server start app, It's tell me configuration.entry does not match the schema:

```sh
 - configuration.entry should be one of these:
   object { <key>: non-empty string | [non-empty string] } | non-empty string | [non-empty string] | function
   The entry point(s) of the compilation.
   Details:
    * configuration.entry should be an object.
    * configuration.entry should be a string.
    * configuration.entry[2] should be a string.
    * configuration.entry should be an instance of function
      function returning an entry object or a promise..
```

I think it maybe not works with the function type entry point. [./lib/util/addDevServerEntrypoints.js#L20
](https://github.com/webpack/webpack-dev-server/blob/master/lib/util/addDevServerEntrypoints.js#L20)

Anything help ❤️

yuffiy:
fix #801

### DIFF
--- a/lib/util/addDevServerEntrypoints.js
+++ b/lib/util/addDevServerEntrypoints.js
@@ -16,7 +16,7 @@ module.exports = function addDevServerEntrypoints(webpackOptions, devServerOptio
 				Object.keys(wpOpt.entry).forEach(function(key) {
 					wpOpt.entry[key] = devClient.concat(wpOpt.entry[key]);
 				});
-			} else if(typeof wpOpt.entry === 'function') {
+			} else if(typeof wpOpt.entry === "function") {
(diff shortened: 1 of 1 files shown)


SpaceK33z:
@yuffiy could you sign the CLA?

yuffiy:
@SpaceK33z 
Oh, I forget, thanks 😘

SpaceK33z:
Thank you!

Issue #792, opened 2017-02-17
self.postMessage is not a function under webpack-dev-server


didi0613:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**
<img width="520" alt="screen shot 2017-02-16 at 6 27 48 pm" src="https://cloud.githubusercontent.com/assets/10135897/23050226/d1477524-f475-11e6-9264-f494500d244d.png">
<img width="1031" alt="screen shot 2017-02-16 at 6 27 56 pm" src="https://cloud.githubusercontent.com/assets/10135897/23050229/d4b86e5c-f475-11e6-9183-2073a45b443c.png">
Self is referring to a wrong object `serviceworkerglobalscope` instead of `window`.
PostMessage is not defined under `serviceworkerglobalscope`

**If the current behavior is a bug, please provide the steps to reproduce.**

Clone https://github.com/didi0613/electrode/tree/webpack-2.0
checkout branch webpack-2.0
npm i && npm run bootstrap
goto samples/universal-react-node app
npm install
gulp dev

**What is the expected behavior?**
There should be no error and pwa should work

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention your webpack and Operating System version.**
webpack 2.2.0
wepack-dev-server 2.2.0
os: mac

dangerousdan:
Related to #745

hccde:
I've found out how this bug is happening,it's not easy,this is because we use `iframe.postMessage` to refresh.But unfortunately, in the `Worker` there is also a method called `postMessage` ,that's ` self.postMessage ` .When ' worker.js'  as an entry in 'webpack.config.js',usually we have to do that,'worker.js' will be packaged with 'webpack-dev-server/client',As previously mentioned，we use 'iframe.postMessage' to refresh，in code,that's:
`                   if(typeof self !== "undefined") {
			// broadcast update to window
			self.postMessage("webpackHotUpdate" + currentHash, "*");
		     }`
'worker.js'  which is packaged with 'webpack-dev-server/client' also contains this code, pay attention, at this time,  `self` is not `window`  but `[DedicatedWorkerGlobalScope](https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage)`,this is the key point. 'webworker' is in a special context  different from 'window'.so we need to distinguish between these two objects.And i have initiated a 'pr' to fix this. #813

shellscape:
[...]

Issue #745, opened 2017-01-17
Second argument of postMessage is not a transfert list


cecilemuller:
**Do you want to request a *feature* or report a *bug*?**

A bug.

**What is the current behavior?**

There is an error (only in webpack-dev-server, not in regular webpack builds) in the console when there is a target "webworker".

Chrome error:
> Uncaught TypeError: Failed to execute 'postMessage' on 'DedicatedWorkerGlobalScope': The 2nd argument is neither an array, nor does it have indexed properties.

Firefox error:
> TypeError: Argument 2 of DedicatedWorkerGlobalScope.postMessage can't be converted to a sequence.

**If the current behavior is a bug, please provide the steps to reproduce.**

Testcase to reproduce it: https://github.com/wildpeaks/issue-webpack-dev-server-postmessage

They seem to point to [this line](https://github.com/webpack/webpack-dev-server/blob/v2.2.0-rc.0/client/index.js#L51) as it appears `"*"` is not a valid [transfert list](https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage).

**What is the expected behavior?**

The expected behavior is no error shown in the console.

**Please mention your webpack and Operating System version.**

 - Node 7.4.0
 - Windows 10
 - Webpack 2.2.0-rc.7
 - Webpack-dev-server 2.2.0-rc.0

SpaceK33z:
Thanks for your bugreport and repo. Would you be interested in making a PR to fix this? It's perhaps useful to add your repo to our `examples/` folder (with a few modifications to make it work like the other examples).

jbms:
I'm actually unclear on what is intended from that code.  What is the purpose of sendMsg?  Is it intended to have an effect only within a webworker, specifically to send a message to the main thread that created it?  That seems like a bad idea, since it will confuse application code that is using that communication channel.

cecilemuller:
Sorry I missed your reply @SpaceK33z and sure, I will prepare a PR to add to the examples.

As for solving the issue however, I'm not familiar with the internals of dev server itself, so I'm not yet sure what side-effects changing it will have.

jakobkylberg:
We ran into the same issue in our application. When we run the application using webpack-dev-server we see the issue and when using another server we have no issues.
 
The example given by cecilemuller describes well what we were trying to do, we have a similar simple web worker in our application. 

We are using: 
webpack-dev-server 2.2.1
webpack 2.2.1

It would be great to find a solution to this problem.

[...]

PR #813, opened 2017-02-24
a bugfix for issue #745 ,webworker error

hccde:
 a bugfix

now,it can  work well without errors.

**Summary**
this bug happened because we use `iframe.postMessage` to communicate with other window.But unfortunately, in the `Worker` there is also a method called `postMessage`,they have diffrent function signature.
#745  #792 

**Other information**
it's a not easy journey


SpaceK33z:
Thanks!

snkhubcom:
Thank you!

PR #827, opened 2017-03-05
Update `http-proxy-middleware` to 0.17.4

EladBezalel:
Updating http-proxy-middleware to it's newest version - [solves many issues](https://github.com/chimurai/http-proxy-middleware/blob/master/CHANGELOG.md)


SpaceK33z:
Thanks!

PR #832, opened 2017-03-07
Support for array of contentBase

Telokis:
Updated the type checking of `contentBase` for `webpack-dev-server.js` so that it can handle multiple `content-base` arguments.

Feature

No. This is minor so I don't think an example is needed.

**Summary**
I made this change because I need it on my personal project. Everything was ready but there was just a missing check so I added it.

No breaking change at all.

**Other information**


SpaceK33z:
Thanks! Tested this by running `webpack-dev-server --content-base first --content-base second`.

PR #835, opened 2017-03-08
Use arrow function if it possible and get rid of .bind in server part

Connormiha:
Refactoring

**Summary**
Current `webpack-dev-server` requires `Node 4+`. So we can feel free and get rid of `bind` method and use arrow functions in server part.

No


SpaceK33z:
Thanks!

PR #836, opened 2017-03-09
Fix to share proxy option between proxy settings when the proxy option is a same object

koba04:

a bugfix

No

**Summary**

Currently, `webpack-dev-server` mutates the proxy options, which is a cause of bug when the proxy options share a same option object.

For an instance, the following option creates different results after this PR.

* A proxy option

```js
const proxyTarget = {
    target: "http://localhost:9000"
};
proxy: {
    "/proxy1": proxyTarget,
    "/proxy2": proxyTarget,
}
```

* Before this PR

```
[ { target: 'http://localhost:9000',
    context: '/proxy2',
    logLevel: 'warn' },
  { target: 'http://localhost:9000',
    context: '/proxy2',
    logLevel: 'warn' } ]
```

* After this PR

```
[ { target: 'http://localhost:9000',
    context: '/proxy1',
    logLevel: 'warn' },
  { target: 'http://localhost:9000',
    context: '/proxy2',
    logLevel: 'warn' } ]
```

I think it should not override the `context` value.

Maybe not. it is a bug fix.

**Other information**

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -159,7 +159,7 @@ function Server(compiler, options) {
 								target: options.proxy[context]
 							};
 						} else {
-							proxyOptions = options.proxy[context];
+							proxyOptions = Object.assign({}, options.proxy[context]);
 							proxyOptions.context = correctedContext;
 						}
 						proxyOptions.logLevel = proxyOptions.logLevel || "warn";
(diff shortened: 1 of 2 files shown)


SpaceK33z:
Thanks again!!

koba04:
Thank you!

PR #841, opened 2017-03-13
Added beforeunload check to index.js (#544)

### LINKED ISSUE #544: [WDS] disconnected! constantly appear on windows 8
One of our team member's system is win8, when he run webpack-dev-server everything seems worked. But every few seconds a `[WDS] disconnected!` error shows in console and page then refresh.

In the meantime, this issue don't appear on win7.

I had tried ways mentioned in [#326](https://github.com/webpack/webpack-dev-server/issues/326), still can't solve the problem.

Here is our webpack.config.js

```
var webpack = require("webpack");
var path = require("path");
var HtmlWebpackPlugin = require('html-webpack-plugin');

var config = {
    entry: {
        "vendor1": ["react", "react-dom", "react-router", "react-router-redux", "react-redux", "redux"],
        "vendor2": ["crypto-js", "fastclick"],
        "app": "./js/index.js"
    },

    output: {
        path: './build/',
        filename: 'js/[name].js',
        publicPath: '/build/',
        chunkFilename: "js/[id].bundle.js"
    },

    module: {
        loaders: [
            {
                test: /\.js$/, 
                include: [
                    path.resolve(__dirname, "js")
                ],
                exclude: [
                    path.resolve(__dirname, "build"),
                    path.resolve(__dirname, "node_modules")
                ],
                loader: 'babel-loader',
                query: {
                    presets: ['es2015', 'react'],
                    plugins: ['transform-object-assign']
                }
            },

            {
                test: /\.css$/,
                loaders: ["style", "css?-minimize"]
            },

            {
        test: /\.(png|jpg|woff|woff2|eot|ttf|svg)(\?v=[0-9]\.[0-9]\.[0-9])?$/,
        loader: 'url?limit=512&&name=[path][name].[ext]?[hash]'
      }
        ]
    },

    plugins: [
        new webpack.optimize.CommonsChunkPlugin({
            names: ["vendor2","vendor1"],
          minChunks: Infinity
        })
    ],

    resolve: {
        modulesDirectories: [
      'node_modules'
    ]
    }
};
module.exports = config;
```

samsch:
Bugfix
No
**Summary**
This fixes #544. In Firefox on Linux and Mac, with a slow proxied server (> 1s), when the page is refreshed (or the server refreshes is for a code change), the web socket reconnection timeout triggers, which causes the webpage to reload (again), causing a loop. This code adds a `beforeunload` event check, so that the reload doesn't happen if the page is already refreshing.
I wouldn't not expect this to cause a BR break, but I'm not super familiar with webpack-dev-server's official API.
**Other information**
This is a naive implementation that just fixes the problem for my limited test case. For example, this would throw in browsers where `addEventListener` is undefined (this probably isn't a problem in 2017, but gives an idea of what I mean).

A test case which should pass with this pull can be found here: https://github.com/samsch/reload-loop-test-webpack-dev-server

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -157,7 +157,15 @@ var socketUrl = url.format({
 
 socket(socketUrl, onSocketMsg);
 
+var isUnloading = false;
+self.addEventListener("beforeunload", function() {
+	isUnloading = true;
+});
+
 function reloadApp() {
+	if(isUnloading) {
+		return;
+	}
 	if(hot) {
 		log("info", "[WDS] App hot update...");
 		var hotEmitter = require("webpack/hot/emitter");


SpaceK33z:
Thanks. Nobody from #544 has tested this yet, but I want to cut a new release and after testing this a bit it doesn't seem to break anything at least.

sarora2073:
i'm on wds v 2.4.1 and still experience this issue sporadically. i'm going to upgrade to 2.4.2 (assuming that has the fix) and will let you know in a few days if it appears to have gone away "for good". It's not reproducible at will for me, hence the need to watch over a few days.

samsch:
@sarora2073 If it's the same issue at fault, then I would suspect that it's sporadic because your server response time varies across the timeout threshold.

sarora2073:
update: this symptom of WDS going into a reload loop appears to have gone away for me after upgrading WDS to 2.4.2. I frequently trigger multiple updates during development (i frequently save one open code file at a time in quick succession, resulting in several interrupted compilation attempts, sometimes leading to the loop). I happen to be using Windows 10 / Chrome, which is different from the other reported environments, so am guessing this affects multiple browsers / OSes, given the related threads / reports of this bug.

Issue #854, opened 2017-03-20
Self signed certificate no longer valid as of Chrome 58


PaulTondeur:
It turns out that the Subject Alt Name property is missing in the certificate, resulting in Chrome that marks the certificate as insecure. The error in chrome is misleading (net::ERR_CERT_COMMON_NAME_INVALID), though according to [this post](https://productforums.google.com/forum/#!topic/chrome/zVo3M8CgKzQ;context-place=topicsearchin/chrome/category$3ACanary%7Csort:relevance%7Cspell:false) it is about the Subject Alternative Name (SAN) that is missing.

johnboxall:
Based on https://github.com/webpack/webpack-dev-server/commit/e97741c84ca69913283ae5d48cc3f4e0cf8334e3 and https://security.stackexchange.com/a/91556 I believe this could be regenerated with something like:

```sh
openssl req \
    -newkey rsa:2048 \
    -x509 \
    -nodes \
    -keyout server.pem \
    -new \
    -out server.pem \
    -subj /CN=localhost \
    -reqexts SAN \
    -config <(cat /System/Library/OpenSSL/openssl.cnf \
        <(printf '[SAN]\nsubjectAltName=DNS:localhost')) \
    -sha256 \
    -days 3650
```

On OSX.

ream88:
Apparently this fix works also for Chrome 57. Thx @johnboxall!

jesstelford:
I also had to set `-extensions SAN` to get @johnboxall's command to work:

```bash
openssl req \
    -newkey rsa:2048 \
    -x509 \
    -nodes \
    -keyout server.pem \
    -new \
    -out server.pem \
    -subj /CN=localhost \
    -reqexts SAN \
    -extensions SAN \
    -config <(cat /System/Library/OpenSSL/openssl.cnf \
        <(printf '[SAN]\nsubjectAltName=DNS:localhost')) \
    -sha256 \
    -days 3650
```

stephanvierkant:
I've used that command, but still getting ERR_CERT_AUTHORITY_INVALID on Chrome 58. Adding it to Chrome doesn't work because of "Not a Certification Authority" error message.

Any idea how to fix this?

ianfitzpatrick:
@stephanvierkant 

I don't know what platform you are on, but assuming you are on OSX, you need to follow these instructions to manually trust self signed certificates.

https://github.com/webpack/webpack-dev-server/issues/854

[...]

PR #857, opened 2017-03-21
Add subjectAltName to development cert

johnboxall:
Resolves https://github.com/webpack/webpack-dev-server/issues/854

Regenerates the included development cert with alt names for both `localhost` and `127.0.0.1` and adds a script to recreate the cert with the configuration used to create what is committed.

**Summary**

In Chrome 58, support for commonName matching in certificates was dropped:

https://www.chromestatus.com/features/4981025180483584

On OSX, for folks who have added the existing committed cert to their keychain, they'll now see issues as detailed here:

https://textslashplain.com/2017/03/10/chrome-deprecates-subject-cn-matching/

Constantly clicking through this warning is old, so we'll want to add the SAN to the dev cert.

To verify the existing behaviour, try adding the current cert to your Keychain (Always Trust!) and browse to a running Webpack server configured for HTTPS. Note that you get an error :|

Now switch over to this version, delete the old cert from your Keychain, add the new one and try again. It works!

Because the cert is updated, folks who have used the Keychain work around will now need to remove the old entry and add the new one.

image

image
pasted image at 2017_03_20 06_48 pm
pasted image at 2017_03_20 06_46 pm

If you want to verify the values in the cert you can paste it into something like https://certificatetools.com/ or just serve it locally yourself and verify in your browser.


SpaceK33z:
Just a note that I haven't forgotten this, I'm very busy these days but I hope I have some time in the weekend to review.

thrandre:
Friendly bump. Chrome 58 started rolling out in Norway this morning.

johnboxall:
If we wanted to extend this to Windows – we could generate the `.p7b` file format that Windows likes from the `.pem` using:

```bash
openssl crl2pkcs7 \
    -nocrl \
    -certfile server.pem
    -out server.p7b
```

sidshetye:
@shellscape : what's the final solution ? Seems this is closed and #942 is in limbo.

shellscape:
@siddhadev use `devcert`. I'm not longer a maintainer on this project.

PR #867, opened 2017-04-02
Use `PORT` environment variable

jgierer12:
Feature

**Summary**
Some tools, e.g. [hotel](https://github.com/typicode/hotel), use an environment variable called `PORT` to store what port they will use. We should use it instead of the default port (`8080`) if it is set.

No; if `PORT` is not set, `8080` will still be used. Overriding via the `--port` CLI option is still possible.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -61,7 +61,7 @@ const BASIC_GROUP = "Basic options:";
 // Taken out of yargs because we must know if
 // it wasn't given by the user, in which case
 // we should use portfinder.
-const DEFAULT_PORT = 8080;
+const DEFAULT_PORT = process.env.PORT || 8080;
 
 yargs.options({
 	"lazy": {
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -40,6 +40,7 @@ function Server(compiler, options) {
 	this.clientOverlay = options.overlay;
 	this.disableHostCheck = !!options.disableHostCheck;
 	this.publicHost = options.public;
+	this.allowedHosts = options.allowedHosts;
 	this.sockets = [];
 	this.contentBaseWatchers = [];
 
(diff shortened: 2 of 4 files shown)


jgierer12:
@SpaceK33z could you review this?

shellscape:
For posterity, I'd like there to be good record in conversation for how hotel works and why this is a necessary change before we move forward.

jgierer12:
Sure. hotel is basically a process manager that assigns local URLs (aka Virtual Hosts, e.g. `http://my-project.dev`) to commands. You can, for example, set it up so that whenever you visit `http://my-project.dev`, it automatically runs `npm start` (or any other command) and serves the website. Since `hotel` uses a random port, exposed in the `PORT` env variable, it would be great to use that in `webpack-dev-server` so it's easier to set up. I know I can pass the port manually via a CLI option, but it would be much easer and more intuitive to set up that way.
Also, there are a bunch of other tools using `PORT`. From the top of my head I can think of Heroku and now, although I'm not sure why you'd want to use `webpack-dev-server` there. But I'm sure ther are other useful tools that expose the `PORT` variable.

shellscape:
`--port $PORT` on the command line, or while adding a command to hotel would work then, right? 

I'll wait for others to weigh in on this. I'm not sure changing the default behavior at this point for a savings of a few keystrokes (not meant to be facetious at all) is a good trade off.

jgierer12:
That's what I was using previously, but it started getting problematic when I wanted to use it in a multi-platform project... overall it would just be nicer to have it use the env variable by default. But I totally get your concerns

shellscape:
I sat on this one for a bit (albeit longer than I should have, apologies there) and don't feel that it's a solid enough case for adding to the codebase. https://github.com/kentcdodds/cross-env should be able to help out with those cross-platform envar blues. Really do appreciate the PR and the time that you put into it.

PR #878, opened 2017-04-18
upgrade jquery from 1.8.1 to 3.2.1-slim

ysangkok:
retire keeps complaining about the dependencies I have that include webpack-dev-server, since this old jQuery version has security issues. I upgraded to a version without ajax, since it is not used in live.js, as far as I can see.


shellscape:
closing because stale. please do ping if you have the chance to circle back to this

PR #881, opened 2017-04-22
Do not show warnings @ 'overlay' unless explicitly set

dcarral:
This change tweaks the feature behavior (implemented @ #790), so it matches the intended behavior (described @ #789):

> overlay: true would be the same as overlay: { errors: true, warnings: false }.

This PR would supersede https://github.com/survivejs/webpack-book/pull/262.

---

It wasn't needed, since the current example explicitly forces _warnings_ to be shown/captured.

Projects using the `overlay: true` option would stop showing _warnings_, unless their _Webpack_ config gets updated to explicitly show them.


jamieathans:
Is it possible for this PR to be merged and a new webpack-dev-server bug-fix release be push to npm ASAP?

I ask because this issue/behaviour is affecting projects built with the latest version of the angular-cli - 1.0.1 (angular-cli projects hard-code `overlay:true` in development mode).

PR #888, opened 2017-04-24
Use correct index when parsing publicHost

phairoh:
2.4.3 introduced checking of the host header and options for setting the public host however when parsing the publicHost it was still using the index of `:` from the host header hostname. This corrects that.


phairoh:
Whoops, used the wrong value in the test. Fixing.

phairoh:
Added a few tests for the `checkHost` method. `Validation.test.js` seemed like the proper place for these but I wasn't entirely sure.

sokra:
Thanks

PR #890, opened 2017-04-25
fix: disableHostCheck should be defaulted to true

### LINKED ISSUE #882: --host 0.0.0.0 Not working
I run the webpack-dev-server with params `--host 0.0.0.0`, then visit the page through my ip, the page shows "Invalid Host header". 
I look into the code, it seems the `Server.prototype.checkHost` in file `lib/Server.js` is wrong, it don't deal with the params of `--host 0.0.0.0`

nick-woodward:

Fixes a breaking change in 29578537c1ad38f29a445237c57a52c373183e75 by defaulting `disableHostCheck` to true.

No.

**Summary**

Previously `webpack-dev-server` would allow connections from any host 29578537c1ad38f29a445237c57a52c373183e75 fixed this.

However it introduced a breaking change by enabling this functionality
by default. 

Instead of breaking a current subset of users with this change, we should 
disable it by default until the next major release of `webpack-dev-server`.

See https://github.com/webpack/webpack-dev-server/issues/882#issuecomment-297088982.

It introduces a breaking change for users dependent on `disableHostCheck` being false.

However this won't really prevent their servers from running, 
instead it just won't verify the `Host` header.

**What issues does this close?**

resolves #882

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -35,7 +35,8 @@ function Server(compiler, options) {
 	this.headers = options.headers;
 	this.clientLogLevel = options.clientLogLevel;
 	this.clientOverlay = options.overlay;
-	this.disableHostCheck = !!options.disableHostCheck;
+	// TODO: Default this to false when the next major version is released
+	this.disableHostCheck = options.disableHostCheck === undefined ? true : !!options.disableHostCheck;
 	this.publicHost = options.public;
 	this.sockets = [];
 	this.contentBaseWatchers = [];
--- a/test/Compress.test.js
+++ b/test/Compress.test.js
@@ -10,7 +10,8 @@ describe("Compress", function() {
 
 	before(function(done) {
 		server = helper.start(config, {
-			compress: true
+			compress: true,
+			disableHostCheck: false,
 		}, done);
 		req = request(server.app);
 	});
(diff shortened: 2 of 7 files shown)


evan-scott-zocdoc:
Alternatively, I think it makes sense to autoenable this if setting the host to `0.0.0.0` e.g. bind to all IPs.

nick-woodward:
@evan-scott-zocdoc Totally agree, my primary focus for this PR was to revert the breaking change.

However at this point I'm sure a large enough number of people have worked 
around it that implementing this wouldn't really make sense anymore.

nick-woodward:
**Closing: due to the age of this PR merging it would likely break more users then it would help.**

evan-scott-zocdoc:
@sokra how to do you feel about the change described in https://github.com/webpack/webpack-dev-server/pull/890#issuecomment-305040442? I'd be happy to submit a PR.

sokra:
> @sokra how to do you feel about the change described in #890 (comment)? I'd be happy to submit a PR.

no. The default should be secure. Binding to `0.0.0.0` is insecure.

evan-scott-zocdoc:
It's a dev server though... are you recommending people use this in production?

sokra:
no, don't use it in production.

That's not what this attack is about. It's about attacking the dev-server from a website you visit while it's running.

See https://medium.com/webpack/webpack-dev-server-middleware-security-issues-1489d950874a

PR #899, opened 2017-05-01
Add 'allowedHosts' option

orteth01:
feature/bugfix

no

**Summary**
Host checking was added by default in v2.4.3 for security reasons. Users now have to specify the host/URL being used to access the dev-server [more info here](https://medium.com/webpack/webpack-dev-server-middleware-security-issues-1489d950874a) 

This PR allows users to specify multiple hosts that that will pass the check. Recommended by @edmorley here: https://github.com/webpack/webpack-dev-server/issues/882#issuecomment-296799201

no

**Example usage in webpack config:**

```js
devServer: {
    allowedHosts: [
        'exampleHost.com',
        'differentExampleHost.com',
        ...
    ]
}
```

**subdomain wildcard:**

mimicking django's `ALLOWED_HOSTS`, a value beginning with "." can be used as a subdomain wildcard. '.example.com' will match example.com, www.example.com, and any other subdomain of example.com.

```js
devServer: {
    allowedHosts: [
        '.example.com'
    ]
}
```

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -445,7 +445,19 @@ Server.prototype.checkHost = function(headers) {
 	if(hostname === "127.0.0.1" || hostname === "localhost") return true;
 
 	// allow if hostname is in allowedHosts
-	if(this.allowedHosts && this.allowedHosts.indexOf(hostname) >= 0) return true;
+	if(this.allowedHosts && this.allowedHosts.length) {
+		for(let hostIdx = 0; hostIdx < this.allowedHosts.length; hostIdx++) {
+			const allowedHost = this.allowedHosts[hostIdx];
+			if(allowedHost === hostname) return true;
+
+			// support "." as a subdomain wildcard
+			// e.g. ".example.com" will allow "example.com", "www.example.com", "subdomain.example.com", etc
+			if(allowedHost[0] === ".") {
+				if(hostname === allowedHost.substring(1)) return true; // "example.com"
+				if(hostname.endsWith(allowedHost)) return true; // "*.example.com"
+			}
+		}
+	}
 
 	// allow hostname of listening adress
 	if(hostname === this.listenHostname) return true;
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -40,6 +40,7 @@ function Server(compiler, options) {
 	this.clientOverlay = options.overlay;
 	this.disableHostCheck = !!options.disableHostCheck;
 	this.publicHost = options.public;
+	this.allowedHosts = options.allowedHosts;
 	this.sockets = [];
 	this.contentBaseWatchers = [];
 
@@ -442,6 +443,9 @@ Server.prototype.checkHost = function(headers) {
 
 	// always allow localhost host, for convience
 	if(hostname === "127.0.0.1" || hostname === "localhost") return true;
+
+	// allow if hostname is in allowedHosts
+	if(this.allowedHosts && this.allowedHosts.indexOf(hostname) >= 0) return true;
 
 	// allow hostname of listening adress
 	if(hostname === this.listenHostname) return true;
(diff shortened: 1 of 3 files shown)


mgol:
[Django docs](https://docs.djangoproject.com/en/1.11/ref/settings/#allowed-hosts) say:

> A value beginning with a period can be used as a subdomain wildcard: '.example.com' will match example.com, www.example.com, and any other subdomain of example.com.

Could this be added as well?

orteth01:
@mgol I didn't realize that there were already a few tests for `Server.prototype.checkHost` 
https://github.com/webpack/webpack-dev-server/blob/master/test/Validation.test.js#L65-L118

I can either add tests for `allowedHosts` there or I can remove those tests because the tests that I added cover those cases. Which do you prefer?

mgol:
@orteth01 I guess it makes sense to nest that under existing tests (a separate inner `describe` sounds good).

Note, though, that I'm just a user, not a member of Webpack's core team so you might want to wait for someone on the team to decide. I'm just providing my own feedback.

orteth01:
> I guess it makes sense to nest that under existing tests (a separate inner describe sounds good).

Yea that makes more sense, I think. I've gone with that option. thanks, @mgol

orteth01:
I've signed the CLA. not sure why that check is still waiting for the status to be reported....\

orteth01:
There seem to be a few open issues that this PR would remedy. Any thoughts, @SpaceK33z ??

ssilve1989:
What's the status on this?

orteth01:
> What's the status on this?

@ssilve1989 waiting on review from one of Webpack's core team members

prencher:
@SpaceK33z @sokra I hate to be that guy, but this is preventing us from moving to a more secure setup; The point of the change this inspired. Can we get movement on this PR?

I am happy to assist if I can.

Justus-Maier:
Can I put an IP address with wildcards in here too? I sometimes get new IP from DHCP and I would like to share my current work with my collegues (just open a browser). Can this be done without me registering a local private domain?

orteth01:
> Can I put an IP address with wildcards in here too?

Currently, you can add an ip address as an allowed host but not with wildcards.

Only the subdomain wildcard is supported, meaning that only the first part of your allowed host can be dynamic. In theory I guess you could have something like this
```
devServer: {
    allowedHosts: [
        '.123.123'
    ]
}
```
which would allow any IP's that take the form `*.*.123.123`. However, I think that would not be recommended, as it allows **any** ip that matches that pattern to access your dev server.

mgol:
@orteth01 For IPs I'd expect a wildcard going in a different direction, e.g. `192.168.`.

edmorley:
As I mentioned in some of the other threads, it should be safe to whitelist **all** IP-addresses-like hosts names by default, which would save people having to whitelist them manually.

orteth01:
> it should be safe to whitelist all IP-addresses-like hosts names by default

@edmorley it seems like a separate PR would be appropriate for that. I'll create an issue.

shellscape:
@orteth01 merged the PR, as its a solid feature and has been in waiting for quite some time, but we still need example/readme updates for this. we've got a little bit of time before we release, would be great if you would be able to add that.

orteth01:
> we still need example/readme updates for this

@shellscape I'm more than happy to help with documentation! I'm assuming that I should update [this readme](https://github.com/webpack/webpack.js.org/blob/master/content/configuration/dev-server.md). Are there any other places where I should add information regarding this feature?

shellscape:
@orteth01 awesome. yes, please do update that readme. It's also suggested to create an example directory to demo the option.

Justus-Maier:
@orteth01 please also add some documentation for the disabling flag of the allowedHosts feature:
`config.devServer.disableHostCheck = true`
With security warnings obviously ;D

Mario-Eis:
Can this be set via command line?

orteth01:
> Can this be set via command line?

@Mario-Eis not currently. i'll try to get a PR open by the end of the week.

PR #907, opened 2017-05-15
Allow IPv6 loopback address

Timer:

bugfix

N/A

**Summary**

localhost resolves to `::1` for IPv6, `127.0.0.1` for IPv4. We shouldn't forget about IPv6 users.

No

**Other information**

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -439,7 +439,7 @@ Server.prototype.checkHost = function(headers) {
 	// we don't care about port not matching
 	const hostHeader = headers.host;
 	if(!hostHeader) return false;
-	const hostname = url.parse(hostHeader).hostname;
+	const hostname = url.parse("http://" + hostHeader).hostname;
 
 	// always allow localhost host, for convience
 	if(hostname === "127.0.0.1" || hostname === "localhost" || hostname === "::1") return true;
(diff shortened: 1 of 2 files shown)


Timer:
Hi! I made these changes on github so I hope they work. 😅
In a nutshell, there was no support for IPv6... this remedies that!

lib/Server.js
```diff
@@ -405,19 +406,17 @@ Server.prototype.checkHost = function(headers) {
...
 
 	// always allow localhost host, for convience
-	if(hostname === "127.0.0.1" || hostname === "localhost") return true;
+	if(hostname === "127.0.0.1" || hostname === "localhost" || hostname === "::1") return true;
 
 	// allow hostname of listening adress
 	if(hostname === this.listenHostname) return true;
 
 	// also allow public hostname if provided
 	if(typeof this.publicHost === "string") {
```
Timer: It would be great if `publicHost` could be an array so we could provide an IPv6 and IPv4 address.

shellscape:
@Timer looks like there are some issues with the tests to resolve before we can merge this. Please have a look

lib/Server.js
```diff
@@ -489,7 +485,7 @@ Server.prototype.listen = function(port, hostname) {
 	});
 	sockServer.on("connection", (conn) => {
 		if(!conn) return;
-		if(!this.checkHost(conn.headers)) {
+		if(!this.checkHost(conn.hostname)) {
```
Timer: We just have to make sure this works.

shellscape: it looks obvious as to why this was changed to `hostname` but what I'm wondering is if there was reason it was using `headers`. I think we may need tests around this to be certain.

shellscape:
ping @Timer re: tests for headers/hostname change

shellscape:
@Timer last ping before we close this one as abandoned. please do have a look at the conflicts and adding a test around the `hostname` change!

PR #917, opened 2017-05-23
Adding page argument to the Open option

TroyHisted:
A feature to allow a page to be specified when using the --open-page option from the CLI. 

Yes, added cli-open-page

**Summary**
Some applications have multiple "pages" in them and it's useful during development to be able to specify the exact page you'd like to open up as part of the command line argument. This saves a little bit of time having to navigate through the app. 
This feature was also described in this comment: https://github.com/webpack/webpack-dev-server/issues/311#issuecomment-279315230

No

**Other information**


bin/webpack-dev-server.js
```diff
@@ -310,8 +310,8 @@ function processOptions(wpOpt) {
 	if(argv["compress"])
 		options.compress = true;
 
-	if(argv["open"])
-		options.open = true;
+	if(argv["open"] !== undefined)
```
hxuanhung: You don't need to change this, `if(argv["open"])` included undefined check

TroyHisted: It's not entirely obvious but there are three potential values that could come from argv["open"]. It could be undefined (meaning the user did not specify the --open option), it could be an empty string (meaning the user added the --open option without a file path), or it could be a string containing a path (meaning the user added the --open option with a file path). The intent here is to only capture the value when it's a blank string or a string containing characters. That's not entirely obvious so I've changed it to the truthy check or a check for a blank string. As I'm thinking about this I could use a typeof string check instead. I can switch it over if you'd prefer.

hxuanhung: I didn't get it at first. It's ok for me to keep it like that.

bin/webpack-dev-server.js
```diff
@@ -310,8 +314,15 @@ function processOptions(wpOpt) {
 	if(argv["compress"])
 		options.compress = true;
 
-	if(argv["open"] || argv["open"] === "")
-		options.open = argv["open"];
+	if(argv["open"]) {
```
shellscape: do we need the ` || argv["open"] === ""` from the original logical check? (pretty sure we do, but could be wrong)

lib/optionsSchema.json
```diff
@@ -185,6 +185,10 @@
       "type": "boolean"
     },
     "open": {
+      "description": "Let the CLI open your browser.",
+      "type": "boolean"
+    },
+    "openPage": {
       "description": "Let the CLI open your browser with the URL.",
```
shellscape: "Let the CLI open your browser to a specific page on the site."

bin/webpack-dev-server.js
TroyHisted: I added it when I switched it from a boolean to a string, but now that it's back to a boolean we don't need that check.

But that did make me think about the same check on the open-page arg, so I changed it around so that I don't need that check anymore by making the open-page argument require a value. I think it's much cleaner now.

lib/optionsSchema.json
TroyHisted: Sounds good. I'll update.

PR #918, opened 2017-05-23
fix #752: allow --port 0 again

### LINKED ISSUE #752: --port 0 argument is ignored

Bug

When running `webpack-dev-server --port 0`, the port argument is ignored and the default, 8080, is used.

Port 0 is used to have the operating system assign a free ephemeral port. https://www.dnorth.net/2012/03/17/the-port-0-trick/ has a good explanation.

This worked in webpack-dev-server prior to #685.

**If the current behavior is a bug, please provide the steps to reproduce.**

Run any webpack project with `webpack-dev-server --port 0`. Port 8080 will be used if it is free.

With`--port 0` (or `options.port = 0`), `server.listen()` should be called with port 0, allowing the OS to assign a free port.

If this isn't possible, `--port 0` should cause an error rather than being ignored.

**Please mention your webpack and Operating System version.**

webpack-dev-server master (7e18f6a733162af022423db68a519b5186bdd6e4), webpack 2.2.0, macOS Sierra 10.12.2.

also:

Bugfix for #752 

**Summary**

Fixes the check for `options.port` and `argv.port` to allow `0`. I rely on the operating system to assign a port so that it is guaranteed to be available and doesn't ever bind to the default `8080`. I can't update beyond v2.1.0-beta.10, because after #685 `0` is treated the same as not passing the option and triggers portfinder.

It should only change behavior when `0` is passed as the port option.

**Other information**

The behavior when using port 0 could still be improved. This simply brings back the ability to specify port 0 which existed prior to v2.1.0-beta.11.


bin/webpack-dev-server.js
```diff
@@ -317,8 +319,8 @@ function processOptions(wpOpt) {
 	// that wouldn't throw errors. E.g. both argv.port and options.port
 	// were specified, but since argv.port is 8080, options.port will be
 	// tried first instead.
-	options.port = argv.port === DEFAULT_PORT ? (options.port || argv.port) : (argv.port || options.port);
-	if(options.port) {
+	options.port = argv.port === DEFAULT_PORT ? defaultTo(options.port, argv.port) : defaultTo(argv.port, options.port);
+	if(options.port != null) {
```
hxuanhung: You should not add !=null here because it's not consistant with the current code style.
Should keep `if(options.port)`

hxuanhung: How about `if(options.port || options.port === 0)` ?

Phoenixmatrix: It makes sense to me that if you're doing a truthy check, you do it as per your code style...but in this case, you specifically want "if not null and not undefined", for which the most idiomatic way is `!= null`.

Why obfuscate the code? This isn't about stylistic (I'm all about style consistency), it's essentially about correctness in this very particular scenario.

shellscape: I'm in agreement with @Phoenixmatrix and @also

bin/webpack-dev-server.js
```diff
@@ -39,6 +39,8 @@ function colorError(useColor, msg) {
 	return msg;
 }
 
+const defaultTo = (v, def) => v == null ? def : v;
```
shellscape: no sense in obfuscating parameters here and one-letter variables make me stabby. please do change `v` to `value`.

Issue #855, opened 2017-03-20
BUG: Errors and Warnings Ignore clientLogLevel


IdanCo:
With the latest version on webpack-dev-server, errors and warnings are printed to the browser (client) console, regardless of the _clientLogLevel_ value.

A PR is being cooked right now...

shellscape:
This has been resolved in recent versions.

PR #921, opened 2017-05-26
Browser console messages should respect clientLogLevel

jameslai:
Removes console calls with internal log function which respects clientLogLevel configuration values.

Moves warnings and errors to the warning and error logs, respectively. Also utilizes constants for fewer string comparisons.

Resolves issue #855

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -1,6 +1,7 @@
 /* global __resourceQuery WorkerGlobalScope */
 var url = require("url");
 var stripAnsi = require("strip-ansi");
+var log = require("loglevel")
 var socket = require("./socket");
 var overlay = require("./overlay");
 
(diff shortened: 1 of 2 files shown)


jameslai:
Absolutely, I can implement that shortly.

jameslai:
I've rebased the change into a single commit to keep the history clean.

Some defaults we were previously setting were able to be removed in lieu of loglevel's native default methods. Additionally a `none` setting will now utilize loglevel's `disableAll`.

We we need to do a little funny business when receiving the log level from configuration, as Webpack's documentation specifies using the string `warning` while loglevel expects `warn`.

An error will now be emitted if the configuration is attempting to set an unknown value for `clientLogLevel`.

client/index.js
```diff
@@ -32,18 +33,16 @@ if(typeof __resourceQuery === "string" && __resourceQuery) {
...
-		return console.log(msg);
-	if(["info", "warning"].indexOf(logLevel) >= 0 && level === "warning")
-		return console.warn(msg);
-	if(["info", "warning", "error"].indexOf(logLevel) >= 0 && level === "error")
-		return console.error(msg);
-}
+var INFO = "info";
+var WARNING = "warning";
+var ERROR = "error";
+var NONE = "none";
```
shellscape: this is duplicating the same functionality as the `quiet` option though, isn't it? https://webpack.js.org/configuration/dev-server/#devserver-quiet-

jameslai: The specific lines you highlighted are not duplicating the `quiet` functionality, but absolutely our consumption of a log level of `none` (see https://webpack.js.org/configuration/dev-server/#devserver-clientloglevel for the documentation specifying the use of `none`) most certainly seems like duplicate functionality. I think this is a good question for the Webpack team, being that we technically have two options which will do roughly the same thing - should one of these options be deprecated?

A proposed solution is to deprecate the use of `none` since that is technically not a log level (which the name `clientLogLevel` suggests), and leaving `quiet` for completely silencing output.

PR #926, opened 2017-06-01
Set HMR log level.

lbogdan:
Second take at fixing the HMR logging issue (first was [here](https://github.com/webpack/webpack-dev-server/pull/925)) - not being able to turn off HMR's console logging. This change makes HMR respect `webpack-dev-server`'s `clientLogLevel`.
It uses `webpack/hot/log` merged [here](https://github.com/webpack/webpack/pull/4960).

This is a bugfix.

None needed.

**Summary**
See https://github.com/webpack/webpack-dev-server/pull/925.

This needs a `webpack` version with https://github.com/webpack/webpack/pull/4960 merged, otherwise `require("webpack/hot/log")` will error. That's why the tests failed.


lbogdan:
Maybe we should check if `webpack/hot/log` exists?
```javascript
if(hotCtx.keys().length > 0) {
  hotCtx("./log").setLogLevel(level);
}
```

shellscape:
Looks good, but we need to wait for a pending major release before we merge. Didn't want to leave you hanging there.

lbogdan:
A webpack release, you mean?

shellscape:
We're actually going to be doing a webpack-dev-server major release soon. It may correspond with webpack.

lbogdan:
Makes sense. It would be nice to be able to get rid of the `webpack/hot/log` check, like in the initial commit. Thanks for the heads up!

shellscape:
> This needs a webpack version with webpack/webpack#4960 merged, otherwise require("webpack/hot/log") will error. That's why the tests failed.

@lbogdan since this module is going to continue to support older versions of webpack, we're going to need some error catching around this change. We can't just let it throw cause it'll affect those unable to upgrade webpack. 

once we get that in place I think we'll be good to go on this

lbogdan:
@shellscape The last commit has error checking (more like "feature" checking, actually). That's why the tests pass. Can you please check?

shellscape:
@lbogdan cool, looks good to me. there's a conflict that needs resolving and then we can merge

lbogdan:
@shellscape Fixed conflicts, everything looks good.

Also, I saw that with https://github.com/webpack/webpack-dev-server/pull/921 `webpack-dev-server` is now using the `loglevel` package for logging. Maybe use it in `webpack`, too?

shellscape:
Unfortunately I'm not privy to the workings of webpack main. It might be worth raising an issue there.

lbogdan:
Oh, ok, I was under the impression that `webpack` and `webpack-dev-server` have pretty much the same maintainers. :smiley:

shellscape:
I think there's some cross-polination between them all, but it's not 1:1

PR #929, opened 2017-06-05
Strongly check client isn't running on WebWorker for sendMsg

### LINKED ISSUE #745: Second argument of postMessage is not a transfert list

jhen0409:
**What kind of change does this PR introduce?** bugfix

**Summary**

Fix #745 again.

For some reason I've defined `self.window` in WebWorker, so #813 still doesn't fixed for me.

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -1,4 +1,4 @@
-/* global __resourceQuery */
+/* global __resourceQuery WorkerGlobalScope */
 var url = require("url");
 var stripAnsi = require("strip-ansi");
 var socket = require("./socket");
@@ -47,7 +47,11 @@ function log(level, msg) {
 
 // Send messages to the outside, so plugins can consume it.
 function sendMsg(type, data) {
-	if(typeof self !== "undefined" && self.window) {
+	if(
+		typeof self !== "undefined" &&
+		(typeof WorkerGlobalScope === "undefined" ||
+		!(self instanceof WorkerGlobalScope))
+	) {
 		self.postMessage({
 			type: "webpack" + type,
 			data: data


client/index.js
```diff
@@ -47,7 +47,7 @@ function log(level, msg) {
 
 // Send messages to the outside, so plugins can consume it.
 function sendMsg(type, data) {
-	if(typeof self !== "undefined" && self.window) {
+	if(typeof self !== "undefined" && typeof WorkerGlobalScope === "undefined") {
```
shellscape: I'd like to amend this slightly:

```js
if(typeof self !== "undefined" && typeof WorkerGlobalScope === "undefined" && !self instanceOf WorkerGlobalScope) {
```

If we find another reason to check to see if the script is executing within a WebWorker, we'll want to move that out to a shared var.

jhen0409: I think the correct check should be this:

```js
typeof self !== "undefined" &&
(typeof WorkerGlobalScope === "undefined" ||
!(self instanceof self.WorkerGlobalScope))
```

Use `instanceof` only if it not undefined. I changed it. :)

client/index.js
```diff
@@ -47,7 +47,11 @@ function log(level, msg) {
 
 // Send messages to the outside, so plugins can consume it.
 function sendMsg(type, data) {
-	if(typeof self !== "undefined" && self.window) {
+	if(
+		typeof self !== "undefined" &&
+		(typeof WorkerGlobalScope === "undefined" ||
+		!(self instanceof self.WorkerGlobalScope))
```
shellscape: I don't understand why you're using `self.WorkerGlobalScope` here. `self instanceof WorkerGlobalScope` should be correct.

jhen0409: Just passing eslint. Anyway I moved it to line 1.

PR #930, opened 2017-06-06
zeroconf dns (bonjour) service publishing

samuraisam:
**What kind of change does this PR introduce?** Feature

**Did you add or update the `examples/`?** Yes

**Summary**
This creates a zeroconf dns service when the `--bonjour` flag is provided to the dev server command. it publishes an `http` service with the subtype of `webpack`. This is highly useful when debugging the live server across the network which can often be necessary when developing for network attached devices.

No, but it does add a dependency on the `bonjour` package.

**Other information**
I had thought to add customization to the service type and subtypes but the defaults seem sufficient enough to serve the use case without bloating the interface.


samuraisam:
I'm actually not sure why travis would fail here.

bin/webpack-dev-server.js
```diff
@@ -401,8 +409,10 @@ function startDevServer(wpOpt, options) {
 			});
 		});
 	} else {
+		console.log("shits");
```
shellscape: well I'm not sure we want to be telling users about this 😄

shellscape:
OK, you need to add the option to https://github.com/webpack/webpack-dev-server/blob/master/test/Validation.test.js#L51. See https://github.com/webpack/webpack-dev-server/commit/ab889c31a584e193814b2e952410307d071ede61#diff-46ade4578831d079715c6efd9c726d48 for an example

shellscape:
I think you missed the test file in that last commit :)

samuraisam:
Wow this has to be my personal record for worst managed PR ever. Thanks for being understanding!

shellscape:
@samuraisam no worries :) PR is solid, but the CLA needs signing (https://github.com/webpack/webpack-dev-server/pull/930#issuecomment-306384164) before we can merge it. Please give that a look.

samuraisam:
It looks like I have signed it (the CLA check passes)

PR #942, opened 2017-06-13
Ssl path

shellscape:
Removes the in-repo SSL Cert

**Summary**
Replaces the use of the in-repo ssl certificate in favor of certs generated on a 30 day cycle, per-instance of webpack-dev-server.

Negatory

**Other information**
See @TheLarkInn


shellscape:
(should've squashed, my bad)

perlun:
Coming in late here - doesn't this mean that you have to re-trust the localhost cert (i.e. add it to Keychain Access if using Chrome) _every month_ if you want to get rid of SSL warnings in the browser? (According to my April 2018 experience, Chrome doesn't trust localhost self-signed certificates unless you add it to the CA root list on the machine, or enable the `chrome://flags/#allow-insecure-localhost` setting.)

shellscape:
Heh, yeah nearly a year late. See: https://medium.com/@mikenorth/webpack-preact-cli-vulnerability-961572624c54. After much discussion between security experts and the webpack team, this was the solution arrived upon. Not here to debate the merits at this point, (and I no longer maintain this repo) only sharing info.

perlun:
@shellscape Thanks for the link and the clarification. Hehe, I'm not even a webpack user myself. 😉 But is implementing SSL support in our Ruby-based application framework and stumbled on this post. I respect the decisions taken by the people you refer to, just thought about the implications. But, the PR also suggests to "manage your own certificate" which makes sense.

PR #943, opened 2017-06-17
Multiple devServer.publicPath

reda-alaoui:
Add the multiple publicPath feature.

Yes

**Summary**
Fix webpack/webpack-dev-server#945.

No

**Other information**
This PR depends on https://github.com/webpack/webpack-dev-middleware/pull/198.


shellscape:
@reda-alaoui sorry that no one followed up on your PR. please do know that it's appreciated. we're going to slate this for v3, which will begin work soon. we'll release a new minor version of webpack-dev-middleware with your PR on that module to support this change when we're closer to releasing v3 of WDS.

shellscape:
After much contemplation and consideration we've decided not to merge this into the codebase. For a more in-depth explanation and further discussion, please see #945.

reda-alaoui:
So you are saying that running as many webpack dev servers as paths is just ok?
Because this ugly workaround exists, you won't accept it?
Because nobody uses a Tomcat or Java container except me?
Because the world is so complex, you don't want to manage this complexity?

I guess I will manage my own complexity and fork until modules are natively supported by all browsers.

PR #970, opened 2017-07-04
Don't output startup info if quiet is set to true

larsbs:
Use `quiet` option to decide if webpack should print startup info or not.

No

**Summary**
As stated by @jonathanconway in https://github.com/webpack/webpack/issues/1191 there's no way to fully disable all logs. Also, I've been investigating and didn't find where the `noInfo` option is used so I decided to use it to disable the startup info.

No

**Other information**


larsbs:
@shellscape but then the `noInfo` option is useless again. At least when I've searched for it in the code is used nowhere. For me it's ok, I can change the code to use `quiet`, but why keep an unused option around?

shellscape:
It might be there for legacy reasons, not entirely sure.

Issue #972, opened 2017-07-05
Webpack open config option opens to /undefined


coltonw:
**Do you want to request a *feature* or report a *bug*?**
Reporting a bug introduced by PR #917 

**What is the current behavior?**
The page opened by webpack-dev-server is /undefined

**If the current behavior is a bug, please provide the steps to reproduce.**

If you use a webpack config file using `devServer.open = true` but have not set `devServer.openPage`, the the opened page will by /undefined

**What is the expected behavior?**
It should open to just /

**Please mention your webpack and Operating System version.**
webpack 2.5 Mac OS X El Capitan

**Also**
PR in progress

lerte:
As a quick fix, you can add openPage: '' to your options:
openPage: '' // <== Add this
https://github.com/webpack/webpack-dev-server/issues/960

shellscape:
I believe we have this fixed in the latest version. And the workaround that @lerte mentioned is valid for older versions.

PR #973, opened 2017-07-05
Set undefined openPage to empty string when open option is true

coltonw:

bugfix

No

**Summary**

#972 Bug with new openPage option

No

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -338,8 +338,11 @@ function processOptions(wpOpt) {
 
 	if(argv["open"] || argv["open-page"]) {
 		options.open = true;
-		options.openPage = argv["open-page"] || "";
+		options.openPage = argv["open-page"];
 	}
+
+	if(options.open && !options.openPage)
+		options.openPage = "";
 
 	if(argv["useLocalIp"])
 		options.useLocalIp = true;


bin/webpack-dev-server.js
```diff
@@ -331,9 +331,12 @@ function processOptions(wpOpt) {
 
 	if(argv["open"] || argv["open-page"]) {
 		options.open = true;
-		options.openPage = argv["open-page"] || "";
```
shellscape: Maybe I'm having an off day, but I don't see how this is any different..

```js
var a = {};
a['b'] || ''  // <== ""
```

Is `argv["open-page"]` returning `"undefined"` as a String here for some reason?

coltonw: No, what is happening is that this statement isn't hit at all, because argv is only for command line arguments.

In the case of a config file, argv will be mostly empty but options.open and options.openPage may already have been populated by the input config file, so we STILL need to set openPage to empty string regardless of how options.open got set to true.

shellscape: I'm not following you. You're saying that this statement won't be hit because argv is for command line use, yet the statement that my comment was for is within a logical block where the checks are against argv. I'm not sure your reply addresses the comment.

coltonw: The change I made to that line only makes sense in the context that it goes hand right along with the next if statement.

Before my change, if you go through this function with argv for open, you get options.open set to true and options.openPage set to either the open-page command line arg or to empty string.  After my change, this logic is identical except the setting openPage to empty string is now a few lines down instead of being inline.  Basically, the line you are concerned about is irrelevant to the change: I removed an unnecessary condition from the code for a tiny bit of code simplification but, if I had left the line in, the behavior would be the same by the time you finish the function.

Before my change, if you go through this function with options.open already set to true and no argv, you would have options.openPage set to undefined because you literally never hit the line of code you are asking about. **This** is what is causing the bug.  After my change, I check for options.open independent of whether you have argv.open or not and set undefined options.openPage to empty string.

Is that clearer?  I am sorry if I am not being clear. Written communication is not my strongpoint.

Run through the code in your head before and after with argv set to `{}` but options.open set to true.  That is the only case I am changing at all and is the case I am fixing.
The reason it matters if openPage is undefined or empty string is because [later in the file](https://github.com/coltonw/webpack-dev-server/blob/ebd9d9db571a5a173a77ab4a52c6071be09371f4/bin/webpack-dev-server.js#L452) we string concatenate the uri with openPage.

ankurk91:
@coltonw 
Can you rebase ?

coltonw:
Rebased!

Issue #941, opened 2017-06-13
WDS Disconnecting with HTTPS


callensm:
Running WDS version: 2.4.5

I'm currently running my `React` application from the `webpack-dev-server` using the following configuration file:

    // webpack.config.js
    /* eslint no-var: 0 */
    
    var path = require('path')
    var webpack = require('webpack')
    var HtmlWebpackPlugin = require('html-webpack-plugin')
    var ExtractTextPlugin = require('extract-text-webpack-plugin')
    
    module.exports = {
      entry: {
        'controller-bundle': path.resolve(__dirname, './src/js/sym-controller.js'),
        'app-bundle': path.resolve(__dirname, './src/js/sym-app.js'),
        bundle: path.resolve(__dirname, './src/js/index.js')
      },
      output: {
        path: path.resolve(__dirname, './dist'),
        filename: '[name].js'
      },
      resolve: {
        extensions: ['.js', '.jsx']
      },
      plugins: [
        new webpack.HotModuleReplacementPlugin(),
        new ExtractTextPlugin('style.css'),
        new HtmlWebpackPlugin({
          template: './src/html/controller.html',
          filename: 'controller.html',
          chunks: ['controller-page']
        }),
        new HtmlWebpackPlugin({
          template: './src/html/app.html',
          filename: 'app.html',
          chunks: ['app-page']
        })
      ],
      module: {
        loaders: [
          {
            test: /\.js(x)?$/,
            exclude: /(node_modules|bower_components)/,
            loader: 'babel-loader'
          },
          {
            test: /\.css$/,
            use: ExtractTextPlugin.extract({
              fallback: 'style-loader',
              use: 'css-loader'
            })
          }
        ]
      },
      devServer: {
        contentBase: path.resolve(__dirname, './dist'),
        host: '0.0.0.0',
        port: 4000,
        https: true,
        headers: {
          'Access-Control-Allow-Origin': '*'
        }
      }
    }

The problem I'm running into is that while the `https` flag for `WDS` is set to `true`, the application continuously throws an error:

    [WDS] Disconnected!

and immediately refreshes the page. (maybe once a second or two)

When `https` is set to `false` this doesn't happen, but I need the application to run with `https` on. I've seen this as an issue on the github issues section for `webpack` but cannot seem to find a clear, straight answer as the why this is happening and how to fix it.

[...]

PR #986, opened 2017-07-11
include `subjectAltName` field in self-signed cert

cwalv:
fixes issue with reconnect on Chrome 58+
#941

bugfix for #941 

**Summary**

**Other information**


cwalv:
I don't understand why, but the failures seem to be intermittent.  I was able to get a green cert with the code @veke provided [here](https://github.com/angular/angular-cli/issues/4839#issuecomment-314608490).  I'll close this PR and create a new one with that

PR #1007, opened 2017-07-24
Always allow requests with IP-address as host in checkHost()

### LINKED ISSUE #931: Whitelist all IP-address-like hosts names by default in checkHost
[the problem](https://github.com/webpack/webpack-dev-server/pull/899#issuecomment-306752341):

Essentially, users desire the ability to whitelist certain IP addresses so that colleagues can access their dev server to view their work. However, they'd prefer to not have to specify each individual allowed IP.

[the solution](https://github.com/webpack/webpack-dev-server/pull/899#issuecomment-306942984) from @edmorley:

"it should be safe to whitelist **all** IP-addresses-like hosts names by default, which would save people having to whitelist them manually."

usefulthink:

enhancement

no.

**Summary**

This patch will allow any requests made using an IP-address to always pass the
checkHost-test.

IP-addresses are not susceptible to a dns-rebind like attack so it would make
sense to not block them to make local-network development possible without
needing to disable the host-checks entirely.

no

**Other information**

fixes #931

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -7,6 +7,7 @@ const express = require("express");
 const fs = require("fs");
 const http = require("http");
 const httpProxyMiddleware = require("http-proxy-middleware");
+const ip = require("ip");
 const serveIndex = require("serve-index");
 const historyApiFallback = require("connect-history-api-fallback");
 const path = require("path");
@@ -441,11 +442,11 @@ Server.prototype.checkHost = function(headers) {
 	const idx = hostHeader.indexOf(":");
 	const hostname = idx >= 0 ? hostHeader.substr(0, idx) : hostHeader;
 
-	// always allow localhost host, for convience
-	if(hostname === "127.0.0.1" || hostname === "localhost") return true;
-
 	// always allow requests with explicit IP-address
-	if(/^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\.|$)){4}$/.test(hostname)) return true;
+	if(ip.isV4Format(hostname)) return true;
+
+	// always allow localhost host, for convience
+	if(hostname === "localhost") return true;
 
 	// allow if hostname is in allowedHosts
 	if(this.allowedHosts && this.allowedHosts.length) {
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -444,6 +444,9 @@ Server.prototype.checkHost = function(headers) {
 	// always allow localhost host, for convience
 	if(hostname === "127.0.0.1" || hostname === "localhost") return true;
 
+	// always allow requests with explicit IP-address
+	if(/^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\.|$)){4}$/.test(hostname)) return true;
+
 	// allow if hostname is in allowedHosts
 	if(this.allowedHosts && this.allowedHosts.length) {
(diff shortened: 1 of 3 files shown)


usefulthink:
first-time contributor here. Please let me know if there are any docs I need to validate/update or anything else I should think of.

lib/Server.js
```diff
@@ -444,6 +444,9 @@ Server.prototype.checkHost = function(headers) {
 	// always allow localhost host, for convience
 	if(hostname === "127.0.0.1" || hostname === "localhost") return true;
 
+	// always allow requests with explicit IP-address
+	if(/^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\.|$)){4}$/.test(hostname)) return true;
```
edmorley: Should we add a few more testcases, for both the valid and invalid IP cases?

Also, I wonder if a simplified regex that just checked that each octet was numeric would be simpler to read/faster/still as secure? [RFC3696 section 2](https://tools.ietf.org/html/rfc3696#section-2) says that top level domain names aren't allowed to be all-numeric, so eg 999.999.999.999 still won't be treated as a domain, so can't be used for DNS rebinding.

For example:
```
// Requests to explicit IP-addresses can't be exploited by DNS rebinding.
// For simplicity the regex matches numeric ranges that aren't valid IPs, but
// this is still secure since top level domain names can never be all-numeric:
// https://tools.ietf.org/html/rfc3696#section-2
if(/^([0-9]{1,3}(\.|$)){4}$/.test(hostname)) return true;
```

usefulthink: I like the simplicity and probably would want to do that. We don't validate other aspects of IP adress as well (reserved, multicast, etc) and I can't think of a way to have an invalid or unrouteable IP-address in the browser anyway.

shellscape: I'd like to see https://www.npmjs.com/package/ip and `ip.isV4Format` used here, rather than a regex.

usefulthink: I thought about that as well, but didn't want to introduce any new dependencies for something that could be put in a simple (ok, thats arguable) regex. Is there any policy regarding adding dependencies?

shellscape: If the dependencies were being packaged for a production environment, I'd agree with you. Since this is strictly a development tool, tried and true dependencies to perform utility functions are acceptable. You'll see that pattern throughout the commit history. (for example; loglevel over a custom implementation).

usefulthink:
Thanks! 🎊

lib/Server.js
```diff
@@ -441,8 +442,11 @@ Server.prototype.checkHost = function(headers) {
 	const idx = hostHeader.indexOf(":");
 	const hostname = idx >= 0 ? hostHeader.substr(0, idx) : hostHeader;
 
+	// always allow requests with explicit IP-address
+	if(ip.isV4Format(hostname)) return true;
```
trygveaa: Shouldn't it allow IPv6-adresses as well?

shellscape: that's true, good catch

usefulthink: Should I do another PR for this? Did anyone ever use IPv6 in such a context? 
Using `http://[::1]:1234/` is absolutely possible, but that will already break in the lines before as a colon no longer works to split out the port-number from the host-header.

shellscape: It may be worthwhile to solve. I haven't heard of anyone using IPv6 in that way, but the call out by @trygveaa is valid. Totally up to you if you'd like to create a followup PR

Issue #1011, opened 2017-07-26
Feature proposal: requestCert field for authenticating client certificate


dbk91:
I was using create-react-app, which relies on webpack-dev-server to deliver content in development, and wanted to propose adding the requestCert configuration field. Client certificate authentication (2-way PKI) is a must have for an application I am working on. While I can get this working for production with Nginx as the proxy server, it would be really nice to have the same or similar behavior for development. In order to do so, I believe webpack-dev-server would have to support the requestCert field as an https option so that it could grab the client's certificate for authentication.

If this would be an unnecessary feature due to the fact that http-proxy-middleware already supports this, please let me know. I have yet to find a way to get it working by means of passing configuration to http-proxy-middleware.

Here are the changes in my forked repository: https://github.com/webpack/webpack-dev-server/compare/master...dbk91:request_cert_support

Thanks!

shellscape:
We'd happily review a pull request to add this feature.

PR #1029, opened 2017-08-09
Add requestCert support to the server

dbk91:
Add this configuration flag to request a client's certificate for 2-way PKI authentication.
Issue #1011

The motivation for this feature proposal is to allow create-react-app to enable the instance of webpack-dev-server to request the client certificate for 2-way PKI auth.

The issue actually doesn't lie in webpack-dev-server itself (configuration is much easier when using webpack-dev-server by itself), but rather that create-react-app only sends a boolean to webpack-dev-server. Here is the line I am referencing: https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/config/webpackDevServer.config.js#L81

From what I can tell, this means that the configuration options for webpack-dev-server are limited to the key, cert, ca, and passphrase when using create-react-app. Minimizing configuration seems to be create-react-app's goal (which I totally understand) but a feature like this would be nice to have in development.

I simply just added the requestCert field to the configuration if webpack-dev-server receives https as a boolean rather than an object. Defaults to false if it is no specified.

### DIFF
--- a/lib/optionsSchema.json
+++ b/lib/optionsSchema.json
@@ -145,6 +145,10 @@
       "description": "The passphrase to a (SSL) PFX file.",
       "type": "string"
     },
+    "requestCert": {
+      "description": "Enables request for client certificate.",
+      "type": "boolean"
+    },
     "inline": {
       "description": "Enable inline mode to include client scripts in bundle (CLI-only).",
       "type": "boolean"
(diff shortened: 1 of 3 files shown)


lib/optionsSchema.json
```diff
@@ -145,6 +145,10 @@
       "description": "The passphrase to a (SSL) PFX file.",
       "type": "string"
     },
+    "requestCert": {
+      "description": "Enables request for client certificate.",
```
shellscape: Please add "This is passed directly to the https server" to the description so folks know what the purpose is, since there are some mixed-use child properties in the `https` property.

Issue #974, opened 2017-07-05
Show compilation status on overlay


Wernerson:
**Do you want to request a *feature* or report a *bug*?**
*feature*

**What is the current behavior?**
Error messages get display on the overlay and in the terminal. Compilation status only appears in terminal.

**What is the expected behavior?**
Display a little popup when webpack recompiles in watch mode.

**If this is a feature request, what is motivation or use case for changing the behavior?**
You can see the compilation status on the website. This would be great during development because you wouldn't have to switch to the console over and over again :)

Wernerson:
I have a first version which just displays the text you can see in the client console in my forked repo.

shellscape:
@Wernerson please do feel free to open a Pull Request with your proposed changes from your fork and we can go from there.

PR #1032, opened 2017-08-09
New feature (compilation) status window

Wernerson:
feature

yes

**Summary**

When Webpack is compiling a little window appears and shows the compilation status (like building modules, optimizing chunks, etc.) and a progress bar indicating the progress (who would have thunk). The title of the browser tab also displays the progress in percent.

With this feature, you don't have to check your Terminal for the compilation status anymore. You see any important information right on the screen.

First idea in issue: #974 

If the webpack.config.js isn't modified there *shouldn't* be any breaking changes (as far as I tested) for existing projects.

**Other information**
 - Tested in a real-life project (thou without the changing title feature, going to test this tomorrow).
 - added example
 - added status option to optionsSchema.json (Validation test for this runs as well)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ We appreciate all help! Check out [CONTRIBUTING.md](CONTRIBUTING.md) for more in
         <img width="150" height="150"
         src="https://avatars.githubusercontent.com/SpaceK33z?v=3">
         <br />
-        <a href="https://github.com/">Kees Kluskens</a>
+        <a href="https://github.com/SpaceK33z">Kees Kluskens</a>
       </td>
     </tr>
   </tbody>
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -8,7 +8,6 @@ const net = require("net");
 const portfinder = require("portfinder");
 const addDevServerEntrypoints = require("../lib/util/addDevServerEntrypoints");
 const createDomain = require("../lib/util/createDomain");
-const bonjour = require("bonjour")();
 
 // Local version replaces global one
 try {
@@ -185,6 +184,11 @@ yargs.options({
 		describe: "The port",
 		group: CONNECTION_GROUP
 	},
+	"disable-host-check": {
+		type: "boolean",
+		describe: "Will not check the host",
+		group: CONNECTION_GROUP
+	},
 	"socket": {
 		type: "String",
 		describe: "Socket to listen",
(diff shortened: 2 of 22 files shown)


shellscape:
Thanks for the PR! Could you provide a screenshot of what this feature looks like? We'll also need you to sign the CLA before we can move forward.

Wernerson:
I have tested the code in our real-life app. Worked as well.

Here is the screenshot (the window appears in the middle of the screen):
screenshot

Wernerson:
I resolved the conflicts but **couldn't test them so far** (can't update webpack at my workplace...).

shellscape:
> I resolved the conflicts but couldn't test them so far (can't update webpack at my workplace...).

That's an interesting restriction they have on you. Please let us know when you're able to test and confirm your conflict resolutions were correct.

Wernerson:
Found an error from the merge. The code is tested and working. Ready to merge from my side :)

shellscape:
Thanks for staying on top of this. We're taking a look at how best to move forward with this type of addition and feature on the client. We haven't forgotten about you :)

Wernerson:
Just had an idea. Maybe add an option to display the webpack console output also in the browser console? We could do this the same way with the overlay option. Either true (just displaying the status window) or an object with window and console properties... What do you think?

shellscape:
@Wernerson that's a good idea. we've been wrestling with how to handle additional features for the client like this - we've got the error overlay and now the proposal for the progress overlay. their styles are very different, but we don't have a unified guide for that. now it's clear that we may have other additions that folks may want to add to the client as well, so it presents an interesting question as to how to handle contributions for those moving forward. and your PR here was the catalyst for that 👍 

all that being said, I don't think merging this PR _right this moment_ is the right call. I do think that sending the progress info to the console _is_ the right call right now, as it gives the user some valuable feedback immediately. however, I'd like to leave this PR open and in a pending state. that's because...

...what I'd like to do in the near future is to provide a mechanism for overlay plugins that contain their own business logic, move the warning and errors overlay into their own plugins, get this progress overlay into a plugin (I'd be happy to help with that), unify their styles (to match https://webpack.js.org/ as a guide), and release. this would ultimately open the door for any number of overlays on demand that won't need any kind of approval. it should also slim down the codebase for webpack-dev-server and overall be a win.

Wernerson:
@shellscape yeah that makes sense... I'm gonna think of a possible solution as well and will rewrite the progress feature as a plugin as soon as it's possible. Maybe we can close the PR and sort of label it? And maybe make an issue so we could gather opinions or possible solutions from others? 😄

shellscape:
@Wernerson we'll close this one, but please do take this branch and create a feature branch from it that adds the status to console messages as you suggested. I think that's a feature we can introduce immediately.

PR #1036, opened 2017-08-10
No longer generating ssl cert when one is already specified

mrsharpoblunto:
Small startup time optimization

**Summary**
Skip the generation of a self signed SSL cert if an existing cert is already specified in the https options. The self signed SSL cert never gets used if you specify a cert in the options anyway, so theres not much point generating it in this case

No 

**Other information**


shellscape:
@mrsharpoblunto thanks for the PR! please sign the CLA and we can move forward

mrsharpoblunto:
@shellscape Signed the CLA

Issue #877, opened 2017-04-17
CLI option --hot-only is confusing


agoldis:
Using 
* `webpack-dev-server` `2.4.2`
* `webpack` `2.4.1`

According to the [documentation](https://webpack.js.org/configuration/dev-server/#devserver-hotonly-cli-only) CLI option `--hot-only` should:

> Enable Hot Module Replacement (see devServer.hot) without page refresh as fallback in case of build failures.

However, HMR won't work unless `--hot` option is specified as well. The reason is that [webpack/bin/convert-argv.js](https://github.com/webpack/webpack/blob/master/bin/convert-argv.js#L385) only injects `HotModuleReplacementPlugin` if `--hot` option is set.

CLI help message is a bit better, but still worth mentioning that it must be `--hot` option set:

```bash
./node_modules/.bin/webpack-dev-server -h
...
--hot-only                 Do not refresh page if HMR fails          [boolean]
...
```

May be it's better to implicitly apply `--hot` if `--hot-only` mentioned

benwiley4000:
Agree with this - was about to open the same. Documentation implies `--hot-only` would enable hot module replacement and requiring `--hot` **in addition** seems  needlessly verbose. I can submit a PR to correct this behavior if desired?

benwiley4000:
Looks like this is actually handed [inside of webpack](https://github.com/webpack/webpack/blob/57825153e531a079eb0dbd2528e47f4cbafcc7f2/bin/convert-argv.js#L385) (confusing). So there are two options:

1. Assign `argv['hot']` to `true` before passing `argv` to `webpack/bin/convert-argv` if `argv['hot-only']` is true (seems messy).
2. Check additionally for the `hot-only` arg inside of `webpack/bin/convert-argv.js` (since we're already checking for `hot` this seems ok?).

I prefer option 2.

agoldis:
@benwiley4000 Thank for supporting this request, i am certain that option 2 + documentation fix will make it clearer.
It would be good, thought, to get an input from contributors :)

arjunkathuria:
this bug / issue still around ?

can get started on my contributions : )

arjunkathuria:
@shellscape i have a long weekend coming, will look into it : )

shellscape:
Closing issue as abandoned as it's more than 90 days stale

limitedAtonement:
[...]

PR #1039, opened 2017-08-13
explicit hot when hotOnly in options.

arjunkathuria:
### What kind of change does this PR introduce?
bugfix / refactor confusing behaviour

### Did you add or update the examples/?
No.

### Summary

When `hot-only` option is supplied, the `hot` option also needs to be set to true for the `hot-only` option to work as intended which for some users, creates confusion. 

This PR removes the need to turn both `hot` AND `hot-only` on for devServer, when specifying hot-only as it sets hot to true internally for the user.  

### Does this PR introduce a breaking change?
No.

### Other information
cc- @shellscape, were we looking for something like this ?

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -261,8 +261,13 @@ function processOptions(wpOpt) {
 	if(!options.hot)
 		options.hot = argv["hot"];
 
-	if(!options.hotOnly)
+	if(!options.hotOnly) {
 		options.hotOnly = argv["hot-only"];
+		// if hotOnly is true, explicitly set `hot` true.
+		if(options.hotOnly) {
+			options.hot = true;
+		}
+	}
 
 	if(!options.clientLogLevel)
 		options.clientLogLevel = argv["client-log-level"];


shellscape:
@arjunkathuria thanks for taking the time to put this together. please note that pull request templates typically aren't optional and it's generally not considered good form to remove their content. please edit the PR description using the template (https://github.com/webpack/webpack-dev-server/blob/master/.github/PULL_REQUEST_TEMPLATE.md) and we'll have a look at the PR afterward :)

arjunkathuria:
@shellscape updated.

shellscape:
@arjunkathuria thanks

I'm looking at [Server.js#L39](https://github.com/webpack/webpack-dev-server/blob/da188a6359b372d7885bde1fe204b5b835ff3466/lib/Server.js#L39) and [addDevServerEntrypoints.js#L9](https://github.com/webpack/webpack-dev-server/blob/85de417fd5592efc5448758bf550d7d34e670f60/lib/util/addDevServerEntrypoints.js#L9) and within webpack-dev-server there seems to be an explicit difference between the two. Overriding the value of `options.hot` doesn't seem wise, as it'll execute code the user may not have wanted. It could also be considered a breaking change.

According to [this comment](https://github.com/webpack/webpack-dev-server/issues/877#issuecomment-310925554) the real issue lies within webpack itself. Given all of this, it seems that there are two fairly clear solutions that lie outside of webpack-dev-server;

1. add `HotModuleReplacementPlugin` manually
2. file a pull request with the [main webpack repo](https://github.com/webpack/webpack/) to account for the `hot-only` command line arg with regard to injection of `HotModuleReplacementPlugin`

But it doesn't look like the best solution is going to be in this project.

arjunkathuria:
hupmh, alright, guess we could close this then ?

shellscape:
Sure, can do. If you find more info that suggests the solution should live here, we can reopen.

Issue #906, opened 2017-05-09
Missing [missing_subjectAltName] in ssl certificate (since Chrome 58)


kasperpeulen:
There has something changed since Chrome 58. The certificate misses subjectAltName.

```
This server could not prove that it is localhost; its security certificate is from [missing_subjectAltName]. This may be caused by a misconfiguration or an attacker intercepting your connection. Find out more.
```

![image](https://cloud.githubusercontent.com/assets/1035299/25848570/33a0371c-34bb-11e7-8b58-5ea9a07a81d0.png)

I made my own certificate with openssl, and providing subjectAltName, and now chrome accepts the certificate again:

```
openssl req \
    -newkey rsa:2048 \
    -x509 \
    -nodes \
    -keyout server.key \
    -new \
    -out server.crt \
    -subj /CN=dev.mycompany.com \
    -reqexts SAN \
    -extensions SAN \
    -config <(cat /System/Library/OpenSSL/openssl.cnf \
        <(printf '[SAN]\nsubjectAltName=DNS:localhost')) \
    -sha256 \
    -days 3650
```

kasperpeulen:
Seems to be fixed by:
https://github.com/webpack/webpack-dev-server/pull/857/files

shellscape:
Indeed this one has been resolved.

PR #1047, opened 2017-08-16
Add subjectAltName to self-signed certificate for https

### LINKED ISSUE #854: Self signed certificate no longer valid as of Chrome 58

paulca:
This fixes #854 and #906 by adding a subjectAltName matching
the commonName for the self-signed cert.

I've been using `webpack-dev-server` with Rails 5 and it's been quite frustrating to have to manually click "proceed" when using Chrome, even after trusting the self-signed cert.

This PR adds the requisite subjectAltName to the self-signed cert so that Chrome no longer displays the error message described at https://github.com/webpack/webpack-dev-server/issues/854 and https://github.com/webpack/webpack-dev-server/issues/906

Simply adds a config option to the self-signed cert.

No, but tested using the https example.

**Summary**

Using webpack-dev-server for Rails 5 development and repeatedly having to visit the server manually in Chrome before I can continue development.

No.

**Other information**
This PR shouldn't have any side-effects, other than fixing https in Chrome.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -386,7 +386,17 @@ function Server(compiler, options) {
 
 		if(!certExists) {
 			console.log("Generating SSL Certificate");
-			const attrs = [{ name: "commonName", value: "localhost" }];
+			const attrs = [
+				{ name: "commonName", value: "localhost" },
+				{
+					name: "subjectAltName",
+					value: {
+						altNames: [
+							{ type: 6, value: "localhost" }
+						]
+					}
+				}
+			];
 			const pems = selfsigned.generate(attrs, {
 				algorithm: "sha256",
 				days: 30,


shellscape:
Really appreciate taking the time to open a PR :heart: Unfortunately this is a duplicate of #987. Please check out the comments on that one.

Issue #1049, opened 2017-08-22
Uncaught Error: Cannot find module './log'


johan-lejdung:
**What is the current behavior?**
We recieve a warning in the console:
```
Uncaught Error: Cannot find module './log'.
    at webpackContextResolve (bundle.js:84181)
    at webpackContext (bundle.js:84176)
    at Object.log-level (bundle.js:78566)
    at SockJS.sock.onmessage (bundle.js:80519)
    at SockJS.EventTarget.dispatchEvent (bundle.js:43017)
    at bundle.js:82337
    at Array.forEach (<anonymous>)
    at SockJS._transportMessage (bundle.js:82335)
    at WebSocketTransport.EventEmitter.emit (bundle.js:3748)
    at WebSocketTransport.ws.onmessage (bundle.js:80623)
```

**What is the expected behavior?**
No Warning, we are not using hot-reloading.

**What we think is happening**
The warning is generated from this [line](https://github.com/webpack/webpack-dev-server/commit/6da2f38bb24b99fa0eceb10d5ee63259d02e6c8a#diff-895656aeaccff5d7c0f56a113ede9662R81). We are **not** using hot-reloading but it looks like the code think we are. 

If we call `hotCtx.keys()` on line 80 in the commit above we get this array:
```
["../package.json", "./console", "./formatters"]
```

**Please mention your webpack and Operating System version.**
MacOs Sierra
Webpack-dev-server: 2.7.1
Webpack: 3.5.5

johan-lejdung:
PR here: https://github.com/webpack/webpack-dev-server/pull/1050

PR #1050, opened 2017-08-22
Proposed fix for ./log module not found

johan-lejdung:
Bugfix of https://github.com/webpack/webpack-dev-server/issues/1049

No

**Summary**
This is a proposed fix of the https://github.com/webpack/webpack-dev-server/issues/1049
Instead of just assuming that the module ./log is loaded there is a check to see if it actually is.

It should not!

**Other information**

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -78,8 +78,9 @@ var onSocketMsg = {
 		sendMsg("StillOk");
 	},
 	"log-level": function(level) {
-		var hotCtx = require.context("webpack/hot", false, /^\.\/log$/);
-		if(hotCtx.keys().length > 0 && hotCtx.keys()["./log"]) {
+    var hotCtx = require.context("webpack/hot", false, /^\.\/log$/);
+    var contextKeys = hotCtx.keys();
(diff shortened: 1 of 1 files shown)


client/index.js
```diff
@@ -79,7 +79,7 @@ var onSocketMsg = {
 	},
 	"log-level": function(level) {
 		var hotCtx = require.context("webpack/hot", false, /^\.\/log$/);
-		if(hotCtx.keys().length > 0) {
+		if(hotCtx.keys().length > 0 && hotCtx.keys()["./log"]) {
```
shellscape: I'd like to see a small modification here please. 

```js
var contextKeys = hotCtx.keys();
if(contextKeys.length && contextKeys["./log"]) {
```

shellscape:
@johan-lejdung thanks for the PR! we have one small change request and then we'll be happy to merge.

client/index.js
johan-lejdung: Absolutely!

shellscape:
Thanks!

Issue #998, opened 2017-07-19
Request the running port number using port 0


gmschouten:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
``` javascript
server.listen(0, '127.0.0.1', () => {
 // port will automaticly assign to a free port.
});
```

**What is the expected behavior?**
``` javascript
server.listen(0, '127.0.0.1', () => {});

server.once('listening', () => {
   // while running we want to request the used port number.
   const port = server.address().port;
});

```
**If this is a feature request, what is motivation or use case for changing the behavior?**
I use webpack-dev-server to serve my application for browser-test. I want webpack to use a free port and after serving to tell my browser-test where to find the test-pages. 

**Please mention your webpack and Operating System version.**
webpack: 3.3.0
webpack-dev-server: 2.5.1

shellscape:
We'd welcome a PR to implement this feature.

shellscape:
this should be resolved by #1054. if it's not, please give me a ping.

PR #1053, opened 2017-08-24
Include the listening port number in the server object

doughamlin:
feature

no

**Summary**
Exposes the listening port number in the server object. 
Implements request in #998.

No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -484,6 +484,7 @@ Server.prototype.checkHost = function(headers) {
 
 // delegate listen call and init sockjs
 Server.prototype.listen = function(port, hostname) {
+	this.listenPort = port;
 	this.listenHostname = hostname;
 	const returnValue = this.listeningApp.listen.apply(this.listeningApp, arguments);
 	const sockServer = sockjs.createServer({


doughamlin:
I just realized this doesn't fully address the request in #998 because there is no `listening` event. It is easy enough to extend Server from EventEmitter and emit the `listening` event when the server starts listening, but I question the need for that when the server object is accessible inside the `server.listen` callback.

shellscape:
Much appreciate you taking the time to open a PR 🍻  however #998 seems to me to request automatic port assignment, rather than just exposing the port assigned. I'm not sure you're on the same page as the issue author here.

doughamlin:
@shellscape Yeah, I don't know why I thought portfinder was already being used in lib/Server.js already. It is used in bin/webpack-dev-server.js and could be added to Server.js. Would you want to remove it from webpack-dev-server.js in that case?

shellscape:
closing in favor of #1054. thanks again for the PR.

@doughamlin I _think_ portfinder is still needed to check to see if the specified port is in use.

Issue #106, opened 2015-02-13
CLI doesn't behave like Node API


gaearon:
A lot of times when somebody has problems with webpack-dev-server CLI, I suggest them use Node API instead and their problem disappears. Here's two recent issues I've seen.

Take this [sample project](https://github.com/edygar/react-hot-test) from https://github.com/gaearon/react-hot-loader/issues/77.

Run `node ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --hot --port 8080`. Change a file. HMR will work _but_ will output `Uncaught RangeError: Maximum call stack size exceeded` on every update.

Then add a `server.js` that should by all means be identical to the call above:

``` js
var webpack = require('webpack');
var WebpackDevServer = require('webpack-dev-server');
var config = require('./webpack.config');

new WebpackDevServer(webpack(config), {
  hot: true
}).listen(8080, 'localhost');
```

Magically, this doesn't work **at all**. The reason is because WebpackDevServer doesn't read `publicPath` from Webpack config. I always end up passing it manually. Why is this necessary?

Okay, let's pass it manually:

``` js
var webpack = require('webpack');
var WebpackDevServer = require('webpack-dev-server');
var config = require('./webpack.config');

new WebpackDevServer(webpack(config), {
  hot: true,
  publicPath: config.output.publicPath
}).listen(8080, 'localhost');
```

Voila! `RangeError` is gone.
1. Why does CLI and Node API behave differently? One gives `RangeError`, another one works.
2. Why won't CLI infer `publicPath` from Webpack config?

sokra:
The node.js API has no access to the webpack options. You just pass an `Compiler` object.

This has the effect that you have to do many things manually, i. e. `publicPath` or adding the `HotModuleReplacementPlugin`.

The CLI can do this things because it has access to the webpack options.

gaearon:
Is there a reason Compiler can't expose its options?
Is there a reason we don't let WebpackDevServer accept options instead of compiler, and create compiler itself?

sokra:
> Is there a reason Compiler can't expose its options?

It does expose it, but by design you shouldn't access it.

> Is there a reason we don't let WebpackDevServer accept options instead of compiler, so it creates compiler itself (and has access to options)?

by design. This way we can write other compilers that work with the webpack-dev-server, i. e. the MultiCompiler...

gaearon:
We can make `publicPath` required so people don't omit it by mistake in dev server config

[...]

PR #1055, opened 2017-08-28
following #106, makes options.publicPath required via the API. alphabetizes optionsSchema

shellscape:

**BREAKING CHANGE** Consistency update following comments on #106.

No need

**Summary**
#106 outlines a discrepancy of expectations in using the module API directly versus the CLI with regard to `publicPath`. This changes the options schema to make `publicPath` a required option and will fail gracefully if the option is not provided.

This PR also reorganizes the optionsSchema.json file to alphabetize the `properties` property and put more relevant meta at the beginning of the file.

**YES** Users previously not specifying a `publicPath` in API options will now have to - though I'm not sure if works without doing so to begin with.

**Other information**

This will likely need to release in a major version update as it may hose some implementations.


shellscape:
Considering waiting on this one until 3.0.0.

PR #1061, opened 2017-08-31
Add "mimeTypes" option to schema

xiaoyann:

Bugfix

No need

**Summary**

mimeTypes is a valid option from webpack-dev-middleware, but it can't  pass validation

No

**Other information**

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -18,6 +18,7 @@ const historyApiFallback = require('connect-history-api-fallback');
 const selfsigned = require('selfsigned');
 const sockjs = require('sockjs');
 const spdy = require('spdy');
+const mime = require('mime');
 const webpack = require('webpack');
 const webpackDevMiddleware = require('webpack-dev-middleware');
 const OptionsValidationError = require('./OptionsValidationError');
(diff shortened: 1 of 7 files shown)


shellscape:
Please update the tests and CLI options as well.

xiaoyann:
ok

xiaoyann:
need not to update the CLI options

shellscape:
@xiaoyann why don't you think we need to add this option to the CLI?

xiaoyann:
@shellscape Because "mimeTypes" should be an object like "headers" option. So, you can't use by CLI. you should define it in "devServer".

shellscape:
That's a rather subjective opinion. Thanks very much for taking the time to put together a PR for this, but we're going to pass on adding this feature at the moment. Instead, we'll be implementing a pass-through API option that will allow people to pass any options to sub-components that they'd like without having to modify the options schema.

xiaoyann:
Thank you for your guidance and look forward to this feature.

olsonpm:
@shellscape 

> Instead, we'll be implementing a pass-through API option that will allow people to pass any options to sub-components that they'd like without having to modify the options schema

Is this already being worked on?  I don't see relevant code in v3, another branch nor a PR.  I will start the work if nobody's started.  I got here due to needing access to the mimeTypes option.

PR #1063, opened 2017-08-31
Print webpack console to browser console

Wernerson:
feature

yes

**Summary**
As discussed in issue #1032, the webpack console now gets displayed in the browser console if the progress option is set to true.

It shouldn't ;)

**Other information**
None


Wernerson:
There were some merge issues. Will check them this weekend and make the branch ready to merge 👍

shellscape:
I'd also remove the clearing of the console. Best to leave that up to the users to do manually.

Wernerson:
Done and ready to merge 👍

shellscape:
@Wernerson please update the PR description to remove the bit about clearing the console and we'll get this merged.

lib/Server.js
```diff
@@ -52,8 +53,15 @@ function Server(compiler, options) {
   const invalidPlugin = () => {
     this.sockWrite(this.sockets, 'invalid');
   };
+  const progressPlugin = new webpack.ProgressPlugin((percent, msg, addInfo) => {
```
shellscape: should probably only load this plugin if `options.progress` is true as well

Wernerson:
@shellscape Condition added and description edited. Codecov fails though...

shellscape:
I'm going to bypass the code coverage this time. The refactor for v3 should help boost those numbers, and the config for that needs to ignore client files until we figure out a good way to test the client. Thanks for getting this in.

lib/Server.js
```diff
@@ -52,6 +53,15 @@ function Server(compiler, options) {
   const invalidPlugin = () => {
     this.sockWrite(this.sockets, 'invalid');
   };
+  if (this.progress) {
+    const progressPlugin = new webpack.ProgressPlugin((percent, msg, addInfo) => {
+      percent = Math.floor(percent * 100);
+      if (percent === 100) msg = 'Compilation competed';
```
sokra: `Compilation competed` -> `Compilation completed`

lib/optionsSchema.json
```diff
@@ -97,6 +97,14 @@
...
       ]
     },
+    "progress": {
+      "description": "Shows compilation progress in browser console.",
+      "anyOf": [
+        {
+          "type": "boolean"
+        }
+      ]
+    },
```
sokra:
``` js
"progress": {
  "description": "Shows compilation progress in browser console.",
  "type": "boolean"
 },
```

client/index.js
```diff
@@ -122,6 +123,14 @@ const onSocketMsg = {
       }
     }
   },
+  progress(progress) {
+    if (typeof document !== 'undefined') {
+      useProgress = progress;
+    }
+  },
+  'progress-update': function progressUpdate(data) {
+    if (useProgress) log.info(`[WDS] ${data.percent}% - ${data.msg}.`);
```
sokra: Isn't this super spammy?

shellscape: given that it's an opt-in and the browser console doesn't have control characters, I don't believe it is.

lib/Server.js
shellscape: I'll clean up that typo in a bit

pelotom:
I was looking forward to this feature but it doesn't seem to be working, what all is needed to enable it? I set `progress: true` in both the dev server config and in the webpack config, but no dice. FWIW I'm using an ejected `create-react-app`.

shellscape:
@pelotom I just checked the example app for progress and it appears to be working for me.

Wernerson:
@pelotom Whats your loglevel set to? Has to be on info at least.

g0t4:
I'd like to use this feature but on fast builds it is super chatty, and so I disabled it and will probably forget about it when the build starts slowing down :)

How about throttle the console messages if they come in rapidly? 

- That way if the build is slow you get a good indication of progress. 
- And if the build is somewhat slow you get fewer messages. 
- And if the build is super fast you maybe see one or two of these or maybe none. 

Or maybe a simple way to hook into this as a user and apply my own filtering?

Wernerson:
@g0t4 Good idea, I'll take a look into it.

PR #1068, opened 2017-09-04
Add feature to disable hotReloading with query string

edi9999:
Feature

No, but you could use any sample and use the queryParam : `?disableAutoreload=true` to 

**Summary**

See corresponding issue : https://github.com/webpack/webpack-dev-server/issues/936

This option can be used if you want to use the same platform for developping and to run automated tests (the tests should not be impacted by a code change whenever they are running). 

No

**Other information**

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -22,6 +22,11 @@ function getCurrentScriptSource() {
 }
 
 let urlParts;
+let hotReload = true;
+if (typeof window !== 'undefined') {
+  const qs = window.location.search.toLowerCase();
+  hotReload = qs.indexOf('hotreload=false') === -1;
+}
 if (typeof __resourceQuery === 'string' && __resourceQuery) {
   // If this bundle is inlined, use the resource query to get the correct url.
   urlParts = url.parse(__resourceQuery.substr(1));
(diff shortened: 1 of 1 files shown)


edi9999:
Any updates ?
ping @shellscape

shellscape:
@edi9999 please do us a solid and wait about a week before pings 😄 I promise we see all activity here and will respond/comment as soon as possible.  I'm currently preparing for hurricane irma and will be returning next week.

edi9999:
Ok ! I hope everything will be OK for you

client/index.js
```diff
@@ -22,6 +22,10 @@ function getCurrentScriptSource() {
 }
 
 let urlParts;
+let hotReloading = true;
+if (typeof window !== 'undefined') {
+  hotReloading = window.location.search.indexOf('disableAutoReload') === -1;
```
shellscape: not a fan of negative-options, meaning a positive value will result in the negative reassignment of another variable. let's also keep the variable names the same as the querystring parameter. so;

```js
let hotReload = true;
...
hotReload = !qs.indexOf('hotreload=false');
```

the code was trimmed for brevity, but those are the changes I'd like to see

edi9999: Fixed :-)

Issue #1064, opened 2017-09-01
LogLevel conflict. Always setting level to INFO


attenzione:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**
I'm using `loglevel` in project and it conflicts with `webpack-dev-server`, because of `setLevel`.
i'm setting different level, but webpack-dev-server always set it to INFO.

**If the current behavior is a bug, please provide the steps to reproduce.**

in `main.js` (common chunk)
```
import Logger from 'loglevel';
Logger.setLevel(Logger.levels.TRACE); 
console.log(Logger.getLevel()); # => 0, TRACE
# after some time, probably websocket get connection
setTimeout(function () { console.log(Logger.getLevel()); }, 3000); # => 2, INFO
```

**What is the expected behavior?**
webpack-dev-server must not change global `loglevel`, instead must use `log.getLogger('webpack-dev-server')` to create own logger

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention your webpack and Operating System version.**
webpack 3.5.5
webpack-dev-server 2.7.1

shellscape:
closed by #1070

PR #1070, opened 2017-09-05
Fixes issue #1064 by switching to a named logger

bdurrer:
bugfix

no

**Summary**
This solves issue #1064 
So far, the client JS of webpack-dev-server used the root logger and also changed the configuration of the root logger. This results in webpack overwriting the application's logging configuration.
The PR switches to a named logger, as suggested in the issue description.

I don't think it could break an app.

However, if an application uses loglevel and expects webpack-dev-server to configure the root logger, it might have an different log level than expected. 
The solution for the app would be to call [loglevel's `setLevel `or `setDefaultLevel`](https://github.com/pimterry/loglevel#documentation) (which is how you should do it anyway).

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -5,7 +5,7 @@
 
 const url = require('url');
 const stripAnsi = require('strip-ansi');
-const log = require('loglevel');
+const log = require('loglevel').getLogger('webpack-dev-server');
 const socket = require('./socket');
 const overlay = require('./overlay');
 


shellscape:
@bdurrer looks good. please sign the CLA (see first comment) and we can move forward

bdurrer:
@shellscape 
I did sign the CLA, but I used the wrong email. Adding them to the GitHub account did not help, nor can I change the js-cla data anymore after submitting 😒

bdurrer:
Resetting and commiting the same thing again worked.

shellscape:
ah perfect, thanks!

PR #1079, opened 2017-09-13
Get the port from the options object

binhqx:
bugfix

No

**Summary**
When listeningApp is not supplied, use the port from the options object. I think this functionally was broken by PR:  https://github.com/webpack/webpack-dev-server/pull/1054/files#diff-07cfeb70272594b38ac9139434c94502

No

**Other information**

### DIFF
--- a/lib/util/createDomain.js
+++ b/lib/util/createDomain.js
@@ -6,7 +6,7 @@ const internalIp = require('internal-ip');
 
 module.exports = function createDomain(options, listeningApp) {
   const protocol = options.https ? 'https' : 'http';
-  const appPort = listeningApp ? listeningApp.address().port : 0;
+  const appPort = listeningApp ? listeningApp.address().port : options.port.toString();
   const port = options.socket ? 0 : appPort;
   const hostname = options.useLocalIp ? internalIp.v4() : options.host;
 


lib/util/createDomain.js
```diff
@@ -6,7 +6,7 @@ const internalIp = require('internal-ip');
 
 module.exports = function createDomain(options, listeningApp) {
   const protocol = options.https ? 'https' : 'http';
-  const appPort = listeningApp ? listeningApp.address().port : 0;
+  const appPort = listeningApp ? listeningApp.address().port : options.port.toString();
```
shellscape: this will break dynamic port assignment, port `0` was intentional there for that reason.

shellscape:
What side effects is this causing?

binhqx:
@shellscape 
I am using https://github.com/JeffreyWay/laravel-mix to configure webpack in a laravel project. The hot module loader is failing to connect back to the dev server because the port number is missing.

image

shellscape:
@binhqx this can be resolved by stubbing the `listeningApp` parameter in `addDevServerEntryPoints.js`:

```js
{
  address() {
    return { port: options.port };
  }
}
```

And should allow everything else to function normally. The CLI is doing a few things a little out of order. If we find that we have to stub in more than one place, we'll refactor. But since that's a util function it shouldn't be a big deal.

shellscape:
@binhqx would you like to do that in your PR or would you like me to?

binhqx:
@shellscape You are much more familiar with the code. You can fix it your way.

PR #1088, opened 2017-09-14
Fix Promise version of webpack config

richtera:

When using a Promise as the webpack config the dev server version 2.8.1 doesn't startup due to using the Promise inside of StartDevServer instead of the resolved webpack config.

No

**Summary**

Server doesn't work when using a Promise as the webpack config.

**Other information**


richtera:
This is to patch the latest official npm registry entry for 2.8.1 which is why it's not on the master branch. I have not tested master, but that part of the code is quite different and probably working there.

shellscape:
@richtera this snuck in just before your PR 8f897c575b3c98c3d08b876a4cc4c23c7b48157b. thanks a bunch for putting this together though!

PR #1096, opened 2017-09-16
Fix check for webpack/hot/log when setting HMR log level

lbogdan:
Bugfix.

Not needed.

**Summary**
Some time ago I authored [PR #926](https://github.com/webpack/webpack-dev-server/pull/926). Which, strangely enough, caused [issue #1049](https://github.com/webpack/webpack-dev-server/issues/1049).

The original code was
```js
  if(hotCtx.keys().length > 0) {
    hotCtx("./log").setLogLevel(level);
  }
```
Somehow I was under the impression that `hotCtx.keys()` can only return `[]` or `["./log"]`, and with this assumption, the check `hotCtx.keys().length > 0` is correct. For some strange reason, as per issue #1049, it returned `["../package.json", "./console", "./formatters"]`, which doesn't match the regex at all, so I don't really know what happened there. In this case, of course, the check becomes wrong.

Now, this is the fix for #1049 in [PR #1050](https://github.com/webpack/webpack-dev-server/pull/1050)
```js
  if(hotCtx.keys().length > 0) {
    var contextKeys = hotCtx.keys();
    if(contextKeys.length && contextKeys["./log"]) {
      ...
```
which doesn't work, because `contextKeys` is an array, so `contextKeys["./log"]` will always be `false`.

So the correct check would be `if (hotCtx.keys().indexOf('./log') !== -1) { ...`.

No.

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -92,8 +92,7 @@ const onSocketMsg = {
   },
   'log-level': function logLevel(level) {
     const hotCtx = require.context('webpack/hot', false, /^\.\/log$/);
-    const contextKeys = hotCtx.keys();
-    if (contextKeys.length && contextKeys['./log']) {
+    if (hotCtx.keys().indexOf('./log') !== -1) {
       hotCtx('./log').setLogLevel(level);
     }
     switch (level) {


lbogdan:
Cc @johan-lejdung.

johan-lejdung:
Ah damn, of course it's an array.. Don't know what I was thinking :)  If I could, I would approve this change!

The build seems to fail, but not related to this file. There seem to be a linting error in Server.js:

```
/home/travis/build/webpack/webpack-dev-server/lib/Server.js
  142:17  error  Strings must use singlequote  quotes
✖ 1 problem (1 error, 0 warnings)
  1 error, 0 warnings potentially fixable with the `--fix` option.
```

lbogdan:
What's even stranger is that it fails only on node 6 and 4, 8 is ok. :smile:

johan-lejdung:
Quickly looked through the changelog for node 8 and there were a lot of Array changes, maybe they allow such access now.

lbogdan:
That's an eslint error, so I don't see any reason why it would work differently in node 8 vs. node 4 or 6. Anyway, fixed it in https://github.com/webpack/webpack-dev-server/pull/1098.

lbogdan:
Any news on this?

shellscape:
@lbogdan you're a rock star for getting these fixes in. very much appreciated! but I beg of you some additional patience please 😄  we're all human on this side, and to say the last two weeks have been personally trying would be an understatement. please note that it may take a week or so to get a patch release out. I'll be out of town and I don't like to release unless I'm around to put out any fires that pop up.

lbogdan:
@shellscape Thank you for your kind words, they might not necessarily reflect the reality! :smile:
That was just a follow up, sorry if I came off as impatient, I was just curious about an update. I know OSS is made out of people's personal time, and I really appreciate these people and their hard work!

PR #1097, opened 2017-09-16
Make error overlay translucent

lbogdan:
Error overlay translucence.

Not needed.

**Summary**
Being used to `webpack-hot-middleware`'s error overlay, which has an [alpha of 0.85](https://github.com/glenjamin/webpack-hot-middleware/blob/master/client-overlay.js#L6), it is a bit misleading when getting a completely opaque overlay. This change makes the error overlay's background the same as `webpack-hot-middleware`'s.

No.

### DIFF
--- a/client/overlay.js
+++ b/client/overlay.js
@@ -50,7 +50,7 @@ function addOverlayDivTo(iframe) {
   div.style.bottom = 0;
   div.style.width = '100vw';
   div.style.height = '100vh';
-  div.style.backgroundColor = 'black';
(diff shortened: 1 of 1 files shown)


shellscape:
@lbogdan we can merge this, but as an FYI the overlay and such is going to get a complete makeover for 3.0.0.

lbogdan:
@shellscape Nice, I hope it will be translucent!

It would be nice if you could merge this, so that I don't have to change it in `node_modules` in every new project. :smile:

PR #1098, opened 2017-09-16
Fix quotes

lbogdan:
Quotes fix.

Not needed.

**Summary**
Double quotes in `Server.js` made eslint fail (strangely enough, only in node 4 and 6).

No.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -103,9 +103,8 @@ function Server(compiler, options) {
 
   app.get('/webpack-dev-server', (req, res) => {
     res.setHeader('Content-Type', 'text/html');
-    /* eslint-disable quotes */
     res.write('<!DOCTYPE html><html><head><meta charset="utf-8"/></head><body>');
-    const outputPath = this.middleware.getFilenameFromUrl(options.publicPath || "/");
+    const outputPath = this.middleware.getFilenameFromUrl(options.publicPath || '/');
     const filesystem = this.middleware.fileSystem;
 
     function writeDirectory(baseUrl, basePath) {
@@ -141,7 +140,6 @@ function Server(compiler, options) {
       });
       res.write('</ul>');
     }
-    /* eslint-enable quotes */
     writeDirectory(options.publicPath || '/', outputPath);
     res.end('</body></html>');
   });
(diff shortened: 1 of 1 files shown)


lbogdan:
Please merge so I can rebase https://github.com/webpack/webpack-dev-server/pull/1096 and https://github.com/webpack/webpack-dev-server/pull/1097, whose tests are currently failing because of this.

shellscape:
@lbogdan this is really strange, as neither Travis nor my local install fail on this. the `quotes` rule is disabled for that block at [line 110](https://github.com/webpack/webpack-dev-server/pull/1098/files#diff-15fb51940da53816af13330d8ce69b4eR110) and enabled on [line 144](https://github.com/webpack/webpack-dev-server/pull/1098/files#diff-15fb51940da53816af13330d8ce69b4eR144)

shellscape:
@lbogdan OK so you stumbled across a false-positive here. Your change fixed the immediate issue, but your change doesn't address the actual issue at play here: Node 4 and Node 6 don't respect package-lock.json. Node 4/6 were installing eslint@4.7.0, which fails as you're seeing. Node 8 however, doesn't fail because it's respecting the package-lock.json versions, which is eslint@4.6.1. 

The work to fix this should address the changes in eslint between 4.7.0 and 4.6.1 in terms of config, before changing the code.

lbogdan:
@shellscape I knew this doesn't fix the root problem, but I didn't have time to investigate further.

Even so, for the sake of consistency, I think the `eslint-disable` directives can be removed and the quotes replaced, like in this PR.

shellscape:
@lbogdan I understand what you're saying, but I'm not a fan of band-aiding when a deeper issue exists. Another collaborator may disagree (and I'd be cool with that) but I can't approve and merge this until the greater issue with ESLint is resolved. Holding up a few PRs until they release a patch is OK by me.

lbogdan:
@shellscape I am completely with you on fixing the ESLint issue, I didn't say for a second to not fix it, I hope it didn't come through that way.

Now, back to that block of code, I really don't see the reason for those two `eslint-disable/enable` directives and using single and double quotes interchangeably.

I mean, can't we fix both?

shellscape:
@lbogdan yes, and I'm glad you arrived at that conclusion. to merge this (and to make this a full fix) the disable/enable directives should be removed and the other double quote on Line 131 should be cleaned up. As I practice I'll defer to you as to whether or not you'd like to add that to this PR (please do!).

shellscape:
Thanks for getting that in.

lbogdan:
@shellscape Thanks! I also rebased my two other PRs.

PR #1108, opened 2017-09-20
feat: deprecate setup in favor of before and after hooks

### LINKED ISSUE #869: Is the middlewares ordering correct?
As per this line
https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L320

the middlewares ordering is `["setup", "headers", "middleware"]`. Due to this if i have a middleware for * in setup, it will consume all wepack output routes as well. For e.x

`setup: function(app) {
    app.use('*', function(req, res) {
           res.render('index');
   })
}`

this configuration will server index html even for all webpack bundle paths.But if the middleware ordering is  `[ "headers", "middleware", "setup"]`, this problem won't occur.

Wepack: 2.2.1
Webpack-dev-server: 2.3.0

mrtnbroder:

feature

yes, `examples/node-api-middleware`

**Summary**

fixes #869

Not a breaking change, but a deprecation.

**Other information**
see #869

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -321,9 +321,17 @@ function Server(compiler, options) {
       }
     },
 
+    before: () => {
+      if (typeof options.before === 'function') { options.before(app, this); }
+    },
+
     middleware: () => {
       // include our middleware to ensure it is able to handle '/index.html' request after redirect
       app.use(this.middleware);
+    },
+
+    after: () => {
+      if (typeof options.after === 'function') { options.after(app, this); }
     },
 
     headers: () => {
@@ -335,11 +343,13 @@ function Server(compiler, options) {
     },
 
     setup: () => {
+      log('Using "setup" is deprecated and will be removed in the next major version. Please use the "before" and "after" hooks instead.');
+      log('If "setup" was working fine for you until now, simply replace it with "before"');
       if (typeof options.setup === 'function') { options.setup(app, this); }
     }
   };
 
-  const defaultFeatures = ['setup', 'headers', 'middleware'];
+  const defaultFeatures = ['before', 'setup', 'headers', 'middleware'];
   if (options.proxy) { defaultFeatures.push('proxy', 'middleware'); }
   if (contentBase !== false) { defaultFeatures.push('contentBaseFiles'); }
   if (options.watchContentBase) { defaultFeatures.push('watchContentBase'); }
@@ -351,6 +361,7 @@ function Server(compiler, options) {
   if (contentBase !== false) { defaultFeatures.push('contentBaseIndex'); }
   // compress is placed last and uses unshift so that it will be the first middleware used
   if (options.compress) { defaultFeatures.unshift('compress'); }
+  if (options.after) { defaultFeatures.push('after'); }
 
   (options.features || defaultFeatures).forEach((feature) => {
     features[feature]();
--- a/examples/node-api-middleware/server.js
+++ b/examples/node-api-middleware/server.js
@@ -9,7 +9,7 @@ const server = new WebpackDevServer(compiler, {
   stats: {
     colors: true
   },
-  setup(app) {
+  before(app) {
     app.use((req, res, next) => {
       console.log(`Using middleware for ${req.url}`);
       next();
--- a/lib/optionsSchema.json
+++ b/lib/optionsSchema.json
@@ -274,6 +274,14 @@
       "description": "Exposes the Express server to add custom middleware or routes.",
       "instanceof": "Function"
     },
+    "before": {
+      "description": "Exposes the Express server to add custom middleware or routes before webpack-dev-middleware will be added.",
+      "instanceof": "Function"
+    },
+    "after": {
+      "description": "Exposes the Express server to add custom middleware or routes after webpack-dev-middleware got added.",
+      "instanceof": "Function"
+    },
     "stats": {
       "description": "Decides what bundle information is displayed.",
       "anyOf": [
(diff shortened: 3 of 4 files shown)


lib/Server.js
```diff
@@ -323,7 +323,9 @@ function Server(compiler, options) {
 
     middleware: () => {
       // include our middleware to ensure it is able to handle '/index.html' request after redirect
+      if (typeof options.before === 'function') { options.before(app, this); }
```
shellscape: preferably we'd follow the same pattern established on line 342 with the `setup` middleware. `'before'` would be inserted into that array before `setup` and `after` would be added to the array before line 355

mrtnbroder: Fixed!
I've changed the example for middleware as well, saw that one was using the `setup` hook which is now the `before` hook.

lib/Server.js
```diff
@@ -335,16 +343,18 @@ function Server(compiler, options) {
...
     setup: () => {
+      log('Using "setup" is deprecated and will be removed in the next major version. Please use the "before" and "after" hooks instead.');
+      log('If "setup" was working fine for you until now, simply replace it with "before"');
       if (typeof options.setup === 'function') { options.setup(app, this); }
     }
   };
 
-  const defaultFeatures = ['setup', 'headers', 'middleware'];
-  if (options.proxy) { defaultFeatures.push('proxy', 'middleware'); }
+  const defaultFeatures = ['before', 'setup', 'headers', 'middleware', 'after'];
```
shellscape: `after` shouldn't be here. same goes for the changes on lines 353, 357. rather, it should be pushed to `defaultFeatures` before line 365 - just before the `forEach`

mrtnbroder: Fixed.

lib/Server.js
```diff
@@ -335,11 +343,13 @@ function Server(compiler, options) {
...
 
     setup: () => {
+      log('Using "setup" is deprecated and will be removed in the next major version. Please use the "before" and "after" hooks instead.');
+      log('If "setup" was working fine for you until now, simply replace it with "before"');
       if (typeof options.setup === 'function') { options.setup(app, this); }
     }
   };
 
-  const defaultFeatures = ['setup', 'headers', 'middleware'];
+  const defaultFeatures = ['before', 'setup', 'headers', 'middleware'];
```
bdwain: @shellscape @mrtnbroder doesn't setup being in this array mean that everyone will see the deprecation warning no matter what? Shouldn't it not be in there but one of the things below this say `if(options.setup) defaultFeatures.push('setup')` or something? (or maybe keep it in but if options.setup is absent then remove it).

It seems that the loop on line 367 will always run the deprecation warning. 

I'm asking because when i upgraded to 2.9 i noticed the deprecation warning even though I don't use this hook at all (though it may just be that a dependency of mine needs upgrading. I'm not totally sure)

bdwain: either that or putting the logs on lines 346-347 inside the if on line 348

bdwain: I made #1120 for this

cletusw:
It took me a while to track down where this error was coming from. Please add a link to the changelog in the error message (or at least include the text `webpack-dev-server` somewhere).

cletusw:
@shellscape Sorry, I didn't realize that you were also on this issue. Sorry for the double post.

cletusw:
As for error vs. log, I guess my original message wasn't entirely clear. What I meant was, I saw the log text, but I had no idea how to fix it since it wasn't at all clear which `setup` it was referring to. My mocha test files? Some random `setup` method in my code? A third-party library? My ask was that the deprecation message at least include the text `webpack-dev-server` so I know *where* the message is coming from. Ideally, it would include a link to a page somewhere on the web explaining what the deprecation entails. Sorry I wasn't clear originally.

shellscape:
> logging in the module will be getting an overhaul in v3.

cletusw:
> logging in the module will be getting an overhaul in v3.

Not sure what you mean. Are you saying the deprecation warning will be ambiguous until v3 at which time it will change to say `webpack-dev-server: "setup" is deprecated...`?

bdwain:
agreed about the log prefix. most log messages I've seen from webpack dev server start with [WDS]. I spent about 20 minutes this morning googling and searching to see where the warning came from because it's just a random error in the console.

It seems like a simple enough change that it shouldn't have to wait until V3's log overhaul. People are going to start seeing the deprecation warning soon and have no idea where it's coming from. @shellscape Would a PR to add [WDS] be approved?

shellscape:
@bdwain very much appreciate the initiative, but we're about to announce a new feature freeze to focus on v3.0.0, and logging throughout the entire module needs improvement as it has the same/similar shortcomings. you can follow progress on v3 [here](https://github.com/webpack/webpack-dev-server/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Version+3%22). I won't comment yet on what that will entail because we haven't had time to vet all of the available options. in the meantime we ask for patience.

cletusw:
I would suggest removing this message then. As it is it is unactionable and
will generate lots of confusion.

shellscape:
@cletusw I think you've made your position abundantly clear 😄  The change will stand as-is for the moment. I'll likely abstain from commenting on this PR moving forward, but please feel free to continue the discussion.

Vandivier:
following the newly updated [example](https://github.com/webpack/webpack-dev-server/blob/master/examples/node-api-middleware/server.js) I receive OptionsValidationError on snippet 1 below while snippet 2 is fine.

Snippet 1
```
const options = {
  contentBase: './dist',
  hot: true,
  stats: {
    colors: true
  },
  before(app) {
    app.use('/src/static', express.static('static'));
  }
};
```

Snippet 2
```
const options = {
  contentBase: './dist',
  hot: true,
  stats: {
    colors: true
  },
  setup(app) {
    app.use('/src/static', express.static('static'));
  }
};
```

Issue #1116, opened 2017-09-26
public option doesn't allow explicitly to set protocol


robertaistleitner:
* Operating System: irrelevant
* Node Version: irrelevant
* NPM Version: irrelevant
* webpack Version: 3.6.0
* webpack-dev-server Version: 2.8.2

- [ ] This is a **feature** request
- [x] This is a **bug**

### Code

```js
// webpack.config.js
config.devServer = {
  ...,
  inline: true,
  public: 'https://dev.local/sockjs-node',
  publicPath: 'https://dev.local/'
};
```

### Expected Behavior
I'm injecting an entrypoint to a iframe which has `about:blank` as source so that I have full control of injected DOM elements from the parent window. Problem is that therefore `window.location.protocol` has `about:` as value, which is obviously not `https:` as the parent window. I tried to force the socket url by using the `public` parameter in the options but it's missing support to force the protocol in the URL as well.

### Actual Behavior
Protocol in `public` option is not supported (it's infered from the `https` property). Since I'm utilizing a nginx reverse proxy which does the https stuff, I don't want to enable the `https` options in dev-server.

### For Bugs; How can we reproduce the behavior?
I'll create a PR in a few minutes which should show what's the problem and provide a fix for it..

shellscape:
FWIW this isn't a bug, the behavior is by design so it really doesn't qualify as a bug. Rather, this is a request for a modification of the current behavior. I'll update the issue template to provide for that option in the future.

> I'll create a PR in a few minutes which should show what's the problem and provide a fix for it..

Your PR didn't really explain what the problem here is, but we can infer the issue in code from the PR changeset. This issue is kind of all over the place. The bit about `about:blank` doesn't provide a whole lot of context but...

>  I tried to force the socket url by using the public parameter in the options but it's missing support to force the protocol in the URL as well.

...is where the meat of this issue lies. I'll header over to your PR (#1117) and continue the discussion. Moving forward, if you're planning on creating a PR immediately after your issue, you can skip the issue and just add all info to the PR.

shellscape:
Closed in #1117

PR #1117, opened 2017-09-26
allow explicitly setting the protocol from the public option

robertaistleitner:
`public` options now allows forcing the protocol for the socket connection.

No.

**Summary**
There are issues if using entrypoints in iframes, more infos in issue https://github.com/webpack/webpack-dev-server/issues/1116.

No.


lib/util/createDomain.js
```diff
@@ -11,9 +11,13 @@ module.exports = function createDomain(options, listeningApp) {
   const hostname = options.useLocalIp ? internalIp.v4() : options.host;
 
   // the formatted domain (url without path) of the webpack server
-  return options.public ? `${protocol}://${options.public}` : url.format({
-    protocol,
-    hostname,
-    port
-  });
+  if (options.public) {
+    return options.public.indexOf('://') > 0 ? `${options.public}` : `${protocol}://${options.public}`;
```
shellscape: this is not a good way to check a url for a protocol. please use [`url.parse`](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost) instead, and examine the result for a protocol.

robertaistleitner:
Hey guys, just checked why the tests failed (they work on my machine), and it seems the timeout with 2 seconds is a little too low, is there a possibility to increase it? The test case which fails is generating a ssl certificate (option `https` is set to true for server).

client/index.js
```diff
@@ -219,6 +219,15 @@ function reloadApp() {
     }
   } else {
     log.info('[WDS] App updated. Reloading...');
-    self.location.reload();
+    let rootWindow = self;
+    // use parent window for reload (in case we're in an iframe with no valid src)
+    do {
```
shellscape: this is almost never a good idea in the browser - on the off chance that the `while` condition is not met, this will lock the browser forcing a close. please use `setInterval` in conjunction with `clearInterval`

lib/util/createDomain.js
```diff
@@ -10,8 +10,12 @@ module.exports = function createDomain(options, listeningApp) {
   const port = options.socket ? 0 : appPort;
   const hostname = options.useLocalIp ? internalIp.v4() : options.host;
 
+  // use explicitly defined public url (prefix with protocol if not explicitly given)
+  if (options.public) {
+    return /^[a-zA-Z]+:\/\//.test(options.public) ? `${options.public}` : `${protocol}://${options.public}`;
```
shellscape: sorry, but this is not what was requested in previous comments

> this is not a good way to check a url for a protocol. please use [`url.parse`](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost) instead, and examine the result for a protocol.

client/index.js
robertaistleitner: I don't really get what you mean with using `setInterval` and `clearInterval` here and how this will improve the code.

+1 for moving the log down

lib/util/createDomain.js
robertaistleitner: I wrote in the commit comment that `url.parse` is not appropriate here.. I did it with `url.parse` before but the new tests I provided showed it's not working as it should..

> found a bug due to tests: `url.parse` doesn't properly parse a provided `public` option since it counts everything before the port colon as the protocol which is just wrong. I replaced it with a protocol regex.

robertaistleitner:
Regarding the too complex example: there's no other way to show why this is even needed (reverse proxy running on https + iframes + dev-server) so I don't get how the simple scenario should show how and why it works.

shellscape:
@robertaistleitner thanks for all your attention and work on this so far. it seems we're at odds with what I'd like to see in this PR and how you'd like to see it. I understand that you had issues with `url.parse` but I don't feel they're blocking as you say, and I feel strongly that a regex is not appropriate in this case. the differences between a do/while loop and set/clearInterval in the browser are significant, and do/while's ability to block/lock the browser in an errant condition is very widely known. examples also exist to demonstrate options, but not to demonstrate all-encompassing solutions, and shouldn't involve external tools that require additional user action; if you'd prefer, you can remove your example and I'll add one after the fact.

in the PR's current form I just can't merge it, citing my comments. apologies for being hard on this one but I can't budge on that I'm afraid. if you'd like to abandon, we'll schedule these changes for the v3 and cite your issue for credit. if you'd like to move forward with the requested changes, we'd be happy to continue the review.

robertaistleitner:
Sorry for not writing back sooner, I definitely want to get this merged and I'll propose solutions for the requested stuff as soon as possible..

lib/util/createDomain.js
robertaistleitner: Sorry, but I looked up url.parse documentation again, as well as searching for issues with what I'm experiencing, and there's no easy solution possible by using url.parse.

Why? The `public` parameter can be a non-valid url like "localhost:8080" which is parsed to somethin like this:
```javascript
> url.parse("localhost:8080")
Url {
  protocol: 'localhost:',
  slashes: null,
  auth: null,
  host: '8080',
  port: null,
  hostname: '8080',
  hash: null,
  search: null,
  query: null,
  pathname: null,
  path: null,
  href: 'localhost:8080' }
```
As you can see, the protocol is incorrect and I would need to do more processing to find out what's going on. I think the solution using regex is the better option here.

Also have a look at https://github.com/nodejs/node/issues/5755, where they're discussing the same issue on parsing urls.

shellscape: great documentation of the why. that'll certainly be beneficial in the future

shellscape:
Going to merge this one in now, thanks for all the work on it and apologies it took so long. Please note that it may take a while to get to release. The version 3 effort is in full swing and there's a pending [refactor of the client code](https://github.com/webpack/webpack-dev-server/pull/1137) that I'll need to manually merge your changes into.

robertaistleitner:
Thanks, I'm happy it got merged!

Issue #616, opened 2016-09-21
Minimalize differences between CLI and Node.js API and move CLI in `webpack-cli`


SpaceK33z:
There are some differences between the CLI and Node.js API that make webpack-dev-server a lot harder to use for people. Most differences are documented in the wiki, but users often don't look at this. They also shouldn't have to.

The first problem is that the CLI in `inline` modus (which is the default since v2), automatically adds an entry to the webpack config. When you use the API, you need to do this manually. What makes it more confusing is that there is is a `inline` option, but that only has effect when using the CLI.

The second problem is that, when using `--hot`, the CLI adds an entry to the webpack config _and_ adds the `HotModuleReplacementPlugin`. In the API, you need to add those two things yourself, as well as set `hot: true`.

The idea was that the API shouldn't mutate the webpack config. This is indeed iffy, but perhaps there are other ways to do this without mutating the config. Or if there is not, maybe it's worth it to mutate the config? The current situation causes many issues with users.

Related to #106, but that discussion is quite old and not all points are valid anymore.

cc @bebraw

axe312ger:
You marked it for version 3. Might it be possible to earlier make more clear what the CLI attaches/changes at the configuration?

Still have issues with public path, my webpack config works fine with CLI but not with the Node API :(

SpaceK33z:
@axe312ger, #615 will introduce config validation. Feel free to join that discussion if you have more ideas for validation. We're also working on better docs, which will be released with v2.

jeremejevs:
This definitely needs to be addressed.

For example, look at [the official docs on `React + HMR`](https://webpack.js.org/guides/hmr-react/) and [the `react-hot-boilerplate`](https://github.com/gaearon/react-hot-boilerplate/blob/next/webpack.config.js): both load the WDS client twice. This means that even those intimate with Webpack don't have full understanding of the difference between CLI and Node API.

Also, speaking of config differences, I can see people starting out with CLI and then switching to programmatic use, doing something like `new WebpackDevServer(compiler, webpackConfig.devServer)`, and then not understanding why the client isn't being inlined. Or vice versa, going to CLI and not understanding why the client is being inlined twice (or not even noticing it).

wesleymostien:
any progress ?

shellscape:
[...]

PR #1129, opened 2017-10-03
Set addDevServerEntrypoints default to 'localhost'

caryli:

Bug fix.

**Did you add or update the examples/?**

No.

**Summary**

The default host injected by `addDevServerEntrypoints` when using webpack-dev-server with the Node.js API  is `undefined` instead of `'localhost'` as specified at https://webpack.js.org/configuration/dev-server/#devserver-host where "by default this is `localhost`".

Reproducible from the code given up to the Hot Module Replacement step at the official guide at https://webpack.js.org/guides/getting-started/.

No.

**Other information**

When using webpack-dev-server with the Node.js API in the following scenario:
1. using `addDevServerEntrypoints` to inject webpack-dev-server into the entry point in `webpack.config.js` instead of manually (i.e. `entry: ['./app.js', '../../client/index.js?http://localhost:5000/']`)
2. the options object passed as the argument to the `devServerOptions` parameter of `addDevServerEntrypoints` does not have a `host` property nor a `useLocalIp` property as in the below example used at https://webpack.js.org/guides/hot-module-replacement/#via-the-node-js-api:

   *dev-server.js*
   ```
   const webpackDevServer = require('webpack-dev-server');

   const config = require('./webpack.config.js');
   const options = {
     contentBase: './dist',
     hot: true
   };

   webpackDevServer.addDevServerEntrypoints(config, options);
   ```

because of:

*webpack-dev-server/lib/util/createDomain.js*
```
11  const hostname = options.useLocalIp ? internalIp.v4() : options.host;
```

`hostname` will be set to `options.host` which is currently `undefined`.

This is not the case when using the webpack-dev-server CLI as defaults are set in `webpack-dev-server/bin/webpack-dev-server.js`. That step is skipped when calling `addDevServerEntrypoints` directly through the Node.js API.

Because of this, `devClient`:

*webpack-dev-server/lib/util/addDevServerEntrypoints.js*
```
18  const devClient = [`${require.resolve('../../client/')}?${domain}`];
19  console.log(devClient); // Added to print devClient.
```

will incorrectly resolve to and inject into `webpack.config.js`:

```
$ [ '/Users/someFolder/node_modules/webpack-dev-server/client/index.js?http:' ] // console.log(devClient)
```

instead of the correct:

```
$ [ '/Users/someFolder/node_modules/webpack-dev-server/client/index.js?http://localhost' ] // console.log(devClient)
```

This will eventually cause the error:

```
Uncaught SyntaxError: The URL 'http:/sockjs-node' is invalid
```

when starting the server:

`$ node dev-server.js`

and navigating to, in this case, `localhost:5000` because `socksjs-client` is called with a url without a `host` property:

*sockjs-client/lib/main.js*
```
67  var parsedUrl = new URL(url);
68  if (!parsedUrl.host || !parsedUrl.protocol) {
69    throw new SyntaxError("The URL '" + url + "' is invalid");
70  }
```

The behavior of the webpack-dev-server CLI is different and correctly defaults to localhost:

```
$ ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --content-base dist/
[ '/Users/someFolder/node_modules/webpack-dev-server/client/index.js?http://localhost:8080' ] // console.log(devClient)
Project is running at http://localhost:8080/
```

because the options object passed as the argument to the `devServerOptions` parameter of `addDevServerEntrypoints` is given the defaults `host: 'localhost'` and `port: 8080`:

*webpack-dev-server/bin/webpack-dev-server.js*
```
 72  const DEFAULT_PORT = 8080;
...
 74  yargs.options({
...
210    host: {
211      type: 'string',
212      default: 'localhost',
213      describe: 'The hostname/ip address the server will bind to',
214      group: CONNECTION_GROUP
215    },
...
221  });
222
223  const argv = yargs.argv;
...
244  if (argv.host !== 'localhost' || !options.host) { options.host = argv.host; }
...
344  options.port = argv.port === DEFAULT_PORT ? defaultTo(options.port, argv.port) : defaultTo(argv.port, options.port);
345  if (options.port != null) {
346    startDevServer(webpackOptions, options);
...
359  function startDevServer(webpackOptions, options) {
360    addDevServerEntrypoints(webpackOptions, options);
```

**Note**

The example at https://webpack.js.org/guides/hot-module-replacement/ forgets to pass `'localhost'` as the second argument of the `server.listen` call:

*dev-server.js*
```
server.listen(5000, () => {
  console.log('dev server listening on port 5000');
});
```

instead of:

*dev-server.js*
```
server.listen(5000, 'localhost', () => {
  console.log('dev server listening on port 5000');
});
```

But neither will fix the above as this `'localhost'` is passed to the server, `webpack-dev-server/lib/Server.js`, and not the client, `webpack-dev-server/client/index.js`, that is injected into the entry point.

### DIFF
--- a/lib/util/createDomain.js
+++ b/lib/util/createDomain.js
@@ -8,7 +8,7 @@ module.exports = function createDomain(options, listeningApp) {
   const protocol = options.https ? 'https' : 'http';
   const appPort = listeningApp ? listeningApp.address().port : 0;
   const port = options.socket ? 0 : appPort;
-  const hostname = options.useLocalIp ? internalIp.v4() : options.host;
+  const hostname = options.useLocalIp ? internalIp.v4() : options.host || 'localhost';
 
   // the formatted domain (url without path) of the webpack server
   return options.public ? `${protocol}://${options.public}` : url.format({


shellscape:
@caryli this is probably the most thorough PR description I've run across, and for that I want to say thank you, it's really very much appreciated 🎉 

Now for what's probably going to be a bummer; this isn't a change we're going to merge, mostly for a lot of the reasons you pointed out. We've got a functional API example here: https://github.com/webpack/webpack-dev-server/tree/master/examples/node-api-simple, which demonstrates how to properly use the API to start a server. Unfortunately if you're going to use the API method, you're going to have to specify `host` or `useLocalIp`in your config. 

(Future versions may consolidate the options between the CLI and the API, however. see: #616)

This looks like a case of documentation confusion more than anything. Fortunately, the docs can be updated via a PR on this repo: https://github.com/webpack/webpack.js.org

Thanks again for taking the time to create this.

Issue #1133, opened 2017-10-04
Make Array.prototype.includes polyfill not enumerable


zwug:
* Operating System: macOS Sierra 10.12.6
* Node Version: 4.8.4
* NPM Version: 2.15.11
* webpack Version: 2.7.0
* webpack-dev-server Version: 2.9.1

- [ ] This is a **bug**
- [ ] This is a **feature** request
- [x] This is a **modification** request

My webpack build fails on stylus-loader, because stylus-loader uses `for .. in` loop to iterate through an array of files. And because `includes` property is enumerable, there is an iteration with it which causes failure. I have already created an [issue in stylus-loader](https://github.com/shama/stylus-loader/issues/181), but i think webpack-dev-server should change the polyfill property setting too. I can make a pull request with `Object.defineProperty` to fix this.

### Code

```js
...
{
  test: /\.styl$/,
  use: [
     { loader: 'style-loader', options: { sourceMap: true } },
     { loader: 'css-loader', options: { sourceMap: true } },
     { loader: 'postcss-loader', options: { sourceMap: true } },
     { loader: 'stylus-loader', options: { sourceMap: true } },
   ],
},
...
  // webpack.config.js
```

code in stylus-loader:

```js
function normalizePaths(paths) {
  for(var i in paths) {
    paths[i] = path.normalize(paths[i]);
  }
  return paths;
}
```

webpack-dev-server polyfill setting:

```js
if (!Array.prototype.includes) {
  Array.prototype.includes = require('array-includes');
}
```

### Expected Behavior
Array.prototype.includes should not be enumerable

### Actual Behavior
Array.prototype.includes is enumerable

### For Bugs; How can we reproduce the behaviour?
node v4.8.4 + stylus-loader v3.0.1 should do it

shellscape:
@zwug thanks for checking in. could you elaborate a bit on what you mean by "enumerable" in this case? I'm not 100% sure I understand how polyfills.js is conflicting with stylus.

zwug:
Here is a sample snippet with the problem I am facing

```
if (!Array.prototype.someFunc) {
  Array.prototype.someFunc = function() {
    return 'something';
  };
}

const items = ['potato', 'bread', 'milk'];

for (let index in items) {
  console.log(items[index]);
}

// potato
// bread
// milk
/*
ƒ () {
    return 'something';
  }
*/
```

[...]

PR #1134, opened 2017-10-04
Changed property descriptor for Array.includes polyfill

zwug:
Changed the way of adding Array.includes polyfill for node.js <6.

No need to.

**Summary**

This is a fix for [#1133](https://github.com/webpack/webpack-dev-server/issues/1133).

No

### DIFF
--- a/lib/polyfills.js
+++ b/lib/polyfills.js
@@ -6,10 +6,5 @@
 // internal-ip@2.x uses [].includes
 if (!Array.prototype.includes) {
   const includes = require('array-includes');
-  Object.defineProperty(Array.prototype, 'includes', {
-    enumerable: false,
-    configurable: false,
-    writable: false,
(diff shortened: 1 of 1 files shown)


lib/polyfills.js
```diff
@@ -5,5 +5,11 @@
 
 // internal-ip@2.x uses [].includes
 if (!Array.prototype.includes) {
-  Array.prototype.includes = require('array-includes');
+  const includes = require('array-includes');
+  Object.defineProperty(Array.prototype, 'includes', {
```
shellscape: would probably be better to use `includes.shim();` here, no? https://github.com/ljharb/array-includes/blob/master/shim.js

zwug:
It seems like there is no need for `if (!Array.prototype.includes) {` because shim checks it anyway

Issue #1147, opened 2017-10-16
Runtime error when `node.global` is set to false.


filipesilva:
* Operating System: Windows 10
* Node Version: 6.11.1
* NPM Version: 5.3.0
* webpack Version: 3.7.1
* webpack-dev-server Version: 2.9.1

- [x] This is a **bug**
- [ ] This is a **feature** request
- [ ] This is a **modification** request

### Code

A reproduction can be found at https://github.com/filipesilva/webpack-dev-server-global.

```js
  // webpack.config.js
const path = require('path');
const webpack = require('webpack');
const HtmlWebpackPlugin = require('html-webpack-plugin');


module.exports = {
  entry: './src/main.js',
  devtool: 'source-map',
  output: {
    path: path.resolve(__dirname, './dist'),
    filename: 'bundle.js'
  },
  plugins: [
    new HtmlWebpackPlugin({
      template: './src/index.html'
    })
  ],
  node: {
    global: false
  }
};

```

```js
  // main.js
console.log('hello');
```

### Expected Behavior

Using `webpack-dev-server` on a browser app with `node.global = false` should not break at runtime.

### Actual Behavior

`webpack-dev-server` has a runtime error:
```
Uncaught ReferenceError: global is not defined
    at Object.<anonymous> (browser-crypto.js:3)
    at __webpack_require__ (bootstrap 46c9e79f3cb14d538cb2:19)
    at Object.<anonymous> (random.js:4)
    at __webpack_require__ (bootstrap 46c9e79f3cb14d538cb2:19)
    at Object.<anonymous> (event.js:3)
    at __webpack_require__ (bootstrap 46c9e79f3cb14d538cb2:19)
    at Object.<anonymous> (websocket.js:3)
    at Object.<anonymous> (websocket.js:99)
    at __webpack_require__ (bootstrap 46c9e79f3cb14d538cb2:19)
    at Object.<anonymous> (transport-list.js:5)
```

This happens because `webpack-dev-server` is bundling `sockjs-client` from source, and `sockjs-client` sources assume a node environment with `global` available.

For more details about how this behaviour is problematic, see angular/angular-cli#5804,  https://github.com/mapbox/mapbox-gl-js/issues/4359#issuecomment-303880888 and https://github.com/sockjs/sockjs-client/issues/401#issuecomment-336904363.

### For Bugs; How can we reproduce the behavior?

```
git clone  https://github.com/filipesilva/webpack-dev-server-global
[...]

PR #1148, opened 2017-10-16
Use built version os `sockjs-client`

filipesilva:

Bugfix

Yes.

**Summary**

`webpack-dev-server` bundles `sockjs-client` from source. `sockjs-client` sources assume a node environment however, and will fail to run if `global` is not available.

By bundling `sockjs-client` from source, the node environment requirement from `sockjs-client` will also be required of the app being served and thus it is not possible to have `node.global = false` in the user app being bundled.

This fix uses the built `sockjs-client` instead of the sources.

No.

**Other information**
Fix https://github.com/webpack/webpack-dev-server/issues/1147

### DIFF
--- a/examples/node-false/README.md
+++ b/examples/node-false/README.md
@@ -1,0 +1,9 @@
+# CLI - node false
+
+```shell
+node ../../bin/webpack-dev-server.js --open
+```
+
+## What should happen
+
+In the app you should see "It's working."
--- a/client/socket.js
+++ b/client/socket.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const SockJS = require('sockjs-client');
+const SockJS = require('sockjs-client/dist/sockjs');
 
 let retries = 0;
 let sock = null;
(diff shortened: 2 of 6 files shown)


examples/node-false/README.md
```diff
@@ -0,0 +1,9 @@
+# CLI - node false
```
shellscape: these are more for "how to" for users, but there's no harm in adding this. the tests that matter are in `/test`. no worries if you can't get to adding one there, we can do it after merging.

filipesilva:
@shellscape I removed the example, I don't think it's a real user example.

Regarding the real tests in `/test`, to test `node: false` doesn't throw a runtime error, a e2e test would be needed. It looks like there is no test where a browser is actually used. Instead I added `node: false` to the simple config to ensure that stuff in general still works with it, and that it's not needed for normal webpack-dev-server functioning. WDYT?

shellscape:
Yeah good point; we would actually need a browser for this. I'd leave the example in there as a means to test that change at the moment, even though it's not a traditional example. We can look into headless browser testing for the client in the future.

filipesilva:
Done, re-added it.

shellscape:
Thanks much :)

filipesilva:
Thank you for the quick review!

PR #1157, opened 2017-10-21
fixes #860: failure to exit on SIGINT race condition

### LINKED ISSUE #860: webpack-dev-server does not stop on Ctrl+C
Hi. I noticed that the latest version v2.4.2 have an issue around process signal handling. Pressing Ctrl+C to stop `webpack-dev-server` does not work **while compiling** assets. (NOTE: Ctrl+C works well after finished compilation.)

Also, I reverted the change of the PR #787 to workaround the issue for now.

OS: macOS 10.12.3
webpack-dev-server 2.4.2
webpack 2.3.2

shellscape:
Properly calls `process.exit` only after `server.close` has completed. Probable fix for reported edge case whereby the server process doesn't actually exit when SIGINT is sent.

No need

**Summary**
See above

No

**Other information**
Added new test to confirm SIGINT detection works properly.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -390,8 +390,9 @@ function startDevServer(webpackOptions, options) {
 
   ['SIGINT', 'SIGTERM'].forEach((sig) => {
     process.on(sig, () => {
-      server.close();
-      process.exit(); // eslint-disable-line no-process-exit
+      server.close(() => {
+        process.exit(); // eslint-disable-line no-process-exit
+      });
     });
   });
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1997,9 +1997,9 @@
       }
     },
     "execa": {
-      "version": "0.7.0",
-      "resolved": "https://registry.npmjs.org/execa/-/execa-0.7.0.tgz",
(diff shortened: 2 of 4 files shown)


bin/webpack-dev-server.js
```diff
@@ -390,8 +390,9 @@ function startDevServer(webpackOptions, options) {
 
   ['SIGINT', 'SIGTERM'].forEach((sig) => {
     process.on(sig, () => {
-      server.close();
-      process.exit(); // eslint-disable-line no-process-exit
+      server.close(() => {
```
michael-ciniawsky: `process.nextTick()` ? But not 💯

shellscape: @michael-ciniawsky hm? please elaborate :)

michael-ciniawsky: Hmm I think it is bananas from my side tbh :), the intend was 

```js
server.close()

process.nextTick(() => process.exit())
```

to ensure `server.close()` finishes before `process.exit()` is called by pushing it to the next EventLoop 'tick'.  But a callback to `server.close()` is definitely cleaner and more expressive/logical, so forget what I said 😛

shellscape: haha no worries. something good to keep in mind though!

PR #1159, opened 2017-10-23
Fix https in homepage url

winterbe:
Use `https` instead of `http` URL to Github.


shellscape:
@winterbe thanks for the typo fix! please sign the CLA so we can merge this

PR #1163, opened 2017-10-27
rebase to webpack-cli

ev1stensberg:


shellscape:
@ev1stensberg we need to get some input from collabs on what kind of version bump this change should warrant. initial thinking is that it should be minor, since we're changing a large dependency. should also probably hold off on this until `webpack` accepts https://github.com/webpack/webpack/pull/5869

ev1stensberg:
Yep. What you could try, is removing webpack and only go for the cli dep. It should work!

shellscape:
@ev1stensberg everything seems to work on my end. just waiting for `webpack` to merge that change.

shellscape:
@ev1stensberg might want to change this over to the `next` branch instead of master. it's more likely to drop after v3 is released stable, and there's quite a few differences.

ev1stensberg:
Opening this in a new PR, too many rebase conflicts with empty commits that make the final rebase fail

PR #1174, opened 2017-11-03
Don't mutate stats configuration (webpack 3.8.0+ compatibility)

BenoitZugmeyer:

Bugfix

No

**Summary**

The webpack configuration `stats` property has the same format as the `devServer.stats` property, so the same object is frequently reused to define the same stats output.

Since webpack v3.8.0, the `stats` property is validated more strictly by webpack (see https://github.com/webpack/webpack/commit/e0d4501fabb7fb452284dd4aec1a1938170a20b0#diff-c6d9e330d5284ccaee0ea220907a2280 and https://github.com/webpack/webpack/commit/576187541ae4a1e6320203a1d89ae69d6ddd4c47#diff-c6d9e330d5284ccaee0ea220907a2280).  webpack-dev-server is mutating this `stats` object by adding a `colors` property, which is an object and doesn't match the webpack configuration schema, causing a fatal error:

```
 - configuration.stats should be one of these:
   object { context?, hash?, version?, timings?, performance?, depth?, assets?, env?, colors?, maxModules?, chunks?, chunkModules?, modules?, children?, cached?, cachedAssets?, reasons?, source?, warnings?, errors?, warningsFilter?, excludeAssets?, excludeModules?, exclude?, entrypoints?, errorDetails?, chunkOrigins?, modulesSort?, moduleTrace?, chunksSort?, assetsSort?, publicPath?, providedExports?, usedExports?, optimizationBailout? } | boolean | "none" | "errors-only" | "minimal" | "normal" | "detailed" | "verbose"
   Used by the webpack CLI program to pass stats options.
   Details:
    * configuration.stats.colors should be a boolean.
    * configuration.stats.colors has an unknown property 'level'. These properties are valid:
      object { bold?, red?, green?, cyan?, magenta?, yellow? }
    * configuration.stats.colors has an unknown property 'hasBasic'. These properties are valid:
      object { bold?, red?, green?, cyan?, magenta?, yellow? }
    * configuration.stats.colors has an unknown property 'has256'. These properties are valid:
      object { bold?, red?, green?, cyan?, magenta?, yellow? }
    * configuration.stats.colors has an unknown property 'has16m'. These properties are valid:
      object { bold?, red?, green?, cyan?, magenta?, yellow? }
    * configuration.stats.colors should be one of these:
      boolean | object { bold?, red?, green?, cyan?, magenta?, yellow? }
    * configuration.stats should be a boolean.
    * configuration.stats should be one of these:
      "none" | "errors-only" | "minimal" | "normal" | "detailed" | "verbose"
```

The easy fix is to *not* mutate the `stats` object, so webpack isn't breaking with a fatal error.  But we may also consider setting a valid `colors` property.

No

**Other information**
Please see this minimal project to reproduce: https://gist.github.com/BenoitZugmeyer/29810768cc0f7a4413265d285a160027


bin/webpack-dev-server.js
```diff
@@ -294,7 +294,9 @@ function processOptions(webpackOptions) {
     };
   }
 
-  if (typeof options.stats === 'object' && typeof options.stats.colors === 'undefined') { options.stats.colors = argv.color; }
+  if (typeof options.stats === 'object' && typeof options.stats.colors === 'undefined') {
+    options.stats = Object.assign({}, options.stats, { colors: argv.color });
```
shellscape: I may be having a brain fart here, so please correct me if I'm wrong. this doesn't look like it'll avoid mutating the `stats` property. there's a larger issue here (which is resolved in the v3 beta branch) of the options passed being mutated on the whole:

```
var a = { stats: {} };

function doit (o) {
  o.stats = Object.assign({}, a.stats, { color: '0' });
}

doit(a);
a;

// returns
// {stats: {…}}
//   stats: {color: "0"}
//   __proto__: Object
```

I could be missing something, but I think the solution is to prevent the top level options object passed to devServer from being mutated.

BenoitZugmeyer: You are right, but it's the `stats` property *value* that shouldn't be mutated.

```
var stats = {};
var a = { stats: stats };

function doit (o) {
  o.stats = Object.assign({}, a.stats, { color: '0' });
}

doit(a);
stats;

// {}
```

shellscape: I see, I see. Hadn't considered that case. Could you put a simple test for this in `/test`?

BenoitZugmeyer: Sadly I cannot write a good test case because when running `test/cli.test.js`, the package `supports-color` detect that it doesn't run in a tty, so it returns `false` (which webpack sees as valid).

shellscape:
@BenoitZugmeyer thanks for your patience on this one. It's now merged.

PR #1175, opened 2017-11-03
Upgrade jquery from 1.8.1 to 3.2.1

delabroj:
Upgrade jquery to resolve retire.js security warnings, as mentioned here: https://github.com/webpack/webpack-dev-server/pull/878


shellscape:
@delabroj thanks for the PR. please be aware that this might take some time to get around to - we have to manually test each and every example in `/examples` to make sure this won't break anything. and since the beta of v3 is getting released soon, our immediate priorities are to that.

shellscape:
@delabroj I neglected to mention that the Pull Request Template _is not_ optional, and removing it is generally frowned up. please don't do that in the future. since jQuery is going to be removed in the next version, we're going to close this PR. thanks again for taking the time to contribute.

Issue #1177, opened 2017-11-06
Upgrade sockjs to 0.3.19


fwielstra:
I was running [auditjs](https://www.npmjs.com/package/auditjs) on my project which came across one vulnerability in the version of sockjs used by `webpack-dev-server` ([issue](https://github.com/sockjs/sockjs-node/issues/217), [fix](https://github.com/sockjs/sockjs-node/pull/218), which was resolved in the 0.3.19 release (from about a year ago). I've searched and couldn't find this issue having been raised yet.

PR incoming.

SpaceK33z:
Looks to have been fixed a while ago, so closing.

PR #1178, opened 2017-11-06
Update sockjs dependency to fix auditjs security vulnerability warning

fwielstra:

Updates sockjs to 0.3.19

nay

**Summary**

https://github.com/webpack/webpack-dev-server/issues/1177

It shouldn't

**Other information**


shellscape:
@fwielstra thanks for the PR. we'll still have to go through the `/examples` using this sha to make sure that it doesn't break anything, even though it shouldn't. sockjs in this project is a rather fragile pain point, so we have to be extra certain. please understand that may take a while, as our efforts are focused on v3 (which removes SockJS as a dependency) at the moment. it's also worth noting that this vulnerability shouldn't be an issue for webpack-dev-server, as it should never be run in anything but a dev/test environment, and certainly not exposed to the public.

fwielstra:
That's okay, if socksjs is removed completely that will also resolve the issue. We also only use this project in development, and are ignoring the security warning - we'd prefer to just ignore it, but I've not been able to find the vunlerability ID to put in the auditjs whitelist (see [this issue](https://github.com/OSSIndex/auditjs/issues/23)).

So yeah, no worries, take your time or just close this issue when 3.0 is out, I'm sure we'll be able to upgrade soon after. Thanks!

PR #1181, opened 2017-11-08
enable progress from config

ryanwholey:
Enables progress option to be read from the passed config (in addition to the command line)

Extends current feature by allowing option to be read from webpack config

The example passed progress in its config already, so it wasnt working properly. It now works as expected.

**Summary**
I was working on another project which uses webpack dev server heavily. I like the progress option because it gives the user some feedback. On a code review, someone recommended to clean up my package.json I should move the --progress options to the config since I pass it to each webpack-dev-server script. Turned out it was CLI only! 

No it should not!

**Other information**
Very excited to contribute! I really love webpack dev server and I hope I get the opportunity to contribute more in the future.


bin/webpack-dev-server.js
```diff
@@ -368,7 +370,7 @@ function startDevServer(webpackOptions, options) {
     throw e;
   }
 
-  if (argv.progress) {
```
shellscape: a less impactful option would have been to set `options.progress` within the existing `if` block. keep an eye on those kinds of simple changes in the future and give them favor.

shellscape:
@ryanwholey (copied from your related PR) thanks for the PR! we love getting new tests submitted and normally would be quick to add them to the repo. right now we're in a holding pattern on new code and new features while we prep v3 on the [beta](https://github.com/webpack/webpack-dev-server/tree/beta) branch. While we're not looking for outside contributions on that branch _at the moment_, we will be in the near future. we're going to have to put this PR on hold for a bit, but we'll be circling back to it.

bin/webpack-dev-server.js
rwholey-eb: I think I understand what you're saying, since the progress opt is only used here and its not always common to use, it would have been less impactful to set the progress opt inside of the if block so it wouldn't be assigned on every run? 

Thank you for the review! Looking forward to the v3 version :)

shellscape:
Thanks for your patience @ryanwholey and your contributions. We've merged your two PRs.

ryanwholey:
nice!!!!! thanks team, happy holidays!

PR #1209, opened 2017-11-30
feat: `reportTime` option

evilebottnawi:

feature

**Summary**
`webpack-dev-middleware` have option `reportTime`, but i can't setup this option using `webpack-dev-server`

No

**Other information**

No

### DIFF
--- a/lib/optionsSchema.json
+++ b/lib/optionsSchema.json
@@ -303,6 +303,10 @@
       "description": "Customize what the console displays when compiling.",
       "instanceof": "Function"
     },
+    "reportTime": {
+      "description": "Report time before and after compiling in console displays.",
+      "type": "boolean"
+    },
     "noInfo": {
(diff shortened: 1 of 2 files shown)


shellscape:
@evilebottnawi v3 is leveraging `loglevel` https://github.com/webpack/webpack-dev-server/blob/next/lib/log.js, and output from that should follow the same option, otherwise we'll get unbalanced expectations for log output. that means we need to update webpack-dev-middleware (which I am planning on doing soon) to use the same patterns. going that route, we should introduce the ability for users to provide their own log prefix, rather than granular options.

shellscape:
@evilebottnawi We can merge this as 2.x of WDS is still using 1.x of WDM and 1.x has this older option.

_Note: v2 of webpack-dev-middleware was released and removed `reportTime` in favor of `logTime`. v3 of webpack-dev-server supports the same option._

PR #1219, opened 2017-12-07
Allow webpack 4.0.0-alpha releases

lencioni:

peerDependency version change

no

**Summary**

Webpack 4 alpha has been released, and I want to make it easier for people to try it out in more ways.

No

**Other information**
It might be worth considering changing this to `*` to reduce maintenance and to avoid blocking people from updating. It might also be worth adding 4.0.0 to this list so it will work once the full release comes out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "ssl/"
   ],
   "peerDependencies": {
-    "webpack": "^2.2.0 || ^3.0.0"
+    "webpack": "^2.2.0 || ^3.0.0 || ^4.0.0-alpha"
   },
   "dependencies": {
     "ansi-html": "0.0.7",


shellscape:
@lencioni thanks for checking in and the PR. unfortunately I can't approve of this change at this time. I don't feel that alpha - or any version-tagged - releases should be included in a dependency range for a non-version-tagged version of a module. When v3 is released as alpha/beta, webpack@4 will be allowed. Until webpack@4 is released as stable, the 2.x branch of WDS will need to stick with the current peerDep range. Of note, that doesn't prevent users from installing and using webpack@4 while it's in alpha.

lencioni:
Ok, thanks!

shellscape:
added this to the alpha release in 03cdce74b2f486a033f52048a80f73a38f149d79

PR #1223, opened 2017-12-09
Rebase cli

ev1stensberg:
I actually think this is sufficient, because any time you make a require call for webpack, it will redirect to the cli. This way, avoid extensive refactor/complicated nesting of dependencies. The only thing that needed change was the arg path

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -5,7 +5,7 @@
 require('./lib/polyfills');
 require('loud-rejection/register');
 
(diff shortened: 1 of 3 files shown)


shellscape:
That damn package-log. @ev1stensberg can you address and we'll merge this?

ev1stensberg:
Done @shellscape

PR #1239, opened 2017-12-18
use validateSchema Directly

ev1stensberg:

Enhancement / Bugfix
N/A
**Summary**

Trying to run the devServer makes the validationSchema throw. Maybe webpack has changed their implementation, so I used the module directly. This is probably also better? Feel free to repro and investigate, happy to provide more information

No

**Other information**
devServer version is: 2.9.7

skjermbilde 2017-12-18 kl 21 49 16

### DIFF
--- a/lib/DevServer.js
+++ b/lib/DevServer.js
@@ -11,6 +11,7 @@ const WebSocket = require('ws');
 const app = require('./app');
 const features = require('./features');
 const log = require('./log');
+const validateSchema = require('webpack/lib/validateSchema');
 const OptionsValidationError = require('./OptionsValidationError');
 const optionsSchema = require('./schemas/options.json');
 const plugins = require('./plugins');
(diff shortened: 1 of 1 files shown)


shellscape:
@ev1stensberg PR says you're using 2.9.7 - does the same issue exist for 3.0.0? I'm not really sure what your screenshot is trying to convey, I think it's better to use text than images so we can quote, etc.

ev1stensberg:
@shellscape Ran WDS from a spawned process for the serve feature, I'll get a repro test for you tomorrow, and test it against 3.0.0 👌

ev1stensberg:
Version 3 ( next branch ) ran without errors. I guess this is because the cli and wds depends on different versions of webpack, closing this for now

ev1stensberg:
@shellscape could you ping me a bit before the `next` branch goes into master, so we can get the new cli version out and some warnings during installation away?

Issue #1084, opened 2017-09-14
v2.8.0 / #1058 breaks compatibility with Internet Explorer (template strings)


aloker:
**Bug report**

After an upgrade to 2.8.1, my app does not work in Internet Explorer 11 anymore.
Error:
SCRIPT1014: Invalid character

The error points to  
https://github.com/webpack/webpack-dev-server/blob/e8cbdadd4e7632c1bbf4e6b683d485d254887d45/client/index.js#L68

IE does not support template literals according to https://kangax.github.io/compat-table/es6/

This change and many other template-literals were introduced in https://github.com/webpack/webpack-dev-server/pull/1058

Packages used:
webpack@3.5.6
webpack-dev-server@2.8.1
react@6.26.0 (presets: env, stage-1, react and flow, plugins: react-hot-loader/babel, react-html-attrs, transform-decorators-legacy, syntax-dynamic-import)

I'm running Windows 10 Pro 1703

Rolling back to 2.7.1 fixes the issue. 

Please consider undoing the usage of template literals as this is a show-stopper for testing on IE during development.

shellscape:
@aloker Thanks for the report. Please note that the issue templates are not optional, we ask everyone to fill them out and not remove the contents of the template.

That dang IE though. We tested the changes against Edge, but not IE. I'll queue up a change and release of 2.8.2 to address this today. For the time being you can safely downgrade to 2.7.1.

aloker:
Thanks for the quick response.

I'm sorry about the template, I'll stick to it in the future.

shellscape:
@aloker no worries, that's a nasty one for sure.

please try installing webpack-dev-server from `npm install git://github.com/webpack/webpack-dev-server.git#ie-template-fix` and running against that. please let me know if that resolves the issue for you. I don't have an IE machine/vm to test with at the moment and I'm coming up short finding an ES5 validator that works on the web. so I'm gonna need your help on this one 😄

aloker:
Now it chokes on the shorthand property syntax at https://github.com/webpack/webpack-dev-server/blob/c45fb8bfb210c56cffae8b251bbb1bae318c697b/client/index.js#L69

IE does not support 
```
self.postMessage({
      type: 'webpack' + type,
      data
    }, '*');
``` 
You'll have to use ES5 syntax:
```
self.postMessage({
      type: 'webpack' + type,
      data: data
    }, '*');
```

Basically, there's not much ES6 syntax you can use if you want to support IE...

[...]

Issue #1101, opened 2017-09-17
Build succeeding with webpack fails with webpack-dev-server (Unexpected token)


about-code:
* Operating System: **Ubuntu 16.04 LTS**
* Node Version: **8.0.0**
* NPM Version: **5.4.2**
* webpack Version: **3.6.0**
* webpack-dev-server Version: **2.8.0**

- [ ] This is a **feature** request
- [x] This is a **bug**

### Expected Behavior
*webpack-dev-server* shouldn't break builds

### Actual Behavior
Build succeeding with *webpack* fails with the same config with *webpack-dev-server*. Message
```
ERROR in theme.76a4a23d58ebccf23fb4.js from UglifyJs
Unexpected token: name (urlParts) [theme.76a4a23d58ebccf23fb4.js:1957,4]
```

### How can we reproduce the behavior?

- clone or download [this repo](https://github.com/about-code/slush-ng-monorepo-sample) at version 1.0.0-alpha.6
- `npm install` dependencies
- `npm list webpack` and make sure webpack was installed at 3.6.0
- `npm list webpack-dev-server` and make sure webpack-dev-server was installed at 2.8.0 or 2.8.1
- Have a look at `package.json`. There are two relevant scripts: 
  - `npm run build` builds production bundles using webpack  (succeeds)
  - `npm start` builds with webpack-dev-server using the same config as `build` script  but fails, unexpectedly

The issue must have been introduced with webpack-dev-server 2.8.0. To validate
- `npm install webpack-dev-server@~2.7.0`
- `npm start`   (succeeds)

### Summary of workarounds from the discussion below
- Downgrade to 2.7.1 (see [duhalme](https://github.com/webpack/webpack-dev-server/issues/1101#issuecomment-331651230))
- or install UglifyJS-Plugin separately (see [shellscape](https://github.com/webpack/webpack-dev-server/issues/1101#issuecomment-330528139))
- or temporarily remove UglifyJS-Plugin until webpack includes an es6-capable version
- or transpile webpack-dev-server (see [Turncoat](https://github.com/webpack/webpack-dev-server/issues/1101#issuecomment-330565601))

shellscape:
[...]

PR #1241, opened 2017-12-21
enforce ES5 in all client files, fix usage in ES5 only browsers

yyx990803:

bugfix

No

**Summary**
There were prior issues regarding the client files, after being upgraded to ES6, not working properly in IE11 (#1084). However, the fix (https://github.com/webpack/webpack-dev-server/commit/c9d32f802278a2285f5a0e316b0e60166a3686ed) left `let` and `const` in the client codebase, which works in IE11, but not in other ES5 compliant browsers (IE<11, UC Browser on Android).

Certain regions of the world, in particular China and developing countries, still has a large percentage of browsers that only support ES5 (UC Browser has over 500 million downloads on Google Play), so developers targeting customers in those countries still need to debug their applications in those browsers.

This PR ensures all client files are ES5 only (also enforces this via ESLint parser version).

No.

**Other information**

### DIFF
--- a/client/.eslintrc
+++ b/client/.eslintrc
@@ -1,5 +1,10 @@
 {
+  "parserOptions": {
+    "ecmaVersion": 5
+  },
   "rules": {
+    "no-var": ["off"],
+    "vars-on-top": ["off"],
     "object-shorthand": ["error", "never"],
     "prefer-arrow-callback": ["off"],
     "prefer-template": ["off"]
--- a/client/index.js
+++ b/client/index.js
@@ -3,28 +3,28 @@
 /* global __resourceQuery WorkerGlobalScope self */
 /* eslint prefer-destructuring: off */
 
-const url = require('url');
-const stripAnsi = require('strip-ansi');
-const log = require('loglevel').getLogger('webpack-dev-server');
-const socket = require('./socket');
-const overlay = require('./overlay');
+var url = require('url');
+var stripAnsi = require('strip-ansi');
+var log = require('loglevel').getLogger('webpack-dev-server');
+var socket = require('./socket');
+var overlay = require('./overlay');
 
 function getCurrentScriptSource() {
   // `document.currentScript` is the most accurate way to find the current script,
   // but is not supported in all browsers.
   if (document.currentScript) { return document.currentScript.getAttribute('src'); }
   // Fall back to getting all scripts in the document.
-  const scriptElements = document.scripts || [];
-  const currentScript = scriptElements[scriptElements.length - 1];
+  var scriptElements = document.scripts || [];
+  var currentScript = scriptElements[scriptElements.length - 1];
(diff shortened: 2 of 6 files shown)


shellscape:
@yyx990803 thanks for the PR, I do appreciate the effort. Unfortunately this isn't one we're going to be merging. It's a topic that's been discussed many times in the issues here. The [readme outlines](https://github.com/webpack/webpack-dev-server#caveats) when this was done and what users needing to support unsupported platforms should do (use an older version). If a browser doesn't support `const` and `let`,  webpack-dev-server@2.7.1 should be used. The reasons for not bumping a major are also outlined in the readme, whether or not it was the right call has also been debated. Just as webpack@4 (and subsequent major versions of webpack-family modules) is dropping Node 4 support, older platforms can't be supported forever. We're not leaving those users high-and-dry, they'll just have to lag behind as they're already having to do supporting older platforms.

yyx990803:
This is extremely irresponsible and overly optimistic about the ES feature set being used around the globe. As I pointed out, UC Browser has 500 million downloads on Google Play which equates to millions of users worldwide. You are essentially shutting down all the developers working in those regions from being able to leverage `webpack-dev-server` to debug applications in a browser their users choose to use.

shellscape:
I understand the argument, and as mentioned, this has been discussed at length in other issues.  I would argue it's extremely irresponsible for a browser vendor to continue to distribute a product to 500 million users which is so far beyond modern standards, but there's not much I can say to that effect. These are the same arguments made when China was still predominately on IE6 not long ago. Vendors chose to move forward. At some point the line must be drawn.

If users of WDS are stuck having to support very old (or very behind) browsers, they can use v2.7.1 or use workarounds [such as this one](https://github.com/webpack/webpack-dev-server/issues/1101#issuecomment-330565601).

I also understand the weight your position in the two communities carries, and the water it typically moves. Unfortunately we have a difference of opinion.

yyx990803:
I still don't see a valid argument on why this should not be merged. Can you provide one such argument in prior discussions?

yoyo837:
I think what they mean is to force the browser and its users to improve, and the browser and its users are always easily turned into negative, backward...

michael-ciniawsky:
@shellscape I think this should be revisited since the ES5 compat is a minor change to the code base (mainly `let/const` => `var`) and if the user base is big enough (~500 million (claimed)) and folks have issues it's an fair point to consider imho. (e.g can't the `client` files not be transpiled ?)

wxsms:
Front-end developers can chose Node.js version for themselves, **but NOT browsers**.

Can't you simply babel your client side ES6 code to ES5? You can still move as fast as you wish without left anyone behind!

shellscape:
[A workaround](https://github.com/webpack/webpack-dev-server/issues/1101#issuecomment-330565601) exists for this issue. The webpack core team has taken up discussion of this. Please use that in the meantime. Locking this thread before it goes bananas.

PR #1242, opened 2017-12-21
transpile client bundles with babel

yyx990803:
bugfix

No.

**Summary**
ref: #1241 
The only reason I can think of regarding not merging #1241 is the added overhead of having to use ES5 when working on client files plus differing ESLint config for that part of the codebase.

This can be done by transpiling all the client files with babel beforehand. Since the client files are already bundled, this is simply adding babel-loader. This gets rid of the need for users having to do this workaround in their own config, and allows future iterations on the client files to be done using latest ES features.

This would be a pure win for both maintainers and users: maintainers no longer need to worry about backwards compat when working on client files, users no longer need to worry about dev-server not working in older browsers.

This also fixes the unintended breaking change introduced in 2.8.

No.

**Other information**


shellscape:
As stated [here](https://github.com/webpack/webpack-dev-server/pull/1241#issuecomment-353359148) the webpack core team is discussing this - there are greater issues to consider and patience will be required. Please don't take the "beat the horse until it's reanimated" approach. We get that you want to see this happen.

shellscape:
After much discussion amongst the team, we've arrived at a plan for webpack-dev-server moving forward. We're going to merge this PR and will be releasing 2.10.0 once testing is complete.

Moving forward, webpack-dev-server enters maintenance mode. We won't be accepting any new features or modifications for this module, only fixes. The 2.x branch will be the final major version of webpack-dev-server, and it will be the last to support Node 4, browsers older than the last two major versions, and browsers which don't support WebSockets.

We'll make a formal announcement with this info, and info regarding the future direction of dev servers for webpack in mid-February.

shellscape:
@yyx990803 please remove package-lock.json from the PR and we'll merge this.

yyx990803:
removed `package-lock.json` per requested.

shellscape:
@yyx990803 it would seem that if the entries are added dynamically (https://github.com/webpack/webpack-dev-server/blob/master/lib/util/addDevServerEntrypoints.js) and `inline: true` that the scripts don't transpile properly. IE users are reporting errors with templated strings. See #1268

PR #1243, opened 2017-12-23
support for webpack 4.0.0-alpha.*

donaldpipowitch:

Small fallback to test `webpack-dev-server` with `webpack@4.0.0-alpha.*`.

No.

**Summary**

Now we can use this package with the webpack 4 alphas. See https://github.com/webpack/webpack/issues/6179. Prevents `Cannot find module 'webpack/bin/config-yargs'` errors.

No.

**Other information**

🎄❤️

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -51,7 +51,11 @@ const defaultTo = (value, def) => value == null ? def : value;
 const yargs = require('yargs')
   .usage(`${versionInfo()}\nUsage: https://webpack.js.org/configuration/dev-server/`);
 
-require('webpack/bin/config-yargs')(yargs);
+try {
+  require('webpack/bin/config-yargs')(yargs);
+} catch (err) {
+  require('webpack-cli/bin/config-yargs')(yargs);
+}
 
 // It is important that this is done after the webpack yargs config,
 // so it overrides webpack's version info.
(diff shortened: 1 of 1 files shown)


peter-mouland:
I've just come across the same problem.

As webpack v4 is still in alpha, I guess there is some decisions to be made about how else to handle this i.e. it's probably premature to roll this out now, but perhaps a webpack-dev-server v3-beta might be useful with note around people trying webpack v4?

Should this PR also update the `peerDependencies` from the package.json also?

peter-mouland:
Hi @shellscape (sorry for at'ing you directly!) but I was wondering if it was realised that this prevents people from trying out anything greater than webpack v4-alpha.0? (i.e. alpha.1+ wont' work).

I saw some other issues already mentioning v3-beta but i could not see a branch or release with those tags.  Would v3 fix this issue and Is v3 available for testing?

probablyup:
Any chance this patch could be released as a beta version so it can be tested alongside the webpack 4 beta?

tomas2387:
To test this pull request, you can write directly the patch branch in your package.json
 ```
 "devDependencies": {
    "webpack": "^4.0.0-beta.0",
    "webpack-cli": "^2.0.4",
    "webpack-dev-server": "github:donaldpipowitch/webpack-dev-server#patch-1"
  },
```
----
I executed it and it didn't work. 
```
 $ ./node_modules/.bin/webpack-dev-server
 - configuration.entry should be one of these:
   object { <key>: non-empty string | [non-empty string] } | non-empty string | [non-empty string] | function
   -> The entry point(s) of the compilation.
   Details:
    * configuration.entry should be an object.
      -> Multiple entry bundles are created. The key is the chunk name. The value can be a string or an array.
    * configuration.entry should be a string.
      -> An entry point without name. The string is resolved to a module which is loaded upon startup.
    * configuration.entry[1] should be a string.
      -> A non-empty string
    * configuration.entry should be an instance of function
      -> A Function returning an entry object, an entry string, an entry array or a promise to these things.
```
I dont know if I'm undertanding it correctly but It seems it needs a configuration when webpack4 doesn't have one.

**Comment by SpaceK33z** [comment]
Please test the latest beta, `webpack-dev-server@3.0.0-beta.1`. Also be sure to read [the changelog](https://github.com/webpack/webpack-dev-server/releases/tag/v3.0.0-beta.1).

**Comment by peter-mouland** [comment]
Amazing, thanks guys!!
for anyone else, this now works with my current _test_ setup of:

```
    "webpack": "^4.0.0-beta.1",
    "webpack-cli": "^2.0.6",
    "webpack-dev-server": "^3.0.0-beta.1",
```

PR #1254, opened 2018-01-03
Support disabling index serving

ForbesLindesay:

This is a small feature

no

**Summary**

If you are doing server side rendering, by sending all HTML requests to a proxy, and your webpack config generates an `index.html` file, that file will get served instead of your server-side-rendered content for the root of the domain.  Passing an `index` value of `false` to webpack-dev-middleware prevents it from serving the index file, but it violates the current schema for options passed into webpack dev server.

No, this merely allows an extra value to be passed in for an existing option.

**Other information**

You can work around this issue by passing in `''` which is close enough to `false` to trigger the desired behaviour in webpack-dev-middleware, but also a string so it passes validation.

### DIFF
--- a/lib/optionsSchema.json
+++ b/lib/optionsSchema.json
@@ -320,8 +320,15 @@
       "type": "boolean"
     },
     "index": {
-      "description": "The filename that is considered the index file.",
-      "type": "string"
+      "description": "The filename that is considered the index file, or `false` to proxy the root path",
+      "anyOf": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "boolean"
+        }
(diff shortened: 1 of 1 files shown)


shellscape:
Thanks for the PR and congrats on your first submission to the project. Unfortunately we're not accepting any new features for the foreseeable future. We do appreciate it though!

ForbesLindesay:
I'd appreciate an individual response as to why you think this library is better without the attached change. It's difficult not to be offended by a generic response stating that you do not accept any new features, however small, simple or reasonable they are.

What is your suggestion for how this should be handled?

PR #1263, opened 2018-01-07
use  webpack-cli for `config-yargs` source

peter-mouland:
webpack (alpha-4) no longer contains config-yargs with `/bin`, but webpack-cli does.

- [x] This is a **bugfix**
- [ ] This is a **typo fix**

Would you say that this needs a test?

### Motivation / Use-Case

I am trying out webpack v4 alpha in my own boilerplate project and cam across this today.

As webpack v4 is still in alpha, I guess there is some decisions to be made about how else to handle this i.e. it's probably premature to roll this out now, but perhaps a `webpack-dev-server v3-beta` might be useful with note around people trying webpack v4?

### Breaking Changes

I Imagine this is a breaking change, just because people will now have to ensure they have the new cli installed, which was mot necessary before

### Additional Info

see commit : https://github.com/webpack/webpack/commit/4b0332d3909eea8115d84f9a03da2d52478daa70#diff-c1111bd512b29e821b120b86446026b8

see current v4 alpha: https://github.com/webpack/webpack/tree/v4.0.0-alpha.1/bin

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -51,7 +51,7 @@ const defaultTo = (value, def) => value == null ? def : value;
 const yargs = require('yargs')
   .usage(`${versionInfo()}\nUsage: https://webpack.js.org/configuration/dev-server/`);
 
-require('webpack/bin/config-yargs')(yargs);
+require('webpack-cli/bin/config-yargs')(yargs);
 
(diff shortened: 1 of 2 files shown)


peter-mouland:
will close this in favour or existing PR: https://github.com/webpack/webpack-dev-server/pull/1243/files

PR #1270, opened 2018-01-10
Cleanup client build setup, fix IE<=11 compatibility

### LINKED ISSUE #1268: IE11 fails to load with Invalid character error
* Operating System: Windows 10
* Node Version: 9.3.0
* NPM Version: 5.5.1
* webpack Version: 3.10.0
* webpack-dev-server Version: 2.10.0

- [x] This is a **bug**

### Expected Behavior
Webpack-dev-server allows running my app in IE11

### Actual Behavior
[This code](https://github.com/webpack/webpack-dev-server/blob/6689cb83cf8305ba1f600972487cb223552099d8/client/index.js#L68) makes webpack based development impossible to run in IE11, as it fails with error ``SCRIPT1014: Invalid character, index.js (361,13)``

Start any basic example of webpack-dev-server in IE11

yyx990803:
- [x] This is a **bugfix**
- [ ] This is a **typo fix**

No, because the bug case is CLI usage only and not testable in the current test setup.

### Motivation / Use-Case

Fixes #1268.

#1242 was attempting to fix older browser compatibility by transpiling client bundles with babel. However I didn't notice that the bundles are not used in all cases. It introduced a regression which causes client files to contain ES2015 code when using inline mode via the CLI. This PR fixes that by ensuring the client injected in [lib/util/addDevServerEntrypoints.js](https://github.com/webpack/webpack-dev-server/blob/master/lib/util/addDevServerEntrypoints.js) is the transpiled version as well.

### Breaking Changes

None.

#### Notable Changes

- Previously, source files for all 3 clients and their built bundles are mixed in the same directory, making it very confusing. This PR reorganized client files by moving all client source files in to a separate directory, `/client-src`, and further separating the three clients into their own directories:

  - `/client-src/default`
  - `/client-src/live`
  - `/client-src/sockjs`

- In addition to the bundles, files in `/client-src/default` are transpiled directly via `babel-cli` so that they can be bundled directly in inline mode. (see `transpile:index` script)

- `/client` now holds only built/transpiled bundles and files, and is added to `.gitignore`. Built files and bundles have the exact same file layout with pre-PR state of `/client`, so it should provide complete backwards compat.


.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["env"]
```
TheLarkInn: This would default to CJS module output no? If so, should we be switching this to "modules":false?

yyx990803: The client source is still using `require`, so having `modules: false` has no effect. Or we can convert to use ES modules imports and enable `modules: false`, but that may break compat with webpack 2 (not sure if that's still a requirement)

EloB:
I really hope that you merge this request. It's not just old desktop browsers (that are very easy to upgrade) but other smart devices that has web apps installed on them. That you cannot upgrade. For instance smart tvs/consoles etc.

TheLarkInn:
We have no objections to the methods here, we are just wanting to make sure
it's reviewed by multiple eyes. I've not spent much time in
webpack-dev-server and want to make sure this is good on all fronts.

Thanks for the patience.

sokra:
Thanks

tobilen:
@sokra looks like dependencies are still not being transpiled:

screen shot 2018-01-14 at 15 07 48

PR #1273, opened 2018-01-14
fix: pin strip-ansi to 3.x for ES5 compat

yyx990803:
- [x] This is a **bugfix**

Follow up for https://github.com/webpack/webpack-dev-server/pull/1270#issuecomment-357514454

`strip-ansi` which is required by the clients, uses ES6 syntax. Pinning it to 3.x keeps ES5 compat. ([The only difference between 3.x and 4.0 is the syntax change](https://github.com/chalk/strip-ansi/commits/master))

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -129,9 +129,9 @@
       "integrity": "sha1-gTWEAhliqenm/QOflA0S9WynhZ4="
     },
     "ansi-regex": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-3.0.0.tgz",
-      "integrity": "sha1-7QMXwyIGT3lGbAKWa922Bas32Zg="
+      "version": "2.1.1",
+      "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
+      "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8="
     },
     "ansi-styles": {
       "version": "2.2.1",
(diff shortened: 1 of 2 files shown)


graingert:
@sokra if you don't want to play whack-a-mole with deps as the npm ecosystem moves to releasing greenfield JS you should have a look at https://github.com/webpack/webpack-dev-server/pull/1279

julianxhokaxhiu:
Are you planning on merging and releasing this ASAP? I am also impacted by the same identical issue at the moment.

graingert:
@julianxhokaxhiu FYI, i've released drop-in-replacement `@procensus/webpack-dev-server` with my fix in  #1279

julianxhokaxhiu:
Thanks, I'll use that temporary until this gets merged :)

graingert:
> this gets merged :)

Hopefully it won't! ;) publishing `ES${new Date().getFullYear()}` is the future.  See #1279

julianxhokaxhiu:
I tested your drop-in replacement and I still get the same error:
```
Syntax error on L2114:
module.exports = input => typeof input === 'string' ? input.replace(ansiRegex(), '') : input;
```

it seems that the ansi-module is not transpiled from arrow functions.

graingert:
@julianxhokaxhiu replied here: https://github.com/webpack/webpack-dev-server/pull/1279#issuecomment-358612600

yyx990803:
@graingert FYI #1279 does not fix CLI --inline usage because injected entries will be processed with user config.

TheLarkInn:
@yyx990803 I think this looks fine.

TheLarkInn:
Would someone like to submit an issue so that we can upgrade in the future but not break IE? Pinning versions is a sort of a bandaid IMO.

TheLarkInn:
Nevermind, I got it tracked in issues https://github.com/webpack/webpack-dev-server/issues/1286

TheLarkInn:
I've got limited time today to publish this, however I will see if I can get @SpaceK33z or @sokra to run a release for me on this.

Issue #1271, opened 2018-01-12
Page does not refresh on HTML change when HMR is on


andreyvolokitin:
* Operating System: OS X 10.11.6
* Node Version: 9.2.1
* NPM Version: 5.6.0
* webpack Version: 3.8.1
* webpack-dev-server Version: 2.10.1

- [ ] This is a **bug**
- [x] This is a **modification** request

### Code
https://github.com/andreyvolokitin/test-webpack-dev-server

### Expected Behavior
Editing ```./src/index.html``` should cause recompiling and page refresh in the browser
### Actual Behavior
Recompiling is happening, but the page in the browser does not refresh
### For Bugs; How can we reproduce the behavior?
1. ```git clone https://github.com/andreyvolokitin/test-webpack-dev-server && cd test-webpack-dev-server```
2. ```npm install && npm start```
3. Edit ```./src/index.html``` — compiling is happening, but the page in the browser does not refresh
4. Open browser console and edit ```./src/test.js``` — HMR works

andreyvolokitin:
I found that it is possible to get a list of changed files which caused recompilation. Would it be feasible to add some logic [here](https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L70), so that when the list of changed files contains certain extensions — we pass a flag to ```_sendStats()``` to indicate that full page reload is needed? It is unclear by now what extensions should be included: ```.html``` is fine, but what about countless templating engines which could be used instead and would need the same logic applied?

evilebottnawi:
@andreyvolokitin we use this snippet for `php`:
```js
before(app, server) {
      const chokidar = require("chokidar");
      const files = [
        // Refreshing php files
        "**/*.php"
      ];

      chokidar
        .watch(files, {
          alwaysStat: true,
          atomic: false,
          followSymlinks: false,
          ignoreInitial: true,
          ignorePermissionErrors: true,
          ignored,
          interval: typeof poll === "number" ? poll : null,
          persistent: true,
          usePolling: Boolean(poll)
        })
        .on("all", () => {
          server.sockWrite(server.sockets, "content-changed");
        });
    }
```

evilebottnawi:
@shellscape what do you think about add to documentation example above?

andreyvolokitin:
[...]

PR #1276, opened 2018-01-15
fix(Server): disable page reload when `hot` is not specified (`options.hot`)

### LINKED ISSUE #1251: Don't enable HMR by default
> **Please note that this template is not optional.** If you proceed with this form,
please fill out _all_ fields, or your issue may be closed as "invalid."
Please do not delete this template. Please ask questions on StackOverflow or the
webpack Gitter (https://gitter.im/webpack/webpack). _General questions, how-to
questions, and support requests will be closed._
**Please do remove this header to acknowledge this message.**

* Operating System: OSX 10.11.6
* Node Version: v8.9.1
* NPM Version: 5.5.1
* webpack Version: 3.10.0
* webpack-dev-server Version: 2.9.7

- [x] This is a **bug**
- [ ] This is a **feature** request

### Code

```js
  // webpack.config.js
{
  "entry": {
    "app": [
      "/Users/project/src/index.js",
    ]
  },
  "output": {
    "filename": "[name]-[chunkhash].js",
    "path": "/Users/project/build",
    "publicPath": "/"
  },
  "devServer": {
    "contentBase": "./build",
    "disableHostCheck": true,
    "historyApiFallback": true,
    "host": "127.0.0.1",
    "port": 8080
  },
  "resolve": {
    "symlinks": false,
    "modules": [
      "/Users/project/src",
      "node_modules"
    ],
    "alias": {
      "react": "/Users/project/node_modules/react"
    }
  },
  "module": {},
  "plugins": [],
  "devtool": "inline-source-map"
}
```

```bash
  # command used to run webpack dev server
  ./node_modules/.bin/webpack-dev-server --config webpack.config.js --progress --colors
```

### Expected Behavior

When no options to activate HMR is provided, HMR should NOT be enabled. When modifying a source file watched by webpack, webpack-dev-server should not try to hot reload a module and it should not reload the full page either.

### Actual Behavior

Even though I did not provided --hot (or --hot --inline) on the command line used to run webpack-dev-server and even though my devServer config does not contain hot: true or false, or hotOnly: true or false, when modifying a source file watched by webpack, it reloads the full page. Note that trying to provide hot: false, hotOnly: false to the devServer config does not disable HMR either.

To reproduce the issue, clone this repo: `https://github.com/happypoulp/webpack3-issue/`, go to branch `issue-hmr-disable`, build and start app and you will see the issue when changing and saving `src/index.js`.

byzyk:
- [x] This is a **bugfix**
- [ ] This is a **typo fix**

No

### Motivation / Use-Case

Aimed to fix #1251. As per the issue, the page keeps reloading even if `hot` option is not specified or `false`. This PR contains fix which disables page reload when `hot: false` or not specified or `hotOnly: true`. To enable page refresh along with HMR one has to explicitly set `hot: true`.

### Breaking Changes

No

### Additional Info

I think we can consider adding one more option, something like `livereload` for use-cases when people don't wanna use HMR but instead prefer to go with a regular page refresh. I didn't do it as I feel it's more like a new feature while I tried to keep this PR as simple as possible.

### DIFF
--- a/client-src/default/index.js
+++ b/client-src/default/index.js
@@ -43,6 +43,7 @@ if (!urlParts.port || urlParts.port === '0') {
 }
 
 let hot = false;
+let hotOnly = false;
 let initial = true;
 let currentHash = '';
 let useWarningOverlay = false;
@@ -72,8 +73,9 @@ function sendMsg(type, data) {
 }
 
 const onSocketMsg = {
-  hot() {
+  hot(isHotOnly) {
     hot = true;
+    hotOnly = isHotOnly;
     log.info('[WDS] Hot Module Replacement enabled.');
   },
   invalid() {
@@ -217,21 +219,23 @@ function reloadApp() {
       // broadcast update to window
       self.postMessage(`webpackHotUpdate${currentHash}`, '*');
     }
-  } else {
-    let rootWindow = self;
-    // use parent window for reload (in case we're in an iframe with no valid src)
-    const intervalId = self.setInterval(() => {
-      if (rootWindow.location.protocol !== 'about:') {
-        // reload immediately if protocol is valid
-        applyReload(rootWindow, intervalId);
-      } else {
-        rootWindow = rootWindow.parent;
-        if (rootWindow.parent === rootWindow) {
-          // if parent equals current window we've reached the root which would continue forever, so trigger a reload anyways
+
+    if (!hotOnly) {
+      let rootWindow = self;
+      // use parent window for reload (in case we're in an iframe with no valid src)
+      const intervalId = self.setInterval(() => {
+        if (rootWindow.location.protocol !== 'about:') {
+          // reload immediately if protocol is valid
           applyReload(rootWindow, intervalId);
+        } else {
+          rootWindow = rootWindow.parent;
+          if (rootWindow.parent === rootWindow) {
+            // if parent equals current window we've reached the root which would continue forever, so trigger a reload anyways
+            applyReload(rootWindow, intervalId);
+          }
         }
-      }
-    });
+      });
+    }
   }
 
   function applyReload(rootWindow, intervalId) {
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -40,7 +40,8 @@ function Server(compiler, options) {
     throw new Error("'filename' option must be set in lazy mode.");
   }
 
-  this.hot = options.hot || options.hotOnly;
+  this.hot = options.hot;
+  this.hotOnly = options.hotOnly;
   this.headers = options.headers;
   this.clientLogLevel = options.clientLogLevel;
   this.clientOverlay = options.overlay;
@@ -601,7 +602,7 @@ Server.prototype.listen = function (port, hostname, fn) {
 
       if (this.clientOverlay) { this.sockWrite([conn], 'overlay', this.clientOverlay); }
 
-      if (this.hot) this.sockWrite([conn], 'hot');
+      if (this.hot || this.hotOnly) { this.sockWrite([conn], 'hot', this.hotOnly); }
 
       if (!this._stats) return;
       this._sendStats([conn], this._stats.toJson(clientStats), true);


benstaker:
@byzyk I have checked out your fork by defining the following in my **package.json**:
`git+ssh://git@github.com:byzyk/webpack-dev-server.git#88b89bbd4054f94a18a67c4f834a99a3eac5769e`

And I am receiving the following error when running webpack-dev-server:
```
module.js:471
    throw err;
    ^

Error: Cannot find module '../../client/'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.resolve (internal/module.js:27:19)
    at addDevServerEntrypoints (C:\workplace\feefo\feefo-admin-ui\node_modules\webpack-dev-server\lib\util\addDevServerEntrypoints.js:18:35)
    at startDevServer (C:\workplace\feefo\feefo-admin-ui\node_modules\webpack-dev-server\bin\webpack-dev-server.js:363:3)
    at processOptions (C:\workplace\feefo\feefo-admin-ui\node_modules\webpack-dev-server\bin\webpack-dev-server.js:350:5)
    at Object.<anonymous> (C:\workplace\feefo\feefo-admin-ui\node_modules\webpack-dev-server\bin\webpack-dev-server.js:495:1)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:383:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:496:3
```

Versions (working combinations):
```
"webpack": "2.7.0"
"webpack-dev-server": "2.11.0" or "2.5.1"
```

Which **webpack-dev-server** version did you base your fix from? And which version of **webpack** do you have?

byzyk:
@benstaker I tested it with `webpack: "3.10.0"` and fix is based on the latest WDS release which is `2.11.0`.

This might be related to code builds. As per [contributing guide](https://github.com/webpack/webpack-dev-server/blob/master/CONTRIBUTING.md#testing-a-pull-request), try this step before testing:

> Go to the `webpack-dev-server` module (`cd node_modules/webpack-dev-server`), and run `npm run prepublish`.

Let me know if that helped

happypoulp:
@benstaker I had the same issue at first.

I then did:
```
cd node_modules/webpack-dev-server/
yarn
yarn transpile:index
```

And I was able to test the fix.

benstaker:
@byzyk @happypoulp thank you both. I have tried running `npm run -s transpile:index` instead as I do not have yarn installed, inside of `node_modules/webpack-dev-server` and I am receiving the following error:
```
$ npm run -s transpile:index
client-src/default doesn't exist
```
**Side note:** The same error is outputted when running `npm run prepublish`.

It seems that I am having a different result to you both, may be due to the way I have defined and installed byzyk's forked copy (see my previous comment re. package.json). It's missing the `client-src/**` folder. I will download a zip copy of your changes and report back later.

happypoulp:
@benstaker did you run `npm install` after `cd node_modules/webpack-dev-server/` (that's what the `yarn` command does for me)?

I used the URL you provided in my package.json, to test this PR on my side so this part should be fine.

benstaker:
@happypoulp I have tried running `npm install` inside of `node_modules/webpack-dev-server` and as expected, I am receiving the same error because the `client-src/**` folder does not exist.

I have now downloaded the zip of branch `fix/hmr-disable`, deleted the contents of `node_modules/webpack-dev-server`, then copied the contents of the zip into it. Then ran `npm install` and `npm run prepublish` inside of that folder and it no longer errors!

I have tested running `webpack-dev-server` and it correctly disables HMR however, I am receiving the following errors in the browser console. Do you see the same errors?
image

happypoulp:
@benstaker I did not had those error in my JS console. Maybe the webpack config we use to test this PR differs somehow...
FTR here's the webpack config I am using:
```js
const path = require("path")
const webpack = require("webpack")
const HtmlWebpackPlugin = require("html-webpack-plugin")

module.exports = (settings) => {
    return {
        entry: {
            app: [
                path.join(process.cwd(), "src/index.js"),
            ],
        },
        output: {
            filename: "[name]-[chunkhash].js",
            path: path.join(process.cwd(), "build"),
            publicPath: "/",
        },
        devServer: {
            contentBase: "./build",
            disableHostCheck: true,
            historyApiFallback: true,
            host: "127.0.0.1",
            port: 8181,
        },
        resolve: {
            symlinks: false,
            modules: [path.join(process.cwd(), "src"), "node_modules"],
        },
        module: {},
        plugins: [
            new HtmlWebpackPlugin({
                filename: "index.html",
                template: "index.html",
                inject: true,
                minify: false,
            }),
        ],
    }
}
```

benstaker:
@happypoulp quite possibly, thank you for checking!

This may be due to the v2 webpack and latest webpack-dev-server I am using. We're using this combination because v3 webpack is currently too slow for incremental builds, so have reverted to webpack v2.

It seems that this issue is now solved, so please continue 🥇

byzyk:
@shellscape seems like this PR resolved the issue. Would you be able to review and give your feedback? I see that some tests are failing which is weird since overall coverage increased.

andreyvolokitin:
Note that full page reload ("live reload") and HMR are two distinct features, see here: https://github.com/webpack/webpack-dev-server/issues/1251#issuecomment-358935478.  They are used in different scenarios and one can not completely replace another. I.e. you can update JS/CSS/images with HMR, but if you developing static HTML page, you need to update its HTML with a full page reload ("live reload") because this is the only way to do it and HMR will not work in that case. Moreover, currently due to a bug full page reload is actually *only* works with HMR turned off, see #1271 . So your proposed solution will completely disable HTML refreshing in the described scenario, with HMR turned on or off.

The correct way to proceed on this is to implement a new option for "live reload", like you mentioned. The source files changes need to be delivered to the browser somehow. The simplest way to do it is a "live reload". On top of it you can use HMR, then JS/CSS/images will use it, but HTML changes still will need to be refreshed with "live reload". And if you don't want any of this you just turn off HMR *and* "live reload", but not turn off "live reload" by turning off HMR which is what you proposing

byzyk:
@andreyvolokitin I agree, I also feel like `HMR` and `livereload` should be 2 separated features. That's exactly what I stated in PR description. The only reason I haven't implemented it in this PR is that I wanted to confirm this with the project maintainers first.

However, the proposed solution won't disable `livereload` completely. It'll be still available when `hot: true` is explicitly set.

andreyvolokitin:
Live reload for changes in HTML is currently not working with ```hot: true``` (#1271). With the proposed solution it also won't work with ```hot: false```, so for changes in HTML it will completely disable live reload.

byzyk:
@TheLarkInn would you be able to review and give a feedback whether this PR is ok to merge?

andreyvolokitin:
- why are you calling a "bugfix" what in fact is a feature removal (removal of live reload)? Live reload is a core feature of ```webpack-dev-server``` and not a bug, and HMR is a different feature on top of it. Why are you trying to disable live reload by disabling HMR?
- like you said, the new option is a way to go. Why are you still trying to merge a PR which will break all those using live reload and not HMR?

For additional info, see: https://github.com/webpack/docs/wiki/webpack-dev-server#automatic-refresh

```inline: false``` might be solving your issue already

byzyk:
@andreyvolokitin please refer to PR description. Suggested changes don't remove so-called `live-reload`. You'll still have it when explicitly set `hot: true`, the same way as before. 

As per #1251 , the issue was that page kept reloading even when `hot: false` or not specified (HMR is disabled by default) or even with `hotOnly: true`, which is clearly a bug, isn't it?

As we already discussed, currently there is no option to explicitly enable `livereload` only. It's rather a fallback for HMR, but it would be nice to have it as an option as well. That's why I suggested to add it as a standalone feature and think it's out of the scope of this PR, although I will be more than happy to work on that once confirmed with project maintainers.

Does that address your concerns or did I miss something?

andreyvolokitin:
Yes, you keep missing the point of live-reload. Live-reload is not an addition to HMR. It is the other way around: HMR is an addition to live-reload. And all in all, they are two separate features. Live-reload is always enabled and can not be disabled currently (again, ```inline: false``` might be worth looking at, as it removes page reloading JS from the client and the page itself might also be available at ```http://«host»:«port»/«path»``` so that there is no auto-reloading happening at all). HMR can be enabled additionally.

> Suggested changes don't remove so-called live-reload. You'll still have it when explicitly set hot: true, the same way as before.

Your changes remove live-reload as a basic page refreshing mechanism, which is currently always enabled and is also active with ```hot: false``` (because by design ```hot``` option is not controlling live-reload). ```hot: true``` is an option which enables HMR, conceptually it has nothing to do with live-reload, and should not control live-reload, but in your changes it does. So it will be not the same way as before: live-reload will stop working for those who are not using HMR.

> As per #1251 , the issue was that page kept reloading even when hot: false or not specified (HMR is disabled by default)

This is intended behavior, because by design ```hot``` option is not controlling live-reload — there is live-reload always enabled, and on top of it you can enable or disable HMR with ```hot``` option. The actual problem in this case is that there is no way to disable live-reload.

> or even with hotOnly: true, which is clearly a bug, isn't it?

I think it isn't. [```hotOnly``` description](https://webpack.js.org/configuration/dev-server/#devserver-hotonly) states that it «Enables Hot Module Replacement (see devServer.hot) without page refresh **as fallback in case of build failures**.». I am not sure what **build failures** means exactly, but I guess ```hotOnly``` means "do not reload a page to try to keep it updated if there is no way to apply HMR in case of some **build failures or errors**". So I guess ```hotOnly``` is only about this narrow "build-failure" edge case, and not about "always disable live-reload and only use HMR".

> As we already discussed, currently there is no option to explicitly enable livereload only. 

Actually, there is: live-reload is always enabled, and if HMR is disabled with ```hot: false```, then live-reload is the only thing enabled. So the thing is that *there is no option to explicitly **disable** livereload*

> It's rather a fallback for HMR, but it would be nice to have it as an option as well.

 As stated in the [docs](https://github.com/webpack/docs/wiki/webpack-dev-server#automatic-refresh), live-reload is not a fallback for HMR, but HMR is an additional feature on top of live-reload. Live-reload is called "automatic refresh" there and has two modes: "inline" and "iframe".

byzyk:
@andreyvolokitin thanks for your comment.

Let me clarify a bit. I can't remember that I've ever said `livereload` being an addition to HMR. From the very beginning, I stated that those should be 2 distinct features. I'm aware of the fact that `livereload` is enabled by default, and again I know _it's not the same as HMR_. 

Your concern is that this PR is aimed to remove `livereload` feature while again, it's not true. Check the PR description, it will just disable `livereload` by default which was requested in #1251 and approved by @shellscape. I agree with you that it's not the best way to enable page refresh by actually specifying `hot` option as it indeed is not relevant. That's why I expressed the same concern in PR description and feel like this can be added as another option cos some people might find it handy. And that's exactly why I asked for reviews from project maintainers, to see how to approach this issue further or if it makes any sense at all (there is a pretty big chance it isn't, and that's fine).

While I appreciate your input, I don't get it why are you so worried about `livereload` being removed after this is merged, as this PR is not about removing it. However, it might change some default functionality or expected behavior and thank you for pointing it out. But it's not the same as removing feature as a whole.

Just as a side note, you're referencing docs (which might not be that precise and could be outdated as well), while I'm trying to back up my statements from what I've seen in the source code. Thus, I don't agree with you on `hotOnly` option, because [from code perspective](https://github.com/webpack/webpack-dev-server/blob/83c16251a16794b61fa22761d43b6450bfb18207/client-src/default/index.js#L211) it seemed that `livereload` was purely meant to be a fallback for HMR. However, I can be wrong on that.

Anyway, let's wait for the webpack maintainers (@sokra @TheLarkInn @shellscape) to jump in and will see how we should proceed with this one. Otherwise, we can keep this argument going on forever while basically saying almost the same stuff just in different words.

Peace ✌🏽

evilebottnawi:
@byzyk 
> However, it might change some default functionality or expected behavior and thank you for pointing it out. But it's not the same as removing feature as a whole.

Should have major label (and will be in next branch, not in master)

andreyvolokitin:
@byzyk 

> I don't get it why are you so worried about livereload being removed after this is merged, as this PR is not about removing it

Because of this: https://github.com/webpack/webpack-dev-server/pull/1276#issuecomment-359171664

byzyk:
@evilebottnawi thanks, will make those amends when I get a chance.

What is your view on the discussion above? Does it make sense for page reload to be a distinct feature with its own option like `live` or `liveReload` so users can enable/disable it explicitly?

byzyk:
@andreyvolokitin I don't see how #1271 is relevant to this. Originally this PR was meant to address #1251

evilebottnawi:
@byzyk for me yes, i have some project on `php` and `laravel/blade` and now use hack for reloading page https://github.com/webpack/webpack-dev-server/issues/1271#issuecomment-359817498, be good have this option in out of box in next branch, in theory i can use `browser-sync` for this, but for me it is overload tool. Also we already have `chokidar`, just add option for array of glob and refresh page when something changes.

andreyvolokitin:
@byzyk PR is disabling live-reload if ```hot: false```. Also currently with ```hot: true``` live-reload for HTML changes is not working. So with this PR HTML live-reload will not work with ```hot: true``` and with ```hot: false```, though currently it is working with ```hot: false```

byzyk:
@andreyvolokitin actually in this PR page refresh should work with `hot: true`. Have you tested it for your case?

andreyvolokitin:
@byzyk yep, sorry, it is working, my bad. But now there is no HMR for css and js with ```hot: true```, just a full page reload. And with ```hotOnly: true``` it is an old story: HMR and no live-reload for HTML.

andreyvolokitin:
I saw the code and what you changed from the beginning. I still think that you misinterpreted the meaning and intention of "hot" and "hotOnly" options. In your PR the code is doing 1) [HMR](https://github.com/byzyk/webpack-dev-server/blob/88b89bbd4054f94a18a67c4f834a99a3eac5769e/client-src/default/index.js#L214) with ```hotEmitter.emit('webpackHotUpdate', currentHash);```, and then immediately 2) a [full page reload](https://github.com/byzyk/webpack-dev-server/blob/88b89bbd4054f94a18a67c4f834a99a3eac5769e/client-src/default/index.js#L229). There is no sense in it if you try to understand this. And it does only HMR without page reload if ```hotOnly: true``` — this, on the contrary, is logical, and this is actually the current behavior of ```webpack-dev-server``` with ```hot: true``` (which causes the problem of not updating HTML changes). So yeah, I guess maintainers could clear this up.

byzyk:
@evilebottnawi do you still think the approach I took here is the right one to move forward with? Based on the discussions above I feel like we should distinguish between `livereload` and `hmr` features.

evilebottnawi:
/cc @byzyk sorry for big delay, i think you are right, we need:
- `hot` option enable/disable HMR
- `livereload` option for enable/disable auto reload page (HMR still works)

Summary: 
- using `--no-hot` and `--no-livereload` you disable HMR and livereload
- using `--no-hot` you disable only `HMR` (browser now always reload pages)
- using `--no-livereload` you disable livereload (`HMR` still works)
- no options `HMR` and `livereload` works together

Also we deprecated `--hot-only` option because it is misleading

What do you think?

kaiyoma:
+1 for an option to disable the full page refresh. There's a bit of discussion here too: https://github.com/webpack/webpack-dev-server/issues/1744

tl;dr: Our codebase doesn't work with HMR, so I'd like to disable it, but that enables the full page refresh mechanism. Our UI is expensive to load, so reloading the page on every file change is extremely disruptive. It would be incredibly useful to have webpack build on every change, but I control when the page reloads.

byzyk:
Hi @evilebottnawi! 

Sorry for a delay, yep that's something I had in mind originally and I guess that's exactly what other folks are asking as well.

evilebottnawi:
Here answer https://github.com/webpack/webpack-dev-server/pull/1276#issuecomment-473350942, this PR is abandoned so feel free to send a new PR

byzyk:
Hi everyone! 

Sorry guys but unfortunately I won't be able to finalize this one because of the current workload. I'll try to revisit it in a month or so. In the meanwhile feel free to pick up where I left or simply submit new PR since codebase has evolved a lot.

evilebottnawi:
Implemented you can disable hot using `--no-hot` and disable liveReload using `--no-live-reload` (also supported using `devServer` options)

richard-ling:
One or both of `--no-hot` and `-no-live-reload` appear to still be broken (webpack 4.28.3, webpack-dev-server 3.1.14).  Running as follows
`$ webpack-dev-server --progress --port 8080 --verbose --no-live-reload --no-hot

    Running webpack with config from '.build/config.json':
    NODE_ENV=webpack-dev
    version=dev-c0612f2-20200507003035441
    app-url=https://localhost/myapp/
    api-endpoint=prod
    use-dev-claims=false
  
 10% building 2/2 modules 0 activeℹ ｢wds｣: Project is running at http://0.0.0.0:8080/
ℹ ｢wds｣: webpack output is served from /myapp/
ℹ ｢wds｣: Content not from webpack is served from ./build
ℹ ｢wds｣: 404s will fallback to /index.html
ℹ ｢wdm｣:    3620 modules                                                            
ℹ ｢wdm｣: Compiled successfully.
`
[At this point I change and resave a source file...]
`
ℹ ｢wdm｣: Compiling...
ℹ ｢wdm｣:    3620 modules                                                       
ℹ ｢wdm｣: Compiled successfully.
`
Then the browser reloads, losing all state.  Is that as expected?

PR #1279, opened 2018-01-16
no babel-loader node_modules exclude/web_modules  Fixes #1278

### LINKED ISSUE #1278: Version update from 2.9.5 to 2.11.0 broke the build for IE11
Sorry for the lazy report, but I'm a tad busy to figure out what possibly causes this or provide an example project.

If someone could confirm this is broken for IE11 in general, thanks.

We tried update our webpack-dev-server from 2.9.5 to 2.11 and in IE11 it causes the following: 
image

* Operating System: Windows 10
* Node Version: 6.9.1
* NPM Version: 3.10.8
* webpack Version: 3.10.0
* webpack-dev-server Version: 2.11.0

- [X] This is a **bug**

graingert:
and babel@7 so it's safe to do so.

Fixes #1278

also remove ancient unused version of jquery hiding in `web_modules`.


graingert:
I had @nobitagit try this out in IE11 and it works great.

graingert:
thank you @sindresorhus for dragging us into the future!

graingert:
@julianxhokaxhiu are you sure that's not in your own code?

Here's the bundled JS https://unpkg.com/@procensus/webpack-dev-server@2.11.0/client/index.bundle.js

julianxhokaxhiu:
Yes definitely sure. Feel free to check the project here: https://github.com/julianxhokaxhiu/polysticky.js

graingert:
@julianxhokaxhiu you've not switched to @procensus/webpack-dev-server in polysticky.

julianxhokaxhiu:
I tested it locally, and it didn't work. So no need to push a commit that is useless.

Feel free though to try on your side. Remember that you can try locally by using `npm run serve`

yyx990803:
Just to explain why #1273 was needed instead of doing this: the webpack/babel transpilation only affects the bundles used in live mode. When the user uses WDS via the CLI with `--inline` flag, [the client is injected by adding an entry to the users' entire bundle](https://github.com/webpack/webpack-dev-server/blob/master/lib/util/addDevServerEntrypoints.js#L20). So this PR would be an incomplete fix.

SpaceK33z:
Closing because #1273 fixed the same issue.

Issue #1021, opened 2017-08-08
[2.7.0] Recursive SockJS Errors


skipjack:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Start the dev server with HMR enabled. Here's my dev server configuration:

```js
    devServer: {
        hot: true,
        port:8084,
        inline: true,
        compress: true,
        historyApiFallback: true,
        contentBase: 'dist/',
        overlay: {
            errors: true,
            warnings: true
        }
    }
```

I start getting the following errors:

![image](https://user-images.githubusercontent.com/8988697/29075858-b671470a-7c21-11e7-8992-1ce63fff9a9c.png)

Which then cause other errors in the application and seems to recurse back causing to the same set of errors to continually fire.

**What is the expected behavior?**

No errors. With v2.6.1 of this package everything works fine.

**If this is a feature request, what is motivation or use case for changing the behavior?**

N/A

**Please mention your webpack and Operating System version.**

OSX 10.12.5
Webpack 3.5.1

renchap:
I got a similar problem too.

With `publicPath: 'http://localhost:8080/packs/'` HMR is trying to load `http://localhost:8080/packs/sockjs-node/info?t=1502201588847` which returns a 404 error. `http://localhost:8080/sockjs-node/info?t=1502201588847 returns the expected value.

 This is caused by https://github.com/webpack/webpack-dev-server/pull/911

robmoorman:
Same issue over here 2.7.0 seems to be broken

the-spyke:
Not need for enabling HMR by the way. It doesn't work even without it.

shellscape:
We're in a pickle because the updates to socksjs passed all of the tests we presently have before it was merged, and it was reviewed. So it wasn't a bug that was willfully introduced. If anyone can create a small example that devs can use to reproduce, that would be immensely helpful.

karolyi:
+1

jaredlunde:
@shellscape It's not the simplest example but here's a failing app: https://github.com/jaredlunde/webpack-dev-server-failure-example/tree/master

shellscape:
@jaredlunde thanks, that at least got us moving forward. I modified your server options in `server.js` and was able to stem the tide of websocket errors:

```js
new WebpackDevServer(webpack(config), {
  publicPath: config.output.publicPath,
  sockjsPrefix: '/assets',
  historyApiFallback: {
		disableDotRule: true,
[...]

PR #911, opened 2017-05-18
Sockjs prefix config

kellyrmilligan:
**What kind of change does this PR introduce?**
This adds the capability to prefix the url that `sockjs-node` is served from. This is useful for having multiple applications in development being behind a reverse proxy. 

**Did you add or update the `examples/`?**

**Summary**

https://github.com/webpack/webpack-dev-server/issues/909

**Does this PR introduce a breaking change?**

**Other information**

PR #1289, opened 2018-01-22
feat: add `sockPath` option (`options.sockPath`)

cheapsteak:
- [x] This is a **bugfix**
- [ ] This is a **typo fix**

Added new tests for addDevServerEntrypoints

I've added examples referenced from https://github.com/webpack/webpack-dev-server/pull/911

I've tested this against the [breaking example provided in 2.7.0](https://github.com/webpack/webpack-dev-server/issues/1021#issuecomment-321001601) and it doesn't suffer the same failure.

### Motivation / Use-Case

Similar motivations as https://github.com/webpack/webpack-dev-server/pull/911. 

We also have multiple apps being developed behind a reverse proxy (in our case a docker composed nginx). Each app is served from a top-level sub-directory (can't use root path), e.g. `host/app-part-one`, `host/app-another-piece`

Not sure if v3 fixes this and not sure how long until v3 comes out. This should fix our use case and shouldn't break anything else in the mean time

### Breaking Changes

None

### Additional Info

It doesn't rely on the `publicPath` value, but unfortunately the `sockPath` value has to be passed twice, once in the `devServer` option (this is what Server.js uses to determine what prefix to listen at), and once again in the entry resource url as a query parameter (this is what client/index.js uses to determine where to send requests to)

### DIFF
--- a/client-src/default/index.js
+++ b/client-src/default/index.js
@@ -2,7 +2,7 @@
 
 /* global __resourceQuery WorkerGlobalScope self */
 /* eslint prefer-destructuring: off */
-
+const qs = require('querystring');
 const url = require('url');
 const stripAnsi = require('strip-ansi');
 const log = require('loglevel').getLogger('webpack-dev-server');
@@ -194,7 +194,7 @@ const socketUrl = url.format({
   auth: urlParts.auth,
   hostname,
   port: urlParts.port,
-  pathname: urlParts.path == null || urlParts.path === '/' ? '/sockjs-node' : urlParts.path
+  pathname: urlParts.path == null || urlParts.path === '/' ? '/sockjs-node' : (qs.parse(urlParts.path).sockPath || urlParts.path)
 });
 
 socket(socketUrl, onSocketMsg);
--- a/examples/node-api-sock-path/README.md
+++ b/examples/node-api-sock-path/README.md
@@ -1,0 +1,15 @@
+# Node.js API - Simple
+
+```shell
+node server.js
+```
+
+Starts a simple webpack-dev-server setup via the Node API. Open `http://localhost:8080/` to go the app.
+
+## What should happen
+
+In the app you should see "It's working."
+
+In `app.js`, uncomment the code that results in an error and save. This error should be visible in the CLI and devtools.
+
+Then, in `app.js`, uncomment the code that results in a warning. This warning should be visible in the CLI and devtools.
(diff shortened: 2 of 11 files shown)


cheapsteak:
Would anyone be able to help with the codecov CI error? I'm not quite sure what I need to do there

burtek:
@cheapsteak Does this allow to explicitely set hostname for socket? I need it to connect to a dev server on `localhost:3000` from inside another dev site hosted on `:80`

cheapsteak:
@burtek unfortunately that is not part of this PR  

I believe you should be able to reference my changes and add additional fields for what you'd need (e.g. `sockPort` and `sockHost`)

burtek:
@cheapsteak got it, thanks for answer

Jokero:
Also found that I need a way to change the port for sockjs)

Jokero:
Ah, you don't use inline mode. Otherwise there is already `public` option https://webpack.js.org/configuration/dev-server/#devserver-public

cheapsteak:
The `public` option only affects where the client tries to connect to, but the server is hard coded to listen at `'/sockjs-node'`

flyyang:
What's the status of this PR?

iphands:
@flyyang I converted the project I am working on to use webpack-serve. That project seemed to do the "right thing" regarding proxied web sockets (for hot reloading) straight out of the box.

trescenzi:
Hi what's the status of this PR. Seems like it fell out because `webpack-serve` was doing the right thing. But now that `webpack-serve` is deprecated in favor of `webpack-dev-server`. Is there any chance this could get merged?

I'd be happy to help get it up to date with master and make the changes, host/port, mentioned in [this comment](https://github.com/webpack/webpack/pull/8112#issuecomment-425674544).

THernandez03:
> @cheapsteak Does this allow to explicitely set hostname or at least port for socket? I need the socket to connect to a dev server on `localhost:3000` from inside another dev site hosted on `localhost:80`

That should be very nice too, be able to determine the entire url of the socket, not only the path.

evilebottnawi:
Need rebase or create new Pr

trescenzi:
@evilebottnawi I've got it up to date in my fork. Don't want to just commandeer this though. Especially without confirmation it'll be accepted.

cheapsteak:
Commandeer away! I'll close this PR

Issue #1316, opened 2018-02-23
client javascript doesn't reload the page if there's a build-time error on first load


JayAndCatchFire:
* Operating System: MacOS High Sierra, Ubuntu 16.04
* Node Version: 8.9.4
* NPM Version: 5.6.0
* webpack Version: 3.11.0
* webpack-dev-server Version: 2.11.1

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

I've made a repository that minimally recreates this issue.  [JayAndCatchFire/minimal-wds-bug-example](https://github.com/JayAndCatchFire/minimal-wds-bug-example)

#### Scenario 1
1. The user's code being packed with webpack has a build-time error in it.
2. The user starts webpack-dev-server and navigates to the app in their browser.  The app doesn't fully load due to the build-time error.
3. The build-time error in the user's code is corrected and webpack successfully rebuilds the bundle.
4. Webpack-dev-server should reload the app page with the newly built code.

#### Scenario 2
1. The user has been running webpack-dev-server for a bit and introduces a code change in their code that includes a build-time error.
2. The user navigates to their app in a browser with a fresh tab at this point or manually hard-reloads the page at this point.  The app doesn't fully load due to the build-time error.
3. The build-time error in the user's code is corrected and webpack successfully rebuilds the bundle.
4. Webpack-dev-server should reload the app page with the newly built code.

### Actual Behavior
In both of the above scenarios, the fourth step doesn't occur.  The page remains in a broken state.

### For Bugs; How can we reproduce the behavior?
The steps outlined in the scenarios above should outline how to reproduce the bug reliably, although I give more detailed instructions in the README file in my [example repository](https://github.com/JayAndCatchFire/minimal-wds-bug-example).

JayAndCatchFire:
I believe the reason this is happening is because of the `initial` variable in the [client source](https://github.com/webpack/webpack-dev-server/blob/83c16251a16794b61fa22761d43b6450bfb18207/client-src/default/index.js#L46).  I suspect it's the culprit because if you correct the client code that's causing a built-time error, reintroduce the erroneous code, and then remove it a second time, the app will be reloaded as expected.

SpaceK33z:
Hey I'm working on this right now and looking through your issue and PR, and I just want to say: THANK YOU for looking into this so extensively and making an example repo. I'm going through it to first reproduce your bug and then test your fix.

JayAndCatchFire:
[...]

PR #1317, opened 2018-02-23
Fix issue #1316

JayAndCatchFire:
- [x] This is a **bugfix**
- [ ] This is a **typo fix**

I did not.  I absolutely want to, but it appears that the `client-src/default/index.js` file has no tests at the moment.  Or if it does, I couldn't find them in the `test` directory.

I would really love some help and/or guidance on how best to test this code.

### Motivation / Use-Case
See #1316 for the bug this addresses.

### Breaking Changes
There's a _possibility_ that this could cause unnecessary reloads in other situations.  I don't think it will, but I'm not 100% sure it won't, especially since this bit of code doesn't seem to have any tests right now.

### DIFF
--- a/client-src/default/index.js
+++ b/client-src/default/index.js
@@ -156,6 +156,7 @@ const onSocketMsg = {
     sendMsg('Errors', strippedErrors);
     for (let i = 0; i < strippedErrors.length; i++) { log.error(strippedErrors[i]); }
     if (useErrorOverlay) overlay.showMessage(errors);
+    initial = false;
   },
   error(error) {
     log.error(error);


JayAndCatchFire:
What's going on with travis-ci?  It's been stuck in yellow dot status for 15 hours.

JayAndCatchFire:
@sokra @shellscape @SpaceK33z Can someone please give travis-ci a kick?  It's been stalled for 2 days.

SpaceK33z:
Looking into this PR right now.

SpaceK33z:
I don't know what is going on with Travis but I'm going to merge this since I confirmed it works locally.

Thanks again @JayAndCatchFire!

Issue #1318, opened 2018-02-25
Dynamic entry is broken in webpack-dev-server


mikegreiling:
Dynamic entries do not seem to be well supported by webpack-dev-server.

The `DynamicEntryPlugin` is a poorly documented webpack feature, but the webpack `entry` option can be defined as a function that returns either an entry object or one that returns a Promise which resolves to an object.

e.g.

```javascript
// webpack.config.js
export default {
  entry: async () => {
    const entries = {};
    // generate entries
    return entries;
  }
  //...
}
```

This was _sort of_ addressed with #801 and #802 in which the following lines were added in [/lib/util/addDevServerEntrypoints.js](https://github.com/webpack/webpack-dev-server/blob/master/lib/util/addDevServerEntrypoints.js#L27-L29):

```javascript
if (typeof wpOpt.entry === 'function') {
  wpOpt.entry = wpOpt.entry(devClient);
}
```

However this is incorrect for several reasons:

1. For one, the `entry` function is not expected to accept any arguments, so passing `devClient` into the function is a bit confusing.  This won't necessarily add the `devClient` to the resulting entry object unless it was defined by the user in such a way that it uses that argument.  Why not append it to the final `entry` object yourself so you can be sure it is included?

2. This also redefines the `wpOpt.entry` as a static object by running the function, rather than keeping it dynamic by, say, wrapping the function in another closure that can re-run and re-append `devClient` each time it is invoked.

3. Async functions are completely broken.  Since this code resolves the function, an async `wpOpt.entry` function would resolve to a `Promise` object and webpack will throw an error because it does not expect this.

I'd be happy to open a PR to fix this if it would be accepted.

PR #1310, opened 2018-02-17
Allow no publicPath or entry point

rynclark:
- [x] This is a **bugfix**
- [ ] This is a **code refactor**
- [ ] This is a **test update**
- [ ] This is a **typo fix**
- [ ] This is a **metadata update**

### For Bugs and Features; did you add new tests?

Yes

### Motivation / Use-Case

Fixes this issue https://github.com/webpack/webpack-dev-server/issues/1308

Allows for `webpack-dev-server` to run with no config or no configured entry point, keeping the behaviour consistent with the webpack compiler.

### Additional Info

This should eventually be moved out of `webpack-dev-server` as we've currently got two places setting the default entry point (this and `WebpackOptionsDefaulter`).

PR #1319, opened 2018-02-25
Fix support for DynamicEntryPlugin

mikegreiling:
- [x] This is a **bugfix**
- [ ] This is a **typo fix**

Yes.  I have added tests for this change and I have added new tests for prior behavior to prove that this doesn't break anything.

### Motivation / Use-Case

The support for DynamicEntryPlugin added in #802 was incorrect.  This fixes `addDevServerEntrypoints.js` as it applies to dynamic endpoints (functions which return an entry object and functions which return a Promise)

### Breaking Changes

none.

### Additional Info

This pull request targets the `webpack-4` v3.x branch, but I would like to port this back to `2.x` as well if you plan to make any more releases under that major version.  The only conflict would be the default entry point added in #1310.  I could simply exclude that behavior from this change and open up a separate PR targeting the master branch.

Closes: #1318

### DIFF
--- a/lib/util/addDevServerEntrypoints.js
+++ b/lib/util/addDevServerEntrypoints.js
@@ -19,16 +19,22 @@ module.exports = function addDevServerEntrypoints(webpackOptions, devServerOptio
 
     if (devServerOptions.hotOnly) { devClient.push('webpack/hot/only-dev-server'); } else if (devServerOptions.hot) { devClient.push('webpack/hot/dev-server'); }
 
-    [].concat(webpackOptions).forEach((wpOpt) => {
-      if (typeof wpOpt.entry === 'object' && !Array.isArray(wpOpt.entry)) {
-        Object.keys(wpOpt.entry).forEach((key) => {
-          wpOpt.entry[key] = devClient.concat(wpOpt.entry[key]);
-        });
-      } else if (typeof wpOpt.entry === 'function') {
-        wpOpt.entry = wpOpt.entry(devClient);
-      } else {
-        wpOpt.entry = devClient.concat(wpOpt.entry || './src');
+    const prependDevClient = (entry) => {
+      if (typeof entry === 'function') {
+        return () => Promise.resolve(entry()).then(prependDevClient);
       }
+      if (typeof entry === 'object' && !Array.isArray(entry)) {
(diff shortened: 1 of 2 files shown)


lib/util/addDevServerEntrypoints.js
```diff
@@ -19,16 +19,22 @@ module.exports = function addDevServerEntrypoints(webpackOptions, devServerOptio
 
     if (devServerOptions.hotOnly) { devClient.push('webpack/hot/only-dev-server'); } else if (devServerOptions.hot) { devClient.push('webpack/hot/dev-server'); }
 
-    [].concat(webpackOptions).forEach((wpOpt) => {
-      if (typeof wpOpt.entry === 'object' && !Array.isArray(wpOpt.entry)) {
-        Object.keys(wpOpt.entry).forEach((key) => {
-          wpOpt.entry[key] = devClient.concat(wpOpt.entry[key]);
+    const prependDevClient = (entry) => {
+      if (typeof entry === 'function') {
+        return () => Promise.resolve(entry()).then(prependDevClient);
```
mikegreiling: This will handle both functions which return entry objects and functions which return a `Promise` due to the behavior of `Promise.resolve`.  It is the same technique used in the `DynamicEntryPlugin` itself ([link](https://github.com/webpack/webpack/blob/213226ece261a0e89edb9de15fbc6775418f6041/lib/DynamicEntryPlugin.js#L49)).

test/Entry.test.js
```diff
@@ -2,17 +2,112 @@
...
+
+  it('defaults to src if no entry point is given', () => {
     const webpackOptions = {};
     const devServerOptions = {};
 
     addDevServerEntrypoints(webpackOptions, devServerOptions);
 
-    assert(webpackOptions.entry.length, 2);
-    assert(webpackOptions.entry[1], './src');
+    assert.equal(webpackOptions.entry.length, 2);
```
mikegreiling: I've fixed this test also.  By using `assert` and not `assert.equal`, these tests were not doing anything other than ensuring `webpackOptions.entry.length` was not falsey.

SpaceK33z:
Thanks for this fine PR!!

Issue #1311, opened 2018-02-19
Feature request: Add default value for `output.public`


epeli:
* Operating System: any
* Node Version: v8.9.4
* NPM Version: 5.6.0
* webpack Version: 4.0.0-beta.1
* webpack-dev-server Version: 3.0.0-beta.2

- [ ] This is a **bug**
- [x] This is a **modification** request

### Code

```js
  // webpack.config.js
  // No config :)
```

Since #1295 got closed I'm opening a new issue of a feature request I mentioned there.

Could webpack-dev-server provide a default for the `output.publicPath` config? With this webpack-dev-server could be also used without any config like Webpack 4.

I'd propose the config to be

```
output: {
  publicPath: "/dist",
}
```

because it matches the default Webpack 4 ouput directory.

SpaceK33z:
Yeah that might be a good idea. Are you interested in making a PR?

epeli:
Yeah, I can come up with something for sure.

epeli:
Looking at the code there actually seems to be a default for this:

https://github.com/webpack/webpack-dev-server/blob/v3.0.0-beta.2/bin/webpack-dev-server.js#L267

Currently it becomes `/main.js` when there is no `webpack.config.js`.

I but would still like to change it to `/dist/main.js` to match with Webpack 4.

The reasoning is that one would not have to add any config or edit the script when using the webpack-dev-server in a Webpack 4 project without the config file.

Ie.

```
webpack-dev-server
```

Would be basically same as

```
webpack --watch
python -m SimpleHTTPServer
```

but with automatic browser reloads.

PR #1322, opened 2018-02-25
refactor(webpack-dev-server): use default `publicPath`

epeli:
- [ ] This is a **bugfix**
- [ ] This is a **typo fix**

PR for #1311

No... There doesn't seem to be any existing direct tests for the `bin/webpack-dev-server.js` script.

### Motivation / Use-Case

Simple zero configuration projects. 

Example <https://github.com/epeli/0cjs-webpack-dev-server>

It hilights how it should work after this PR.

I'm going to do a small presentation about Webpack and this would make it easier to sell Webpack to my audience. People seem to be quite intimidated by Webpack configurations. This would make the beginning very smooth.

### Breaking Changes

This is a breaking change. Currently the bundle is served from `/main.js` after this it will be `/dist/main.js` which aligns with the default Webpack 4 ouput.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,10 @@ execute the file or command.
 Either method will start a server instance and begin listening for connections
 from `localhost` on port `8080`.
 
+By default the entry file will be `src/index.js` and it will be served at
+`/dist/main.js` when the `--mode` flag is used. These can be customized in
+`webpack.config.js` if needed.
+
 webpack-dev-server is configured by default to support live-reload of files as
 you edit your assets while the server is running.
 
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -264,7 +264,12 @@ function processOptions(webpackOptions) {
     // eslint-disable-next-line
     options.publicPath = firstWpOpt.output && firstWpOpt.output.publicPath || '';
     if (!/^(https?:)?\/\//.test(options.publicPath) && options.publicPath[0] !== '/') {
-      options.publicPath = `/${options.publicPath}`;
+      if (webpackOptions.mode) {
+        // Enable Webpack 4 like default path when the --mode flag is set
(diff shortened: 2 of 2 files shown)


SpaceK33z:
Thanks for your PR. This change should have been in webpack-dev-server v3 which I released this morning. Which makes this kinda difficult to handle since this change makes so much sense but it might be confusing to immediately release a next major version.

Just thinking in some solutions, perhaps we can detect if `mode` is used and only then apply this change.

epeli:
Yeah, sorry I missed that mark...

But is that bad actually since the v3 release can now act as the version with a "deprecation" warning about the change? Bit like React does breaking changes.

Also the default seems to be currently undocumented?

epeli:
> Just thinking in some solutions, perhaps we can detect if mode is used and only then apply this change.

I'm not sure about this. Seems too "clever" which can get very confusing. I would not expect it to change with the mode flag...


Weird. Travis seems to have failed after I added documentation to this PR.

SpaceK33z:
webpack seems to do the same when you don't set `mode`:

```
WARNING in configuration
The 'mode' option has not been set. Set 'mode' option to 'development' or 'production' to enable defaults for this environment.
```

epeli:
Oh right. Makes sense. Not fan of that but I do think webpack-dev-server should work like the `webpack` cli as much as possible.

epeli:
I rebased the pr with this feature.

SpaceK33z:
Thanks, I will try to look at it this week. This change does have some consequences that I need to look into.

epeli:
Did you find anything?

SpaceK33z:
Sorry don't have the time for this yet. I am a bit scared of making this change without releasing a major version so I want to understand the consequences fully first.

michael-ciniawsky:
In case my assumption is wrong here, feel free to reopen :)

PR #1362, opened 2018-04-04
speed up incremental builds by not doing excessive stats.toJSON work

kenotron:
- [x] This is a **bugfix**
- [ ] This is a **typo fix**

There's no need to add tests, as it is just a configuration change to toJSON

### Motivation / Use-Case

The stats.toJSON call is done every time an incremental build finishes. Given a sufficiently large codebase, this is very SLOW (my repo went from ~30s to ~10s just by getting rid of this)

### Breaking Changes

None

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -25,7 +25,7 @@ const createLog = require('./createLog');
 const OptionsValidationError = require('./OptionsValidationError');
 const optionsSchema = require('./optionsSchema.json');
 
-const clientStats = { errorDetails: false };
+const clientStats = { errorDetails: false, chunkSortMode: 'none', chunks: false };
 
 function Server(compiler, options, _log) {
   // Default options
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -25,7 +25,7 @@ const createLog = require('./createLog');
 const OptionsValidationError = require('./OptionsValidationError');
 const optionsSchema = require('./optionsSchema.json');
 
-const clientStats = { errorDetails: false, chunkSortMode: 'none', chunks: false };
+const clientStats = { all: false, assets: true, warnings: true, errors: true, errorDetails: false };
 
 function Server(compiler, options, _log) {
   // Default options


lib/Server.js
```diff
@@ -25,7 +25,7 @@ const createLog = require('./createLog');
 const OptionsValidationError = require('./OptionsValidationError');
 const optionsSchema = require('./optionsSchema.json');
 
-const clientStats = { errorDetails: false };
+const clientStats = { errorDetails: false, chunkSortMode: 'none', chunks: false };
```
sokra: I think you can even go with

``` js
{ all: false, assets: true, warnings: true, errors: true, errorDetails: false }
```

since these are the only fields used.

kenotron: will do!

kenotron: Updated

kenotron:
@sokra - can you please merge after my update?

evilebottnawi:
@kenotron just wait @shellscape review

evilebottnawi:
@shellscape let's do this :+1:

kenotron:
@shellscape would mind taking this change? Not everyone in the world can afford to move to the next new thing. I think this little bit of maintenance mode change would benefit quite a LOT of people who still must stick with patch / minor changes.

shellscape:
@kenotron I don't think you understand - I am not a maintainer of this repo, hence I cannot make these changes. I was only asked for my opinion on the change. I'll be unsubscribing from this thread now, as I don't care for it when users levy unfair accusations at me.

SpaceK33z:
Fix released in webpack-dev-server@3.1.2.

SpaceK33z:
Note that this broke HMR, which was fixed in https://github.com/webpack/webpack-dev-server/commit/e1b263a99d510f31e2ff418a0725352ebb8fb9ea.

SpaceK33z:
@kenotron I am curious if the above fix has any negative performance implications? If so we could perhaps only enable the `hash` option when HMR is used.

kenotron:
@SpaceK33z the hash option didn't significantly do anything to our build times, so I think it's probably okay to leave it be. I'd report another bug if we noticed any anomalies.

PR #1363, opened 2018-04-04
[2.x backport] perf(Server): don't enable all `stats` by default

kenotron:
- [x] This is a **bugfix**
- [ ] This is a **typo fix**

No - tests currently covered the scenario

### Motivation / Use-Case

Merging the same speedy change from v3.x code to v2.x (we still use it since v2 is slightly faster for inc builds still for us)

### Breaking Changes

No breaking changes

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -24,7 +24,7 @@ const webpackDevMiddleware = require('webpack-dev-middleware');
 const OptionsValidationError = require('./OptionsValidationError');
 const optionsSchema = require('./optionsSchema.json');
 
-const clientStats = { errorDetails: false };
+const clientStats = { all: false, assets: true, warnings: true, errors: true, errorDetails: false };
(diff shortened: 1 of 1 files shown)


kenotron:
@SpaceK33z I know this is super in maintenance mode (v2), but would you mind merging this one too?  The v2 branch somehow is faster than v3, so at scale, we're still having to use this.

evilebottnawi:
@kenotron all good, v2 also in `maintenance mode`, because many people use `webpack@3`

evilebottnawi:
@SpaceK33z friendly ping :+1:

michael-ciniawsky:
This isn't really a patch as sadly as it is. I'm not sure if we can backport this tbh

michael-ciniawsky:
Sry but this is a breaking chnage for the `v2.x` line, please upgrade to `v3.x` instead

PR #1383, opened 2018-04-19
Use webpack-dev-middleware ^3.1.3.

jasondavies:
Unclear why a caret is not already in use here (or even a tilde).


jasondavies:
Err, never mind!

PR #1397, opened 2018-05-03
don't use argv outside of processOptions function.

0xorial:
- [ ] This is a **bugfix**
- [ ] This is a **typo fix**

### Motivation / Use-Case
I want to use the webpack-dev-server from my js code (not command line). To do that I will need to separate command line arguments passing. Mostly it is easy to do, but in those 2 places global argv is unfortunately referenced.

### Breaking Changes
None

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "webpack-dev-server",
-  "version": "3.1.3",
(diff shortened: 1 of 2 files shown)


evilebottnawi:
@0xorial please rebase on master and fix CI problems. Thanks!

PR #1398, opened 2018-05-03
don't use argv outside of processOptions function.

0xorial:
- [ ] This is a **bugfix**
- [ ] This is a **typo fix**

### Motivation / Use-Case
I want to use the webpack-dev-server from my js code (not command line). To do that I will need to separate command line arguments passing. Mostly it is easy to do, but in those 2 places global argv is unfortunately referenced.

### Breaking Changes
None

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -351,6 +351,10 @@ function processOptions(webpackOptions) {
 
   if (argv.useLocalIp) { options.useLocalIp = true; }
 
+  if (argv.profile) { options.profile = true; }
+
+  if (argv.color) { options.color = true; }
+
   // Kind of weird, but ensures prior behavior isn't broken in cases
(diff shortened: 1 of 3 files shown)


SpaceK33z:
I don't understand this change. Why would you need files in the `bin/` dir? See for an example to use dev-server in JS: https://github.com/webpack/webpack-dev-server/blob/master/examples/api/simple/server.js

0xorial:
Thanks for the example! I wanted to intercept 'done' event of webpack (but otherwise use webpack-dev-server as it is) and I saw a good amount of code in webpack-dev-server.js. So I was afraid that if I use stuff from 'lib' I will have to end up writing the same code as in webpack-dev-server.js (but probably worse xD)

Now, after looking at the code more, I start to think that options contain only options used in the lib/Server, so my approach is probably wrong :(

PR #1401, opened 2018-05-04
fix(examples/cli): correct template path in `open-page` example

elrumordelaluz:
Otherwise is returning  an _Html Webpack Plugin_ error:

```
Error: Child compilation failed:
Entry module not found: Error: Can't resolve '/(...)/webpack/webpack-dev-server/examples/assets/layout.html
```

- [ ] This is a **bugfix**
- [x] This is a **typo fix**

### Motivation / Use-Case

Trying to test the `--open` behaviour, first approach is to check the _example_
and seems broken finding a wrong `path`.

### DIFF
--- a/examples/cli/open-page/webpack.config.js
+++ b/examples/cli/open-page/webpack.config.js
@@ -11,7 +11,7 @@ module.exports = setup({
   plugins: [
     new HtmlWebpackPlugin({
       filename: 'example.html',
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
Released in [`v3.1.6`](https://github.com/webpack/webpack-dev-server/releases/tag/v3.1.6) 🎉

PR #1403, opened 2018-05-09
patch: Bumps chokidar to 2.0.3

jeremyzahner:
- [ ] This is a **bugfix**
- [ ] This is a **typo fix**

### Motivation / Use-Case

While updating a project leveraging webpack-dev-server to Node v10 we've encountered an issue with the `engine` compatibility definition for [upath](https://www.npmjs.com/package/upath). They have fixed that in their own repo (see https://github.com/anodynos/upath/issues/14).

webpack-dev-server uses upath through [chokidar](https://www.npmjs.com/package/chokidar). They have also fixed this engine requirement (as in locking upath to a certain version before the fix) in a recent patch (see https://github.com/paulmillr/chokidar/issues/710).

### Breaking Changes

None

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1991,9 +1991,9 @@
       "dev": true
     },
     "chokidar": {
-      "version": "2.0.2",
-      "resolved": "https://registry.npmjs.org/chokidar/-/chokidar-2.0.2.tgz",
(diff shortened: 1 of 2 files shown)


evilebottnawi:
@jeremyzahner our semver allow install `2.0.3`, why we should change this in package.json? Just update own lock file and all

jeremyzahner:
@evilebottnawi Nevermind, it seems like i've had an issue with a lockfile beeing cached.

PR #1414, opened 2018-05-21
Option to disable "serve-index"

frenzzy:
- [ ] This is a **bugfix**
- [ ] This is a **typo fix**

### Additional Info

Allow to disable [`serve-index`](https://github.com/expressjs/serve-index) when you need to serve your own middleware on the same urls.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -374,7 +374,7 @@ function Server(compiler, options, _log) {
     if (contentBase !== false) { defaultFeatures.push('contentBaseFiles'); }
   }
   defaultFeatures.push('magicHtml');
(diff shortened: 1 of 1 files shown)


frenzzy:
But in this case I need to add one more dependency (`serve-static`) to my package.json and serve static manually using `before` or `after` option.
Also `contentBaseIndex` option could be used to provide options to `serve-index`...

evilebottnawi:
No new option `webpack-dev-server`, repo in maintenance mode (only bug fixes), please use https://github.com/webpack-contrib/webpack-serve, in near future documentation will be updated

PR #1434, opened 2018-06-28
Update sockjs-client to 1.1.5

franjohn21:
- [x] This is a **bugfix**
- [ ] This is a **typo fix**

N/A - the bug is not in webpack-dev-server but a dependency.

### Motivation / Use-Case
If the response to `sockjs-client` times out, there is a bug in version 1.1.4 that will start an infinite reconnection. We are able to reproduce this timeout and infinite reconnect scenario in webpack-dev-server particularly when we have gzipping enabled and are bundling a really large JS file.

You can read more about the issue here: https://github.com/sockjs/sockjs-client/issues/334#issuecomment-280866701

And upgrading to v1.1.5 will fix it as per this PR: https://github.com/sockjs/sockjs-client/pull/358

### Breaking Changes
N/A

### Additional Info
N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,7 +30,7 @@
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.5.tgz",
       "integrity": "sha1-63d99gEXI6OxTopywIBcjoZ0a9I=",
       "requires": {
-        "mime-types": "2.1.18",
+        "mime-types": "~2.1.18",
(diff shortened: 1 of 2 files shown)


gitors:
My project has encountered the same situation. What should I do about it

franjohn21:
@SpaceK33z @evilebottnawi -- would it be possible to take a look at this PR?

SpaceK33z:
Thanks!

Issue #1449, opened 2018-07-30
Crash on node 10.6.0 and up


Ky6uk:
* Operating System: macOS 10.13.6
* Node Version: 10.6.0, 10.7.0
* NPM Version: 6.1.0
* webpack Version: 4.16.3
* webpack-dev-server Version: 3.1.5

- [x] This is a **bug**
- [x] This is a **modification** request

Webpack crashes sometimes in node `10.6.0` and up. Unfortunately I can't localize that problem, but I think this is relating to https://github.com/nodejs/node/issues/21665.

blackshadev:
@evilebottnawi  this was one of the issues that moved me to webpack-serve since it completely broke webpack-dev-server with https  for me.

evilebottnawi:
Problem still exists?

Ky6uk:
@evilebottnawi I am not sure, because I am using node `v10.5.0` because of the bug. I will try to use the latest node from now. Let's see.

blackshadev:
Good to hear. I will upgrade my setup next week and report back with the
results.

blackshadev:
@evilebottnawi Seems to work just fine. Thanks!

jrop:
I am getting this in an app created with `create-react-app` under node 10.9.0

blackshadev:
@jrop Which version of webpack-dev-server are you running. For me this is absolutely fixed in webpack-dev-server version 3.1.7

jrop:
@blackshadev 

macOS 10.13.6
webpack-dev-server 2.9.4

I created the app using `create-react-app --scripts-version react-scripts-ts`.   As others have noted, rebooting the machine causes the issue to go away for a limited time.  But then it comes right back.  I have switched to using Node 8.11.4 and the issue has ceased to occur.

evilebottnawi:
@jrop you should update `webpack-dev-server`

jrop:
@evilebottnawi I believe the responsible party to upgrading `webpack-dev-server` is the [team maintaining `react-scripts-ts`](https://github.com/wmonk/create-react-app-typescript/blob/master/package.json#L52), which is why I have resorted to using an older version of Node instead

WellspringCS:
So... five hours later, and I've not seen the error yet. I still have the occasional "The expression evaluated to a falsy value" crash, but moving back to 10.5... I count it as a huge win.

For me, at least, webpack-dev-server 3.1.7 proved to be a false hope.

PR #1451, opened 2018-08-01
fix(Server): don't use `spdy` on `node >= v10.0.0` 

addaleax:
- [x] This is a **bugfix**
- [ ] This is a **typo fix**

This does add Node 10 to the test matrix. There do not seem to be any existing tests for `https://` support, though, so I added one, even though it does not function as a regression test (one would actually have to explicitly use spdy/http2 for that).

### Motivation / Use-Case

This fixes the bug reported at:

- https://github.com/nodejs/node/issues/21665
- https://github.com/webpack/webpack-dev-server/issues/1449

### Breaking Changes

This should not affect any user agents that detect protocol support themselves (like they should), in particular not browsers.

### Additional Info

Another option that was considered was reverting the change that led to this in Node 10.

However, the `spdy` module is built in a very fragile manner, and unfortunately it seems unreasonable to expect that anything we can do inside of Node would be more than a stop-gap solution.

(See e.g. https://github.com/spdy-http2/node-spdy/issues/333 for an attempt to reach out to the `spdy` maintainers about this problematic.)


addaleax:
Updated the test with an increased timeout because of the on-the-fly key generation.

Kovensky:
Unfortunately, just changing the module to http2 won't work because of https://github.com/expressjs/express/issues/3388

Express doesn't seem to actually use node's request/response stuff but uses custom ones which are then `setPrototypeOf` to the node versions, and because of that it actually will not work correctly and crashes as soon as an h2 request happens.

addaleax:
@Kovensky Without having looked into it deeper, from https://github.com/expressjs/express/pull/3390#issuecomment-408646774 and below, it looks like there might be some movement in that area? :)

addaleax:
Btw, are the recommendations for trying out pull requests in the CONTRIBUTING.md accurate? See e.g. https://github.com/nodejs/node/issues/21665#issuecomment-410003250

leup:
Any news on this one ?

michael-ciniawsky:
I think this is definitely blocked by https://github.com/nodejs/node/issues/21665 and https://github.com/expressjs/express/issues/3388 atm

addaleax:
> I think this is definitely blocked by nodejs/node#21665 and expressjs/express#3388 atm

@michael-ciniawsky I’m not sure if I’m understanding you correctly, but *this PR* is not blocked on either of these issues. It fixes the first one, and the second one is just something to keep in mind for the future (fully migrating from SPDY to the built-in HTTP/2 module in Node would be dependent on it being solved).

What this PR *may* be blocked on is this issue, reported by somebody who tried this code out: https://github.com/nodejs/node/issues/21665#issuecomment-410208986

I don’t know enough about webpack-dev-server to figure out whether that issue is actually related, and if so, in what way. (And as @Kovensky said, automated testing is hard here & I’m not sure how to do it well.)

michael-ciniawsky:
Should we better remove the dependency on `spdy` entirely ? 🙃 It doesn't seemed to be promoted for `http/2` explicitly nor is there an option to enable `http2`, so we may get away with that :D

michael-ciniawsky:
Released in [`v3.1.7`](https://github.com/webpack/webpack-dev-server/releases/tag/v3.1.7) 🎉

avocadee:
Can we cherry-pick this fix to the latest 2.11.5 and bump up the version to 2.11.6?
Webpack 3 works with 2.x.x, not 3.x.x.

avocadee:
cherry-picked https://github.com/webpack/webpack-dev-server/pull/2316 for webpack-dev-server 2.x.x for webpack3 users using nodejs 10

PR #1464, opened 2018-08-24
refactor(package): migrate to `schema-utils` (`dependencies`)

michael-ciniawsky:
### Type

- [x] Refactor

### Issues

- None

### SemVer

- [x] Patch (?)


lib/options.json
```diff
@@ -157,27 +137,21 @@
...
     "disableHostCheck": {
-      "description": "Disable the Host header check (Security).",
       "type": "boolean"
     },
     "public": {
-      "description": "The public hostname/ip address of the server.",
       "type": "string"
     },
     "https": {
-      "description": "Enable HTTPS for server.",
```
evilebottnawi: Don't remove description, In the future, we can derive more friendly errors based on them

PR #1469, opened 2018-08-25
refactor(examples): remove from repository

michael-ciniawsky:
I'm not willing to maintain them here in the repository long-term, I kept them and need to figure out a sensible migration path to remove them before any further actions are taken, so marked as blocked for now 

### Type

- [x] Refactor

### Issues

- None

### SemVer

- [x] Patch


evilebottnawi:
@michael-ciniawsky May be `Example` section in `READNA.md` is enough for us?

SpaceK33z:
I initially added all the examples way back when there was only one example in the repository, for three reasons:

- Showing what is possible with WDS
- Helping people with complete examples to give them something to copy/paste
- Making maintaining WDS easier because many use-cases could be tested by just going to an example

If the tests were very good (they are not, partially my fault), I'd agree to remove maybe.

evilebottnawi:
Let's do this in major release, some developers use examples

PR #1478, opened 2018-08-29
style(utils/addEntries): cleaner variable naming

michael-ciniawsky:
### Type

- [x] Style

### Issues 

- None

### SemVer

- [x] Patch


lib/utils/addEntries.js
```diff
@@ -1,51 +1,53 @@
...
       address() {
-        return { port: devServerOptions.port };
+        return { port: options.port };
       }
     };
 
-    const domain = createDomain(devServerOptions, app);
-    const devClient = [`${require.resolve('../../client/')}?${domain}`];
+    const domain = createDomain(options, app);
+    const entries = [ `${require.resolve('../../client/')}?${domain}` ];
```
evilebottnawi: Looks your have ` ` here, maybe better run prettier?

michael-ciniawsky: The space is intenentional `[ a, b ]` here, prettier does hopefully leave it :)

evilebottnawi: @michael-ciniawsky hm, we should integrate prettier here :confused:

michael-ciniawsky: Yep, prettier will be included with `webpack-defaults` in `webpack-dev-server >= v4.0.0`, it should still leave whitespace in this case since it is more readable

evilebottnawi: @michael-ciniawsky :+1:

evilebottnawi:
@michael-ciniawsky looks appveyor fail maybe need rebase?

michael-ciniawsky:
Yep :)

PR #1496, opened 2018-09-09
chore(package): add `webpack-cli` as peer dependency (`peerDependencies`)

meteorlxy:
As `webpack-dev-server` requires `webpack-cli` indeed:

https://github.com/webpack/webpack-dev-server/blob/3d37cc5a4a4fc06f0649db8d382ef8484ff73131/bin/webpack-dev-server.js#L84

I couldn't determine which semver of `webpack-cli` is required, and I just copy `^3.1.0` from `devDependencies` for now.

### Type

- [x] Chore

### Issues

- None

### SemVer

- [x] Patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,8 @@
     "ws": "^6.0.0"
   },
   "peerDependencies": {
-    "webpack": "^4.0.0"
+    "webpack": "^4.0.0",
+    "webpack-cli": "^3.1.0"
   },
   "author": "Tobias Koppers @sokra",
   "bugs": "https://github.com/webpack/webpack-dev-server/issues",


SpaceK33z:
I don't understand this PR, there is code specifically for giving you an error when you don't have webpack-cli installed, [see here](https://github.com/webpack/webpack-dev-server/blob/d0725c98fb752d8c0b1e8c9067e526e22b5f5134/bin/webpack-dev-server.js#L69-L78).

meteorlxy:
@michael-ciniawsky Thanks for your explaination 😄 . In fact, `webpack` has no `peerDependency` on `webpack-cli` at all, that's why I got into the error.

@SpaceK33z I notice the code, but that code won't exit the process at once. So the error message will be flushed by the following webpack related logs soon, and may not be noticed by users (I didn't notice the code in my console before I scroll back to the very beginning 😅 ).

SpaceK33z:
The point of separating the CLI in a `webpack-cli` package is mostly to make the original `webpack` package smaller. In dev-server the `webpack-cli` package is only required if you use the `webpack-dev-server` CLI command. If you don't, you'll never need it.

Anyway, I think the real bug here is that the code I mentioned doesn't exit the process fully.

PR #1499, opened 2018-09-10
chore(package): update `webpack-dev-middleware` v3.2.0...3.3.0 (`dependencies`)

michael-ciniawsky:
### Type

- [x] Chore

### Issues

- None

### SemVer

- [x] Patch

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12016,14 +12016,13 @@
       }
     },
     "webpack-dev-middleware": {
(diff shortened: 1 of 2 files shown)


evilebottnawi:
@michael-ciniawsky looks we don't support https://github.com/webpack/webpack-dev-middleware#writetodisk, maybe we can do this here?

PR #1513, opened 2018-09-28
chore: host defaults to 127.0.0.1

Aladdin-ADD:
- [ ] This is a **bugfix**
- [ ] This is a **typo fix**

### DIFF
--- a/bin/options.js
+++ b/bin/options.js
@@ -150,7 +150,7 @@ const options = {
   },
   host: {
     type: 'string',
-    default: 'localhost',
+    default: '127.0.0.1',
     describe: 'The hostname/ip address the server will bind to',
     group: CONNECTION_GROUP
   },


evilebottnawi:
@Aladdin-ADD lookc good, why we should use `127.0.0.1` instead `localhost`?

Aladdin-ADD:
In most cases, there is no difference, but `localhost` can be changed via `hosts`(I encountered it unexpectedly). 😄

LongLiveCHIEF:
localhost goes through loopback resolution, whereas `127.0.0.1` does not. `lib/Server.js` always allows IP based hostnames, so with `localhost` as the default, then using `127.0.0.1` in your browser will _always_ work, however the inverse is _not_ true.

In cases where localhost doesn't include `127.0.0.1`, then you can use the `host: ` config option to set wds to listen to the correct network interface url so that `localhost` can still be used.

Changing this could have impacts to certificates, and there are built-in options provided to support the problem case described by OP.

michael-ciniawsky:
> localhost goes through loopback resolution, whereas 127.0.0.1 does not. lib/Server.js always allows IP based hostnames, so with localhost as the default, then using 127.0.0.1 in your browser will always work, however the inverse is not true.

I'm closing this PR for the time being since as long as this actually doesn't fix a bug, there is no need to change the default here

PR #1531, opened 2018-10-17
fix(Server): set `tls.DEFAULT_ECDH_CURVE` to `'auto'`

nekolab:
- [x] This is a **bugfix**
- [ ] This is a **typo fix**

The default value of tls.DEFAULT_ECDH_CURVE is 'prime256v1',
it breaks the connection when certificate is not compatible
with the default curve since node^8.6.0.

To fix this issue, we need set it to 'auto', makes OpenSSL
select the curve automatically.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -13,6 +13,7 @@ const fs = require('fs');
 const path = require('path');
 
 const ip = require('ip');
+const tls = require('tls');
 const url = require('url');
 const http = require('http');
 const https = require('https');
@@ -39,6 +40,16 @@ const createCertificate = require('./utils/createCertificate');
 
 const validateOptions = require('schema-utils');
 const schema = require('./options.json');
+
+// Workaround for node ^8.6.0, ^9.0.0
+// DEFAULT_ECDH_CURVE is default to prime256v1 in these version
+// breaking connection when certificate is not signed with prime256v1
+// change it to auto allows OpenSSL to select the curve automatically
+// See https://github.com/nodejs/node/issues/16196 for more infomation
+const version = parseFloat(process.version.slice(1));
+if (version >= 8.6 && version < 10) {
+  tls.DEFAULT_ECDH_CURVE = 'auto';
+}
 
 const STATS = {
   all: false,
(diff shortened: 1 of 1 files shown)


evilebottnawi:
@nekolab on `node@10` all works fine, right?

nekolab:
@evilebottnawi 
Yes, node@10 default `tls. DEFAULT_ECDH_CURVE` to `auto`
Seems this change breaks node@6, so I've added some version check

evilebottnawi:
@nekolab thanks!

lib/Server.js
```diff
@@ -40,6 +41,16 @@ const createCertificate = require('./utils/createCertificate');
 const validateOptions = require('schema-utils');
 const schema = require('./options.json');
 
+// Workaround for node ^8.6.0, ^9.0.0
+// DEFAULT_ECDH_CURVE is default to prime256v1 in these version
+// breaking connection when certificate is not signed with prime256v1
+// change it to auto allows OpenSSL to select the curve automatically
+// See https://github.com/nodejs/node/issues/16196 for more infomation
+const verMatch = process.version.match(/^v(\d+).(\d+)/);
```
michael-ciniawsky: Is the version check required here or could this be defaulted to `'auto'` in general? E.g was this being 'reverted' for `node >= 10.0.0` for some reason? If so, why?

```suggestion
const version = process.version.match(/^v(\d+).(\d+)/);
```

michael-ciniawsky: `node` v9.0.0 is EOL

michael-ciniawsky: Just saw https://github.com/webpack/webpack-dev-server/pull/1531#issuecomment-430943980 Hmm....

evilebottnawi: @michael-ciniawsky `node@8` also affected

michael-ciniawsky:
```js
const version = process.version

if (parseFloat(version.slice(1)) >= 8.6) {
  ...
}
```

evilebottnawi: @michael-ciniawsky `node@10` is not affected

nekolab: Thanks for guiding, I've update the code and please take a look again.

There is another place using regex to match the version, I've changed it too

michael-ciniawsky:
Released in [`v3.1.10`](https://github.com/webpack/webpack-dev-server/releases/tag/v3.1.10) 🎉

nekolab:
Thanks for merging!

Can we backport this patch to v2 branch?

michael-ciniawsky:
> I suggest just compare version >= 8 && version < 10, I believe it will not take any side effect because for version from 8 to 8.6, the tls.DEFAULT_ECDH_CURVE is actually not take effect due to node-side bug.

Yes, please

> Can we backport this patch to v2 branch?

Yes

PR #1538, opened 2018-10-23
ci(travis): add windows

michael-ciniawsky:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,37 +2,70 @@ language: node_js
 
 jobs:
   fast_finish: true
+
+  allow_failures:
+    - os: windows
+      node_js: 6
+
   include:
     - stage: Lint
       env: SCRIPT=lint
-      node_js: 'stable'
-    - &linux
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,37 +2,54 @@ language: node_js
...
       env: SCRIPT=lint
       node_js: 'stable'
+
+    - &osx
+      stage: Test
+      os: 'osx'
+      env: SCRIPT=test
+      node_js: 'stable'
+    - <<: *osx
+      node_js: 'lts/*'
```
evilebottnawi: Better use `8`, i think we should use `lts/NAME_OF_RELEASE` or number everywhere than mix they. Also think we should add `node@10` and latest night build to avoid regressions and broken package (for `linux` and `windows`)

evilebottnawi:
Ci on travis is still very unstable, we can revisit this in future, right now i am closing this PR, anyway thanks for work

PR #1553, opened 2018-10-29
feat: add `sockPath` option (`options.sockPath`)

trescenzi:
- [ ] This is a **bugfix**
- [x] This is a **feature**

This is a version of #1289 that's up to date with master.

The tests that were added in #1289 seem to exist in master now. I can add more specific tests if desired.

### Motivation / Use-Case

*copied from #1289; although I have the same motivation*
Similar motivations as #911.

We also have multiple apps being developed behind a reverse proxy (in our case a docker composed nginx). Each app is served from a top-level sub-directory (can't use root path), e.g. host/app-part-one, host/app-another-piece

### Breaking Changes

None

### Additional Info
There was discussion on #1289 about adding `port` and `host` to the socks options. I'm happy to do so if there's still interest. I didn't include it here though because this happens to be all I need.


examples/node-api-sock-path/README.md
```diff
@@ -0,0 +1,15 @@
+# Node.js API - Simple
```
michael-ciniawsky: Please remove all `examples` as we are don't intend to maintain them in the future (#1469)

client-src/default/index.js
```diff
@@ -2,7 +2,7 @@
 
 /* global __resourceQuery WorkerGlobalScope self */
 /* eslint prefer-destructuring: off */
-
+const querystring = require('querystring');
```
michael-ciniawsky:
```suggestion
const qs = require('querystring');
```

client-src/default/index.js
```diff
@@ -196,7 +196,7 @@ const socketUrl = url.format({
   auth: urlParts.auth,
   hostname,
   port: urlParts.port,
-  pathname: urlParts.path == null || urlParts.path === '/' ? '/sockjs-node' : urlParts.path
+  pathname: urlParts.path == null || urlParts.path === '/' ? '/sockjs-node' : (querystring.parse(urlParts.path).sockPath || urlParts.path)
```
michael-ciniawsky:
```suggestion
  pathname: urlParts.path == null || urlParts.path === '/' ? '/sockjs-node' : (qs.parse(urlParts.path).sockPath || urlParts.path)
```

lib/options.json
```diff
@@ -321,6 +324,7 @@
       "filename": "should be {String|RegExp|Function} (https://webpack.js.org/configuration/dev-server/#devserver-filename-)",
       "port": "should be {String|Number} (https://webpack.js.org/configuration/dev-server/#devserver-port)",
       "socket": "should be {String} (https://webpack.js.org/configuration/dev-server/#devserver-socket)",
+      "sockPath": "should be {String} (https://webpack.js.org/configuration/dev-server/#devserver-sockPath)",
```
michael-ciniawsky: This needs to be documented by opening a PR in [docs repo](https://github.com/webpack/webpack.js.org) please. Currently to URL redirects to the dev server 'landing page', which might be confusing for some and ultimately isn't really helpful

client-src/default/index.js
evilebottnawi: Disagree, no need rewrite full package name to short name.

trescenzi: `qs` also conflicts with another variable in the package. That's why it's been renamed from the original PR.

evilebottnawi: @trescenzi you did everything right, do not rename

trescenzi:
I've:

- removed the example
- added tests
- fixed a bug noted while testing

Still need to update the webpack documentation itself with another PR.

Let me know if there's more to do here.

Also do you want these commits squashed or are they ok as is?

trescenzi:
Huh so I was running the tests with a `.only` looks like they fail when run with all tests. I'll investigate that.

evilebottnawi:
@trescenzi feel free to feedback

trescenzi:
So my attempt to use `supertest` to actually hit the socket was flawed. So I removed those particular tests. I'm not sure if there's a good way to actually test that the socket is setup correctly as `supertest` doesn't know anything about the socket and can't mock it. Still testing that we're setting the path to what the option is though.

evilebottnawi:
@trescenzi we need search solution to test this, test value of option is not good test

evilebottnawi:
@trescenzi Also looks `travis` still fail

trescenzi:
@evilebottnawi yea I'll put more time into this tonight. The travis test that's failing is the CLI one. Looks like it times out occasionally?

evilebottnawi:
@trescenzi hm, looks not related to this PR, if you have time please look what is wrong, feel free to feedback

trescenzi:
Updated with better tests.

If I get a chance this evening I'll look into the cli tests.

trescenzi:
The test issues are only in `node 11` it seems. I'm still investigating but there isn't anything obvious in the node 11 release notes that would indicate that child process would behave differently.

evilebottnawi:
@trescenzi can you reproducible test failed locally?

trescenzi:
@evilebottnawi I can reproduce the hanging CLI tests locally. I'm busy over the next few days but I can take a look again on monday.

evilebottnawi:
@trescenzi can you provide details how you can reproduce problem?

trescenzi:
@evilebottnawi all I have to do, on master, is:

```
> brew switch node 11.0.0
> rm -rf node_modules
> npm install
> npm test
```

Sometimes it hangs during the tests, sometimes it segfaults. I don't have issues with node 10.12.0.

evilebottnawi:
@trescenzi thanks!

trescenzi:
My apologies I've been away from this for so long. I've got time over the next few days to figure out what's up and get this out.

trescenzi:
When I was trying to get the tests working before I made so updates to the tests for this change that used supertest more idiomatically instead of relying upon `request`. I've pushed these changes. Should just have to click the update branch button and this will be good to go. Waiting to click though in case it gets out of date again between now and when it's good to merge.

@odinho definitely interested in debugging node. It's been too long since I've attached gdb to a running process haha. Not sure if I have time this moment to do so though. I'll see when I end up today. The issues are still 100% reproducible just by switching to node 11.

trescenzi:
Any updates here? As far as I'm aware it should be good to merge. Let me know if there's anything else I can do.

lib/Server.js
```diff
@@ -85,6 +85,7 @@ function Server (compiler, options = {}, _log) {
 
   this.watchOptions = options.watchOptions || {};
   this.contentBaseWatchers = [];
+  this.sockPath = `/${options.sockPath ? options.sockPath.replace(/^\/|\/$/g, '') : 'sockjs-node'}`;
```
evilebottnawi: Why we need `.replace(/^\/|\/$/g, '')`?

client-src/default/index.js
evilebottnawi: Why we need use `querystring.parse` here?

trescenzi: That I don't know the answer to. It was left over from the initial pr and I kinda just assumed was how stuff was done. Will investigate and come back with an answer.

evilebottnawi: @trescenzi thanks!

zsalzbank:
While waiting for this, I came up with a temporary workaround (assuming you are using webpack). You'll need the file in this gist: https://gist.github.com/zsalzbank/cdee4f852e4a89c116eb7a28491bbf63

Then, add the exported plugin to your babel plugins like so:

```js
...
plugins: [
  require('sockjs-prefix-modification')('/my-prefix-path/').plugin
],
...
```

And before you import `webpack-dev-server` for your server, run the following:

```js
require('sockjs-prefix-modification')('/my-prefix-path/').fixWebpackDevServer();
```

client-src/default/index.js
trescenzi: So what's happening here is that `__resouceQuery` gets the `sockPath` option passed in. This is then parsing that querystring and getting the `sockPath` option.

evilebottnawi: @trescenzi just add this comment to code :+1:

trescenzi:
My apologies for taking forever to update this. It dropped off my radar. It should be good now.

I'm going to put up a PR against the docs site now.

trescenzi:
Looks like the linting hung? It lints fine for me locally.

evilebottnawi:
@trescenzi yep, something wrong with travis ci

Issue #1560, opened 2018-11-05
Setup without warnings (defaults to production mode)


NaNdreas:
* Operating System: Windows 10
* Node Version: 8.9.0
* NPM Version: 6.1.0
* webpack Version: 4.24.0
* webpack-dev-server Version: 3.1.10

- [ ] This is a **bug**
- [x] This is a **modification** request

### Expected Behavior
When following the setup instructions in the readme I want to have `webpack-dev-server` running without any warnings or errors. 

### Actual Behavior
I end up with a warning because `webpack-dev-server` is in production mode. 

`WARNING in configuration
The 'mode' option has not been set, webpack will fallback to 'production' for this value.`

Please note that in readme it says: _"This should be used for **development only**."_

### For Features; What is the motivation and/or use-case for the feature?
I would love one of the following:

- change the readme instructions to include the `--mode=development`. Instead of letting people get a warning because of incorrect setup, and have to figure out a solution themselves (not obvious to everyone that `webpack-dev-server` takes the same flags as `webpack`).

- This was already rejected, but I think it's highly related and deserves to be an option, so please see this as very respectfully and carefully mentioned: Revisit https://github.com/webpack/webpack-dev-server/issues/1327 (Default mode to development?), read the comments, and consider the number of 👍  vs 👎 and see if maybe it's reasonable to reconsider this request. If not, could we please have a slightly more detailed explanation to help us understand the issues? You never know, maybe someone has a different idea that would work for everyone. But to get there, they need to know the problem. 

And of course either way: Thank you for a great plugin, only getting better!

evilebottnawi:
@NaNdreas Feel free to send a PR

NaNdreas:
Thank you @evilebottnawi 

Just to be sure, do you want the documentation fix or setting development by default? 😃

evilebottnawi:
@NaNdreas based on user experience, we need implement `development` by default (don't forget respect `mode` from configuration)

Issue #1327, opened 2018-03-01
Default mode to development?


billyjanitsch:
* Operating System: macOS 10.13.3
* Node Version: 8.9.4
* NPM Version: 5.7.1
* webpack Version: 4.0.1
* webpack-dev-server Version: 3.1.0

- [ ] This is a **bug**
- [x] This is a **modification** request

### Expected Behavior

`webpack-dev-server` should default `mode` to `"development"`, since that will be the developer's intention in the vast majority of cases.

### Actual Behavior

Currently, `mode` does not have a default, which causes webpack to default it to `"production"` and emit a warning.

### For Features; What is the motivation and/or use-case for the feature?

It's understandable for webpack itself to default to production settings if a mode isn't explicitly specified, but since webpack-dev-server is intended only for use in development, it seems reasonable to default to development settings unless the user explicitly overrides them.

evilebottnawi:
@billyjanitsch no, you should set `mode` manually or using `NODE_ENV`, changing `mode` and `env` inside in code is bad practice. Exmaple:
```
cross-env NODE_ENV=development webpack-dev-server
```

astorije:
I'm with @billyjanitsch here, the default should be `development`, and I wouldn't deem bad practice to have sensible defaults.
Overriding the default using `mode`, `--mode`, or `NODE_ENV` should still be expected, but the default shouldn't be production IMO.

paleite:
This would be a really nice enhancement.

`webpack-dev-server` with default mode `production` is unexpected to me.
If there was a `webpack-production-server`, I wouldn't expect it to have default mode `development`, either.

However, if this repo was called `webpack-server`, I could understand the argument of not having a default mode (even though that probably goes against the zero-conf policy).

Jokero:
@evilebottnawi Are you kidding?) It's absolutely illogical to have `mode=production` by default in **dev-server**. Original message now has 48 upvotes and 0 downvotes, it's for a reason.

PR #1562, opened 2018-11-07
set default mode to development

NaNdreas:
- [x] This is a **bugfix**
- [ ] This is a **feature**

If this needs any tests I would appreciate some help what type of test, and where to add it

### Motivation / Use-Case
Set default `mode` to `development`. webpack config and CLI option will override the default if provided. 
See #1560 and #1327 

### Additional Info
As far as I could tell the devServer options (just below the line changed in this PR) are only picked up from the first array item if the webpack config is an array(?). So I used the same principle for detecting `mode` in the config.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -113,6 +113,9 @@ function processOptions (config) {
     ? config[0]
     : config;
 
+  // This updates both config and firstWpOpt
+  firstWpOpt.mode = defaultTo(firstWpOpt.mode, 'development');
+
   const options = config.devServer || firstWpOpt.devServer || {};
 
   if (argv.bonjour) {


NaNdreas:
Hi @ooflorent 

Personally I think what Webpack's warning is really saying (the reason behind it) is "_Hey, I don't know if you want production or development, please specify or the result may be really wrong_". That makes sense in Webpack's case, because from running `webpack` there's no way to tell if the user wants production or development. 

But in this case, from `webpack-dev-server`, we do know that the user wants (or should have) development, so I don't see the reason to ask for it. 

Webpack will still warn when the user wants to build a production build with the `webpack` command

I could definitely be missing something though, maybe you're thinking of a specific scenario?

evilebottnawi:
@NaNdreas 
> EDIT: to be clear it's not my decision in the end, I have no power here hehe, I just added my view on it.
I will look at tests meanwhile

We need feedback from other developers

MatTheCat:
For now there are 78 :+1: on https://github.com/webpack/webpack-dev-server/issues/1327

> With this change, the warning vanishes which is not correct in my opinion.

The warning will appear when `webpack` is used.

I heard a lot about “zero configuration” and “sensible defaults”. Is it not sensible for webpack-**dev**-server to imply a **dev** enviroment?

odinho:
IMHO this sounds like a sane default (so one extra random developer there).

@NaNdreas: If you rebase, Travis should re-run and hopefully not fail.

evilebottnawi:
@NaNdreas we need add tests and we can merge this

NaNdreas:
Hi, 
I will not have access to a laptop in december, if anyone else feel like taking over this PR, feel free. If nobody worked on it when I am back around new years, I can look at it.
Thanks

NaNdreas:
Hi,
I've started looking at the tests and how to effectively test this change, but I'm lost when looking at other tests and don't have the time to deep dive into it. I will have to leave this one as it is now sorry. 

Feel free to take this PR anyone who wants to write the tests. 

Thanks

evilebottnawi:
Done in master

Issue #1491, opened 2018-09-06
AssertionError in spdy-transport consistently causing failures while running in Chrome 69


bdpartridge:
* Operating System: macOS High Sierra (10.13.6)
* Node Version: 8.11.1
* Yarn Version: 1.6.0
* webpack Version: 4.17.1
* webpack-dev-server Version: 3.1.6

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

I'm not including our webpack config for now, but I can include a stripped-down version if it's really needed to reproduce the problem.

### Expected Behavior

*webpack-dev-server* works like it did in Chrome <69, and like it continues to work in other browsers (like Firefox).

### Actual Behavior

After several incremental recompiles, the process errors out due to an AssertionError in the spdy-transport library:

```
assert.js:42
  throw new errors.AssertionError({
  ^

AssertionError [ERR_ASSERTION]: false == true
    at PriorityNode.removeChild (/Users/bpartridge/forward/fwd/gui/node_modules/spdy-transport/lib/spdy-transport/priority.js:72:3)
    at PriorityNode.remove (/Users/bpartridge/forward/fwd/gui/node_modules/spdy-transport/lib/spdy-transport/priority.js:61:15)
    at PriorityTree.add (/Users/bpartridge/forward/fwd/gui/node_modules/spdy-transport/lib/spdy-transport/priority.js:157:23)
    at Stream._initPriority (/Users/bpartridge/forward/fwd/gui/node_modules/spdy-transport/lib/spdy-transport/stream.js:101:25)
    at new Stream (/Users/bpartridge/forward/fwd/gui/node_modules/spdy-transport/lib/spdy-transport/stream.js:75:8)
    at Connection._createStream (/Users/bpartridge/forward/fwd/gui/node_modules/spdy-transport/lib/spdy-transport/connection.js:391:16)
    at Connection._handleHeaders (/Users/bpartridge/forward/fwd/gui/node_modules/spdy-transport/lib/spdy-transport/connection.js:438:21)
    at Connection._handleFrame (/Users/bpartridge/forward/fwd/gui/node_modules/spdy-transport/lib/spdy-transport/connection.js:321:10)
    at Parser.<anonymous> (/Users/bpartridge/forward/fwd/gui/node_modules/spdy-transport/lib/spdy-transport/connection.js:158:10)
    at emitOne (events.js:116:13)
```

[...]

PR #1563, opened 2018-11-08
fix(package): update `spdy` v3.4.1...4.0.0 (assertion error) (#1491)

odinho:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No. Upstream issue.

### Motivation / Use-Case

- Fixes issue https://github.com/webpack/webpack-dev-server/issues/1491 (AssertionError in spdy-transport consistently causing failures while running in Chrome 69)
- Upstream bug: https://github.com/spdy-http2/spdy-transport/issues/47#issuecomment-436956755

### Breaking Changes

No breaking change. Changelog for upstream:
- https://github.com/spdy-http2/node-spdy/compare/v3.4.7...v4.0.0
- (the ALPN changes are just their own tests)
- most other changes are sub-dependency updates

### Additional Info

No.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3129,9 +3129,9 @@
       }
     },
     "detect-node": {
-      "version": "2.0.3",
-      "resolved": "https://registry.npmjs.org/detect-node/-/detect-node-2.0.3.tgz",
-      "integrity": "sha1-ogM8CcyOFY03dI+951B4Mr1s4Sc="
+      "version": "2.0.4",
+      "resolved": "https://registry.npmjs.org/detect-node/-/detect-node-2.0.4.tgz",
+      "integrity": "sha512-ZIzRpLJrOj7jjP2miAtgqIfmzbxa4ZOr5jJc601zklsfEx9oTzmmj2nVpIPRpNlRTIh8lc1kyViIY7BWSGNmKw=="
     },
     "diff": {
       "version": "3.5.0",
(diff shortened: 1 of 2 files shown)


odinho:
If anything https://github.com/webpack/webpack-dev-server/pull/1543 looks suspicious. It was added to `master` with its Travis check failing. Both commits since then have failed. The commit before was cancelled, so it actually never got to run a Travis build.

evilebottnawi:
Can we fix it here?

odinho:
Maybe remove all caches? https://docs.travis-ci.com/user/common-build-problems/#segmentation-faults-from-the-language-interpreter-ruby-python-php-nodejs-etc

I have tried some runs with what looked like the culprit removed (https://github.com/webpack/webpack-dev-server/pull/1564) , but it still crashes (only stable though). So if you could kick Travis a bit, remove its caches, that might fix it?

evilebottnawi:
No cache setup for this package to avoid same problem

odinho:
I've opened a bug with Node.js, hoping they can figure it out. https://github.com/nodejs/node/issues/24835

An alternative (that I've been thinking of), is to change your test to test Node 10 instead of Node/stable (which now is Node 11.3.0, but this does not seem stable :P).

evilebottnawi:
@odinho very thanks for helping, let's disable test and put comment with link on issue :+1:

odinho:
> @odinho very thanks for helping, let's disable test and put comment with link on issue

I've added https://github.com/webpack/webpack-dev-server/pull/1588

evilebottnawi:
@odinho okay, let's merge this

Issue #1572, opened 2018-11-16
Retry port finding


u9520107:
* Operating System: Windows 10
* Node Version: 10.12.0
* NPM Version: 6.4.1 (yarn 1.12.1)
* webpack Version: 4.21.0
* webpack-dev-server Version: 3.1.9

- [ ] This is a **bug**
- [x] This is a **modification** request

### Code

```js
  // webpack.config.js
  // Not really config related issue
```

### Expected Behavior
When I start 2 instances of dev servers without specifying port, both instances can figure out available ports and start up correctly.

### Actual Behavior
During the findPort phase, both instances will find the same port that is not being used, then both instances will try to use the same port to start the server, resulting in one of them inevitably errors out.

### For Features; What is the motivation and/or use-case for the feature?
I am not sure how often people try to fire up multiple instances like this, so I'd consider this a feature request. And I think for a workbench like monorepo setup to quickly prototype ideas, this feature can be helpful.
It is possible add a retry port finding procedure when starting servers, so that when this kind of race condition occurs, the instance that lost the race would find the next available port and try again?
I know is is potentially bad if there is no limit to the retry mechanism, perhaps a new parameter for specifying the allowed port ranges would be helpful?

evilebottnawi:
@u9520107 Feel free to send a PR

PR #1577, opened 2018-11-22
feat: Implement random port retry logic

u9520107:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Port finding itself doesn't seem to have any test written already. If someone can suggest how to structure the test for this scenario, I can try and set something up.

### Motivation / Use-Case

Trying to support #1572.
The main change is to delay port finding to just before trying to run server.listen. Just this change alone reduced the time window between determining the available port and actually listening to it, as previously the compiler is initiated between this time frame. This naturally let most instances attempt to get available port in different moments and reducing the risk of trying to use the same port.
Another change is to listen to error event when port finding is used, when 'EADDRINUSE' error is encountered, retry the port finding again for up to 10 times. Not the most elegant solution, but should be sufficient for most prototyping scenarios.
If retried for 10 times and still cannot find available pot to use, the dev-server will error out like before.

### Breaking Changes
Unless there are people relying on the fact that attempting to start multiple instances without specifying port could result in errors, this should not have breaking changes.

### Additional Info
This is my first contribution here, please comment if there are things that I should add.


u9520107:
Can't tell what caused the segment error on that particular instance. Ran the tests on my mac as well. Any advice?

bin/webpack-dev-server.js
```diff
@@ -418,7 +407,42 @@ function startDevServer(config, options) {
...
+      }
+      throw err;
+    });
+    tryStartListening(options, suffix, log);
   }
 }
 
+function tryStartListening(options, suffix, log) {
+  portRetries += 1;
+  portfinder.getPort((err, port) => {
```
evilebottnawi: Need create `utils` what get port, no need move `listen` logic in this utils

evilebottnawi: @u9520107 we need split the logic of searching free port and starting webserver, also you code doesn't solve problem second port also can be busy

u9520107: I'll see what i can do to separate them. But to your second comment, if the second port is busy as well, it will throw an error with code==='EADDRINUSE' again, then trigger another get port. This modification will allow an instance to try at most 10 times before giving up.

evilebottnawi: @u9520107 let's use `process.env.DEFAULT_PORT_RETRY` to allow people setup this, also let's do it by default `3`

u9520107: OK!

bin/utils.js
```diff
@@ -8,6 +8,7 @@
   space-before-function-paren
 */
 const open = require('opn');
+const portfinder = require('portfinder');
```
evilebottnawi: Why not `get-port` https://github.com/sindresorhus/get-port

u9520107: This was already in use before I made any changes though...

u9520107: Hmm... get-port will simply fall back to random port instead of trying to look for the next available port in sequence. Although I am not sure if seeking ports in sequence was ever a requirement. This was the behavior before I added the retries, so I try to maintain that behavior.

u9520107: well, if you're sure, I can switch to get-port. Just need to confirm because portfinder was introduced by whoever introduced port finding in the first place. And normally I try not to change project dependencies for projects that I don't own.

evilebottnawi: @u9520107 hm, If we used it before, I propose to leave it and maybe change it in the next version. Good work.

u9520107: Thanks!

bin/webpack-dev-server.js
```diff
@@ -97,6 +98,10 @@ const config = require('webpack-cli/bin/convert-argv')(yargs, argv, {
 // we should use portfinder.
 const DEFAULT_PORT = 8080;
 
+// Try to find unused port and listen on it for 10 times,
```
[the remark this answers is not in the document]
u9520107: Lol... Proves that comments can be misleading if not properly maintained.

OK, fixed.

u9520107:
force-pushed after rebase to current master.

bin/webpack-dev-server.js
```diff
@@ -97,6 +98,10 @@ const config = require('webpack-cli/bin/convert-argv')(yargs, argv, {
 // we should use portfinder.
 const DEFAULT_PORT = 8080;
 
+// Try to find unused port and listen on it for 3 times,
+// if port is not specified in options.
+const defaultPortRetry = defaultTo(tryParseInt(process.env.DEFAULT_PORT_RETRY), 3);
```
evilebottnawi: Why don't use `parseInt`?

bin/webpack-dev-server.js
```diff
@@ -418,6 +407,27 @@ function startDevServer(config, options) {
...
       status(uri, options, log, argv.color);
     });
+  } else {
+    // only run port finder if no port as been specified
+    findPort(server, DEFAULT_PORT, defaultPortRetry, (err, port) => {
+      if (err) {
+        throw err;
+      }
+      options.port = port;
+      server.listen(options.port, options.host, (err) => {
```
evilebottnawi: Maybe we can move `server.listen` to separate function, looks it is duplicate code

bin/webpack-dev-server.js
u9520107: Mainly because NaN == null is false, so if parseInt returns NaN, defaultTo will not work properly.

bin/webpack-dev-server.js
u9520107: Indeed, I'll look into that over the weekend. Making sure all the variables are passed in for the subtle differences in each server.listen call would be a bit tedious.

u9520107: Nvm... the server.listen uses for sockets and ports are too different. I just extract the 2 identical server.listens for port uses in to a local function to keep variable references. Not sure if that's aligned with your coding style standards though. Please comment if that also needs to be changed.

bin/webpack-dev-server.js
evilebottnawi: @u9520107 can we add comment in code about this?

bin/webpack-dev-server.js
evilebottnawi: @u9520107 hm, let's leave this as is, and refactor this in future

evilebottnawi:
Ping me when PR was ready for review again, thanks!

u9520107:
@evilebottnawi Done, just waiting for CI to finish tests.

evilebottnawi:
Good job, thanks!

bin/utils.js
```diff
@@ -105,10 +106,44 @@ function bonjour (options) {
...
+function tryParseInt(input) {
+  const output = parseInt(input, 10);
+  if (Number.isNaN(output)) {
+    return null;
+  }
+  return output;
+}
+
+function findPort(server, defaultPort, defaultPortRetry, fn) {
+  let tryCount = 0;
```
mistic: @u9520107 Just a simple NIT here, I think the code could be more readable in the following way, but feel free to not apply the changes!

```
function runPortFinder(tryCount, defaultPort, cb) {
  portfinder.basePort = defaultPort;
  tryCount += 1;
  portfinder.getPort((err, port) => {
    cb(err, port);
  });
}

function findPort(server, defaultPort, defaultPortRetry, fn) {
  let tryCount = 0;

  server.listeningApp.on('error', (err) => {
    if (err && err.code !== 'EADDRINUSE') {
      throw err;
    }

    if (tryCount >= defaultPortRetry) {
      fn(err);
      return;
    }

    runPortFinder(tryCount, defaultPort, fn);
  });

  runPortFinder(tryCount, defaultPort, fn);
}
```

evilebottnawi:
Done in https://github.com/webpack/webpack-dev-server/pull/1692

PR #1588, opened 2018-12-04
ci(travis): Node 11 (on OS X) crashes, use 10 for now

odinho:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No, this is about the tests.

### Motivation / Use-Case

Mac OS X (on Travis) has a unknown segfault running the tests. About a month ago the `node/stable` went from Node 10 to Node 11, and then the project started getting these segfaults on Proxy tests as far as I can see.

### Additional Info

This is just a workaround, but the tests failing like now isn't very useful. I've added a bug with Node:
https://github.com/nodejs/node/issues/24835

This change should be reverted once it is fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,9 @@ jobs:
       stage: Test (MacOS)
       os: 'osx'
       env: SCRIPT=test
-      node_js: 'stable'
+      # Node 10 instead of 'stable' since Node 11 segfaults
+      # https://github.com/webpack/webpack-dev-server/pull/1588
+      node_js: 10
     - <<: *osx
       node_js: 'lts/*'
     - <<: *osx


odinho:
If it is successful, I can add a comment with a link to this, so it's easy to remember removing it.

.travis.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       stage: Test (MacOS)
       os: 'osx'
       env: SCRIPT=test
-      node_js: 'stable'
+      node_js: 10
```
evilebottnawi: Let's add comment here on issue :+1:

odinho: Yeah, it was a good idea I saw it a bit late :) But I also don't have the URL before actually creating the PR, so :D Will update once the test-run is done.

odinho:
Very interesting. It hung itself right before the test that segfaults on `11`. So there's something fishy on 10 as well. :/

evilebottnawi:
Looks tests passed

odinho:
Yes, but the first try did fail (before the comment added). So there is a chance that this will be flaky.

It went from 100% fail, to something less than that, hopefully a very low fail-%, but we only have two data points. :)

evilebottnawi:
@odinho hm, maybe random failure, sometimes it is happens, let's merge this PR as is, if problem appear again we investigate deeply

@odinho Just question: do you want join to `webpack` as contributor `webpack-dev-server`? We don't have enough developers and some PRs stalled due i don't have time on all repos (right now i am one active contributor/issue resolver/etc here).

odinho:
> @odinho hm, maybe random failure, sometimes it is happens, let's merge this PR as is, if problem appear again we investigate deeply, big thanks!

Great, let's see :)

> @odinho Just question: do you want join to `webpack` as contributor `webpack-dev-server`? We don't have enough developers and some PRs stalled due i don't have time on all repos (right now i am one active contributor/issue resolver/etc here).

I can't take that on I'm afraid. I'm amazed at all the work you guys are doing, and you being lone maintainer for this one is quite impressive and sad.

I will have a look at some of the other PRs next week and see why they are stalled and if there's some ways I can help without being a member. :)

evilebottnawi:
@odinho thanks, just help with issues and PRs (answer, review, mark as duplicate/invalid and etc). Any help welcome :+1:

PR #1611, opened 2018-12-25
fix: remove onbeforeunload workaround

eqyiel:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

N/A

### Motivation / Use-Case

The code that is removed here added a `beforeunload` event check in order to prevent the reload from happening if the page is already refreshing.

However, it is not always the case that after `beforeunload` page will continue reloading, for two reasons at least:
- `beforeunload` is cancelable, by spec
- custom protocol use case

In my case, there is an API that writes to `window.location` using a custom protocol.  `window.location` changes but the JS context is preserved, leaving `webpack-dev-server` incapacitated.

This PR fixes that (and should also apply cleanly to the v2 branch).

Ref: https://github.com/webpack/webpack-dev-server/pull/841
Ref: https://github.com/webpack/webpack-dev-server/issues/544

### Breaking Changes

N/A

### Additional Info

N/A

### DIFF
--- a/client-src/default/index.js
+++ b/client-src/default/index.js
@@ -201,13 +201,8 @@ const socketUrl = url.format({
 
 socket(socketUrl, onSocketMsg);
 
-let isUnloading = false;
-self.addEventListener('beforeunload', () => {
-  isUnloading = true;
-});
-
 function reloadApp() {
-  if (isUnloading || !hotReload) {
+  if (!hotReload) {
     return;
   }
   if (hot) {


evilebottnawi:
Also please accept CLA

eqyiel:
@evilebottnawi I'm confused about the change from 

> We need search way to write tests on this, good work, but it is breaking change and should be merge to next major release

to

> No need right now, thanks

If there's something that needs improvement, please tell me about it!

PR #1617, opened 2019-01-02
fixes #1616, allows retries to increase at each reconnect

### LINKED ISSUE #1616: [sockjs-client] fix exponential reconnect time
* Operating System: Debian
* Node Version: v11.6.0
* NPM Version: 6.5.0
* webpack Version: 4.28.3
* webpack-dev-server Version: 3.1.14

- [x] This is a **bug**

### Code

You can see the problem here:
https://github.com/webpack/webpack-dev-server/blob/v3.1.14/client-src/default/socket.js

### Expected Behavior

An exponential increase in time between reconnects with the sockjs server.

### Actual Behavior

As you can see in the code I linked, the method `onopen` resets the `retries` variable to 0, so retries is always 0, and `onclose` always tries to reconnect waiting the minimum time. Therefore the 'infinite logspam' solution doesn't work and people experience what is depicted in https://github.com/webpack/webpack-dev-server/issues/1604

Also, here (from https://github.com/webpack/webpack-dev-server/issues/1604 as well): 
50633098-6354a900-0f4a-11e9-95ba-22cbbc569438

### How can we reproduce the behavior?

Make the sockjs server close the connection (for example because the client has a bad origin header), the client will try to reconnect forever without respecting the intended timeout.

It is obvious looking at the code anyway.

### PR

I've already submitted a PR.

carlosgeos:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

fixes #1616 

### Breaking Changes

None

### DIFF
--- a/client-src/default/socket.js
+++ b/client-src/default/socket.js
@@ -2,15 +2,16 @@
 
 const SockJS = require('sockjs-client/dist/sockjs');
 
+// retries variable must increment after each failed reconnection.
+// reconnect time depends on it
+// See https://github.com/webpack/webpack-dev-server/pull/1617
 let retries = 0;
 let sock = null;
 
 const socket = function initSocket(url, handlers) {
   sock = new SockJS(url);
 
-  sock.onopen = function onopen() {
-    retries = 0;
-  };
+  sock.onopen = function onopen() {};
 
   sock.onclose = function onclose() {
     if (retries === 0) {
--- a/test/Socket.test.js
+++ b/test/Socket.test.js
@@ -3,6 +3,7 @@
 const request = require('supertest');
 const config = require('./fixtures/simple-config/webpack.config');
 const helper = require('./helper');
+const socket = require('../client-src/default/socket.js');
 
 describe('socket options', () => {
   let server;
@@ -25,6 +26,23 @@ describe('socket options', () => {
 
     it('responds with a 200', (done) => {
       req.get('/sockjs-node').expect(200, done);
+    });
+  });
+
+  describe('failing behaviour', () => {
+    beforeEach((done) => {
+      server = helper.start(config, {}, done);
+      req = request('http://localhost:8080');
+    });
+
+    it('should increase reconnect time exponentially', () => {
+      // TODO
+      // Try to connect to websocket server, ideally with a different
+      // origin so it fails to connect
+      socket("http://localhost:8080/sockjs-node/123/456/websocket");
+      // Inspect socket (retries or retryInMs) to make sure it waits
+      // an exponential amount of time and retries variable increases
+      // every time
     });
   });
 


carlosgeos:
No idea how to test this sorry

evilebottnawi:
@carlosgeos hm, can you provide example how we can tests this manually?

carlosgeos:
Yes very easy:

1. Have a simple `webpack` project. Install `webpack-dev-server` version 3.1.11 (the bug in this version makes testing easier)
2. Start the dev server and open the console
3. You will see an infinite and continuous amount of `Invalid Host/Origin header` and `[WDS] Disconnected!`
4. Now change: https://github.com/webpack/webpack-dev-server/pull/1617/files#diff-921405d3577e8069eb36a2f7c859c075 and run the dev server again.
5. You will see a maximum of 10 reconnects (not infinite) and spaced out in time.

I also added some logs and did this and this is what I get:
OLD:
screenshot_20190109_111057

NEW:
screenshot_20190109_111022

evilebottnawi:
@carlosgeos Thanks looks good for me! Can we add some comments about this behavior in code for `retries` variable? It will be good for future developers and avoid regressions, thanks

carlosgeos:
Yes I'll do it

hiroppy:
CI fails.

```
/home/travis/build/webpack/webpack-dev-server/test/Socket.test.js
  6:16  error  `../client-src/default/socket.js` import should occur before import of `./fixtures/simple-config/webpack.config`  import/order
✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

evilebottnawi:
@hiroppy can you reproduce problem with above repo?

hiroppy:
@evilebottnawi  Which repo? this one? https://github.com/webpack/webpack-dev-server/pull/1617#issuecomment-452643037

I tested using `examples/cli/hmr` at master branch, but I didn't meet this error.

evilebottnawi:
@hiroppy Yes, i can't reproduce too, we need reproduce problem

hiroppy:
@carlosgeos Could you submit the reproducible repo because we weren't able to meet this error?

carlosgeos:
Steps to reproduce:

1.- Clone https://github.com/carlosgeos/simple-webpack-repo
2.- `npm i` && `npm run start:dev`
3.- Open browser, go to localhost:8080, BUT mimic accessing the socket server externally from another tab/site. I used https://addons.mozilla.org/en-US/firefox/addon/simple-modify-header/ and changed the Origin header to something else. This way you will trigger the functionality of https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/socket.js
4.- You will see an infinite, 1-second spaced log of:

Screenshot_20190531_124630

It should be an exponential increase which was the intended feature.

I agree, this does not happen often, and it is not the normal use case, but it could happen. It's a small change for a small problem.

---

Options set in `simple-modify-header`:
Screenshot_20190531_125620

Previous `webpack-dev-server` versions (3.1.11) did not check the Origin header correctly so the header change was not even necessary.

evilebottnawi:
Thank, agree it is bug, but nut sure about message output, i think we should rewrite:
```
 if (retries === 0) {
      handlers.close();
    } // Try to reconnect.
```
to
```
handlers.close();
```

For better message

hiroppy:
@carlosgeos Could you rebase from master?

PR #1540, opened 2018-10-23
fix: respect `open` and `bonjour` for `node` API

evilebottnawi:
- [x] This is a **bugfix**
- [ ] This is a **feature**
- [x] This is a **code refactor**
- [ ] This is a **test update**
- [ ] This is a **docs update**
- [ ] This is a **metadata update**

### For Bugs and Features; did you add new tests?

fixes #1510 also `bonjour` is broken too using `node` API

### Motivation / Use-Case

fix bug

### Breaking Changes

No

### Additional Info

It is hard tests for `external`, i think we mock `open` function and just check on call. We need rewrite tests, around 1/3 tests are broken locally due `8000/8080` port can be busy other processes.

Some refactoring included.

We can move around 50% code `Server` to utils. I can start with tests and when refactor code.

PR #1627, opened 2019-01-16
[WIP]: add "open" option to node api to allow opening in browser

amilajack:
- [x] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

Node api does not support open option

### Additional Info

### DIFF
--- a/bin/utils.js
+++ b/bin/utils.js
@@ -7,7 +7,6 @@
   array-bracket-spacing,
   space-before-function-paren
 */
-const open = require('opn');
 
 const colors = {
   info (useColor, msg) {
@@ -69,22 +68,6 @@ function status (uri, options, log, useColor) {
     log.info(
       'Broadcasting "http" with subtype of "webpack" via ZeroConf DNS (Bonjour)'
     );
-  }
-
-  if (options.open) {
-    let openOptions = {};
-    let openMessage = 'Unable to open browser';
-
-    if (typeof options.open === 'string') {
-      openOptions = { app: options.open };
-      openMessage += `: ${options.open}`;
-    }
-
-    open(uri + (options.openPage || ''), openOptions).catch(() => {
-      log.warn(
-        `${openMessage}. If you are running in a headless environment, please do not use the --open flag`
-      );
-    });
   }
 }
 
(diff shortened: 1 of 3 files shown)


amilajack:
Does this conflict with https://github.com/webpack/webpack-dev-server/pull/1540/files?

bin/utils.js
```diff
@@ -70,22 +69,6 @@ function status (uri, options, log, useColor) {
...
-      openOptions = { app: options.open };
-      openMessage += `: ${options.open}`;
-    }
-
-    open(uri + (options.openPage || ''), openOptions).catch(() => {
-      log.warn(
-        `${openMessage}. If you are running in a headless environment, please do not use the --open flag`
-      );
-    });
-  }
```
evilebottnawi: Don't move this from utils

lib/Server.js
```diff
@@ -804,6 +825,12 @@ Server.prototype.listen = function (port, hostname, fn) {
     if (fn) {
       fn.call(this.listeningApp, err);
     }
+
+    if (this.options.open) {
+      const suffix = (this.options.inline !== false || this.options.lazy === true ? '/' : '/webpack-dev-server/');
+      const uri = createDomain({ host: hostname, port }, this.listeningApp) + suffix;
+      openBrowser(uri, this.options, this.log);
```
evilebottnawi: It is should be in listen event and remove logic for open from bin file

amilajack:
@evilebottnawi @michael-ciniawsky it seems like https://github.com/webpack/webpack-dev-server/pull/1540 does the same thing as this PR. Should this PR be closed in favor of that PR?

evilebottnawi:
@amilajack PR was abadoned, so you can continue develop

lib/Server.js
amilajack: > remove logic for open from bin file

I'm not sure what do you mean by this. The previous comment says that open logic should be kept in bin

amilajack: Also socksjs doesn't seem to register a 'listen' event. On what object should i listen for and where should i call it?

evilebottnawi: @amilajack listen `express` events https://github.com/webpack/webpack-dev-server/blob/master/bin/webpack-dev-server.js#L408

amilajack: But if I listen with express what port should i listen on? Also i don't understand the reason for listening on another port with express. Can you please explain this.

hiroppy:
friendly ping @amilajack

PR #1633, opened 2019-01-22
fix: allow single object proxy config

### LINKED ISSUE #1438: devServer.proxy config does not accept object as described in options schema
* Operating System: macOS High Sierra 10.13.5
* Node Version: 10.5.0
* NPM Version: 6.1.0
* webpack Version: 4.13.0
* webpack-dev-server Version: 3.1.4

- [X] This is a **bug**

### Code

```js
const path = require('path');

const configuration = {
    entry: path.resolve(__dirname, 'src/source.js'),
    output: {
        filename: 'bundle.js',
        path: path.resolve(__dirname, 'dist'),
    },
    mode: 'development',
    devServer: {
        contentBase: false,
        host: 'localhost',
        proxy: {
            context: () => true,
            target: 'http://localhost:8080',
        },
        port: 8000,
    },
};

module.exports = configuration;
```

### Expected Behavior
Dev server should proxy request http://localhost:8000/test to http://localhost:8080/test.

### Actual Behavior
Dev server does not recognise the proxy setup and returns `Cannot GET /test`

If the proxy config is changed to an array with the current object as a single value it works as intended. The initial config should be valid though following the [options schema](https://github.com/webpack/webpack-dev-server/blob/master/lib/optionsSchema.json#L244).

Minimal repository to reproduce error: https://github.com/petterm/webpack-dev-server-test

lukebro:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes. 

* I added a new test to test the proxy config.
* I renamed a previous test since the name of the test was too generic after adding a new one.

### Motivation / Use-Case

Fix #1438.  Allow for a simple object proxy config as described in the [options schema](https://github.com/webpack/webpack-dev-server/blob/349e7332a18f46250554ccc5fb5272a3550aad53/lib/options.json#L224) and demonstrated in the [docs](https://webpack.js.org/configuration/dev-server/#devserver-proxy).

```js
module.exports = {
  //...
  devServer: {
    index: '', // specify to enable root proxying
    host: '...',
    contentBase: '...',
    proxy: {
      context: () => true,
      target: 'http://localhost:1234'
    }
  }
};
```

### Breaking Changes

N/A

### Additional Info

I'm not sure if just checking for a `target` prop is too naive on the proxy config.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -307,27 +307,31 @@ class Server {
            * }
            */
           if (!Array.isArray(options.proxy)) {
-            options.proxy = Object.keys(options.proxy).map((context) => {
-              let proxyOptions;
-              // For backwards compatibility reasons.
-              const correctedContext = context
-                .replace(/^\*$/, '**')
-                .replace(/\/\*$/, '');
+            if (Object.prototype.hasOwnProperty.call(options.proxy, 'target')) {
+              options.proxy = [options.proxy];
+            } else {
+              options.proxy = Object.keys(options.proxy).map((context) => {
+                let proxyOptions;
+                // For backwards compatibility reasons.
+                const correctedContext = context
+                  .replace(/^\*$/, '**')
+                  .replace(/\/\*$/, '');
 
-              if (typeof options.proxy[context] === 'string') {
-                proxyOptions = {
-                  context: correctedContext,
-                  target: options.proxy[context],
-                };
-              } else {
-                proxyOptions = Object.assign({}, options.proxy[context]);
-                proxyOptions.context = correctedContext;
-              }
+                if (typeof options.proxy[context] === 'string') {
+                  proxyOptions = {
+                    context: correctedContext,
+                    target: options.proxy[context],
+                  };
+                } else {
+                  proxyOptions = Object.assign({}, options.proxy[context]);
+                  proxyOptions.context = correctedContext;
+                }
 
-              proxyOptions.logLevel = proxyOptions.logLevel || 'warn';
+                proxyOptions.logLevel = proxyOptions.logLevel || 'warn';
 
-              return proxyOptions;
-            });
+                return proxyOptions;
+              });
+            }
           }
 
           const getProxyMiddleware = (proxyConfig) => {
--- a/test/Proxy.test.js
+++ b/test/Proxy.test.js
@@ -11,7 +11,7 @@ const shouldSkipTestSuite = require('./shouldSkipTestSuite');
 const WebSocketServer = WebSocket.Server;
 const contentBase = path.join(__dirname, 'fixtures/proxy-config');
 
-const proxyOption = {
+const proxyOptionPathsAsProperties = {
   '/proxy1': {
     target: 'http://localhost:9000',
   },
@@ -28,8 +28,13 @@ const proxyOption = {
   },
 };
 
+const proxyOption = {
+  context: () => true,
+  target: 'http://localhost:9000'
+};
+
 const proxyOptionOfArray = [
-  { context: '/proxy1', target: proxyOption['/proxy1'].target },
+  { context: '/proxy1', target: proxyOption.target },
   function proxy() {
     return {
       context: '/api/proxy2',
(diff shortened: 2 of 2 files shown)


lukebro:
@evilebottnawi I'm not sure what the issue is with CI, it timed out. 

I ran tests on both Node.js `10.15.0` and everything is green.

It doesn't make much sense because the LTS one passed and it's also `10.15.0`.

Can you retrigger the CI?

evilebottnawi:
@lukebro done, somethings wrong with Travis, let's wait rebuild, feel free to ping me after build

lukebro:
@evilebottnawi failed again by timeout. How do you recommend I debug this?

evilebottnawi:
@lukebro do you have mac os machine? We need run this on this system and look what is happening

lukebro:
@evilebottnawi Yeah I'm on macOS 10.14.2 with Node.js v10.6.0.  I've tried with Node.js v10.5.0 also and I can't reproduce the timeout.

I'm going to setup my own Travis CI build to debug, I'll ping you when I find something.

evilebottnawi:
@lukebro thanks for helping

evilebottnawi:
@lukebro looks something wrong with proxy middleware on mac os, because last build is green https://github.com/webpack/webpack-dev-server/pull/1640

lukebro:
Yeah @evilebottnawi it is the proxy middleware.

When the test helpers call close the middleware just hangs and callback is never called:
https://github.com/lukebro/webpack-dev-server/blob/841c585ba0a2754f91b9be99f8326221cea779c5/lib/Server.js#L829-L831

Should I open an issue there and look into it there?

If I remove my changes and create a test with this config:
```
module.exports = {
  //...
  devServer: {
    index: '', // specify to enable root proxying
    host: '...',
    contentBase: '...',
    proxy: [{
      context: () => true,
      target: 'http://localhost:1234'
    }]
  }
};
```

this should have the same effect (from proxy middleware perspective) and fail.

evilebottnawi:
@lukebro hm, don't know without this fix all works fine, maybe something wrong with tests, need investigate how we can reproduce bug without dev server

lukebro:
The good news is I'm able to now consistently reproduce the timeout locally on my machine, still debugging

evilebottnawi:
@lukebro 
> I can also take a look at the ConcurrentCompilationError errors that the current set of tests are throwing.

For some tests it is normal.

Okey, let's comment tests and add todo with link on this issue and i think we can merge

evilebottnawi:
Need rebase too :+1:

lukebro:
@evilebottnawi sorry, better late then never!  Sorry Apex came out and I got really distracted.

Since the jump to Jest it looks like the failing tests on 10, 8, 6 are [now passing](https://travis-ci.org/webpack/webpack-dev-server/builds/493057063) 😊.  The failure for 11 is extremely weird, and might not be related. Can you quickly take a look and possibly rebuild?

evilebottnawi:
@lukebro Looks something wrong with node on macos :confused: Rebuild doesn't solve problem, maybe related https://github.com/webpack/webpack-dev-server/pull/1660, can you try it locally?

hiroppy:
Node11 is often unstable only on CI 🤔

evilebottnawi:
@hiroppy other on macos, and it is very strange, maybe we can investigate which test break CI?

evilebottnawi:
Somebody can test this locally?

hiroppy:
@evilebottnawi 8 and 10 are good, but 11 is not good.
Ok, I'll investigate flaky tests.

lukebro:
@hiroppy let me know if I can help investigate in any way.

@evilebottnawi For me 8, 10 and 11 all pass locally.   Like Yuta mentioned this seems to be a CI problem with 11.

hiroppy:
fyi: https://github.com/nodejs/node/issues/24835

hiroppy:
@evilebottnawi 

The results of the investigation are as follows.

Proxy.test.js has libuv erros.

```
Assertion failed: (handle->flags & UV_HANDLE_CLOSING), function uv__finish_close, file ../deps/uv/src/unix/core.c, line 253.
```

HistoryApiFallback.test.js has segmentation fault.

```
[1]    40029 segmentation fault  npm
```

evilebottnawi:
@hiroppy Great work! Maybe we can skip test what break CI? Like `if (node === 11 && isMacOS) { return expect(true).toBe(true); }`

lukebro:
Waiting for #1667 to be merged.

lukebro:
@evilebottnawi all set 👍

PR #1647, opened 2019-02-02
introduce prettier

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

I would like to introduce `prettier` for development.
And I picked configure file and command from webpack repository.

### Additional Info

Currently, the webpack-dev-server code is far out of webpack's rule and there are many changes in this PR.
So should we decide on a custom rule? /cc @evilebottnawi

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,15 @@
+module.exports = {
+  printWidth: 80,
+  tabWidth: 2,
+  singleQuote: true,
+  trailingComma: 'es5',
+  arrowParens: 'always',
+  overrides: [
+    {
+      files: '*.json',
+      options: {
+        useTabs: false,
+      },
+    },
+  ],
+};
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,19 +1,19 @@
 {
-	"extends": ["webpack", "prettier"],
-	"globals": {
-		"document": true,
-		"window": true
-	},
-	"parserOptions": {
-		"sourceType": "script"
-	},
-	"rules": {
-		"comma-dangle": ["error", "never"],
-		"consistent-return": "off",
-		"no-param-reassign": "off",
-		"no-underscore-dangle": "off",
-		"prefer-destructuring": ["error", { "object": false, "array": false }],
-		"prefer-rest-params": "off",
-		"strict": ["error", "safe"]
-	}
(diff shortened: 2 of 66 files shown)


montogeek:
Code should have same tab/space width, that is current style.

.prettierrc.js
```diff
@@ -0,0 +1,13 @@
...
+	tabWidth: 2,
+	overrides: [
+		{
+			files: "*.json",
+			options: {
+				useTabs: false
+			}
+		}
+	]
+};
```
evilebottnawi: Use this config:
```
{
  "singleQuote": true,
  "trailingComma": "es5",
  "arrowParens": "always"
}
```

hiroppy: Added,PTAL

hiroppy: @evenstensberg I made a forced push to add `html` to prettier.

file: https://github.com/webpack/webpack-dev-server/blob/0185f6397e0edd0285fc2154d7e8fe94e9103f51/.prettierrc.js

evenstensberg: Looks like a regular config, got no opinion here. @evilebottnawi has worked with this more to have a better argument

hiroppy: @evenstensberg  Oops, Sorry, I mistook the name of the destination 🙇

hiroppy:
@evilebottnawi sorry but do you want spaces or tabs? I’m fine either way!

evilebottnawi:
@hiroppy spaces

hiroppy:
@evilebottnawi I changed spaces from tabs:)

client-src/live/live.html
```diff
@@ -1 +1,17 @@
...
+      content="width=device-width, height=device-height, initial-scale=1.0, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0"
+    />
+    <script
+      type="text/javascript"
+      charset="utf-8"
+      src="/__webpack_dev_server__/live.bundle.js"
+    ></script>
+  </head>
+  <body></body>
+</html>
```
evilebottnawi: Let's don't use prettier here

hiroppy:
@evilebottnawi I removed `html` from prettier. PTAL 🙏

PR #1656, opened 2019-02-08
fix:add webpack-cli as peerDependency

tomasztunik:
- [x] This is a **bugfix**
- [ ] This is a **feature**

This does not require any extra tests

### Motivation / Use-Case

Lack of `webpack-cli` in peerDependencies was causing late discovery of missing dependencies only after trying to start the dev server. Having it here will show unmet peer dependency warning when installing the package.

### Breaking Changes

none

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,8 @@
     "ws": "^6.0.0"
   },
   "peerDependencies": {
-    "webpack": "^4.0.0"
+    "webpack": "^4.0.0",
+    "webpack-cli": "^3.2.1"
   },
   "husky": {
     "hooks": {


evilebottnawi:
Some projects use `webpack-dev-server` without cli, so we should avoid peer-deps

tomasztunik:
This was our case, we did not use webpack-cli for anything but when trying to start webpack-dev-server (configured via webpack.config.js) this is what we saw:

```
The CLI moved into a separate package: webpack-cli
Please install 'webpack-cli' in addition to webpack itself to use the CLI
-> When using npm: npm i -D webpack-cli
-> When using yarn: yarn add -D webpack-cli
internal/modules/cjs/loader.js:582
    throw err;
    ^

Error: Cannot find module 'webpack-cli/bin/config-yargs'    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:580:15)
    at Function.Module._load (internal/modules/cjs/loader.js:506:25)
    at Module.require (internal/modules/cjs/loader.js:636:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/Users/tomasztunik/development/impactMapper-Client/node_modules/webpack-dev-server/bin/webpack-dev-server.js:84:1)
    at Module._compile (internal/modules/cjs/loader.js:688:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:699:10)
    at Module.load (internal/modules/cjs/loader.js:598:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:537:12)
    at Function.Module._load (internal/modules/cjs/loader.js:529:3)
```

tomasztunik:
checked the project and the only place that requires `webpack-cli/bin/config-yargs` is webpack-dev-server in the built file https://github.com/webpack/webpack-dev-server/blob/master/bin/webpack-dev-server.js#L80

evilebottnawi:
You need webpack-cli for run dev server right now, it will be solved in future.

tomasztunik:
So shouldn't we keep the warning in peers until it's not needed anymore to not hide this dep? 

Leaving it up to you - apart of this small inconvenience all is operational on our end 💪

evilebottnawi:
@tomasztunik let's avoid message because `webpack-dev-server` use boilerplate like `create-react-app`, `vue-cli` and etc, it will causes a lot of discontent from other developers

evilebottnawi:
We will use webpack-cli `serve`/`server` command for `4` version. So no need this fix right now, it outputs uninstalled peer deps for many application, anyway thanks for PR

croraf:
When is this planned to be solved? It is a bit inconvenient as the author noted.

Issue #1661, opened 2019-02-12
Feature request to extend use of sockjs


EloB:
I would like a simple way to communicate between clients and devServer.

By using the `devServer.before` property. We could do something like:
```js
{
  devServer: {
    before: (app, server) => {
      // If we could add a new event to the server
      server.on(
        'clientSentMessageToWebpackDevServer',
        message => server.sockWrite(server.sockets, 'custom', message)
      );
    }
  }
}
```

In the code I see that we expose message sent.
https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/index.js#L71

If we could add `custom` to https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/index.js#L81

Like this:
```js
var onSocketMsg = {
  // ...
  custom: function custom(data) {
    sendMsg('custom', data);
  }
};
```

# Why add this?
In my case I use multi compiler (multiple configs) because I develop to multiple devices. What I'm trying to achieve is development tools when I run `webpack-dev-server`.

This is how my **webpack.config.babel.js** looks like
```js
export default (env, argv) => [
  ...devices.map(device => ({
    name: device,
    entry: [`./src/${device}`],
    output: {
      path: path.join(__dirname, 'dist', device),
      publicPath: argv.mode === 'production' ? '' : 'http://${webpackIp}:${webpackPort}/${device}',
    },
    plugins: [
      new HtmlWebpackPlugin({ title: `My app on ${device}` }),
    ],
  })),
  ...(argv.mode === 'production' ? [] : [{
    name: 'Development tools',
    entry: './src/devtools',
    output: {
      publicPath: '/',
    },
    plugins: [
      new HtmlWebpackPlugin({ title: 'Development tools' }),
    ],
  }]),
];
```

So when I visit for instance `http://localhost/` when running `webpack-dev-server` I see some devtools. I would imagine `./src/devtools` to look something like this.
```js
window.addEventListener('message', ({ data }) => {
  if (data.type === 'custom') doSomething(data.data);
});

// If the client script were listening for some message event so we could send info back to devServer
const sendMessageToWebpackDevServer = data => postMessage(
  {
    type: `sendMessageToWebpackDevServer`,
    data,
  },
  '*'
);
```

The question you might be thinking of is why not add this as a separate solution? It's because of the hassle of trying to figure out where the webpack-dev-server is located on the network. You guys already do that (see link below). Then we can extend webpack-dev-server with even more cool tools/plugins :)

[...]

PR #1662, opened 2019-02-12
Expose Sockjs server/client

EloB:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Not yet this is just a draft.

### Motivation / Use-Case

Read more https://github.com/webpack/webpack-dev-server/issues/1661 and https://github.com/webpack/webpack-dev-server/pull/1662#issuecomment-463255123.

### Breaking Changes

No

### Additional Info
This is just a draft.

**webpack.config.js**
```js
module.exports = {
  devServer: {
    before(app, server) {
      // Simple logger what clients sends to server
      server.subscribeClientData((data) => console.log(data));
      server.sockWrite(server.sockets, 'custom', 'Send info to client');
    }
  }
};
```

**/some/entry.js**
```js
import send from 'webpack-dev-server/client/send';
// To be implemented
import listen from 'webpack-dev-server/client/listen';

const unsubscribe = listen((message) => console.log(message));

send({ hello: 'world' });
```


EloB:
> Also will bee great if you create minimum reproducible test repo with this feature (PoC), it is allow to think about how better solve this

I will add an working example. Just throw this PR together. So I can get feedback :)

EloB:
@evilebottnawi I've updated the PR now with an example but haven't written any tests. If you think this feature belongs in webpack-dev-server I will add them :)

lib/Server.js
```diff
@@ -667,6 +669,11 @@ class Server {
     next();
   }
 
+  subscribeClientData(handle) {
+    this.app.on(CLIENT_EVENT, handle);
+    return () => this.app.off(CLIENT_EVENT, handle);
```
evilebottnawi: Looks very misleading

lib/Server.js
```diff
@@ -81,6 +81,8 @@ const STATS = {
   errorDetails: false,
 };
 
+const CLIENT_EVENT = 'webpackDevServerClientData';
```
evilebottnawi: `webpack-dev-server-data`

client-src/default/socket.js
```diff
@@ -4,11 +4,21 @@ const SockJS = require('sockjs-client/dist/sockjs');
...
+}
+
+const socket = function initSocket(url, handlers, messages) {
   sock = new SockJS(url);
 
   sock.onopen = function onopen() {
+    messageQueue.forEach(function dequeue(message) {
+      sock.send(message);
+    });
+    messageQueue = [];
```
evilebottnawi: Breaking change for multi compiler compilation

evilebottnawi: Never remove messages, because you can have two/three/four clients, and some clients can have delay

evilebottnawi:
Not sure it is right solution:

```
import send from 'webpack-dev-server/client/send';
// To be implemented
import listen from 'webpack-dev-server/client/listen';
```

We should export this function or use `window.webpackDevServerSend` because now you use not publish api, and files can be moved in other directory in any time.

Maybe we can search other solution for this case? Code really looks very misleading. 

/cc @hiroppy what do you think

client-src/default/socket.js
EloB: How do you mean? If you lose connection and reconnect you probably don't want to resend all messages again?

EloB: I've added the message queue for messages that are sent before the connection is fully established or else you will get an error from sockjs.

lib/Server.js
EloB: Fixed

lib/Server.js
EloB: How do you want it then? :)

client-src/default/socket.js
evilebottnawi: @EloB open two or three browsers and tests how it works, we already do same before and remove this, because it is break situation when you have more then one client (some client can disconnect and connect again)

lib/Server.js
evilebottnawi: Maybe `subscribe`/`unsubscribe`

client-src/default/socket.js
EloB: That `initSocket` looks really weird in the first place because it's not pure and does side effects on `sock`.

lib/Server.js
EloB: This pattern is pretty known though. Rxjs, redux and react hooks for instance uses it. Where you return the unsubscribe/teardown function. So the consumer don't need to store his handle himself.

EloB: Do you like this better?

```js
subscribeClientData() {
  const { app }  = this;
  app.on(CLIENT_EVENT, handle);
  return function unsubscribeClientData() {
    app.off(CLIENT_EVENT, handle);
  };
}
```

client-src/default/socket.js
EloB: Maybe we could rewrite that `initSocket` to be pure instead?

EloB:
> Not sure it is right solution:
> 
> ```
> import send from 'webpack-dev-server/client/send';
> // To be implemented
> import listen from 'webpack-dev-server/client/listen';
> ```
> We should export this function or use `window.webpackDevServerSend` because now you use not publish api, and files can be moved in other directory in any time.
> 
> Maybe we can search other solution for this case? Code really looks very misleading.
> 
> /cc @hiroppy what do you think

The reason why I did this is because you use `__resourceQuery` and if users require `webpack-dev-server/client` then they might include the client twice.

lib/Server.js
EloB: Fixed

EloB:
More feedback? :)

hiroppy:
I like export style.
How about putting(or create) it in `index.js`?
It can resolve `files can be moved in other directory in any time.`

EloB:
> I like export style.
> How about putting(or create) it in `index.js`?
> It can resolve `files can be moved in other directory in any time.`

```
import { send, listen } from 'webpack-dev-server';

EloB:
I'm blocked that the moment and really need some feedback on the questions :)

EloB:
I'm sorry for pushing but I really want this feature. Please give me more feedback so I continue working with this :)

hiroppy:
@EloB  Sorry for late to my reply. I do not think that `lib/Server.js`(main) will be used at the frontend because it uses node.js native modules like `fs`, `path`.
So concerning the rewrite to ES2015 is no problem.

BTW, I also apologize for having suggested without thinking about `lib/Server.js` is using node.js native modules.

I think this solution is good.

https://github.com/webpack/webpack-dev-server/pull/1662#issuecomment-463570652

EloB:
@hiroppy another solution is to add another file called for instance browser.js and add client methods there (`"browser": "./lib/browser.js"` to the `package.json`). I think webpack will resolve that by default for web builds.

EloB:
I been a bit inactive. This is still on my agenda to complete but have a lot to do at the moment.

EloB:
Still on my agenda but no time atm.

evilebottnawi:
Need rebase

evilebottnawi:
Close in favor https://github.com/webpack/webpack-dev-server/issues/1860, sorry for delay, it will be implemented in next major release

EloB:
@evilebottnawi Is there any example how to make something like this with the solution you guys went with?

evilebottnawi:
@EloB sorry, lost the idea, what you want to achieve?

EloB:
@evilebottnawi Be able to use the websocket within webpack dev server to send and receive data. Everything is already well described in this PR comments. There was also example in the PR and images.

evilebottnawi:
With v4 you can do it

EloB:
Can you pinpoint me in a direction?

evilebottnawi:
It is not fully implemented yet, feel free to ping me when we do stable release and I will help you

Issue #1663, opened 2019-02-13
HMR socket URL should infer not only from location.hostname but also location.port


rlamana:
HMR socket fails to connect when the *dev server* is behind some sort of proxy or layered architecture (nginx, k8s ingress, etc.) exposing a different end **port** than the one used in the server's configuration. In these cases, a common configuration is to set *dev server*'s host to `0.0.0.0` to make it externally accessible and rely on browser's **location object** to infer the HMR socket URL.

Dev server exposes and proxies the socket connection accessible at `/sockjs-node/info?t=XX`.

HMR client code builds the socket URL using`location.hostname` but uses the **port** set in server's configuration. Since the browser is at the other end of the architecture, the proper way to infer socket's URL is to also use `location.port` along with the `location.hostname`.

- Operating System: Docker 18.09 in GKE cluster, MacOS X 10.14.2 Mojave
- Node Version: 10.15.0
- NPM Version: 6.4.1 (Yarn Version: 1.10.1)
- webpack Version: 4.29.0
- webpack-dev-server Version: 3.1.14

I'm using a development environment in a Kubernetes cluster. In this scenario, the dev server runs in a docker container configured to use `0.0.0.0` and port `:9000`. It is exposed externally using a Kubernetes ingress controller with the proper dns setup and exposing standard http port `:80`.

This scenario is becoming more and more common when using development environments in the cloud, either with in-house solutions or using tools like Codenvy, Okteto, Koding or online IDEs like Cloud9.

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

```js
// dev server config in webpack.config.js
devServer: {
    port: 9000,
    host: '0.0.0.0',
    disableHostCheck: true,
    hot: true,
    watchOptions: {
      poll: true // To watch inside Docker container.
    }
  }
```

### Expected Behavior

- Externally exposed server: `http://dev.mydomain.com` (port: 80).
- Internally exposed dev server: 0.0.0.0:9000.
  - *index.html* accessible through: `http://dev.mydomain.com/index.html`
  - *HMR socket* generated URL: `http://dev.mydomain.com/sockjs-node/info?t=XXXXXX` 
- Connection **established**.

### Actual Behavior

- Externally exposed server: `http://dev.mydomain.com` (port: 80).
- Internally exposed dev server: 0.0.0.0:9000.
  - *HMR socket* generated URL: `http://dev.mydomain.com`**`:9000`**`/sockjs-node/info?t=XXXXXX` 
[...]

PR #1664, opened 2019-02-13
Use location.port when location.hostname is used to infer HMR socket URL

rlamana:
- [x] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

On a externally exposed dev server (using host 0.0.0.0), HMR client code builds the socket URL using `location.hostname` but uses the port set in server's configuration. In a more complex and layered architecture, since the browser is at the very end of it, the proper way to infer socket's URL (as exposed and proxied by dev server in `.../sockjs-node/info?t=XX`) is to also use `location.port` along with the `location.hostname`.

More and more we see the scenario where the dev server is used in a cloud-based development environment or an architecture with some sort of proxy in front of it.

- With a dev server externally exposed, host `0.0.0.0` and port `9000` and a proxy exposing it in `http://dev.mydomain.com` (default port 80).
  - Expected HMR socket generated URL:  `http://dev.mydomain.com/sockjs-node/info?t=XX` (default port 80).
  - Actual HMR socket generated URL:  `http://dev.mydomain.com:9000/sockjs-node/info?t=XX` (**wrong** port 9000).

More details in issue: https://github.com/webpack/webpack-dev-server/issues/1663

### Breaking Changes

No.
Those who exposed also that wrong port in their proxies as a way around the issue can now remove that configuration.

### DIFF
--- a/test/Client.test.js
+++ b/test/Client.test.js
@@ -13,7 +13,7 @@ function startProxy(port) {
   proxy.use(
     '/',
     httpProxy({
-      target: 'http://localhost:8080',
+      target: 'http://localhost:9001',
       ws: true,
       changeOrigin: true,
     })
@@ -25,7 +25,7 @@ describe('Client code', () => {
   beforeAll((done) => {
     const options = {
       compress: true,
-      port: 8080,
+      port: 9001,
       host: '0.0.0.0',
       disableHostCheck: true,
       hot: true,
--- a/test/Client.test.js
+++ b/test/Client.test.js
@@ -42,10 +42,10 @@ describe('Client code', () => {
   describe('behind a proxy', () => {
     let proxy;
 
-    jest.setTimeout(20000);
+    jest.setTimeout(30000);
 
     beforeAll(() => {
-      proxy = startProxy(9000);
+      proxy = startProxy(9050);
     });
 
     afterAll(() => {
@@ -53,7 +53,7 @@ describe('Client code', () => {
     });
 
     it('responds with a 200', (done) => {
-      const req = request('http://localhost:9000');
+      const req = request('http://localhost:9050');
       req.get('/sockjs-node').expect(200, 'Welcome to SockJS!\n', done);
     });
 
@@ -63,12 +63,12 @@ describe('Client code', () => {
           .waitForRequest((requestObj) => requestObj.url().match(/sockjs-node/))
           .then((requestObj) => {
             expect(requestObj.url()).toMatch(
-              /^http:\/\/localhost:9000\/sockjs-node/
+              /^http:\/\/localhost:9050\/sockjs-node/
             );
             browser.close();
             done();
           });
-        page.goto('http://localhost:9000/main');
+        page.goto('http://localhost:9050/main');
       });
     });
   });
--- a/test/Client.test.js
+++ b/test/Client.test.js
@@ -1,0 +1,72 @@
+'use strict';
+
+const express = require('express');
+const httpProxy = require('http-proxy-middleware');
+const request = require('supertest');
+const addEntries = require('../lib/utils/addEntries');
+const helper = require('./helper');
+const config = require('./fixtures/client-config/webpack.config');
+const runBrowser = require('./helpers/run-browser');
+
+function startProxy(port) {
+  const proxy = express();
+  proxy.use(
+    '/',
+    httpProxy({
+      target: 'http://0.0.0.0:8080',
+      ws: true,
+      changeOrigin: true,
+    })
+  );
+  return proxy.listen(port);
+}
+
+describe('Client code', () => {
+  beforeAll((done) => {
+    const options = {
+      compress: true,
+      port: 8080,
+      host: '0.0.0.0',
+      disableHostCheck: true,
+      hot: true,
+    };
+    addEntries(config, options);
+    helper.start(config, options, done);
+  });
+
+  afterAll(helper.close);
+
+  describe('behind a proxy', () => {
+    let proxy;
+
+    jest.setTimeout(20000);
+
+    beforeAll(() => {
+      proxy = startProxy(9000);
+    });
+
+    afterAll(() => {
+      proxy.close();
+    });
+
+    it('responds with a 200', (done) => {
+      const req = request('http://localhost:9000');
+      req.get('/sockjs-node').expect(200, 'Welcome to SockJS!\n', done);
+    });
+
+    it('requests websocket through the proxy with proper port number', (done) => {
+      runBrowser().then(({ page, browser }) => {
+        page
+          .waitForRequest((requestObj) => requestObj.url().match(/sockjs-node/))
+          .then((requestObj) => {
+            expect(requestObj.url()).toMatch(
+              /^http:\/\/localhost:9000\/sockjs-node/
+            );
+            browser.close();
+            done();
+          });
+        page.goto('http://localhost:9000/main');
+      });
+    });
+  });
+});
--- a/test/Client.test.js
+++ b/test/Client.test.js
@@ -13,7 +13,7 @@ function startProxy(port) {
   proxy.use(
     '/',
     httpProxy({
-      target: 'http://0.0.0.0:8080',
+      target: 'http://localhost:8080',
       ws: true,
       changeOrigin: true,
     })
@@ -45,7 +45,7 @@ describe('Client code', () => {
     jest.setTimeout(30000);
 
     beforeAll(() => {
-      proxy = startProxy(9050);
+      proxy = startProxy(9000);
     });
 
     afterAll(() => {
@@ -53,7 +53,7 @@ describe('Client code', () => {
     });
 
     it('responds with a 200', (done) => {
-      const req = request('http://localhost:9050');
+      const req = request('http://localhost:9000');
       req.get('/sockjs-node').expect(200, 'Welcome to SockJS!\n', done);
     });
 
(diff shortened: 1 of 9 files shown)


rlamana:
@evilebottnawi How are you testing the logic of the client code? I looked for it but couldn't find it in any of the tests. I guess webdriver is the right solution here but I didn't want to add too much overhead to your current tests. Guide me on how you recommend to do it and I'll be glad to add it.

evilebottnawi:
https://github.com/webpack/webpack-dev-server/blob/master/test/Socket.test.js

rlamana:
@evilebottnawi as I understand those tests are checking that the socket and the server are available, but it is not testing the client code that will be executed on the browser's end (more precisely testing the code in https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/index.js
).

Any suggestions on how to test it without adding the overhead of webdriver?

> https://github.com/webpack/webpack-dev-server/blob/master/test/Socket.test.js

rlamana:
@evilebottnawi @hiroppy I added some e2e tests using `puppeteer` headless Chrome API. This allows testing my fix and enables more e2e testing. 

I created a new `Client.test.js` that contains this client code related tests: https://github.com/webpack/webpack-dev-server/blob/50ef125fbfe04eb51f6788a0c44ada539834ea57/test/Client.test.js

Let me know what you think.

evilebottnawi:
Looks good! Ready to merge?

rlamana:
@evilebottnawi it's ready but for some weird reason the test keeps failing in node 6. The proxy I use fails with a EADDRNOTAVAIL error. I've been trying to repro in a container with node 6.16.0, but it works just fine. Any idea why this could be happening?

evilebottnawi:
@rlamana 
```
 console.info node_modules/http-proxy-middleware/lib/logger.js:77
      [HPM] Proxy created: /  ->  http://0.0.0.0:8080
    console.error node_modules/http-proxy-middleware/lib/logger.js:89
      [HPM] Error occurred while trying to proxy request /sockjs-node from localhost:9050 to http://0.0.0.0:8080 (EADDRNOTAVAIL) (https://nodejs.org/api/errors.html#errors_common_system_errors)
    console.error node_modules/http-proxy-middleware/lib/logger.js:89
      [HPM] Error occurred while trying to proxy request /main from localhost:9050 to http://0.0.0.0:8080 (EADDRNOTAVAIL) (https://nodejs.org/api/errors.html#errors_common_system_errors)
  ● Client code › behind a proxy › responds with a 200
    expected 200 "OK", got 500 "Internal Server Error"
      at Test.Object.<anonymous>.Test._assertStatus (node_modules/supertest/lib/test.js:268:12)
      at Test.Object.<anonymous>.Test._assertFunction (node_modules/supertest/lib/test.js:283:11)
      at Test.Object.<anonymous>.Test.assert (node_modules/supertest/lib/test.js:173:18)
      at localAssert (node_modules/supertest/lib/test.js:131:12)
      at node_modules/supertest/lib/test.js:128:5
      at Test.Object.<anonymous>.Request.callback (node_modules/superagent/lib/node/index.js:756:3)
      at IncomingMessage.parser (node_modules/superagent/lib/node/index.js:944:18)
  ● Client code › behind a proxy › requests websocket through the proxy with proper port number
    Timeout - Async callback was not invoked within the 30000ms timeout specified by jest.setTimeout.
      58 |     });
      59 | 
    > 60 |     it('requests websocket through the proxy with proper port number', (done) => {
         |     ^
      61 |       runBrowser().then(({ page, browser }) => {
      62 |         page
      63 |           .waitForRequest((requestObj) => requestObj.url().match(/sockjs-node/))
      at new Spec (node_modules/jest-jasmine2/build/jasmine/Spec.js:92:20)
      at Suite.it (test/Client.test.js:60:5)
      at Suite.describe (test/Client.test.js:42:3)
      at Object.describe (test/Client.test.js:24:1)
```

Maybe something doesn't handle (like errors) or something starts early then tests

rlamana:
@evilebottnawi For what I have seen so far the dev server is having problems to start with the 0.0.0.0 as host, in node 6.16.0. The proxy then fails to connect with a EADDRNOTAVAIL error. Only happens in that node version and in the appveyor environment. I cannot repro when I run it in a container with the same node version. It doesn't seem to be an async problem either. Any ideas?

evilebottnawi:
Oh, no ideas
/cc @hiroppy

evilebottnawi:
Something wrong with CI :confused:

rlamana:
@evilebottnawi I'm still trying to figure out why the proxy fails to connect to the devserver on the appveyor Windows environment for node 6 when the host is 0.0.0.0. I tried to launch a local instance of the appveyor env in local to repro the issue and try to debug with no luck...

rlamana:
@evilebottnawi @hiroppy, it is now ready to merge.

Apparently, win32/node 6 has issues establishing a connection to `0.0.0.0`. I fixed it by just pointing the test proxy to localhost. The test needs the dev server to be listening on 0.0.0.0, it doesn't really matter how the proxy redirects to it anyway, as long as it reaches the dev server.

With this PR you can now use the `runBrowser` helper I included to add more e2e testing. I hope this helps.

hiroppy:
Sorry for the late reply. Thank you for creating e2e environment.
@rlamana Could you change the port number because `8080` is popular number?

rlamana:
@hiroppy in the test code? sure! I just thought since it should be in an isolated environment it wouldn't matter. I'll just set `9001` for the dev server and `9000` for the proxy.

hiroppy:
Yep, thanks! I'll change the default port number(`test/helper.js`) at other pr.

test/Client.test.js
```diff
@@ -0,0 +1,75 @@
...
+      compress: true,
+      port: 9001,
+      host: '0.0.0.0',
+      disableHostCheck: true,
+      hot: true,
+      watchOptions: {
+        poll: true,
+      },
+    };
+    addEntries(config, options);
```
evilebottnawi: /cc @rlamana Looks we catch bug, when we use `API`, hot plugin should be added too, can you create issue?

rlamana: I assumed that was the API expected behavior, so I didn't think it was a bug. It all makes sense now. I will create the issue then.

rlamana: I created issue #1703.

pedrofurtado:
@rlamana @evilebottnawi This PR, in version 3.3.0, breaks the use of webpack-dev-server with Rails webpacker gem, inside Docker containers. We used to configure rails app to execute in 0.0.0.0:3000, and webpack-dev-server in 0.0.0.0:3035. So, the socket must be "0.0.0.0:3035". But, with this version, the socket URL is generated as "0.0.0.0:3000" (same port of Rails app, instead of 3035).

pedrofurtado:
The version 3.2.1 does not have problem with it :+1:

nick:
This change also broke my setup. I have a static assets server on port 3000 that refers to the webpack-dev-server build on port 8080. Now it is looking for the websocket on port 3000 instead of 8080. This worked fine in 3.2.1.

rlamana:
@pedrofurtado @nick, so as I understand you are not using the dev-server to actually serve your files, but just using it for the hot reload capability?

https://github.com/webpack-contrib/webpack-hot-middleware actually allows you to add hot reloading into an existing server.

rlamana:
I wonder if `client-src/default/index.js` should also take into account the [`devServer.public`](https://webpack.js.org/configuration/dev-server/#devserverpublic) option to override the URL value and support these kind of scenarios where dev server is not use to serve the files.

jcudzich:
@rlamana This change also broke my setup. I using the dev-server the same way as @nick.

chenyong:
This change broke out tools too. We have multiple sites proxied to a single domain with different paths. The original behavior was to get the port the actual Webpack dev server, rather than the one after proxied. Now the port changed and it's very tricky works to update the rest parts of the toolchains.

pedrofurtado:
> @pedrofurtado @nick, so as I understand you are not using the dev-server to actually serve your files, but just using it for the hot reload capability?
> 
> https://github.com/webpack-contrib/webpack-hot-middleware actually allows you to add hot reloading into an existing server.

Thanks for feedback, @rlamana ! Firstly, thanks for you work in maintainance of webpack-dev-server!

Actually, I use webpack-dev-server for serve static files and hot replacement. For this, we here use the Webpacker gem (Integration between Webpack and Rails), that provides the two features mentioned (serving static files and hot replacement). Our setup is Rails application running in localhost:3000, and webpack-dev-server (and its socket) running in localhost:3035

My scenario and scenario of @nick are similar :+1:

evilebottnawi:
It was fixed, please test with latest version

pedrofurtado:
@evilebottnawi Thanks! I will test here. Doubt: The latest version is the 3.3.0 (re-released) or the branch master?

evilebottnawi:
@pedrofurtado `3.3.0`, but we prepare `3.3.1` on today due some regressions

pedrofurtado:
@evilebottnawi Ok, thank you!

pedrofurtado:
@evilebottnawi Even in `3.3.1`, the error is still occuring :cry:  The workaround, for me, is to lock in version `3.2.1` for a while.

hiroppy:
@pedrofurtado Please submit a reproducable repo.

leoliu:
Broke my setup too (similar to https://github.com/webpack/webpack-dev-server/pull/1664#issuecomment-481006338.). `webpack-dev-server` running on a different port is only used to serve assets i.e. js and whatnot.

evilebottnawi:
https://github.com/webpack/webpack-dev-server/pull/1664#issuecomment-481699922

timboslice69:
also broke my setup, this fix is a breaking change, maybe revisit it?

fkranenburg:
This also broke our setup and this is a breaking change for us. How can we fix this?

evilebottnawi:
yep :disappointed: we have `sockPort` for this case now https://github.com/webpack/webpack-dev-server/pull/1792, feel free to send a PR with revert

gtzampanakis:
This broke my setup as well. I have an HTML page that references two bundles. Each of them is served from a different WDS instance, running on a different port.

evilebottnawi:
PR welcome to revert

PR #1665, opened 2019-02-14
fix(Server): cascade stats options

### LINKED ISSUE #1557: No respect, no respect at all! (for stats.warningsFilter 😉)
* Operating System: mac OS Sierra 10.12.6
* Node Version: 6.11.1
* NPM Version: 3.10.10
* webpack Version: 4.19.0
* webpack-dev-server Version: 3.1.8

- [x] This is a **bug**

### Code

**webpack.config_.dev_.js**
```js
  //...
devServer: {
    host: '0.0.0.0',
    port: WEBPACK_PORT,
    watchOptions: {
      aggregateTimeout: 300,
      poll: true
    },
    stats: {
      assets: false,
      children: false,
      colors: true,
      version: false,
      hash: false,
      timings: true,
      chunks: false,
      chunkModules: false,
      warningsFilter: '[mini-css-extract-plugin]'
    },
    headers: {
      'Access-Control-Allow-Origin': '*'
    }
```

Webpack config stats docs I'm going off of:
https://webpack.js.org/configuration/stats/#stats

WDS stats docs:
https://webpack.js.org/configuration/dev-server/#devserver-stats-

### Expected Behavior
While using `stats.warningsFilter` in my configuration, then it would be respected by webpack-dev-server, and that errors of anything from `stats.warningsFilter` would be filtered out and not posted to my browser's console. The closest to addressing logging that I can find in your code is [this line](https://github.com/webpack/webpack-dev-server/blob/55398b5de17b9a845b1ee5aaa90bb2002c25ddfb/bin/webpack-dev-server.js#L179), or several other lines that check for `stats.colors` from the same file.
Note: I _am_ certain that my stats are connected correctly, because setting my `stats.colors` to false turns the colors off.

### Actual Behavior
When adding `warningsFilter: '[mini-css-extract-plugin]'` to my config, errors from Webpack-Dev-Middleware are respected, and I no longer see them:
```
[webpack] ℹ ｢wdm｣: Compiled with warnings.
```
However, I am still getting my warnings in the browser's console:

screen shot 2018-11-01 at 3 41 11 pm

Which is originating from https://github.com/webpack/webpack-dev-server/blob/55398b5de17b9a845b1ee5aaa90bb2002c25ddfb/client-src/default/index.js#L145

Hmm I guess you'd need to set up a warning that you would try to filter out with your webpack config. Open browser console, errors persist with each page reload.

Note: when I update webpack-dev-server to latest version (3.1.10), problem still persists.

Thank you in advance!

yoannmoinet:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Will work on it if really needed.

### Motivation / Use-Case

We get flooded by warnings in the console.

### Additional Info

Fixes #1557

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -69,6 +69,17 @@ if (semver.satisfies(process.version, '8.6.0 - 9')) {
 }
 
 class Server {
+  static get DEFAULT_STATS() {
+    return {
+      all: false,
+      hash: true,
+      assets: true,
+      warnings: true,
+      errors: true,
+      errorDetails: false,
+    };
+  }
+
   constructor(compiler, options = {}, _log) {
     this.log = _log || createLogger(options);
 
@@ -78,15 +89,10 @@ class Server {
       throw new Error("'filename' option must be set in lazy mode.");
     }
 
-    this.stats = options.stats;
-    this.DEFAULT_STATS = {
-      all: false,
-      hash: true,
-      assets: true,
-      warnings: true,
-      errors: true,
-      errorDetails: false,
-    };
+    this.stats =
+      options.stats && Object.keys(options.stats).length
+        ? options.stats
+        : Server.DEFAULT_STATS;
     this.hot = options.hot || options.hotOnly;
     this.headers = options.headers;
     this.progress = options.progress;
@@ -669,10 +675,7 @@ class Server {
   }
 
   getStats(statsObj) {
-    return Object.assign(
-      statsObj.toJson(this.DEFAULT_STATS),
-      statsObj.toJson(this.stats)
-    );
+    return statsObj.toJson(this.stats);
   }
 
   use() {
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -68,15 +68,6 @@ if (semver.satisfies(process.version, '8.6.0 - 9')) {
   tls.DEFAULT_ECDH_CURVE = 'auto';
 }
 
-const STATS = {
-  all: false,
-  hash: true,
-  assets: true,
-  warnings: true,
-  errors: true,
-  errorDetails: false,
-};
-
 class Server {
   constructor(compiler, options = {}, _log) {
     this.log = _log || createLogger(options);
@@ -87,6 +78,15 @@ class Server {
       throw new Error("'filename' option must be set in lazy mode.");
     }
 
+    this.stats = options.stats;
+    this.DEFAULT_STATS = {
+      all: false,
+      hash: true,
+      assets: true,
+      warnings: true,
+      errors: true,
+      errorDetails: false,
+    };
     this.hot = options.hot || options.hotOnly;
     this.headers = options.headers;
     this.progress = options.progress;
@@ -140,7 +140,7 @@ class Server {
       compile.tap('webpack-dev-server', invalidPlugin);
       invalid.tap('webpack-dev-server', invalidPlugin);
       done.tap('webpack-dev-server', (stats) => {
-        this._sendStats(this.sockets, stats.toJson(STATS));
+        this._sendStats(this.sockets, this.getStats(stats));
         this._stats = stats;
       });
     };
@@ -668,6 +668,13 @@ class Server {
     }, this);
   }
 
+  getStats(statsObj) {
+    return Object.assign(
+      statsObj.toJson(this.DEFAULT_STATS),
+      statsObj.toJson(this.stats)
+    );
+  }
+
   use() {
     // eslint-disable-next-line
     this.app.use.apply(this.app, arguments);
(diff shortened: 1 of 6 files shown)


evilebottnawi:
Also please accept CLA

yoannmoinet:
Actually this won't work if we use a string for `options.stats` like `errors-only`.
Will update and add tests.

yoannmoinet:
@evilebottnawi I've refactored the code to also accepts strings.
I tried to write tests, but I'm not sure where to begin.

I tried creating a new `test/Server.test.js` but I can't make any test pass, I think I'm going in the wrong direction.
Would you help please?

yoannmoinet:
Can someone help with the tests (cc @evilebottnawi) ?

Here's what I tried with `Server.test.js` but I've been unsuccessful.

```javascript
'use strict';

const Server = require('../lib/Server');
const config = require('./fixtures/simple-config/webpack.config');

const allStats = [
  {},
  // eslint-disable-next-line
  undefined,
  'errors-only',
  {
    assetsSort: "size"
  }
];

describe('Server', () => {
  allStats.forEach((s) => {
    it('should cascade stats options', (done) => {
      const compiler = webpack(Object.assign({}, config, s && { stats: s }));
      const server = new Server(compiler, {});
      const next = () => {
        const localStats = server.getStats(server._stats);

        if(!s || !Object.keys(s).length) {
          expect(localStats).toBe(server.DEFAULT_STATS);
        } else {
          expect(localStats).not.toBe(server.DEFAULT_STATS);
        }

        server.close(() => {
          done();
        });
      };
      server.listen(8080, 'localhost', (err) => {
        if (err) return next(err);
        next();
      });
    });
  });
});
```

yoannmoinet:
@evilebottnawi from what I understand, what you pointed to is if we want to use webpack's stats config if none is given to devServer's config, as a fallback.
I've added a commit that does just that, but on its own, it won't fix the bug, we still need the change in `Server.js` for it to cascade into the browser's console.

lib/Server.js
```diff
@@ -91,6 +82,15 @@ class Server {
...
 
+    this.stats = options.stats;
+    this.DEFAULT_STATS = {
+      all: false,
+      hash: true,
+      assets: true,
+      warnings: true,
+      errors: true,
+      errorDetails: false,
+    };
```
evilebottnawi: Let's avoid `this.DEFAULT_STATS` (because it should be as private property)

hiroppy:
The snapshot test is currently failing.

lib/Server.js
yoannmoinet: I needed it for the tests.
Should I add them to the Server's prototype then?

evilebottnawi: Can you provide link where you need it for test? You already have `getStats` method

yoannmoinet:
> Loooks good, need add tests for `warningsFilter`

@evilebottnawi I'm not sure what you're asking for.

> The snapshot test is currently failing.

@hiroppy indeed, I cannot reproduce locally though, so I don't know what to look for.

evilebottnawi:
@yoannmoinet we need add tests for `stats` https://webpack.js.org/configuration/stats/ (`warningsFilter` option) to ensure we fix problem (what is written in issue)

yoannmoinet:
@evilebottnawi oh yes, I'm working on this right now 😉 
I wasn't sure we were talking about the same thing 👍

yoannmoinet:
@evilebottnawi here you go, I've added test for the Server.

lib/Server.js
yoannmoinet: [this is where I'm using it](https://github.com/webpack/webpack-dev-server/pull/1665/files#diff-7775e181de5f32a001a1636103e7008fR33)

yoannmoinet:
I'm not sure I understand Travis' errors.

yoannmoinet:
can you help with the failing task in Travis @evilebottnawi please?

evilebottnawi:
@yoannmoinet try to do rebase

yoannmoinet:
I'm not sure what kind of test I'm supposed to add following the `codecov/patch` task failing.
I felt like I added all the tests needed.
cc @evilebottnawi

lib/Server.js
```diff
@@ -668,6 +674,10 @@ class Server {
     }, this);
   }
 
+  getStats(statsObj) {
+    return statsObj.toJson(this.stats);
+  }
```
evilebottnawi: Let's move
```
this.stats =
      options.stats && Object.keys(options.stats).length
        ? options.stats
        : Server.DEFAULT_STATS;
```

logic here, we don't need new property like `stats`

yoannmoinet: Oh, good point, will update.

yoannmoinet: I still need to keep the property to have access to `options.stats`.

yoannmoinet: Not sure it would be useful to move the logic since I still need to keep that `this.stats` so I can use `options.stats` in `getStats()`.
What do you think @evilebottnawi ?

evilebottnawi: Yep, you are right, we should rewrite this in next major release

PR #1667, opened 2019-02-16
test(Proxy): skip Proxy.test.js

hiroppy:
This test is unstable when running on Node v11 on MacOS.

- [ ] This is a **bugfix**
- [ ] This is a **feature**

N/A

### Motivation / Use-Case

see https://github.com/webpack/webpack-dev-server/pull/1633#issuecomment-463982581

### Breaking Changes

no 

### Additional Info

`HistoryApiFallback.test.js` seems to be good when we add skipping code to `Proxy.test.js`.

### DIFF
--- a/test/Proxy.test.js
+++ b/test/Proxy.test.js
@@ -6,6 +6,7 @@ const express = require('express');
 const WebSocket = require('ws');
 const helper = require('./helper');
 const config = require('./fixtures/proxy-config/webpack.config');
+const shouldSkipTestSuite = require('./shouldSkipTestSuite');
 
 const WebSocketServer = WebSocket.Server;
 const contentBase = path.join(__dirname, 'fixtures/proxy-config');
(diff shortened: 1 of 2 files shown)


test/Proxy.test.js
```diff
@@ -62,6 +63,11 @@ function startProxyServers() {
 }
 
 describe('Proxy', () => {
+  // issue: https://github.com/webpack/webpack-dev-server/pull/1633#issuecomment-463938675
+  if (shouldSkipTestSuite()) {
+    return;
+  }
```
evilebottnawi: All tests failed? Maybe we can skip only one test?

hiroppy: As far as I confirm, running two tests seems to fail.

evilebottnawi: @hiroppy can we ship only their, thanks?

hiroppy: The first test(anyone) will succeed but subsequent tests will fail. 
Probably postprocessing would be a problem.

Once we merge this PR and then I think that it is better to investigate the cause separately.(maybe difficult)

evilebottnawi: :+1: Let's do it

PR #1668, opened 2019-02-18
Use internal-ip@4.0.0

patrickhrastnik:
- [ ] This is a **bugfix**
- [ ] This is a **feature**
- [x] This is a **metadata update** (I'm not sure if this category is correct)

### Motivation / Use-Case

_Short conclusion:_ This change makes webpack-dev-server available for AIX systems.

internal-ip got an upgrade, as well as its dependency default-gateway. Both of them now support aix systems. Since internal-ip got a new major release version (because it was upgraded to use node>=6 as well), we need to upgrade the required version of internal-ip here.

We have an AIX system we want to use Angular and/or React on. Since default-gateway was not available on AIX systems, we had to come up with a solution for this system.
The package maintainer of default-gateway politely made the pull request to update internal-ip. By doing this, they made a new major release. That's why the changes in default-gateway and internal-ip are not automatically used when installing webpack-dev-server.

### Additional Info

default-gateway has a new optional dependency idb-connector. This is a package only available for AIX systems. Thus, it's listed in the package-lock.json, but not installed on other platforms. Optional dependencies don't lead to failure if they cannot be installed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2780,7 +2780,6 @@
       "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz",
       "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
       "dev": true,
-      "optional": true,
       "requires": {
         "kind-of": "^3.0.2",
         "longest": "^1.0.1",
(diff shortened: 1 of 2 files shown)


patrickhrastnik:
Thanks for merging the changes. When are they going to be released, to be available via npm?

evilebottnawi:
Today

PR #1676, opened 2019-02-21
fix: allow `ca`, `key` and `cert` will be string (regression)

### LINKED ISSUE #1674: Passing an SSL key or certificate as a raw string results in an exception
- Operating System: Windows 10
- Node Version: 11.1.0
- NPM Version: 6.4.1
- webpack Version: 4.27.1
- webpack-dev-server Version: 3.2.0

- [x] This is a **bug**

### Code

```ts
// startWebpackDevServer.ts
/** `ssl` contains the `key` and `cert` properties as raw strings, rather than filenames */
const { ssl } = getUserConfig();
httpsConfig = {
    https: {
        ...ssl,
    },
};

const server = new WebpackDevServer(
        compiler,
        Object.assign(
            httpsConfig,
            {
                port: 3000,
                inline: true,
                quiet: false,
                noInfo: false,
                headers: {
                    'Access-Control-Allow-Origin': '*',
                    'Access-Control-Allow-Credentials': 'true',
                },
                stats: 'errors-only',
                reporter: reporter,
            },
            config.devServer
        )
    );
```

### Expected Behavior
`webpack-dev-server` operates as it did prior to 3.2.0, without throwing an exception

### Actual Behavior
```
(node:15556) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, lstat '-----BEGIN RSA PRIVATE KEY-----
<<REDACTED>>
-----END RSA PRIVATE KEY-----
'
    at Object.lstatSync (fs.js:861:3)
    at new Server (D:\vso\client-web\node_modules\owa-build\node_modules\webpack-dev-server\lib\Server.js:573:38)
```

Pass the raw SSL key and/or cert, rather than a Buffer or filepath, to the https config

evilebottnawi:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

fixes #1674

### Breaking Changes

No

### Additional Info

It is regression, just not `pfx` can't be string due it is binary format

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -570,8 +570,21 @@ class Server {
         const value = options.https[property];
         const isBuffer = value instanceof Buffer;
 
-        if (value && !isBuffer && fs.lstatSync(value).isFile()) {
-          options.https[property] = fs.readFileSync(path.resolve(value));
+        if (value && !isBuffer) {
+          let stats = null;
+
+          try {
+            stats = fs.lstatSync(value).isFile();
+          } catch (error) {
+            // ignore error
+          }
+
+          if (stats) {
+            // It is file
+            options.https[property] = fs.readFileSync(path.resolve(value));
+          } else {
+            options.https[property] = value;
+          }
         }
       }
 
@@ -610,7 +623,7 @@ class Server {
           const pems = createCertificate(attrs);
 
           fs.writeFileSync(certPath, pems.private + pems.cert, {
-            encoding: 'utf-8',
+            encoding: 'utf8',
           });
         }
 
--- a/test/Https.test.js
+++ b/test/Https.test.js
@@ -18,9 +18,8 @@ const contentBasePublic = path.join(
 describe('HTTPS', () => {
   let server;
   let req;
-  afterEach(helper.close);
 
-  describe('to directory', () => {
+  describe('is boolean', () => {
     beforeAll((done) => {
       server = helper.start(
         config,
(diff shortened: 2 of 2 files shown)


lib/Server.js
```diff
@@ -610,7 +623,7 @@ class Server {
           const pems = createCertificate(attrs);
 
           fs.writeFileSync(certPath, pems.private + pems.cert, {
-            encoding: 'utf-8',
+            encoding: 'utf8',
```
evilebottnawi: `utf8` is right value https://nodejs.org/api/fs.html#fs_fs_writefilesync_file_data_options

lib/Server.js
```diff
@@ -570,8 +570,21 @@ class Server {
...
+
+          try {
+            stats = fs.lstatSync(value).isFile();
+          } catch (error) {
+            // ignore error
+          }
+
+          if (stats) {
+            // It is file
+            options.https[property] = fs.readFileSync(path.resolve(value));
```
alan-agius4: Shouldn’t this have utf8 encoding as well?

evilebottnawi: @alan-agius4 hm, maybe, can you provide example?

evilebottnawi: No need 
https://github.com/nodejs/node/blob/master/lib/fs.js#L392

`buffer.toString` :
```
encoding <string> The character encoding to use. Default: 'utf8'.

buf.toString([encoding[, start[, end]]])
```

evilebottnawi: So default is `utf8`

alan-agius4: thanks for clarifying

PR #1679, opened 2019-02-21
Test refactor and more

evilebottnawi:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

Refactor

### Breaking Changes

No

### Additional Info

Add comments for `v4`

### DIFF
--- a/jest.config.json
+++ b/jest.config.json
@@ -1,6 +1,7 @@
 {
   "testURL": "http://localhost/",
   "collectCoverage": true,
+  "coveragePathIgnorePatterns": ["test"],
   "moduleFileExtensions": ["js", "json"],
   "testMatch": ["**/test/**/*.test.js"],
   "setupFilesAfterEnv": ["<rootDir>/setupTest.js"]
--- a/lib/utils/createConfig.js
+++ b/lib/utils/createConfig.js
@@ -14,12 +14,12 @@ function createConfig(config, argv, { port }) {
     options.bonjour = true;
   }
 
-  if (argv.host !== 'localhost' || !options.host) {
+  if (argv.host && (argv.host !== 'localhost' || !options.host)) {
     options.host = argv.host;
   }
 
-  if (argv['allowed-hosts']) {
-    options.allowedHosts = argv['allowed-hosts'].split(',');
+  if (argv.allowedHosts) {
+    options.allowedHosts = argv.allowedHosts.split(',');
   }
 
   if (argv.public) {
@@ -47,11 +47,11 @@ function createConfig(config, argv, { port }) {
     }
   }
 
-  if (!options.filename) {
+  if (!options.filename && firstWpOpt.output && firstWpOpt.output.filename) {
     options.filename = firstWpOpt.output && firstWpOpt.output.filename;
   }
 
-  if (!options.watchOptions) {
+  if (!options.watchOptions && firstWpOpt.watchOptions) {
     options.watchOptions = firstWpOpt.watchOptions;
   }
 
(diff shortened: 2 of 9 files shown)


jest.config.json
```diff
@@ -1,6 +1,7 @@
 {
   "testURL": "http://localhost/",
   "collectCoverage": true,
+  "coveragePathIgnorePatterns": ["test"],
```
evilebottnawi: Ignore `test` directory for coverage

lib/utils/createConfig.js
```diff
@@ -14,12 +14,12 @@ function createConfig(config, argv, { port }) {
...
   }
 
-  if (argv.host !== 'localhost' || !options.host) {
+  if (argv.host && (argv.host !== 'localhost' || !options.host)) {
     options.host = argv.host;
   }
 
-  if (argv['allowed-hosts']) {
-    options.allowedHosts = argv['allowed-hosts'].split(',');
+  if (argv.allowedHosts) {
```
evilebottnawi: yargs support all `arguments` in `camelCase` format, so we don't need use hack like `argv['allowed-hosts']`

lib/utils/createConfig.js
```diff
@@ -14,12 +14,12 @@ function createConfig(config, argv, { port }) {
     options.bonjour = true;
   }
 
-  if (argv.host !== 'localhost' || !options.host) {
+  if (argv.host && (argv.host !== 'localhost' || !options.host)) {
```
evilebottnawi: Set host only when `CLI` arg specify

lib/utils/createConfig.js
```diff
@@ -47,11 +47,11 @@ function createConfig(config, argv, { port }) {
     }
   }
 
-  if (!options.filename) {
+  if (!options.filename && firstWpOpt.output && firstWpOpt.output.filename) {
```
evilebottnawi: Set only when configuration specify (otherwise it was `undefined`, what doesn't have sense)

lib/utils/createConfig.js
```diff
@@ -64,38 +64,42 @@ function createConfig(config, argv, { port }) {
...
+      options.contentBase = options.contentBase.map((p) => path.resolve(p));
+    } else if (/^[0-9]$/.test(options.contentBase)) {
+      options.contentBase = +options.contentBase;
+    } else if (!/^(https?:)?\/\//.test(options.contentBase)) {
+      options.contentBase = path.resolve(options.contentBase);
     }
+    // It is possible to disable the contentBase by using
+    // `--no-content-base`, which results in arg["content-base"] = false
+  } else if (argv.contentBase === false) {
+    // TODO doesn't work need fix, `false` in this case is string
```
evilebottnawi: Need fix, let's do it in other PR

test/cli.test.js
```diff
@@ -21,62 +21,76 @@ describe('CLI', () => {
...
+  });
+
   it('--https', (done) => {
     runDevServer('--https')
       .then((output) => {
         expect(output.code).toEqual(0);
-        expect(
-          output.stdout.indexOf('Project is running at') >= 0
-        ).toBeTruthy();
+        expect(output.stdout.indexOf('Project is running at') >= 0).toBe(true);
```
evilebottnawi: We should prefer use `toBe(true)` to avoid misleading in tests

lib/utils/createConfig.js
evilebottnawi: Same logic for other arguments (otherwise it is `undefined` and it makes no sense

lib/utils/createConfig.js
evilebottnawi: Same logic for other arguments too

lib/utils/createConfig.js
```diff
@@ -117,10 +122,12 @@ function createConfig(config, argv, { port }) {
     options.lazy = true;
   }
 
+  // TODO remove in `v4`
```
hiroppy: Why will we remove this from v4?

test/cli.test.js
```diff
@@ -21,62 +21,76 @@ describe('CLI', () => {
...
         done();
       })
       .catch(done);
   });
+
+  it('--bonjour', (done) => {
+    runDevServer('--bonjour')
+      .then((output) => {
+        expect(output.code).toEqual(0);
+        expect(output.stdout.indexOf('Bonjour') >= 0).toBe(true);
```
hiroppy: we can use `includes`:)

lib/utils/createConfig.js
evilebottnawi: @hiroppy in favor `logLevel` option or `logger`, we should refactor this place because we have `4` option for output log and stats, i think we should have 1/2 option, so i mark this as TODO on removing, anyway we can revisit this in future

test/cli.test.js
evilebottnawi: :+1:

lib/utils/createConfig.js
hiroppy: Thank you for answering. I understood.

PR #1685, opened 2019-02-22
chore(readme): update styles and information

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

no

### Motivation / Use-Case

Change to the same style as webpack's readme.

### Breaking Changes

no


README.md
```diff
@@ -30,7 +31,7 @@ fast in-memory access to the webpack assets.
 7. [Attribution](#attribution)
 8. [License](#license)
 
-## Getting Started
+<h2 align="center">Getting Started</h2>
```
evilebottnawi: We should avoid using `html`, it is bad practice. Also let's do:
- Remove `Maintainers` section (it is outdated) and no need
- Don't move `div` after description

hiroppy: > * Don't move `div` after description

sorry, I want to get more details.

evilebottnawi: @hiroppy yep L21

hiroppy:
updated

hiroppy:
update

PR #1692, opened 2019-02-25
Random port retry logic

mistic:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Some find port tests were added.

### Motivation / Use-Case
Closes https://github.com/webpack/webpack-dev-server/issues/1572.
The main change is to delay port finding to just before trying to run server.listen. Just this change alone reduced the time window between determining the available port and actually listening to it, as previously the compiler is initiated between this time frame. This naturally let most instances attempt to get available port in different moments and reducing the risk of trying to use the same port.
Another change is to listen to error event when port finding is used, when 'EADDRINUSE' error is encountered, retry the port finding again for up to 10 times. Not the most elegant solution, but should be sufficient for most prototyping scenarios.
If retried for 10 times and still cannot find available pot to use, the dev-server will error out like before.

### Breaking Changes
Unless there are people relying on the fact that attempting to start multiple instances without specifying port could result in errors, this should not have breaking changes.

### Additional Info
This is recreation of https://github.com/webpack/webpack-dev-server/pull/1577 which seems to have any activity. As this is an important feature,  this PR presents the original work with some refactoring and also conflict free state after merged against last master. The original commits were cherry picked and most of the credits should be given to the original author of the PR. 

A manual test of the feature was also performed.

/CC @evilebottnawi


bin/webpack-dev-server.js
```diff
@@ -93,6 +94,15 @@ const config = require('webpack-cli/bin/convert-argv')(yargs, argv, {
 // we should use portfinder.
 const DEFAULT_PORT = 8080;
 
+// Try to find unused port and listen on it for 3 times,
+// if port is not specified in options.
+// Because NaN == null is false, defaultTo fails if parseInt returns NaN
+// so the tryParseInt function is introduced to handle NaN
+const defaultPortRetry = defaultTo(
+  tryParseInt(process.env.DEFAULT_PORT_RETRY),
```
hiroppy: IMO: I think we should leave `process.env.DEFAULT_PORT_RETRY` as a document.

evilebottnawi: @hiroppy can you create issue in webpack docs repo?

hiroppy: I think `DEFAULT_PORT_RETRY` should be added in `devServer` because currently, we don' use `process.env`. see https://webpack.js.org/configuration/dev-server/

@evilebottnawi what do you think?

evilebottnawi: @hiroppy yep, after merge we need open issue here https://github.com/webpack/webpack.js.org :+1:

u9520107:
Thanks for doing this. I must apologize for putting this aside for too long. Got bombarded with work right after our Chinese New Year holidays, and I totally forgot to put this into my reminders.

Issue #1677, opened 2019-02-21
proxy.bypass return false does not work


jsmith-sapient:
- Operating System: MacOS High Sierra
- Node Version: 11.0.0
- NPM Version: 6.4.1
- webpack Version: 4.29.5
- webpack-dev-server Version: 3.2.0

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

does *not* work:
```js
// webpack.config.js

module.exports = {
  devServer: {
    proxy: {
      '**': {
        bypass(req) {
          if (!req.headers.accept.includes('application/json'))
            return false; // ← does nothing
        },
        target: `http://localhost:${yargs.proxyPort}`,
      },
    },
  },
  entry: './app',
}
```

*does* work:
```js
// webpack.config.js

module.exports = {
  devServer: {
    proxy: {
      '**': {
        bypass(req) {
          if (!req.headers.accept.includes('application/json'))
            return req.url;
        },
        target: `http://localhost:${yargs.proxyPort}`,
      },
    },
  },
  entry: './app',
}
```

### Expected Behaviour

Based on the [vague explanation in the docs](https://webpack.js.org/configuration/dev-server/#devserverproxy)

>It must return either `false` or a path that will be served instead of continuing to proxy the request.

I expect `return false` to skip the proxy

### Actual Behaviour

`return false` does nothing.

### For Bugs; How can we reproduce the behaviour?

Minimal example: https://github.com/jsmith-sapient/local-api-mocker

```sh
npm run start:dev
```

evilebottnawi:
Can you try return `null` or something else? Looks like regression in `http-proxy`/`http-proxy-middleware`

evilebottnawi:
Anyway PR welcome

jsmith-sapient:
@evilebottnawi I tried `null`: No difference to `false` (it does nothing).

I'll try to find some time this weekend to take a look. I might be posting as @jshado1 (my personal account, not a catfish).

evilebottnawi:
@jsmith-sapient i will look on this in near future, anyway we glad any help

evilebottnawi:
@jsmith-sapient reproducible test repo works fine, i am use `/test` and it is return `json`. What expected?

jsmith-sapient:
@evilebottnawi see the [commit history](https://github.com/jsmith-sapient/local-api-mocker/commits/webpack-devsrv):

>
>[…]
>
[...]

PR #1696, opened 2019-02-28
Fix proxy bypass return false

mistic:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

Based on the [explanation in the docs](https://webpack.js.org/configuration/dev-server/#devserverproxy)

> It must return either `false` or a path that will be served instead of continuing to proxy the request.

It was expected that `return false` skip the proxy with a 404

### Breaking Changes

None

### Additional Info

Closes https://github.com/webpack/webpack-dev-server/issues/1677

/CC @evilebottnawi 

@jsmith-sapient @jshado1 @evilebottnawi could you please test this manually and confirm that this have the intended behaviour?

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -381,21 +381,22 @@ class Server {
                 }
               }
 
-              // const bypass = typeof proxyConfig.bypass === 'function';
-              // const bypassRawValue = bypass ? proxyConfig.bypass(req, res, proxyConfig) : null;
-              // const bypassUrl = typeof bypassRawValue === 'string'
-              //   ? bypassRawValue
-              //   : (typeof bypassRawValue === 'boolean' && bypassRawValue === false);
-
               const bypass = typeof proxyConfig.bypass === 'function';
+              const bypassRawValue = bypass
+                ? proxyConfig.bypass(req, res, proxyConfig)
+                : null;
               const bypassUrl =
-                (bypass && proxyConfig.bypass(req, res, proxyConfig)) || false;
+                typeof bypassRawValue === 'string'
+                  ? bypassRawValue
+                  : typeof bypassRawValue === 'boolean' &&
+                    bypassRawValue === false;
 
               if (bypassUrl) {
-                // In case bypassURL is a boolean with a true value
-                // (as result from setting a bypassRawValue boolean
-                // with false value) the request will be ignored
-                req.url = bypassUrl;
+                // Only in case bypassURL is a string
+                // the request will be set
+                if (typeof bypassUrl === 'string') {
+                  req.url = bypassUrl;
+                }
                 next();
               } else if (proxyMiddleware) {
                 return proxyMiddleware(req, res, next);
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -381,14 +381,21 @@ class Server {
                 }
               }
 
-              const bypass = typeof proxyConfig.bypass === 'function';
+              // const bypass = typeof proxyConfig.bypass === 'function';
+              // const bypassRawValue = bypass ? proxyConfig.bypass(req, res, proxyConfig) : null;
+              // const bypassUrl = typeof bypassRawValue === 'string'
+              //   ? bypassRawValue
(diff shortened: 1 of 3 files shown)


lib/Server.js
```diff
@@ -382,12 +382,23 @@ class Server {
...
+                ? proxyConfig.bypass(req, res, proxyConfig)
+                : null;
               const bypassUrl =
-                (bypass && proxyConfig.bypass(req, res, proxyConfig)) || false;
+                typeof bypassRawValue === 'string'
+                  ? bypassRawValue
+                  : typeof bypassRawValue === 'boolean' &&
+                    bypassRawValue === false;
 
               if (bypassUrl) {
```
evilebottnawi: Why don't use :
```js
// Remove very strange nested check for `string` and `boolean`?

if (bypassRawValue) {
 // Logic
}
```

lib/Server.js
```diff
@@ -382,12 +382,23 @@ class Server {
...
-
+              const bypassRawValue = bypass
+                ? proxyConfig.bypass(req, res, proxyConfig)
+                : null;
               const bypassUrl =
-                (bypass && proxyConfig.bypass(req, res, proxyConfig)) || false;
+                typeof bypassRawValue === 'string'
+                  ? bypassRawValue
+                  : typeof bypassRawValue === 'boolean' &&
+                    bypassRawValue === false;
```
jshado1: The type check is superfluous because the triple (exactly) equal already asserts the type check.

jshado1: this plus the `if` below is massively convoluted. please don't do this. also code comments 😳

lib/Server.js
```diff
@@ -382,12 +382,23 @@ class Server {
               }
 
               const bypass = typeof proxyConfig.bypass === 'function';
```
jshado1: Setting the default value of proxyConfig.bypass to a no-op function would dramatically simplify the code here (assuming noop default):

```suggestion
              const bypassUrl = proxyConfig.bypass(req, res, proxyConfig);
              
              if (bypassUrl) req.url = bypassUrl;
              else if (bypassUrl === false) req.url = null;
              
              next();
```

It would also help if the flag for `bypass` to return was `null` (since that's apparently what req.url wants), instead of mixing `false` and `null` (but that would probably be a breaking change). I think this is right, but the suggested logic is nigh impossible to follow after re-reading it for 15 minutes:

```js
              const bypassUrl = proxyConfig.bypass(req, res, proxyConfig);
              
              if (bypassUrl || bypassUrl === null) req.url = bypassUrl;
              
              next();
```

mistic:
@evilebottnawi @jshado1 thanks for the suggestions. I tried to make the logic more clear and simple.

lib/Server.js
```diff
@@ -381,25 +381,22 @@ class Server {
...
               }
 
-              const bypass = typeof proxyConfig.bypass === 'function';
-              const bypassRawValue = bypass
+              // - Check if we have a bypass function defined
+              // - In case the bypass function is defined we'll retrieve the
+              // bypassUrl from it otherwise byPassUrl would be null
+              const isByPassFuncDefined =
+                typeof proxyConfig.bypass === 'function';
+              const bypassUrl = isByPassFuncDefined
```
jshado1: `isByPassFuncDefined` wasn't too confusing before, but no harm having it. The new code comments explain it :)

mistic:
@evilebottnawi I think you can merge this 😄

lib/Server.js
```diff
@@ -381,14 +381,22 @@ class Server {
...
+                ? proxyConfig.bypass(req, res, proxyConfig)
+                : null;
+
+              if (typeof bypassUrl === 'boolean') {
```
randing89: This is not aligning to the documentation 
> In the function you get access to the request, response and proxy options. It must return either false or a path that will be served instead of continuing to proxy the request.

evilebottnawi: PR welcome to webpack docs

randing89: @evilebottnawi https://github.com/webpack/webpack.js.org/pull/3211

PR #1698, opened 2019-03-04
do not swallow errors from http server.

0xorial:
- [ ] This is a **bugfix**
- [x] This is a **feature**

I was not sure where I can put a test for it. If someone can guide me I will be happy to add one.

### Motivation / Use-Case

I was using WDS using node.js, made a trivial mistake (typed locahlost instead of localhost) and spent 10 minutes trying to find it. 

### Breaking Changes
None

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -659,6 +659,10 @@ class Server {
       this.listeningApp = http.createServer(app);
     }
 
+    this.listeningApp.on('error', (err) => {
+      this.log.error(err);
+    });
+
     killable(this.listeningApp);
 
     // Proxy websockets without the initial http request


evilebottnawi:
Can you provide example where we can reproduce this problem, thanks for PR

0xorial:
```TypeScript
      const config = require(configPath);

      const log = require("webpack-log");

      const logger = log({
        name: "wds",
        level: "info",
        timestamp: true
      });
      const devServerInstance = new devServer(compiler, {}, logger);
      outputChannel.appendLine("about to listen...");

      // mistyped localhost
      devServerInstance.listen(8080, "locallllhost", (err: any) => {
        // never reached here :(
        if (err) {
          outputChannel.appendLine("listen error!");
          throw err;
        }
        outputChannel.appendLine('listening!');
      });
```

0xorial:
Just for some more context - I am trying to develop an extension for Visual Studio Code to integrate WDS.

evilebottnawi:
@0xorial can you rebase and add tests?

Issue #1713, opened 2019-03-09
Question: What is the status of HTTP/2 support?


Aghassi:
- Operating System: macOS 10.14.3
- Node Version: 10
- Yarn Version: 1.13.0
- webpack Version: 4.29.4
- webpack-dev-server Version: 3.2.1

- [ ] This is a **bug**
- [ ] This is a **modification** request
- [x] This is a question

I spent about an hour digging around yesterday and couldn't find the answer to this question, so I wanted to post here and get some insight.

I read through https://github.com/webpack/webpack-dev-server/issues/516, which seems to indicate there is HTTP/2 supported, but I couldn't find any docs on how to enable it. All I found in the official docs was the line under this section https://webpack.js.org/configuration/dev-server/#devserverhttps saying that it can run with HTTP/2.

In the code in this repo I found this https://github.com/webpack/webpack-dev-server/blob/21687c3bfc95752957d16470b97d281001dc6a07/lib/Server.js#L633-L660

All of which links to issues and other repos which have either been closed or have no work being done on them. And according to this comment https://github.com/webpack/webpack-dev-server/issues/1493#issuecomment-419493497, it is planned for v4, but nothing in the 4.0.0 milestone https://github.com/webpack/webpack-dev-server/milestone/4 seems to support that statement.

Ultimately, I'm confused as to the current situation. Does Webpack-Dev-Server support HTTP/2? If not, will it be supported in the future?

evilebottnawi:
We support `http2` using `spdy` module for backward compatibility with old node version, to enable http2 you should enable `https` option and `spdy` option, but i agree it is misleading, and looks we doesn't respect `spdy` (https://github.com/webpack/webpack-dev-server/blob/master/bin/options.js), thanks for issue. We need:
1. Respect `spdy` (add alias `http2`) in CLI
2. Add tests

For future version use `http2` module instead `node-spdy`

Aghassi:
As an added note, will respecting the `spdy` option in the CLI also reflect if I am using this programmatically by importing `webpack-dev-server`?

evilebottnawi:
@Aghassi no, but fix it very easy

jshado1:
[...]

PR #1721, opened 2019-03-16
New spdy option to enable/disable HTTP/2 with HTTPS

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

I decided to try solving this problem for my first PR: https://github.com/webpack/webpack-dev-server/issues/1713

I added the spdy option for the node API and the CLI, along with an http2 alias option on both which does the same thing. When spdy is enabled, https is enabled, and node version is below 10.0.0 (since spdy is broken beyond that), spdy will be used to create the server.

### Breaking Changes

The dev server will no longer automatically serve over HTTP/2 when HTTPS is enabled and Node version is below 10.0.0, as the spdy/http2 option must be true for this to happen.

### Additional Info
Changes in docs that I made to reflect the new options: https://github.com/Loonride/webpack.js.org/blob/dev-server-spdy-http2/src/content/configuration/dev-server.md#devserverspdy


test/Spdy.test.js
```diff
@@ -0,0 +1,113 @@
...
+          https: true,
+          spdy: true,
+        },
+        done
+      );
+      req = request(server.app);
+    });
+
+    it('Request to index', (done) => {
+      req.get('/').expect(200, /Heyo/, done);
```
hiroppy: Please check `http/2` or not.

```js
    it('Request to index', (done) => {
      req
        .get('/')
        .expect(200, /Heyo/)
        .then(({ res }) => {
          expect(res.httpVersion).toEqual('2.0');
          done();
        });
    });
```

test/Spdy.test.js
```diff
@@ -0,0 +1,113 @@
...
+    });
+
+    afterAll(() => {
+      helper.close();
+    });
+  });
+
+  afterEach(helper.close);
```
hiroppy: why is this necessary?

Loonride: Sorry, mistake from copying HTTPS tests over as a template. I will remove it.

test/Spdy.test.js
Loonride: @hiroppy I've realized this is more difficult than it seems. Supertest sends an HTTP/1.1 request in the test, so the spdy server communicates back with 1.1. I confirmed in the browser that the protocol is h2 and spdy works. The two options I think are: use Node's built-in http2 module for the test and try to communicate with the server using it to confirm HTTP/2 works, or use puppeteer to test with a browser that supports HTTP/2 and confirm that the h2 protocol is used. What do you think?

lib/Server.js
```diff
@@ -87,6 +87,11 @@ class Server {
       throw new Error("'filename' option must be set in lazy mode.");
     }
 
+    if (options.spdy || options.http2) {
+      options.spdy = true;
+      options.http2 = true;
+    }
```
evilebottnawi: Let's separate `spdy` and `http2` options, because it is misleading, `spdy` and `https2` are difference protocols, better don't mix their

Loonride: @evilebottnawi I considered `http2` an alias for `spdy` because `http2` needs `spdy` to work at the moment. As an alternative, I could do this, to allow for future separation of `http2` from `spdy`: 

When ONLY `http2` is enabled, the spdy server will be run (since this is the only way to run HTTP/2 on express at the moment), and these protocols will be provided to spdy:

```js
options.https.spdy = {
  protocols: ['h2', 'http/1.1'],
};
```
Then the exact same thing will happen as above if `http2` is enabled and `spdy` is enabled.

If ONLY `spdy` is enabled, the `options.https.spdy` will not be set unless the user specifies it themselves. If the user does not specify it in their options, it will go to the default spdy value as specified [here](https://www.npmjs.com/package/spdy):

```js
protocols: ['h2','spdy/3.1', 'spdy/3', 'spdy/2','http/1.1', 'http/1.0']
```

Do you like this alternative better?

evilebottnawi: Ok, let's use the `http2` option instead `spdy`. And we can introduce `spdy` under other option in future because `spdy` !== `http2`, also we need throw error if developer doesn't defined `https` when use `http2: true`

Loonride: Ok, that makes sense. Should user still be allowed to provide spdy options like this?:

```js
{
  contentBase: './dist',
  watchContentBase: true,
  https: {
    spdy: {
      protocols: ['h2', 'http/1.1' ...]
    }
  },
  http2: true
}
```

evilebottnawi: @Loonride i thought about it, let's do this in other PR

Loonride:
I have removed the `spdy` option and instead just have the `http2` option. I also made a test which depends on Node's experimental `http2` module, since the only way to test an HTTP/2 server is to make an HTTP/2 request. I thought this was a reasonable choice since it will only force testers to have a Node version high enough for that, not users. If you think that is a problem, let me know. My alternative solution would be to dig into puppeteer and find the protocol in the test, similar to the way Chrome DevTools can show the protocol of a request.

lib/Server.js
```diff
@@ -87,9 +87,8 @@ class Server {
       throw new Error("'filename' option must be set in lazy mode.");
     }
 
-    if (options.spdy || options.http2) {
-      options.spdy = true;
-      options.http2 = true;
+    if (options.http2 && !options.https) {
+      throw new Error("'https' option must be enabled to use HTTP/2");
```
evilebottnawi: I was wrong, it is poor DX, better set `https: true` in this case, no one browsers don't support `http2` without `ssl` so, we can generate `certificate` sefely

lib/Server.js
```diff
@@ -644,14 +643,12 @@ class Server {
...
         this.listeningApp = https.createServer(options.https, app);
       } else {
-        if (!options.https.spdy) {
-          options.https.spdy = {
-            protocols: ['h2', 'http/1.1'],
-          };
-        }
+        options.https.spdy = {
+          protocols: ['h2', 'http/1.1'],
+        };
```
evilebottnawi: Unfortunately, it is breaking change, we can't remove `spdy` option in this version, spdy, how we should rewrite this part:
```
const isHttp2 = options.http2 || options.spdy
// Log deprecation `spdy` option.

if (semver.gte(process.version, '10.0.0') || !isHttp2) {
        this.listeningApp = https.createServer(options.https, app);
      } else {
        options.https.spdy = {
          protocols: ['http/1.1'],
        };

        // Allow to using https with `http1.1`
        if (isHttp2) { 
          options.https.spdy.protocols.unshift('h2');
        }

        /* eslint-disable global-require */
        // The relevant issues are:
        // https://github.com/spdy-http2/node-spdy/issues/350
        // https://github.com/webpack/webpack-dev-server/issues/1592
        this.listeningApp = require('spdy').createServer(options.https, app);
        /* eslint-enable global-require */
      }
```

lib/options.json
```diff
@@ -162,9 +162,6 @@
         }
       ]
     },
-    "spdy": {
-      "type": "boolean"
-    },
     "http2": {
```
evilebottnawi: Don't remove `spdy`, we just new output deprecation as i written above

lib/options.json
```diff
@@ -327,8 +324,7 @@
       "disableHostCheck": "should be {Boolean} (https://webpack.js.org/configuration/dev-server/#devserver-disablehostcheck)",
       "public": "should be {String} (https://webpack.js.org/configuration/dev-server/#devserver-public)",
       "https": "should be {Object|Boolean} (https://webpack.js.org/configuration/dev-server/#devserver-https)",
-      "spdy": "should be {Boolean} (https://webpack.js.org/configuration/dev-server/#devserver-spdy)",
-      "http2": "should be {Boolean} (https://webpack.js.org/configuration/dev-server/#devserver-spdy)",
+      "http2": "should be {Boolean} (https://webpack.js.org/configuration/dev-server/#devserver-http2)",
       "contentBase": "should be {Array} (https://webpack.js.org/configuration/dev-server/#devserver-contentbase)",
```
evilebottnawi: Don't remove, also we need create issue about documentation this (not related to PR right now, we should do this after merge)

lib/utils/createConfig.js
```diff
@@ -136,8 +136,8 @@ function createConfig(config, argv, { port }) {
     options.https = true;
   }
 
-  if (argv.spdy) {
-    options.spdy = true;
+  if (argv.http2) {
+    options.http2 = true;
```
evilebottnawi: Same as above

test/CreateConfig.test.js
```diff
@@ -559,28 +559,6 @@ describe('createConfig', () => {
...
-        devServer: { https: true, spdy: true },
-      }),
-      argv,
-      { port: 8080 }
-    );
-
-    expect(config).toMatchSnapshot();
-  });
-
   it('http2 option', () => {
```
evilebottnawi: Don't remove this test, just move this in `Http2` test

test/Http2.test.js
```diff
@@ -0,0 +1,99 @@
...
+const semver = require('semver');
+const helper = require('./helper');
+const config = require('./fixtures/contentbase-config/webpack.config');
+
+const contentBasePublic = path.join(
+  __dirname,
+  'fixtures/contentbase-config/public'
+);
+
+describe('HTTP2', () => {
```
evilebottnawi: `http2`

test/__snapshots__/CreateConfig.test.js.snap
```diff
@@ -494,6 +494,7 @@ exports[`createConfig http2 option 1`] = `
 Object {
   "hot": true,
   "hotOnly": false,
+  "http2": true,
   "https": true,
   "noInfo": true,
   "port": 8080,
```
evilebottnawi: Let's add `spdy` and `http2`

test/CreateConfig.test.js
Loonride: I think this would stay in `CreateConfig.test.js` since it is only confirming that the config option appears after passing through the config helper, right? `Http2` test should be for actual functionality of the server, not just confirming that the config options get through to the server

lib/Server.js
Loonride: @evilebottnawi I added the `spdy` option earlier in this PR, it wouldn't be a breaking change to remove it. Currently it looks like this:
```js
if (!options.https.spdy) {
        options.https.spdy = {
          protocols: ['h2', 'http/1.1'],
        };
      }

      if (semver.gte(process.version, '10.0.0')) {
        this.listeningApp = https.createServer(options.https, app);
      } else {
        /* eslint-disable global-require */
        // The relevant issues are:
        // https://github.com/spdy-http2/node-spdy/issues/350
        // https://github.com/webpack/webpack-dev-server/issues/1592
        this.listeningApp = require('spdy').createServer(options.https, app);
        /* eslint-enable global-require */
      }
    } else {
      this.listeningApp = http.createServer(app);
    }
```



Actually to avoid breaking changes, I think it should be like this:
```js
const isHttp2 = options.http2 || options.https.spdy;
if (options.https.spdy) {
// log deprecation warning, this means they provided it as part of https option
}

if (semver.gte(process.version, '10.0.0') || !isHttp2) {
        this.listeningApp = https.createServer(options.https, app);
      } else {

        if (!options.https.spdy) {
          options.https.spdy = {
            protocols: ['h2', 'http/1.1'],
          };
        }

        /* eslint-disable global-require */
        // The relevant issues are:
        // https://github.com/spdy-http2/node-spdy/issues/350
        // https://github.com/webpack/webpack-dev-server/issues/1592
        this.listeningApp = require('spdy').createServer(options.https, app);
        /* eslint-enable global-require */
      }
```

So basically, user should still be allowed to use this: `options.https.spdy`. It was never in the docs, but it was possible to use it before. The only breaking change now would be that they used to get http2 automatically if they were below Node 10.0.0. Now, they have to set `http2` to true to get http2.

Loonride:
@evilebottnawi It didn't pass CI for Node 6 because I used experimental `http2` module for the test added in Node 8. Would it be fine to skip that test below the version where `http2` was added, or should I find a new solution to test that HTTP/2 server works?

Loonride:
Made a few changes. As I said before, `options.spdy` did not exist before this PR, so I don't think we should add it if it will just instantly be considered deprecated. Instead I think the `options.https.spdy` should be described as deprecated, even though it was never in the docs.

lib/Server.js
jshado1: @evilebottnawi i think it would help if the `http2` option was nested under `https` because it makes it more obvious that http2 requires https. Perhaps even when only

```js
{
    https: {
        http2: true,
    },
}
```

Treat that as if both https (with self-signed cert) & http2 are enabled?

evilebottnawi: hm, http2 !== https2, but no one browsers don't support http2 without https, but for avoid misleading better contain 2 option (i think), new developers can start think what http2 === https, but i am open for feedback

lib/Server.js
```diff
@@ -630,7 +635,21 @@ class Server {
...
+
+      // note that options.spdy never existed. The user was able
+      // to set options.https.spdy before, though it was not in the
+      // docs. Keep options.https.spdy if the user sets it for
+      // backwards compatability, but log a deprecation warning.
+      if (options.https.spdy) {
+        this.log.warn(
+          'Providing custom spdy server options is deprecated and will be removed in the next major version.'
+        );
+      } else {
```
evilebottnawi: `spdy` should enable `h2` (don't do braking change, even when option doesn't exists in docs) 

Deprecation message is ok :+1:

lib/Server.js
```diff
@@ -630,7 +635,21 @@ class Server {
       options.https.key = options.https.key || fakeCert;
       options.https.cert = options.https.cert || fakeCert;
 
-      if (!options.https.spdy) {
+      // options.https.spdy is here for slightly better backwards compatability,
+      // since a user will probably provide this deprecated option if they
+      // are expecting the spdy server to be used.
+      const isHttp2 = options.http2 || options.https.spdy;
```
evilebottnawi: What do you think should we enable `http2` be default when user use `https: true` (without `http2` option) for better DX?

lib/Server.js
jshado1: Actually, I like [your suggestion below](#discussion_r267269542): assume `https: true` when `http2: true` and no https config is specified (and maybe output something in verbose logging about the assumption).

jshado1: Oh, duh. That's exactly what this code is doing 🤦‍♂️

lib/utils/createConfig.js
jshado1: Suggestion for a later/separate cleanup: There are a lot of `if (argv.foo) options.foo = true`. For the ones that are a simple copy:

```js
const simpleOverrides = _.pick(argv, [
  'http2',
  // other simple if argv then set options
]);

_.merge(options, simpleOverrides);
```

lib/Server.js
Aghassi: I would prefer this since h2 is the standard for most CDNs going forward (or should be). Would be good to mimic the greater movement across the web. I originally opened the issue because I wrongly assumed I was getting h2 by default.

lib/Server.js
Loonride: This is not a breaking change, if `options.https.spdy` is set by the user, it will not be changed at all currently: https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L650

lib/Server.js
Loonride: I changed it so that h2 is only prevented if http2 is explicitly set to false.

Loonride:
There should be no breaking changes at this point. I added another warning when someone tries to explicitly enable `http2`, but Node version is 10.0.0 is greater (where it does not work). I'm not sure what the second option is that @evilebottnawi was suggesting above. I think only `http2` option should be added for now.

I think what you meant to say earlier is https !== http2. What I interpret as the misleading thing you are mentioning is that if a user wants to use `http2`, they have to put their key/cert stuff into `options.https`. Maybe the solution is that `http2` should also take an object in the same way that `options.https` does. But I feel we should hold off on that until Express supports HTTP/2.

jshado1:
Putting the certs stuff inside `http2` would align with [node's http2](https://nodejs.org/api/http2.html#http2_server_side_example):

```js
const server = http2.createSecureServer({
  key: fs.readFileSync('localhost-privkey.pem'),
  cert: fs.readFileSync('localhost-cert.pem')
});
```
```js
{
  http2: {
    key:  // …
    cert: // …
  }
}
```

So then `https` could just relate to `http/1.1`, and maybe both could have a `true` option that generates & uses a self-signed cert.

But then what if someone includes both?

```js
{
  https: {
    key:  // …
    cert: // …
  },
  http2: true,
}
```

Throw and say use one or the other? In the above, you _could_ detect that `https` has certs and use those for `http2` instead of generating new ones (and then warn that `https` and `http2` options are mutually exclusive). That sounds like a can of worms though.

lib/Server.js

test/CreateConfig.test.js
evilebottnawi: :+1:

Issue #1734, opened 2019-03-23
Routes Test relies on compilation to finish, but doesn't wait for it


Loonride:
- Operating System: Windows 7
- Node Version: 8.10.0
- NPM Version: 6.4.1
- webpack Version: 4.29.0
- webpack-dev-server Version: 3.2.1

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

The request made to /webpack-dev-server by Routes.test.js relies on the webpack compilation to complete: https://github.com/webpack/webpack-dev-server/blob/master/test/Routes.test.js#L58

### Expected Behavior

Compilation should finish before the test starts. As stated here: https://github.com/webpack/webpack-dev-server/issues/847 the server callback can be called before compilation is done.

### Actual Behavior

If the compilation takes a long time, this test fails, since it does not wait for compilation to finish before testing.

### For Bugs; How can we reproduce the behavior?

Bring the test here: https://github.com/webpack/webpack-dev-server/blob/master/test/Routes.test.js#L58 to the top, and the test will probably run before compilation is done. The only reason it has been succeeding is because there are many tests before it, and compilation has time to finish. 

### Additional

I ran into this while working on a different fix which extended the compilation time. I will make a PR with the fix for this test. I think this is the only test where this mistake is made, but I'm not 100% sure.

evilebottnawi:
Fixed https://github.com/webpack/webpack-dev-server/pull/1735

Issue #1703, opened 2019-03-07
Using API doesn't add HMR plugin entry


rlamana:
Hot Module Replacement plugin entry doesn't seem to be added when using the API. The new tests on the PR #1664 use the dev server API via the `test/helper.js` and `addEntries` function in `lib/utils/addEntries.js` had to be manually called to make it work.

webpack Version: 4.29.0
webpack-dev-server Version: 3.1.14

- [x] This is a **bug**
- [ ] This is a **modification** request

### How can we reproduce the behavior?

Remove `addEntries` call in [test/Client.test.js](https://github.com/webpack/webpack-dev-server/pull/1664/files#diff-b2d4a9632f66a16163b3edfdf960185eR36) and the e2e test will fail because HMR is not loaded.

### Expected behavior

When using the API and `hot` is enabled, HMR plugin entry should be also added.

Loonride:
@evilebottnawi I messed with this issue briefly here: https://github.com/Loonride/webpack-dev-server/blob/hmr-entry-fix/lib/Server.js#L131. It would be nice if we had the ability to create the webpack compiler within Server.js. The problem I ran into was that users can use the compiler after they pass it in, so redefining it in Server.js does not work: https://github.com/webpack/webpack-dev-server/blob/master/test/Server.test.js#L25

evilebottnawi:
@Loonride yep, we should think how better solve this without break other code, feel free to experiment and send a PR

PR #1735, opened 2019-03-23
Fix Routes test and add new server creation helpers

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No, but the bug is in a test.

### Motivation / Use-Case

Solves this issue: https://github.com/webpack/webpack-dev-server/issues/1734.

I ran into this issue while creating a fix for this: https://github.com/webpack/webpack-dev-server/issues/1703. The fix I made for that problem extends the initial compilation time and makes the Routes test fail, so I thought it would be good to solve this before going further.

### Breaking Changes
None

### DIFF
--- a/test/Routes.test.js
+++ b/test/Routes.test.js
@@ -21,7 +21,7 @@ describe('Routes', () => {
 
   describe('without headers', () => {
     beforeAll((done) => {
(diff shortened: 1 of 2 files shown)


evilebottnawi:
Thanks!

PR #1738, opened 2019-03-23
HMR plugin insertion for hot option and inline entry insertion in Server

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes, added Hot.test.js to test addition of the HMR plugin using the Server API, added new test in Entry.test.js to check that it prevents duplicate entries on successive calls.

### Motivation / Use-Case

Fixes: https://github.com/webpack/webpack-dev-server/issues/1703

It uses the compiler provided to the server to add inline entries (if they don't exist already) and to apply the HMR plugin to the compiler when the hot option is provided.

Edit: also relates to https://github.com/webpack/webpack-dev-server/issues/616. I think it is impossible to avoid mutating the webpack config for this problem. It might be nice to add a feature where you can pass the webpack config into the Server, and then the Server itself creates the compiler.

### Breaking Changes

None

### Additional Info

It uses some slightly internal compiler functionality like compiler.options. I think it is all safe to use though.


Loonride:
Hmm I can't figure out why it's failing on OSX. Seems to be something to do with chromium writing files as a result of my new Hot tests.

lib/utils/addEntries.js
```diff
@@ -35,13 +36,22 @@ function addEntries(config, options, server) {
...
 
         return clone;
       }
 
-      return entries.concat(entry);
+      // in this case, entry is a string or an array.
+      // make sure that we do not add duplicates.
+      const entriesClone = entries.slice(0);
+      [].concat(entry).forEach((newEntry) => {
+        if (entriesClone.indexOf(newEntry) === -1) {
```
hiroppy: nit

```javascript
-  if (entriesClone.indexOf(newEntry) === -1) {
+  if (!entriesClone.includes(newEntry)) {
```

hiroppy:
Sorry, but I will check again later as I don't have enough time to check this pr. (a little busy)

hiroppy:
libuv is occurring errors on CI.

```
Assertion failed: (0), function uv__finish_close, file ../deps/uv/src/unix/core.c, line 280.
/Users/travis/.travis/functions: line 104:  3332 Abort trap: 6           npm run $JOB_PART
```

hiroppy:
I have this error at my local pc(mac).
node(v11.11.0)

```
 FAIL  test/Client.test.js (6.966s)
  ● Console

    console.info node_modules/http-proxy-middleware/lib/logger.js:77
      [HPM] Proxy created: /  ->  http://localhost:9001

  ● Client code › behind a proxy › requests websocket through the proxy with proper port 
number

    expect(received).toMatch(expected)

    Expected pattern: /^http:\/\/localhost:9000\/sockjs-node/
    Received string:  "http://localhost:9001/sockjs-node/info?t=1553426774616"

      61 |           .waitForRequest((requestObj) => requestObj.url().match(/sockjs-node/))
      62 |           .then((requestObj) => {
    > 63 |             expect(requestObj.url()).toMatch(
         |                                      ^
      64 |               /^http:\/\/localhost:9000\/sockjs-node/
      65 |             );
      66 |             browser.close().then(done);

      at toMatch (test/Client.test.js:63:38)

```
@Loonride Could you confirm this error?

Loonride:
Not sure how to solve the libuv CI problem, any suggestions? I was changing around various other tests because I suspect what I did is making two changes in behavior with existing tests:
1. The inline script was not actually inserted by default until now when using Server API.
2. Adding the inline script makes first compilations take longer than usual because I have to call the entry hook late in the compilation process.

lib/Server.js
```diff
@@ -89,6 +89,61 @@ class Server {
...
+        compilersWithoutHMR.forEach((compiler) => {
+          // addDevServerEntrypoints above should have added the plugin
+          // to the compiler options
+          const plugin = findHMRPlugin(compiler.options);
+          if (plugin) {
+            plugin.apply(compiler);
+          }
+        });
+      }
+    }
```
evilebottnawi: I think better create helper/util for this and move this code from `Server`

Loonride: @evilebottnawi moved to `utils/updateCompiler.js`

anchengjian:
It changed the order of my entries.

``` javascript
{
  entry: {
    index: ['src/polyfill.js', '(webpack)-dev-server/client?http://0.0.0.0:8081', '(webpack)/hot/only-dev-server.js', './index.tsx']
  }
}
```

My environment requires Polyfill to be run first, But after `updateCompiler`, my entries is

``` javascript
{
  entry: {
    index: ['(webpack)-dev-server/client?http://0.0.0.0:8081', '(webpack)/hot/only-dev-server.js', 'src/polyfill.js', './index.tsx']
  }
}
```
@Loonride @hiroppy @evilebottnawi

Ivaylo-Lafchiev:
I believe this PR has broken our use of the dev-server. When trying to view the site we now get the following error, which prevents the site from loading:

```
sockjs.js:678 Uncaught SyntaxError: The URL 'localhost.[company].com:20000:20000/sockjs-node' is invalid
    at new SockJS (sockjs.js:678)
    at initSocket (socket.js:9)
    at Object.eval (VM63 client:253)
    at eval (VM63 client:299)
    at Object.../node_modules/webpack-dev-server/client/index.js?localhost.[company].com:20000/ (index-webpack.js:1295)
    at __webpack_require__ (index-webpack.js:724)
    at fn (index-webpack.js:101)
    at eval (client:3)
    at Object.0 (index-webpack.js:4084)
    at __webpack_require__ (index-webpack.js:724) 
```

I believe this issue stems from the HMR entries that have been added, however I'm not entirely sure why the port is duplicated in that URL.
 
This error doesn't exist in webpack-dev-server@3.2.1. 
Tested across Node10/Npm 6, Win 7/10 and several permutations of Webpack 4/webpack-dev-server versions.

devServerConfig looks something like this: 
```
        publicPath: '/',
        contentBase: path.join(buildConfig.config.destDir, 'distro'),
        port: 20000,
        host: 'localhost.[company].com',
        hot: true,
        historyApiFallback: true,
        inline: true,
        quiet: false,
        noInfo: false,
        disableHostCheck: true,
        open: true,
```

Has anyone faced a similar issue or have any idea what the problem might be?

evilebottnawi:
`localhost.[company].com` is invalid url so it can't be parsed normally

Ivaylo-Lafchiev:
> `localhost.[company].com` is invalid url so it can't be parsed normally

apologies, [company] is just me redacting the url. The real URL does not contain brackets.

evilebottnawi:
Please open new issue with reproducible test repo

Issue #1743, opened 2019-03-27
Uncaught SyntaxError: The URL 'http:/sockjs-node' is invalid


hiroppy:
- Operating System: MacOS
- Node Version: 10.15.1
- NPM Version: 6.4.1
- webpack Version: 4.29.5
- webpack-dev-server Version: 3.2.1

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

no code

### Expected Behavior

no errors

### Actual Behavior

```
Uncaught SyntaxError: The URL 'http:/sockjs-node' is invalid
    at new SockJS (sockjs.js:678)
    at initSocket (socket.js:9)
    at Object.eval (client:227)
    at eval (client:273)
    at Object.../../../client/index.js?http: (bundle.js:97)
    at __webpack_require__ (bundle.js:20)
    at eval (client:1)
    at Object.0 (bundle.js:396)
    at __webpack_require__ (bundle.js:20)
    at bundle.js:84
```

### For Bugs; How can we reproduce the behavior?

```sh
$ cd examples/api/middleware 
$ node server.js
```

And then you can see the error in browser's console.
Occurs at `api/middleware` and `api/simple`.

### Info

related https://github.com/webpack/webpack-dev-server/pull/1741

PR #1749, opened 2019-03-30
fix:Uncaught SyntaxError: The URL 'http:/sockjs-node' is invalid

EslamHiko:
- [ x] This is a **bugfix**
- [ ] This is a **feature**

Tests already exist

### Motivation / Use-Case
to solve this issue : https://github.com/webpack/webpack-dev-server/issues/1743

### Breaking Changes
I just replaced `socketUrl` with '/sockjs-node' and removed unused code.

### Additional Info
I'm new to open source contributing if there is anything wrong about my pull request or any advice please tell me I'll really appreciate it.

### DIFF
--- a/client-src/default/index.js
+++ b/client-src/default/index.js
@@ -2,7 +2,6 @@
 
 /* global __resourceQuery WorkerGlobalScope self */
 /* eslint prefer-destructuring: off */
-const querystring = require('querystring');
 const url = require('url');
 const stripAnsi = require('strip-ansi');
 const log = require('loglevel').getLogger('webpack-dev-server');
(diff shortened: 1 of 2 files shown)


Loonride:
Oops, it looks like we were working on the same fix at the same time https://github.com/webpack/webpack-dev-server/pull/1750. The issue with your fix is that it disregards the `sockPath` option, which allows users to change that path from `/sockjs-node` to something else https://webpack.js.org/configuration/dev-server/#devserversockpath. Also, you should not ignore if the user requests a specific hostname, nor should you ignore if a user wants https. That is what the block of code is doing that you removed. Be sure to consider other option cases, even if it works in one particular case!

This does reveal that more tests are needed, since your changes passed CI. Maybe something more is needed here https://github.com/webpack/webpack-dev-server/blob/master/test/Socket.test.js or here https://github.com/webpack/webpack-dev-server/blob/master/test/Client.test.js to confirm that the client actually connects to the sockPath that the server provides.

EslamHiko:
@Loonride Thank you, I understand now what was wrong....  
about the testing part, I noticed here: https://github.com/webpack/webpack-dev-server/blob/master/test/Socket.test.js
at line 27 `req.get('/sockjs-node').expect(200, done);` should it be `req.get(server.sockPath).expect(200, done);` or should i add something else ?

Loonride:
That test works how it should, since it is confirming that the default behavior of the server is to use `/sockjs-node` when the the `sockPath` option is not provided. In the tests, hard coding things like that is fine, because the goal of the tests is to confirm that the server is following some strict behavior, and if it is not, the test should fail.

I added a test in my PR which confirms that the client connects to the right sockPath: https://github.com/webpack/webpack-dev-server/pull/1750/files#diff-b2d4a9632f66a16163b3edfdf960185e. The test runs a browser, then waits for the simulated client to make a request to the correct hostname and sockPath.

It may be redundant at this point for you to fix your PR, but I recommend you look at the test above and make sure you understand why what you wrote would not pass that test. If you have any other questions, I would be happy to answer them!

EslamHiko:
@Loonride I got the idea about what the test is testing that was really helpful for me, I'm going to learn more about tests to get a full understanding of exactly what's going on in the test.
I really appreciate your comments, Thank you.

EslamHiko:
@evilebottnawi Yes sure no problem at all 😄

Issue #1745, opened 2019-03-28
Option to disable serveIndex middleware


jamesknelson:
- Operating System: Mac OS 10.14.3
- Node Version: 11.1.0
- NPM Version: 6.7.0
- webpack Version: 4.29.6
- webpack-dev-server Version: 3.2.1

- [ ] This is a **bug**
- [x] This is a **modification** request

### Code

It'd be useful to be able to disable the express serve-index middleware, which generates directory listings when you view a directory that doesn't have an index.html file.

```js
// webpack.config.js
module.exports = {
  // ...
  devServer: {
    serveIndex: false,
  }
}
```

### For Features; What is the motivation and/or use-case for the feature?

This option is required to be able to do SSR for the root page with webpack-dev-server in an `after()` middleware using the `serverSideRender` option, as otherwise an automatically-generated index page will be shown instead.

I'd be happy to put a PR together for this change if it's likely to be accepted.

evilebottnawi:
@jamesknelson PR welcome :+1:

EslamHiko:
hello I want to work on this feature 
as I understand 
this option will disable `serveIndex` 
in server.js
```
contentBaseIndex: () => {
        if (Array.isArray(contentBase)) {
          contentBase.forEach((item) => {
              app.get('*', serveIndex(item));
          });
        } else if (
          !/^(https?:)?\/\//.test(contentBase) &&
          typeof contentBase !== 'number'
        ) {
            app.get('*', serveIndex(contentBase));
        }
      }
```
the new code will be
```
contentBaseIndex: () => {
   if(options.devServer.serveIndex){
        if (Array.isArray(contentBase)) {
          contentBase.forEach((item) => {
              app.get('*', serveIndex(item));
          });
        } else if (
          !/^(https?:)?\/\//.test(contentBase) &&
          typeof contentBase !== 'number'
        ) {
            app.get('*', serveIndex(contentBase));
        }
     }
   }
```
instead of directory listing the files in folders without `index.html` it'll show the message `Cannot GET /folder/` 

so am I getting it right?
and one more thing I don't know how to add a new option? please, can you tell me how would I add the new option?

excuse me I'm new to WebPack.

evilebottnawi:
Look on other options and just send a PR

EslamHiko:
@evilebottnawi I added the new option and I tested it on one of the examples in the `examples/api/simple` folder
[...]

PR #1752, opened 2019-04-01
Feature: add option "serveIndex" to enable/disable serveIndex middleware

EslamHiko:
- [ ] This is a **bugfix**
- [x] This is a **feature**

I used the examples in the example folder

### Motivation / Use-Case
add this feature https://github.com/webpack/webpack-dev-server/issues/1745

### Breaking Changes
it disables/enables serveIndex


EslamHiko:
if I got it right

I tried to create a test 
the logic of the test is very simple 
` 1 - create a folder with a simple file in it`
` 2 - set the option to true and check if it's listing the files in the directory (status:200)`
` 3 - set the option to false and check if it's not listing the files in the directory (status:404)`
` 4 - remove the folder`
I managed to create the folder and remove it but whenever I test `/folder` route it returns `status:301 Moved Permanently` if it exists or not.
I need some help here to implement the test.

EslamHiko:
@evilebottnawi here is a minimum reproducible test repo : https://github.com/EslamHiko/webpack-dev-server/tree/serve-index-option-test
`True scenarios :` 
`default settings testing`
`1 - go to examples/api/simple`
`2- node server`
`3- go to http://localhost:8080/folder/ it'll directory list simple.txt`
`serveIndex:true testing`
`1- edit webpack.config.js in the examples/api/simple folder`
`2- change the settings like here:`
```
module.exports = setup({
  context: __dirname,
  entry: ['./app.js', '../../../client/index.js?http://localhost:8080/'],
  output: {
    filename: 'bundle.js',
  },
  devServer:{
    serveIndex:true
  }
});
```
`3- node server`
`4- go to http://localhost:8080/folder/ it'll directory list simple.txt`

`False scenarios :`
`1- edit webpack.config.js in the examples/api/simple folder`
`2- change the settings like here:`
```
module.exports = setup({
  context: __dirname,
  entry: ['./app.js', '../../../client/index.js?http://localhost:8080/'],
  output: {
    filename: 'bundle.js',
  },
  devServer:{
    serveIndex:false
  }
});
```
`3- node server`
`4- go to http://localhost:8080/folder/ it'll show you the message Cannot GET /folder/`

I wanted to write a jest test but i got `status:301 Moved Permanently` error because tests run in the root folder.
if there is any ideas to create a jest test for it i'll be so happy to know about them.

lib/Server.js
```diff
@@ -555,7 +553,10 @@ class Server {
 
     defaultFeatures.push('magicHtml');
 
-    if (contentBase !== false) {
+    if (
+      contentBase !== false &&
+      (options.serveIndex || typeof options.serveIndex === 'undefined')
```
evilebottnawi: Don't use `typeof`, we prefer using `!`

EslamHiko:
```resolved
      (options.serveIndex || options.serveIndex === undefined)
```
@evilebottnawi done

lib/Server.js
```diff
@@ -555,7 +555,7 @@ class Server {
 
     if (
       contentBase !== false &&
-      (options.serveIndex || typeof options.serveIndex === 'undefined')
+      (options.serveIndex || options.serveIndex === undefined)
```
evilebottnawi: Just simplify:
```
const shouldHandleServeIndex = contentBase || options.serveIndex
```

`serveIndex` by default should be true

EslamHiko: @evilebottnawi i've simplified the condition & i added a comment to elaborate why i'm comparing with `undefined` 
```
          if (contentBase !== false) {
    // checking if it's set to true or not set (Default : undefined => true)
    options.serveIndex = options.serveIndex || options.serveIndex === undefined;

    const shouldHandleServeIndex = contentBase && options.serveIndex;

    if (shouldHandleServeIndex) {
      defaultFeatures.push('contentBaseIndex');
    }

```

lib/Server.js
```diff
@@ -553,12 +551,15 @@ class Server {
 
     defaultFeatures.push('magicHtml');
 
-    if (
-      contentBase !== false &&
-      (options.serveIndex || options.serveIndex === undefined)
-    ) {
+    // checking if it's set to true or not set (Default : undefined => true)
+    options.serveIndex = options.serveIndex || options.serveIndex === undefined;
```
evilebottnawi: Better avoid modify `options` because code below also can use options and modification original options is not good idea better keep this in separate variable, it is good practice.
I think we can rewrite this to:
```
const serveIndexValue = options.serveIndex ? options.serveIndex : true;
```

If it is `null/undefined` we use `true` otherwise we get original value.

EslamHiko: @evilebottnawi thank you for the advice, I set the option value to a separate variable `this.serveIndex`

EslamHiko:
@evilebottnawi here is a minimum reproducible test repo with the new code: https://github.com/EslamHiko/webpack-dev-server/tree/serve-index-option-test2

testing is the same as this comment :

> @evilebottnawi here is a minimum reproducible test repo : https://github.com/EslamHiko/webpack-dev-server/tree/serve-index-option-test
> `True scenarios :`
> `default settings testing : `
> `1 - go to examples/api/simple`
> `2- node server`
> `3- go to http://localhost:8080/folder/ it'll directory list simple.txt`
> 
> `serveIndex:true testing : `
> `1- edit webpack.config.js in the examples/api/simple folder`
> `2- change the settings like here:`
> 
> ```
> module.exports = setup({
>   context: __dirname,
>   entry: ['./app.js', '../../../client/index.js?http://localhost:8080/'],
>   output: {
>     filename: 'bundle.js',
>   },
>   devServer:{
>     serveIndex:true
>   }
> });
> ```
> `3- node server`
> `4- go to http://localhost:8080/folder/ it'll directory list simple.txt`
> 
> `False scenarios :`
> `1- edit webpack.config.js in the examples/api/simple folder`
> `2- change the settings like here:`
> 
> ```
> module.exports = setup({
>   context: __dirname,
>   entry: ['./app.js', '../../../client/index.js?http://localhost:8080/'],
>   output: {
>     filename: 'bundle.js',
>   },
>   devServer:{
>     serveIndex:false
>   }
> });
> ```
> `3- node server`
> `4- go to http://localhost:8080/folder/ it'll show you the message Cannot GET /folder/`
> 
> I wanted to write a jest test but i got `status:301 Moved Permanently` error because tests run in the root folder.
> if there is any ideas to create a jest test for it i'll be so happy to know about them.

if there is any way to use jest to browse folders it'll be great to create a jest test for it.

evilebottnawi:
We have test for contentBase, so test should be same

EslamHiko:
@evilebottnawi I added valid tests testing True,False & default settings, you wouldn't blieve what caused the `301 Error` i was doing `req.get('/assets')` instead of `req.get('/assets/')` that what prevented me from submitting my first test 😅 in the end it's working now without any issues.

test/ContentBase.test.js
```diff
@@ -64,7 +64,81 @@ describe('ContentBase', () => {
...
+
+    afterAll((done) => {
+      helper.close(() => {
+        done();
+      });
+    });
+
+    it('should list the files inside the assets folder (200)', (done) => {
+      req.get('/assets/').expect(200, done);
+    });
```
hiroppy: Please add the following test here because it is confirmed in `serveIndex:false`.

```javascript
    it('should show Heyo. because bar has index.html inside it (200)', (done) => {
      req.get('/bar/').expect(200, /Heyo/, done);
    });

```

test/ContentBase.test.js
```diff
@@ -64,7 +64,81 @@ describe('ContentBase', () => {
...
+  });
```
hiroppy: ditto

EslamHiko:
@hiroppy Done

EslamHiko:
@hiroppy You're Welcome 😄

Issue #1753, opened 2019-04-01
Upgrade to webpack-cli v.3.3.x


evenstensberg:
# Just a heads up

When doing this, remember to update the location of `config-yargs.js`, it has changed. Maybe some other sources has changed too, so check where you import directly from the library.

EslamHiko:
Hello @evenstensberg 
I did a PR to update the paths of both config-yargs & convert-argv
I upgraded webpack-cli to the master branch and I run `npm test cli.test` on my machine and it passes the test however in AppVeyor testing it doesn't pass the test 
on my machine, if I returned the paths to their origin it won't pass the test
I tried to upgrade webpack to it's latest version but it's still not passing the test 
I got an idea to check if the new module with the new path exists or not and if it doesn't exist fall back to the original path so it can work with all versions v3.3.x and v3.2.x but I got that error `85:1   error  Calls to require() should use string literals  import/no-dynamic-require`
so what did I miss here?

PR #1754, opened 2019-04-03
Upgrade webpack-cli to v3.3x

EslamHiko:
- [ ] This is a **bugfix**
- [ ] This is a **feature**
- [x ] This is a **code refactor**

NO

### Motivation / Use-Case
make webpack-dev-server work with webpack-cli v3.3.x 
https://github.com/webpack/webpack-dev-server/issues/1753

### Breaking Changes
Changed the paths of convert-argv & config-yargs

### Additional Info
NO

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,8 +88,8 @@
     "style-loader": "^0.23.1",
     "supertest": "^3.0.0",
     "url-loader": "^1.1.1",
-    "webpack": "^4.29.0",
-    "webpack-cli": "^3.2.1",
+    "webpack": "^4.29.*",
+    "webpack-cli": "github:webpack/webpack-cli#master",
     "ws": "^6.0.0"
   },
   "peerDependencies": {
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -75,7 +75,7 @@ yargs.usage(
 );
 
 // eslint-disable-next-line import/no-extraneous-dependencies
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -88,8 +88,8 @@
     "style-loader": "^0.23.1",
     "supertest": "^3.0.0",
     "url-loader": "^1.1.1",
-    "webpack": "^4.29.0",
-    "webpack-cli": "^3.2.1",
+    "webpack": "^4.29.6",
+    "webpack-cli": "github:webpack/webpack-cli#master",
```
evilebottnawi: It is break webpack when developers don't have `git`

evilebottnawi:
@evenstensberg can we don't change location right now? Or just create file with import/export, it is really looks like breaking change and in theory can break not only webpack-dev-server

EslamHiko:
@evilebottnawi I got the idea to check if the new module with the new path exists or not and if it doesn't exist fall back to the original path so it can work with all versions v3.3.x and v3.2.x but I got that `error 85:1 error Calls to require() should use string literals import/no-dynamic-require` when committing 
I can create a pull request but we will have to disable `import/no-dynamic-require` middleware

evilebottnawi:
@EslamHiko just add `// eslint-disable-next-line import/no-dynamic-require` and add comment what in `webpack-cli@3.3 change path to ...`

EslamHiko:
@evilebottnawi you're Welcome 😄

PR #1760, opened 2019-04-04
feat: add webpack as argument to before and after options

evilebottnawi:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

Allow to get information from webpack in `before` and `after`

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -497,7 +497,7 @@ class Server {
       },
       before: () => {
         if (typeof options.before === 'function') {
-          options.before(app, this);
+          options.before(app, this, compiler);
         }
       },
       middleware: () => {
@@ -507,7 +507,7 @@ class Server {
       },
       after: () => {
         if (typeof options.after === 'function') {
-          options.after(app, this);
+          options.after(app, this, compiler);
         }
       },
       headers: () => {
--- a/test/BeforeAndAfter.test.js
+++ b/test/BeforeAndAfter.test.js
@@ -1,0 +1,76 @@
+'use strict';
+
+const request = require('supertest');
+const helper = require('./helper');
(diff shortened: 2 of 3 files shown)


hiroppy:
Which do you prefer modifying webpack-dev-server's eslint config or eslint-config-webpack?

evilebottnawi:
@hiroppy `eslint-config-webpack` https://github.com/webpack-contrib/eslint-config-webpack

hiroppy:
@evilebottnawi eslint-config-prettier overwrite this rule.

I found that eslint-config-prettier was overriding the rules.(curly: off)

So, we have to change the order https://github.com/webpack/webpack-dev-server/blob/master/.eslintrc#L2 or write this rule to webpack-dev-server's eslintrc.

evilebottnawi:
hm, it is strange, need investigate why prettier disable `curly`, because prettier don't add `{` and `}`

hiroppy:
eslint-config-prettier set 0 😞

https://github.com/prettier/eslint-config-prettier/blob/dd8511b107e5adaaa107a2f8d83d9298018e895b/index.js#L9

evilebottnawi:
@hiroppy https://github.com/prettier/eslint-config-prettier/blob/dd8511b107e5adaaa107a2f8d83d9298018e895b/README.md#curly, we need change order, yep

hiroppy:
Yeah, but we are setting `"extends": ["webpack", "prettier"]` now, so the setting of eslint-config-webpack is overwritten because the setting of curly is written in eslint-config-prettier

evilebottnawi:
@hiroppy we need fix :+1:

hiroppy:
@evilebottnawi for now, I created #1761.

Issue #1724, opened 2019-03-18
Cannot set property 'wasm' of undefined. Mismatched `mime` versions?


tannerlinsley:
- Operating System: MacOS 10.14
- Node Version: 10.5.0
- NPM Version: 6.4.1
- webpack Version: 4.29.6
- webpack-dev-server Version: 3.2.1

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

```bash
TypeError: Cannot set property 'wasm' of undefined
  
  - Server.js:160 new Server
    [basic]/[webpack-dev-server]/lib/Server.js:160:36
  
  - index.js:248 _callee7$
    ../../../src/static/webpack/index.js:248:15
  
  - runtime.js:62 tryCatch
    [basic]/[regenerator-runtime]/runtime.js:62:40

```

This is currently in a local WIP branch of [React Static](https://react-static.js.org) v7.
This bug may be related to #1580?

The line of code `[basic]/[webpack-dev-server]/lib/Server.js:160:36` is referring to is `https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L160`

### Expected Behavior
The dev server should start without errors. 

### Actual Behavior
When attempting to start the dev server, `Cannot set property 'wasm' of undefined` is logged and the process hangs.

### For Bugs; How can we reproduce the behavior?
Working on a reproduction link, will edit to include asap.

evilebottnawi:
Yep, reproducible test repo will be great

tannerlinsley:
I realize this is extremely involved, so I admire your patience should you choose to help me with this. React Static is no walk-in-the-park, but here we go!

- `git clone https://github.com/nozzle/react-static.git`
- Checkout the `v7` branch
- Run `yarn`
- Run `yarn link`
- Run `yarn start` to compile
- In `packages/react-static/templates/basic`
  - Run `yarn`
  - Run `yarn link react-static`
  - Run `yarn start`
  - The error is displayed

Given that this is an alpha branch, any number of things could go wrong, so if you hit a roadblock, let me know and I'll do my best to guide you to the issue.

Points of interest:
- https://github.com/nozzle/react-static/tree/v7/packages/react-static/src/static/webpack contains all of our webpack files. In this case, `webpack.dev.js` is what is being used to compile.

evilebottnawi:
Thanks for repo, i will see on this in near future, anyway PR welcome

tannerlinsley:
Thanks.

[...]

PR #1765, opened 2019-04-05
fix(Server): validate express.static.mime.types

hiroppy:
fixes: #1724

- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes.

A test which Checks `application/wasm` already exists.

https://github.com/webpack/webpack-dev-server/blob/master/test/ContentBase.test.js#L301

### Motivation / Use-Case

Validate express object.

### Breaking Changes

no

### Additional Info

no

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -89,12 +89,12 @@ class Server {
     if (options.lazy && !options.filename) {
       throw new Error("'filename' option must be set in lazy mode.");
     }
-    
+
     // if the user enables http2, we can safely enable https
     if (options.http2 && !options.https) {
       options.https = true;
     }
-    
+
     updateCompiler(compiler, options);
 
     this.stats =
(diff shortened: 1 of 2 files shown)


lib/Server.js
```diff
@@ -174,7 +174,14 @@ class Server {
 
     // ref: https://github.com/webpack/webpack-dev-server/issues/1575
     // remove this when send@^0.16.3
-    express.static.mime.types.wasm = 'application/wasm';
+    if (
+      express.static &&
+      express.static.mime &&
+      express.static.mime.types &&
+      !express.static.mime.types.wasm
```
evilebottnawi: I think better to remove `!express.static.mime.types.wasm` because in some old cases it can be `text/html` and it is invalid, we need use right mime for `wasm`

hiroppy:
@evilebottnawi PTAL 🙏

hiroppy:
Fixed.

evilebottnawi:
God job!

PR #1766, opened 2019-04-05
test: compress option

evilebottnawi:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

yep

### Motivation / Use-Case

Avoid skip test

### Breaking Changes

No

### Additional Info

/cc @hiroppy I think we should enable `compress` by default in next major for better DX, 99% developers set it to true

### DIFF
--- a/test/Compress.test.js
+++ b/test/Compress.test.js
@@ -7,29 +7,77 @@
 
 const request = require('supertest');
 const helper = require('./helper');
-const config = require('./fixtures/simple-config/webpack.config');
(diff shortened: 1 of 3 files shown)


evilebottnawi:
Added to https://github.com/webpack/webpack-dev-server/projects/1

hiroppy:
@evilebottnawi  I want to know how much the build speed, etc. will be affected if enabling `compress`. (if we make `compress` a default)

PR #1774, opened 2019-04-08
Only add entries after compilers have been created

DylanPiercey:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No...

### Motivation / Use-Case

Currently `addEntries` is called twice when using the `cli`. First it is called [here](https://github.com/webpack/webpack-dev-server/blob/master/bin/webpack-dev-server.js#L143) in the cli bin, and then always called again when we create the dev server instance [by calling updateCompiler](https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L98) then [here](https://github.com/webpack/webpack-dev-server/blob/master/lib/utils/updateCompiler.js#L47).

This extra call is not needed, and actually causes an issue with my specific setup. In my Webpack config the `entry` is set dynamically via a plugin once the plugin has initialized. Since the first call to `addEntries` is ran on the raw config before any plugins run it is inaccurate and thinks that I have no entries, causing some issues. This would cause issues for any plugin which mutates the entries assuming it is used with the `webpack-dev-server` cli.

### Additional Info

If a test case is required for this then I'd be happy to work with someone to figure out the best way to integrate such a test into this suite. It looks like the CLI tests are fairly basic and separated from the server  tests.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -20,7 +20,6 @@ const webpack = require('webpack');
 const options = require('./options');
 const Server = require('../lib/Server');
 
-const addEntries = require('../lib/utils/addEntries');
 const colors = require('../lib/utils/colors');
 const createConfig = require('../lib/utils/createConfig');
 const createDomain = require('../lib/utils/createDomain');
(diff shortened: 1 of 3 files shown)


hiroppy:
Oops, I'll fix `delimiter` for windows.

hiroppy:
@evilebottnawi PTAL;)

PR #1441, opened 2018-07-11
feat(Server): exclude client scripts for `node` targets

EloB:
```bash
git clone git@github.com:EloB/ssr.git
cd ssr/node_modules/webpack-dev-server # <- This is using my branch
npm install # Install all dependencies
npm run prepublish # To build all client scripts
cd ../../
npm run dev
```

### Type 

- [x] Feature

### Issues

- Fixes #1256 
- Fixes #1338

### SemVer

- [x] Minor

PR #1338, opened 2018-03-08
feat(addDevServerEntryPoints): exclude client scripts for`node` targets

sanonz:
### Type 

- [x] Feature

### Issues

- Fixes #1256 
- Fixes #1441

### SemVer

- [x] Minor

Issue #1256, opened 2018-01-05
webpack Multiple Targets


sanonz:
### Code

```js
// webpack.config.js
module.exports = [
  {
    entry : {
      app: './src/app'
    },
    output: {
      path: path.resolve(__dirname, 'dist/')
      filename: 'assets/js/app.js',
      chunkFilename: `assets/js/${getAssetName('[hash]', '[name]')}.js`
    },
    devServer: {
      hot: true
    }
  },
  {
    entry : {
      app: './src/app.server'
    },
    output: {
      path: path.resolve(__dirname, 'dist/server')
      filename: 'app.js',
      chunkFilename: '[name].js',
      libraryTarget: 'commonjs2'
    },
    target: 'node'
  }
];
```
### Run
```shell
npx webpack-dev-server
```

### Expected Behavior
- ~~_The client app entry add `webpack/hot/only-dev-server` `webpack/hot/dev-server` etc._~~
- The client app entry add https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/index.js file, from at line: https://github.com/webpack/webpack-dev-server/blob/master/lib/util/addDevServerEntrypoints.js#L23
- The server app entry no add.

### Actual Behavior
All added.

### For Bugs; How can we reproduce the behavior?
Example edit: https://github.com/webpack/webpack-dev-server/blob/master/lib/util/addDevServerEntrypoints.js#L22
```diff
    [].concat(webpackOptions).forEach((wpOpt) => {
+     if (wpOpt.target === 'node') return;
      if (typeof wpOpt.entry === 'object' && !Array.isArray(wpOpt.entry)) {
        Object.keys(wpOpt.entry).forEach((key) => {
          wpOpt.entry[key] = devClient.concat(wpOpt.entry[key]);
        });
      } else if (typeof wpOpt.entry === 'function') {
        wpOpt.entry = wpOpt.entry(devClient);
      } else {
        wpOpt.entry = devClient.concat(wpOpt.entry);
      }
    });
```

### For Features; What is the motivation and/or use-case for the feature?
server side rendering

sanonz:
Turned out to be such, no add:

// webpack.config.js
```diff
{
+    entry: (devClient) => {
+      return {
+        app:  './src/setup.server',
+      };
+    },
    output: {
      path: path.resolve(__dirname, 'dist/server')
      filename: 'app.js',
      chunkFilename: '[name].js',
      libraryTarget: 'commonjs2'
    },
[...]

PR #1775, opened 2019-04-09
fix(addEntries): only add client entry to web targets

mlrawlings:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

When using an universal/isomorphic multicompiler setup and using the default setting of `inline:true`, the client runtime gets injected into the entries for each compilation.  However, as far as I can tell, the client runtime is only designed to work in the browser (`target:"web"` or `target:"webworker"`).  

With `target:"node"`, the following error is thrown:

```
urlParts.port = self.location.port;
                ^
ReferenceError: self is not defined
```

From

https://github.com/webpack/webpack-dev-server/blob/99e8db0f8cc441bb8f6acc60d786e33c38b8fc1d/client-src/default/index.js#L46

### Changes

This PR updates `addEntries` to check the compilation target for each compiler.  Since the client runtime is only designed for web targets, it is only prepended to those compilations.  The HMR runtimes are still prepended for all compile targets (when `options.hot` or `options.hotOnly` are set), because as far as I can tell, these should work regardless of the compilation target.

### Breaking Changes

None, if my assumptions are correct.

### Additional Info

Related: #1256 #1338 #1441

### DIFF
--- a/lib/utils/addEntries.js
+++ b/lib/utils/addEntries.js
@@ -18,26 +18,31 @@ function addEntries(config, options, server) {
     const domain = createDomain(options, app);
     const sockPath = options.sockPath ? `&sockPath=${options.sockPath}` : '';
     const sockPort = options.sockPort ? `&sockPort=${options.sockPort}` : '';
-    const entries = [
-      `${require.resolve('../../client/')}?${domain}${sockPath}${sockPort}`,
-    ];
+    const clientEntry = `${require.resolve(
+      '../../client/'
+    )}?${domain}${sockPath}${sockPort}`;
+    let hotEntry;
+
     if (options.hotOnly) {
-      entries.push(require.resolve('webpack/hot/only-dev-server'));
+      hotEntry = require.resolve('webpack/hot/only-dev-server');
     } else if (options.hot) {
-      entries.push(require.resolve('webpack/hot/dev-server'));
+      hotEntry = require.resolve('webpack/hot/dev-server');
     }
 
-    const prependEntry = (entry) => {
-      if (typeof entry === 'function') {
-        return () => Promise.resolve(entry()).then(prependEntry);
+    const prependEntry = (originalEntry, additionalEntries) => {
+      if (typeof originalEntry === 'function') {
+        return () =>
+          Promise.resolve(originalEntry()).then((entry) =>
+            prependEntry(entry, additionalEntries)
+          );
       }
 
-      if (typeof entry === 'object' && !Array.isArray(entry)) {
+      if (typeof originalEntry === 'object' && !Array.isArray(originalEntry)) {
         const clone = {};
 
-        Object.keys(entry).forEach((key) => {
+        Object.keys(originalEntry).forEach((key) => {
           // entry[key] should be a string here
-          clone[key] = prependEntry(entry[key]);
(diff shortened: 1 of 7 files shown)


lib/utils/addEntries.js
```diff
@@ -56,7 +60,17 @@ function addEntries(config, options, server) {
 
     // eslint-disable-next-line no-shadow
     [].concat(config).forEach((config) => {
-      config.entry = prependEntry(config.entry || './src');
+      const webTarget =
+        config.target === 'web' ||
+        config.target === 'webworker' ||
+        config.target == null;
```
evilebottnawi: I am afraid what some developers can have very difficult setup (example testing hmr using jsdom and using webpack, it is very weird, but can be in theory). Maybe we can implement option to allow developers filter configurations (by default modify configuration only for `web`/`webworker` and null)

mlrawlings: Addressed in latest commit

mlrawlings:
@evilebottnawi I see your concern and I also realized the target can also be a [function](https://webpack.js.org/configuration/target/#function) which complicates things.

I've just pushed a commit that adds two options (plus tests): `inlineClient` and `inlineHot`.  These options control whether, when `inline` is true, which configurations get the client runtime and hot runtimes inlined.

```js
new DevServer(compiler, {
  /* default: web targets only */
  inlineClient: (config) => (
    config.target === 'web' ||
    config.target === 'webworker' ||
    config.target == null
  ) 
});
```

```js
new DevServer(compiler, {
  /* add to all compilers */
  inlineClient: true 
});
```

```js
new DevServer(compiler, {
  /* add to no compilers */
  inlineClient: false 
});
```

This would allow users to choose to inline based on any config value.  The [name](https://webpack.js.org/configuration/other-options/#name) option in particular could be useful to target specific compilations.

```js
new DevServer({
  /* add to all compilers (as a function) */
  inlineClient: (config) => config.name !== 'Server' 
});
```

evilebottnawi:
Looks good, but `inlineClient` and `inclideHot` is very misleading, maybe `injectClient` and `injectHot`?

mlrawlings:
Updated.  I like `inject` better as well.

hiroppy:
@mlrawlings Could you rebase?

hiroppy:
BTW, I'll add an example for this after this pr is merged.

evilebottnawi:
Thanks great job!

EECOLOR:
Awesome, it took a few hours to see this was the problem, but I was very happy to see it has already been solved and merged. Any indication on when this will be released?

Thanks!

For future reference, the error I received:

```
 ReferenceError: self is not defined
```

EECOLOR:
Does it make sense to still have the `HotModuleReplacementPlugin` injected for non web targets?

FallingSnow:
I'm running into this issue while using web audio workers at
https://github.com/webpack/webpack-dev-server/blob/540c43852ea33f9cb18820e1cef05d5ddb86cc3e/client-src/index.js#L149-L151
```js
ReferenceError: self is not defined
```

I have tried the following in my webpack config but the issue remains.
```js
output: {
  globalObject: "globalThis"
}
```

I also tried setting the target to `webworker` but the error persists.

snitin315:
Cam you please create an issue with proper steps to reproduce?

PR #1794, opened 2019-04-12
exclude /node_modules/ from _watch by default

### LINKED ISSUE #1781: exclude `node_modules` by default in `_watch` for content base
- Operating System: no matter
- Node Version: no matter
- NPM Version: no matter
- webpack Version: no matter
- webpack-dev-server Version: 3.3.0

- [x] This is a **bug**

### Code

no matter

### Actual Behavior

Looks it was bug also in previous webpack-dev-server version, but we watch only on depth, no we relax this so some developers get problem with watching and fs

https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L301

Just run webpack-dev-server without configuration

EslamHiko:
- [x] This is a **bugfix**
- [ ] This is a **feature**

I'm not sure how exactly to write a test for this except testing it manually.

### Motivation / Use-Case
fixes #1781

### Breaking Changes
No

### Additional Info
I had a very strange behavior when I try to change the value of `this.watchOptions` when I do 
```
this.watchOptions = options.watchOptions || {};
this.watchOptions.ignored = this.watchOptions.ignored || /node_modules/;
```
chokidar always reads the ignored option as undefined even if i try to edit the value of this.watchOptions just right before setting it's options.
```
this.watchOptions.ignored = this.watchOptions.ignored || /node_modules/;

const options = {
      ignoreInitial: true,
      persistent: true,
      followSymlinks: false,
      atomic: false,
      alwaysStat: true,
      ignorePermissionErrors: true,
      ignored: this.watchOptions.ignored,
      usePolling,
      interval,
    };
```
it works only with me the way I did the change
So what's wrong about it? and if I can write a test for this how would I test it.


lib/Server.js
```diff
@@ -115,7 +115,14 @@ class Server {
 
     this.sockets = [];
 
-    this.watchOptions = options.watchOptions || {};
+    if (!options.watchOptions) {
+      options.watchOptions = {};
+    }
+    // ignoring node_modules folder by default
+    options.watchOptions.ignored =
+      options.watchOptions.ignored || /node_modules/;
```
evilebottnawi: I think we need use here `[/node_modules/]`. also need add tests to avoid future regressions

EslamHiko:
@evilebottnawi Done

hiroppy:
plz check Node6

```
FAIL test/cli.test.js (25.375s)
  ● CLI › should use different random port when multiple instances are started on different processes
    expect(received).toBe(expected) // Object.is equality
    Expected: true
    Received: false
      175 |       runtime.cp2.done = true;
      176 |       if (runtime.cp.done) {
    > 177 |         expect(runtime.cp.port !== runtime.cp2.port).toBe(true);
          |                                                      ^
      178 |         done();
      179 |       }
      180 |     });
      at ChildProcess.toBe (test/cli.test.js:177:54)
      at ChildProcess.cp.emit (node_modules/cross-spawn/lib/enoent.js:34:25)

```

EslamHiko:
@hiroppy You're Welcome 😄

hiroppy:
Users using html-webpack-plguin are many and this change causes many impacts.
FYI: https://github.com/webpack/webpack-dev-server/issues/1934

/cc @evilebottnawi

PR #1824, opened 2019-04-26
chore(deps): adds webpack-cli to the peer dependencies

arcanis:

### Motivation / Use-Case

This package [requires `webpack-cli`](https://github.com/webpack/webpack-dev-server/blob/b31cbaaf49417eb4adf8949cd87c36ca82653b25/bin/webpack-dev-server.js#L59-L107) to work properly, but the dependency isn't listed anywhere. My best guess is that it's intended to be a peer dependency, so I added it to the manifest listing.

### Breaking Changes

n/a

### Additional Info

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "ws": "6.2.1"
   },
   "peerDependencies": {
+    "webpack-cli": "^3.3.0",
     "webpack": "^4.0.0"
   },
   "husky": {


evilebottnawi:
We can't do this, because CLI apps like `CRA`, `vue-cli` and etc no need webpack-cli

arcanis:
@evilebottnawi can it be declared as an [optional peer dependency](https://yarnpkg.github.io/berry/configuration/manifest#optional) then? Those are supported by both Yarn and pnpm.

evilebottnawi:
@arcanis it is architecture bug, we can't do this, we already working on this and in `webpack-dev-server@4` no need webpack-cli (we move binary files into `webpack-cli`)

arcanis:
Sounds good 👍

PR #1839, opened 2019-05-01
Refactor Server.js

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

no

### Motivation / Use-Case

`Server.js` has more than 1000 lines and it is hard to test.
So, we split the code.

### Breaking Changes

no


lib/Server.js
```diff
@@ -201,349 +198,22 @@ class Server {
...
-      },
-    };
-
+    const features = createFeatures.call(this, {
+      app,
+      compiler,
+      options,
+      contentBase,
+      websocketProxies,
+    });
```
evilebottnawi: Maybe better do this as method `setupFeatures` (we need `this` in context so better use SOLID practice here and do this as method)

hiroppy:
@evilebottnawi PTAL 🙏 https://github.com/webpack/webpack-dev-server/pull/1839/commits/d30d2621692031474e420047463fcccdfed45877

lib/Server.js
evilebottnawi: Apparently I didn’t explain well :smile: I meant to do method `Server.setupFeatures` like:
```js
class Server {
  setupFeatures() {
    // Logic here
  }
}
```

hiroppy: like this?

```javascript
    const features = this.setupFeatures({
      app,
      compiler,
      options,
      contentBase,
      websocketProxies,
    });

...

  setupFeatures(...args) {
    return createFeatures.call(this, args);
  }

```

evilebottnawi: put code from `setupFeatures` inside method (SOLID), we use this and using `call` with `this` is not good idea

hiroppy: hmm... but `createFeatures` has more than 280 lines... and L591 is already used  `call`.

hiroppy: ok, fixed https://github.com/webpack/webpack-dev-server/pull/1839/commits/1ea70ce11d777388895cba111e51ee59f3d607c1

evilebottnawi:
Good job!

PR #1847, opened 2019-05-06
refactor: server

evilebottnawi:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Not required

### Motivation / Use-Case

Refactor code (part 1 of 3)

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -83,20 +83,23 @@ class Server {
   }
 
   constructor(compiler, options = {}, _log) {
-    this.log = _log || createLogger(options);
-
-    validateOptions(schema, options, 'webpack Dev Server');
-
     if (options.lazy && !options.filename) {
       throw new Error("'filename' option must be set in lazy mode.");
     }
+
+    validateOptions(schema, options, 'webpack Dev Server');
+
+    updateCompiler(compiler, options);
+
+    this.compiler = compiler;
+    this.options = options;
+
+    this.log = _log || createLogger(options);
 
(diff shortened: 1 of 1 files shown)


lib/Server.js
```diff
@@ -324,7 +213,9 @@ class Server {
 
             certExists = false;
           }
-        } else {
+        }
+
+        if (!certExists) {
```
evilebottnawi: Regression in previously refactor, need add tests (will be done in other PR)

lib/Server.js
```diff
@@ -378,38 +269,122 @@ class Server {
...
+
+  setupDevMiddleware() {
+    // middleware for serving webpack bundle
+    this.middleware = webpackDevMiddleware(
+      this.compiler,
+      Object.assign({}, this.options, { logLevel: this.log.options.level })
+    );
+  }
+
+  setupFeatures() {
```
evilebottnawi: No options, we in class so we can use `this`

lib/Server.js
```diff
@@ -625,15 +608,61 @@ class Server {
...
+      defaultFeatures.unshift('compress');
+    }
+
+    if (this.options.after) {
+      defaultFeatures.push('after');
+    }
+
+    (this.options.features || defaultFeatures).forEach((feature) => {
+      features[feature]();
+    });
```
evilebottnawi: Also we initialize features in this method

Issue #1777, opened 2019-04-09
V3.3.0 live reload broken when set hostname to `0.0.0.0:xxxx`


njugray:
- Operating System: MacOS
- Node Version: v10.15.0
- NPM Version: 6.4.`
- webpack Version: 4.29.6
- webpack-dev-server Version: 3.3.0

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

https://github.com/webpack/webpack-dev-server/blob/99e8db0f8cc441bb8f6acc60d786e33c38b8fc1d/client-src/default/index.js#L210

When start a node server like 0.0.0.0:8080, and load resources from `0.0.0.0:9000`, Port was reset to `self.location.port `, which actually is 8080.

### For Bugs; How can we reproduce the behavior?

- Start webpack-dev-server and listen at e.g. 0.0.0.0:9000
- Load  resources from another local server like: 0.0.0.0:8080

### Expected Behavior

Live reload should listen 0.0.0.0:9000

### Actual Behavior

Live reload listen 0.0.0.0:8080

evilebottnawi:
@njugray Please create minimum reproducible test repo

njugray:
@evilebottnawi 

here: https://github.com/njugray/webpack-live-reload

njugray:
1. Install dependencies and `npm run dev`

2. Visit: http://127.0.0.1:8080/

3. See network inspect, will get:
![image](https://user-images.githubusercontent.com/5938871/55861557-30552d80-5ba9-11e9-9a0c-74eba9ddebbb.png)

evilebottnawi:
Thanks!

ryank311:
This one also affects us because we run the dev server on a different port than our node server that is serving the page.  It would be huge for us if the sockjs host was configurable/override-able, as it would let us do things like run the webpack-dev-server on a different host or in a container separate from our node server.

evilebottnawi:
Maybe `sockHost` and `sockPort`? Now we already support `sockPath`

ryank311:
Yeah, I took a look and if we could set the sockHost and sockPort, that would work for us.  Unfortunately with just sockpath, it doesn't give us enough control over the domain.  Another alternative would be to utilize the publicPath, publicHost and publicPort that are already part of the config, but either would unblock us from upgrading.

Also, we should probably document the sockHost, sockPort and sockPath as I was only able to reason about them by stepping through the code.

evilebottnawi:
[...]

PR #1858, opened 2019-05-09
Added sockHost, changed how sock URL is built

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

As mentioned here: https://github.com/webpack/webpack-dev-server/issues/1777, there could potentially be a need for a `sockHost` which is different from the `host` option.

I also realized that when `sockPort` was added, a problem could occur if `sockPort` was included but `sockPath` was not. Specifically, they both check `urlParts.path === '/'` to see if the query string should be parsed or not, but that method of checking only works if a single option is passed in through `urlParts.path`. I can make a minimum reproducible test repo for this problem if needed. This is the reasoning behind partially changing how the sock url is built.

### Breaking Changes

None

### DIFF
--- a/client-src/default/index.js
+++ b/client-src/default/index.js
@@ -225,9 +225,9 @@ if (
 let sockHost = hostname;
 let sockPath = '/sockjs-node';
 let sockPort = urlParts.port;
-// eslint-disable-next-line no-undefined
 if (
   urlParts.path !== null &&
+  // eslint-disable-next-line no-undefined
   urlParts.path !== undefined &&
   urlParts.path !== '/'
 ) {
--- a/client-src/default/index.js
+++ b/client-src/default/index.js
@@ -220,21 +220,30 @@ if (
 ) {
   protocol = self.location.protocol;
 }
+
+// default values of the sock url if they are not provided
+let sockHost = hostname;
+let sockPath = '/sockjs-node';
+let sockPort = urlParts.port;
+if (urlParts.path != null && urlParts.port !== '/') {
+  const parsedQuery = querystring.parse(urlParts.path);
+  // all of these sock url params are optionally passed in through
+  // __resourceQuery, so we need to fall back to the default if
+  // they are not provided
+  sockHost = parsedQuery.sockHost || sockHost;
+  sockPath = parsedQuery.sockPath || sockPath;
+  sockPort = parsedQuery.sockPort || sockPort;
+}
+
 const socketUrl = url.format({
   protocol,
   auth: urlParts.auth,
-  hostname,
-  port:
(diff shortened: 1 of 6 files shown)


client-src/default/index.js
```diff
@@ -220,21 +220,30 @@ if (
 ) {
   protocol = self.location.protocol;
 }
+
+// default values of the sock url if they are not provided
+let sockHost = hostname;
+let sockPath = '/sockjs-node';
+let sockPort = urlParts.port;
+if (urlParts.path != null && urlParts.path !== '/') {
```
hiroppy: Please use strict equality operator. Or are you intentionally doing so to include `undefined`?

lib/options.json
```diff
@@ -288,6 +288,16 @@
     "setup": {
       "instanceof": "Function"
     },
+    "sockHost": {
+      "anyOf": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "null"
```
hiroppy: I think we don't have to write explicitly.

client-src/default/index.js
Loonride: That's how it was here: https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/index.js#L235. I kept it that way under the assumption that it might need to catch `undefined`.

lib/options.json
Loonride: You mean the `"type": "null"`? The `host` option has the same format.

client-src/default/index.js
hiroppy: I see. I didn't know that it already used...

lib/options.json
hiroppy: Yes, we discussed here, and I also don't think it's necessary.

fyi: https://github.com/webpack/webpack.js.org/pull/2995#discussion_r281509754

client-src/default/index.js
evilebottnawi: @Loonride i think we should fix using `!= (`==` and etc) in our code base

Loonride: @evilebottnawi should I fix it here?

hiroppy: I want to change `!==`, but its change is the breaking change, so please rewrite as below.

`if ((urlParts.path !== null || urlParts.path !== undefined) && urlParts.path !== '/') {`

Loonride: I think `urlParts.path != null` would turn into `(urlParts.path !== null && urlParts.path !== undefined)`

hiroppy: You're right. My mistake.

PR #1877, opened 2019-05-15
refactor: setupFeatures method

evilebottnawi:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

no need

### Motivation / Use-Case

Refactor

### Breaking Changes

No

### Additional Info

No


lib/Server.js
```diff
@@ -329,251 +335,299 @@ class Server {
...
+        this.setupMagicHtmlFeature();
       },
       setup: () => {
         if (typeof this.options.setup === 'function') {
-          this.log.warn(
-            'The `setup` option is deprecated and will be removed in v4. Please update your config to use `before`'
-          );
-
-          this.options.setup(app, this);
+          this.setupSetupFeature();
```
evilebottnawi: Weird name, but we remove this in future so we can leave this as is

lib/Server.js
```diff
@@ -329,251 +335,299 @@ class Server {
...
+      typeof contentBase === 'number'
+    ) {
+      throw new Error('Watching remote files is not supported.');
+    } else if (Array.isArray(contentBase)) {
+      contentBase.forEach((item) => {
+        this._watch(item);
+      });
+    } else {
+      this._watch(contentBase);
+    }
```
evilebottnawi: I see a bug here, when you use string we check on remove, but if you use array we don't check this :trollface: Let's fix it in other PR

hiroppy: Can you create an issue? (or already existed?)

evilebottnawi: @hiroppy done https://github.com/webpack/webpack-dev-server/issues/1879

PR #1889, opened 2019-05-16
feat(server): add liveReload option to enable/disable live reloading

EslamHiko:
- [ ] This is a **bugfix**
- [x] This is a **feature**
- [ ] This is a **code refactor**
- [ ] This is a **test update**
- [ ] This is a **docs update**
- [ ] This is a **metadata update**

### For Bugs and Features; did you add new tests?
yes

### Motivation / Use-Case
fix this pr : https://github.com/webpack/webpack-dev-server/pull/1812

### Breaking Changes
no

### Additional Info
I hope everything works fine this time.

Issue #1860, opened 2019-05-10
GSoC Refactor: serverMode, clientMode, replace sockjs with ws


hulkish:
- we should not define `sockjs` as a direct dependency, by default this will be replaced with `ws`
- `clientMode: 'ws'`, by default
- user must install `sockjs` themselves, then set `clientMode: require.resolve('sockjs')` if they want to support older browsers with dicey websocket support.

Loonride:
If we do that `require` method to allow `sockjs` as a `clientMode` then if the user wanted to make their own server implementation they would need to implement in the same way as the [sockjs API](https://www.npmjs.com/package/sockjs), specifically the parts the Dev Server uses (they would need a `createServer` method, event listeners, etc.). We could also alter this API slightly as we like and create adapters for essential servers like `sockjs`. Then we can make an adapter for `ws` and simply implement it as the default in that way. (Should this be called `serverMode` instead since it is the server implementation?)

We need a similar method for the client. I'm not sure how to do this best so that the user can pass their implementation into the Node server. Maybe I am missing an easier way. My thoughts are:
- user specifies a path to their client implementation script/package
- script/package is added to the Webpack compilation as an entry
- dev server client uses user's client implementation

Loonride:
Tasks to complete for client/server refactor:

**Server**
- [ ] Encapsulate SockJS server (started above)
- [ ] Deal with any SockJS workarounds that have been added to the Dev Server over time
- [ ] Finalize a server wrapper API that is simple, such that users can implement their own wrappers (prototype [here](https://github.com/webpack/webpack-dev-server/pull/1904/files#diff-5ee4a67d6e2c856789b602a3551afd60))
- [ ] Use the wrapper to encapsulate ws
- [ ] Switch to native WebSockets on client for simple proof of concept with ws
- [ ] Add `serverMode` option which takes a string for existing/default implementations, or a class (extending some class like [BaseServer](https://github.com/webpack/webpack-dev-server/pull/1904/files#diff-371258c1b34003b889c7d8c2bddd10ad)) which the user would have to implement themselves

**Client**
- [ ] Create client wrapper API, similar to that of server
- [ ] Encapsulate SockJS client, native WebSockets using the wrapper
- [ ] Insert client implementation as an entry in the Webpack compilation, making it easier for the user to supply their own implementation (tentative, see my comment above)
- [ ] Add `clientMode` option, again takes a string for existing/default implementations or a class that the user implements

**Extra**
[...]

PR #1904, opened 2019-05-17
Encapsulate sockjs, introduce ws and build base for users to make own server implementation

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

Refactored to encapsulate `sockjs` better so that it is easier to implement `ws` or other user defined server implementations.

Issue: https://github.com/webpack/webpack-dev-server/issues/1860

### Breaking Changes

I am unsure if moving `sockjs` out breaks this: https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L41

Otherwise there should be no breaking changes.

### Additional Info

Tests were failing locally, I think it is related to a snapshots though: https://github.com/webpack/webpack-dev-server/pull/1889


lib/Server.js
```diff
@@ -32,6 +31,7 @@ const createDomain = require('./utils/createDomain');
 const runBonjour = require('./utils/runBonjour');
 const routes = require('./utils/routes');
 const schema = require('./options.json');
+const SockjsServer = require('./utils/server/sockjsServer');
 
 // Workaround for sockjs@~0.3.19
```
evilebottnawi: Let's move workaround to sockjsServer

lib/Server.js
```diff
@@ -674,20 +674,15 @@ class Server {
...
+    const socket = new SockjsServer({
+      error: this.log.error.bind(this),
+      debug: this.log.debug.bind(this),
+      server: this.listeningApp,
+      path: this.sockPath,
     });
+    this.socketServer = socket;
 
-    socket.on('connection', (connection) => {
+    socket.onConnection((connection) => {
```
evilebottnawi: Better this.sockerServer everywhere

lib/utils/server/sockjsServer.js
```diff
@@ -0,0 +1,43 @@
...
+
+  close(connection) {
+    connection.close();
+  }
+
+  // f should return the resulting connection
+  onConnection(f) {
+    this.socket.on('connection', f);
+  }
+};
```
evilebottnawi: Let's move this to `lib/servers/sockjsServer.js`, it it not util

package.json
```diff
@@ -62,6 +62,7 @@
     "url": "^0.11.0",
     "webpack-dev-middleware": "^3.7.0",
     "webpack-log": "^2.0.0",
+    "ws": "^6.2.1",
     "yargs": "12.0.5"
   },
```
evilebottnawi: No need this right now, it is increase package size

lib/Server.js
```diff
@@ -32,6 +31,7 @@ const createDomain = require('./utils/createDomain');
 const runBonjour = require('./utils/runBonjour');
 const routes = require('./utils/routes');
 const schema = require('./options.json');
+const SockjsServer = require('./utils/server/sockjsServer');
```
evilebottnawi: rename `sockjsServer` to `SockJSServer` because it is class

lib/Server.js
```diff
@@ -674,20 +674,15 @@ class Server {
...
-          this.log.error(line);
-        } else {
-          this.log.debug(line);
-        }
-      },
+    const socket = new SockjsServer({
+      error: this.log.error.bind(this),
+      debug: this.log.debug.bind(this),
+      server: this.listeningApp,
+      path: this.sockPath,
```
evilebottnawi: Better always pass all options (this.options)

lib/Server.js
Loonride: You mean this workaround? https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L41

evilebottnawi: Yes

lib/Server.js
Loonride: @evilebottnawi Are you suggesting, in addition to what is already there, `options: this.options`? Because `this.listeningApp` is not part of the options, along with the `error`, `debug` callbacks.

Users may be creating server implementations of their own, so they would be interacting with the options passed in. Do you think it's a good idea to expose all the options to the user?

evilebottnawi: @Loonride 
> Do you think it's a good idea to expose all the options to the user?

Yes, we because behavior can be changed based on options (not only) in theory. 

Let's do it like:
`new SockJSServer(this);`

Custom implementation should have access to compiler, application, server, options and devServer.

Also it is not available for developers right now, so we can change this in future.

lib/servers/sockjsServer.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
```
hiroppy: The file name should be changed from `sockjsServer.js` to `SockJSServer`.

lib/servers/wsServer.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
```
hiroppy: The file name should be changed from `wsServer.js` to `WsServer`.

lib/servers/baseServer.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
```
hiroppy: The file name should be changed from `baseServer.js` to `BaseServer`.

lib/servers/wsServer.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 // const ws = require('ws');
-const BaseServer = require('./baseServer');
+const BaseServer = require('./BaseServer');
 
 // ws server implementation under construction
 // will need changes in the client as well to function
-module.exports = class WSServer extends BaseServer {};
+module.exports = class WsServer extends BaseServer {};
```
evilebottnawi: /cc @hiroppy What do you think is better:
- WebsocketServer
- WSServer
- WsServer

hiroppy: I like WebsocketServer. For example, the plugin name of wasm(WebAssemblyPlugin) which webpack has is not omitted. But this example is webpack, not webpack-dev-server. In any case, I think that it is easier to understand if it is not omitted.

evilebottnawi: :+1: let's use `WebsocketServer`

/cc @Loonride

Loonride:
@evilebottnawi, @hulkish says just to continue work in this pr, are you fine with that or do you want to get the base changes merged first?

hiroppy:
Plese don't continue to work in this pr. Because don't give PR more than one content, as it will be difficult to review.

hulkish:
@hiroppy

> Plese don't continue to work in this pr. Because don't give PR more than one content, as it will be difficult to review.

I disagree, this is all part of the same gsoc effort. What things do u think deviate enough to be considered separate?

hulkish:
@Loonride please add tests which validate support for `clientMode` & `serverMode` (as we discussed on slack)

hiroppy:
We have `next` branch for the major version. If this pr is changed to `next` branch, this problem will be resolved? Or do you want a branch dedicated to GSoC? I oppose continuing development on this branch as a maintaioner. You should create a dedicated branch and submit PR there.

hiroppy:
If you want to need `gsoc` branch, I'll create it.

hulkish:
@hiroppy How about:

- create `next` branch, if not exists already.
- point the target branch for this pr to `next`
- merge this pr to `next` once code review passes
- go from there?

Loonride:
I like the idea of having `gsoc / next` branch. No chance of adding a breaking change in `master` as I implement things, and also I will be more efficient if I can do more work per PR.

hulkish:
@Loonride i think that may be beside the point - your code will be merged once stable (with tests)

then, the `next` branch will act as the living alpha or beta, until the collection of changes have all made it in and are stable

hiroppy:
> this is all part of the same gsoc effort.

I agree with this idea. I think it is a good idea to put them into one branch so that you can easily visualize the results.

I like the idea of creating gosc branch. because maybe I'll use next branch for webpack5. Don't forget that you need to periodically rebase from master branch.

Loonride:
Should I add tests on the new files to pass codecov?

PR #1907, opened 2019-05-19
test(client): check console.log

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

yes

### Motivation / Use-Case

Just testing `console.log` of the client's code.

### Breaking Changes

no

### Additional Info

Maybe liveReload has a bug... see the snapshot.

### DIFF
--- a/test/Client.test.js
+++ b/test/Client.test.js
@@ -229,3 +229,71 @@ describe('Client complex inline script path with sockHost', () => {
     });
   });
 });
+
+describe('Client console.log', () => {
+  jest.setTimeout(30000);
+
+  const baseOptions = {
+    port: 9000,
+    host: '0.0.0.0',
+  };
+  const cases = [
+    {
+      title: 'hot disabled',
+      options: {
+        hot: false,
+      },
+    },
+    {
+      title: 'hot enabled',
+      options: {
+        hot: true,
+      },
+    },
+    {
+      title: 'liveReload disabled',
+      options: {
+        liveReload: false,
+      },
+    },
+    {
+      title: 'liveReload enabled',
+      options: {
+        liveReload: true,
+      },
+    },
+  ];
+
+  for (const { title, options } of cases) {
+    it(title, () => {
+      const res = [];
+      const testOptions = Object.assign({}, baseOptions, options);
+
+      // TODO: use async/await when Node.js v6 support is dropped
+      return Promise.resolve()
+        .then(() => {
+          return new Promise((resolve) => {
+            testServer.startAwaitingCompilation(config, testOptions, resolve);
+          });
+        })
+        .then(runBrowser)
+        .then(({ page, browser }) => {
+          return new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


test/Client.test.js
```diff
@@ -229,3 +229,71 @@ describe('Client complex inline script path with sockHost', () => {
...
+        .then(({ page, browser }) => {
+          return new Promise((resolve) => {
+            page.goto('http://localhost:9000/main');
+            page.on('console', ({ _text }) => {
+              res.push(_text);
+            });
+            setTimeout(() => {
+              expect(res).toMatchSnapshot();
+              browser.close().then(resolve);
+            }, 3000);
```
hiroppy: Maybe here is flaky tests.

test/__snapshots__/Client.test.js.snap
```diff
@@ -0,0 +1,29 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Client console.log hot disabled 1`] = `
+Array [
+  "Hey.",
+  "[WDS] Live Reloading enabled.",
```
hiroppy: Should it be output this line?

test/__snapshots__/Client.test.js.snap
```diff
@@ -0,0 +1,29 @@
...
+
+exports[`Client console.log liveReload disabled 1`] = `
+Array [
+  "Hey.",
+]
+`;
+
+exports[`Client console.log liveReload enabled 1`] = `
+Array [
+  "Hey.",
```
hiroppy: We should expect `[WDS] Live Reloading enabled.`

test/__snapshots__/Client.test.js.snap
evilebottnawi: Yes

test/__snapshots__/Client.test.js.snap
evilebottnawi: Yes, need investigate what is problem

hiroppy:
client-src depends on browsers variables, e.g `window`, `postMessage`. So this code should be tested on browsers. I propose to divide the tests into different types as npm tasks.

evilebottnawi:
@hiroppy jest support jsdom and windows and other tools avaliable, but maybe it is bad idea try to use jsdom because many features are unavailable, so i think it is ok using this approach

evilebottnawi:
Anyway need investigate why some value invalid in tests as i written above

hiroppy:
>  jest support jsdom and windows and other tools avaliable, but maybe it is bad idea try to use jsdom because many features are unavailable, so i think it is ok using this approach

Yeah, I agree with you.

> Anyway need investigate why some value invalid in tests as i written above

I would like to investigate after merging this PR.

hiroppy:
I just fixed lint. PTAL /cc @evilebottnawi

hiroppy:
And I'll start to investigate client code, so I need this pr.

PR #1915, opened 2019-05-22
fix(dependency): replaced opn with open

jamesgeorge007:
- [x] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

[opn](https://npmjs.com/package/opn) has been deprecated and renamed to [open](https://npmjs.com/package/open)

### Breaking Changes

Existing applications should be made to use `open` instead.

### DIFF
--- a/lib/utils/status.js
+++ b/lib/utils/status.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const open = require('opn');
+const open = require('open');
 const colors = require('./colors');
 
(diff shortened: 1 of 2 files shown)


hiroppy:
Thank you for your contribution. However, this pr is duplicated https://github.com/webpack/webpack-dev-server/pull/1865. And we cannot upgrade to opn now.

PR #1514, opened 2018-09-30
fix(webpack-dev-server): patch the output of webpack with devServer options

krzystof:
- [x] This is a **bugfix**
- [ ] This is a **code refactor**
- [ ] This is a **test update**
- [ ] This is a **typo fix**
- [ ] This is a **metadata update**

### For Bugs and Features; did you add new tests?

I am fairly new to webpack's codebase, and not sure how to test this. With a few pointers, I will be happy to add them.
I am not sure how to test that a file change would trigger a change to the `hot-update.json` manifest and that this one is requested at the right url.

### Motivation / Use-Case

When serving an app from a different domain and/or port, the `hot-update.json` request result in a 404.

This has been discussed here:
https://github.com/webpack/webpack-dev-server/issues/1385
https://github.com/webpack/webpack/pull/8112

Here is a minimal reproduction repository for the issue https://github.com/krzystof/webpack-issue-repro

In addition to this change, the `webpack.config.js` devServer options must include `public: 'domain:port'` to allow the dev server to fetch the manifest at the correct location.

### Breaking Changes

When using the dev server, I believe that the original output config is ignored. If that's right, there is no breaking change.

Issue #1385, opened 2018-04-20
hot-update.json 404 (Not Found)


josh08h:
Reproducible test repos:
- https://github.com/krzystof/webpack-issue-repro
- https://github.com/SetTrend/webpack-issue

Shortly: publichPath in webpack-dev-server doesn't override publisPath in webpack

Need search way how better handle publicPath

-----

https://github.com/webpack/webpack-dev-server/issues/1591#issuecomment-468081433

https://github.com/webpack/webpack-dev-server/issues/1591#issuecomment-448995873

* Operating System: El Capitan 10.11.6
* Node Version: 8.9.3
* NPM Version: 5.5.1
* webpack Version: 4.6.0
* webpack-dev-server Version: 3.1.3

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

```js
const path = require('path');
const fs = require("fs");
const dir = d => path.join(process.cwd(), d);

module.exports = {
  entry: {
    application: './src/application',
    test: './src/test'
  },
  output: {
    path: dir('../public/assets/client'),
    filename: "[name].bundle.js"
  },
  devServer: {
    stats: 'errors-only',
    https: true,
    key: './path_to_pemfile.pem',
    cert: './path_to_pemfile.pem',
    headers: {
      'Access-Control-Allow-Origin': '*'
    },
    port: 9092,
    host: 'broker-gateway-webpack.firstbanco.dev'
  },
  module: {
    rules: [
      {
        test: /\.js$/,
        exclude: /node_modules/,
        use: {
          loader: "babel-loader"
        }
      }
    ]
  }
};



"scripts": {
    "start": "webpack-dev-server --hot --mode development"
}
```

### Expected Behavior
I would expect HMR to work.

### Actual Behavior
When I save a change in a js file a request is made for `application.bundle.js:33 GET https://broker-gateway.firstbanco.dev/1a00a5ecc216afcb2106.hot-update.json 404 (Not Found)`.

As you can see the correct port number has not been specified for the requested resource.

If I manually execute a GET request to https://brokergateway.firstbanco.dev:9092/1a00a5ecc216afcb2106.hot-update.json (note I have added correct port) then I get the resource fine.

### For Bugs; How can we reproduce the behavior?
As explained.. use above webpack config and change a js file under application or test entry points.

### For Features; What is the motivation and/or use-case for the feature?
Ease of development.

hachtr:
[...]

PR #1918, opened 2019-05-23
fix(webpack-dev-server): patch the output of webpack with devServer options

arthurgeron:
revival of #1514 as this seems to also be related to `webpack-hot-middleware`: [webpack-contrib/webpack-hot-middleware#338](https://github.com/webpack-contrib/webpack-hot-middleware/issues/338)

>     * [x]  This is a **bugfix**
> 
>     * [ ]  This is a **code refactor**
> 
>     * [ ]  This is a **test update**
> 
>     * [ ]  This is a **typo fix**
> 
>     * [ ]  This is a **metadata update**
> 
> 
> ### For Bugs and Features; did you add new tests?
> 
> I am fairly new to webpack's codebase, and not sure how to test this. With a few pointers, I will be happy to add them.
> I am not sure how to test that a file change would trigger a change to the `hot-update.json` manifest and that this one is requested at the right url.
> ### Motivation / Use-Case
> 
> When serving an app from a different domain and/or port, the `hot-update.json` request result in a 404.
> 
> This has been discussed here:
> #1385
> [webpack/webpack#8112](https://github.com/webpack/webpack/pull/8112)
> 
> Here is a minimal reproduction repository for the issue https://github.com/krzystof/webpack-issue-repro
> 
> In addition to this change, the `webpack.config.js` devServer options must include `public: 'domain:port'` to allow the dev server to fetch the manifest at the correct location.
> ### Breaking Changes
> 
> When using the dev server, I believe that the original output config is ignored. If that's right, there is no breaking change.
> ### Additional Info


bin/webpack-dev-server.js
```diff
@@ -136,6 +138,9 @@ function processOptions(config) {
 function startDevServer(config, options) {
   const log = createLogger(options);
 
+  addEntries(config, options);
+  patchPublicPath(config, options);
```
evilebottnawi: Invalid, no utils will be here

bin/webpack-dev-server.js
```diff
@@ -136,6 +138,9 @@ function processOptions(config) {
 function startDevServer(config, options) {
   const log = createLogger(options);
 
+  addEntries(config, options);
```
evilebottnawi: Also it is breaking change and can break setup

bin/webpack-dev-server.js
arthurgeron: So that means that a fix for #1385 and possible fix for webpack-contrib/webpack-hot-middleware#338 is going out on next major release?

arthurgeron: If not how could I make this not a breaking change? Or is this completely outdated?

evilebottnawi: @arnihermann answered in original issue, it is completely outdated and we should solve this in other way

PR #1920, opened 2019-05-23
test(server): refactor Server.test.js

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

just refactoring

### Motivation / Use-Case

~~use test-serever.js~~

At first, I was thinking of using AAA and creating PR. But I thought it would be better not to use test-server.js, as we can not test to the details of Server.js

### Breaking Changes

no

### Additional Info

Sorry, I forgot to delete ` test/helpers/run-server.js`.


hiroppy:
Don't merge, I'll add new tests.

hiroppy:
I think codecov is broken now, because just updating dependencies's pr has the same result. 

fyi: https://github.com/webpack/webpack-dev-server/pull/1921#issuecomment-495429452

So, this PR does not lower the coverage.

PR #1925, opened 2019-05-26
chore(commitlint): update scope-case rule

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

no

### Motivation / Use-Case

lib and tests have files beginning with a capital letter, so I accept camel-case and pascal-case.

### Breaking Changes

no

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -2,4 +2,7 @@
 
 module.exports = {
   extends: ['@commitlint/config-conventional'],
+  rules: {
+    'scope-case': [2, 'always', ['lower-case', 'camel-case', 'pascal-case']],
+  },
 };


hiroppy:
I don't think changing rules in our repository is a bad practice. We have already overwritten eslint too.  Also, we shouldn't force this rule to common rules. It is a breaking changes.

evilebottnawi:
@hiroppy Why am I against it?
- `conventionalcommits` is standard (https://www.conventionalcommits.org/en/v1.0.0-beta.4/)
- for better maintenance we should use one configuration for all webpack and webpack-contrib projects (we don't do this right now), but many repos already updated and have one configuration (eslint, prettier, commitlint and etc) and in future we should migrate on one configuration for **all** repos
- we do not quite understand the scope (our scopes - cli, client, server and etc, see labels), we should not use path as scope
- using non lowercase scope was broken when you generate CHANGELOG in past and also don't recommend

hiroppy:
Thank you for the explanation. ok, I'll clsoe

PR #1927, opened 2019-05-27
refactor(cli): code

evilebottnawi:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

not required

### Motivation / Use-Case

Just refactor

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/lib/utils/processOptions.js
+++ b/lib/utils/processOptions.js
@@ -1,0 +1,27 @@
+'use strict';
+
+const createConfig = require('./createConfig');
+const defaultPort = require('./defaultPort');
+
+function processOptions(config, argv, callback) {
+  // processOptions {Promise}
+  if (typeof config.then === 'function') {
+    config.then(processOptions).catch((err) => {
+      // eslint-disable-next-line no-console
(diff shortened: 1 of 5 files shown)


lib/utils/processOptions.js
```diff
@@ -0,0 +1,27 @@
...
+      process.exit();
+    });
+
+    return;
+  }
+
+  // Taken out of yargs because we must know if
+  // it wasn't given by the user, in which case
+  // we should use portfinder.
+  const options = createConfig(config, argv, { port: defaultPort });
```
hiroppy: IMO: I think util shouldn't call util. Is this util?

evilebottnawi: In theory it is bin util, but we remove this after migrate on webpack-cli, right now it is just refactor to reduce bin size, so i think it is normal have this in utils

evilebottnawi: Also it is normal when one utils call other utils, example - `getCertificate`:
1. Certificate exists - return existing certificate
2. Certificate not exists - call `createCertificate` util

PR #1930, opened 2019-05-28
feat: `onListening` option

### LINKED ISSUE #1569: listeningApp is defined late and cannot be used in before/after functions
* Operating System: `Windows 10`
* Node Version: `v8.11.2`
* NPM Version: `5.6.0`
* webpack Version: `4.25.1`
* webpack-dev-server Version: `3.1.10`

- [ ] This is a **bug**

### Code

Fails: `Cannot read property 'address' of undefined`
```webpack.config.js
// webpack.config.js

module.exports = {
  devServer: {
    port: 9001,
    host: '0.0.0.0',
    before(app, server) {
      const port = server.listeningApp.address().port;

      console.log('Listening on port:', port);
    },
  },
  // ...
};
```

Yes, in this simplistic example, I recognize that I could hard code or add a variable `port` that is shared between the `devServer` configuration object and my `before` function. However I'm trying to support other cases where this is not practical. 

### Problem
[`listeningApp` is set](https://github.com/webpack/webpack-dev-server/blob/927a2b38d4d3a5d8fd50dfce0343634d46fa8a92/lib/Server.js#L597-L602) [after `before` function is executed](https://github.com/webpack/webpack-dev-server/blob/927a2b38d4d3a5d8fd50dfce0343634d46fa8a92/lib/Server.js#L517-L519).

#476 Almost addressed this, however creating the server instance is not controllable in the `features` list and is always done last.

I think it would be a relatively small change in the startup sequence to add creating the server to the `features` list and allowing it to be controlled. I don't see how it would affect other parts of the code. If this change is undesirable, I still have a workaround that works for me but I bet others might like this change aswell.

### Workaround

Use `setImmediate()` to delay execution inside `before` function. This happens to work for my particular use case but this introduces other annoyances.

```webpack.config.js
// webpack.config.js

module.exports = {
  devServer: {
    port: 9001,
    host: '0.0.0.0',
    before(app, server) {
      setImmediate(() => {
        const port = server.listeningApp.address().port;

        console.log('Listening on port:', port);
      });
    },
  },
  // ...
};
```

evilebottnawi:
- [ ] This is a **bugfix**
- [x] This is a **feature**

yes

### Motivation / Use-Case

fixes #1569 

### Breaking Changes

No

### Additional Info

- Also added deprecation for next major release
- This option allow to developers run own scripts after run webpack-dev-server

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -414,6 +414,8 @@ class Server {
   }
 
   setupBeforeFeature() {
+    // Todo rename onBeforeSetupMiddleware in next major release
+    // Todo pass only `this` argument
     this.options.before(this.app, this, this.compiler);
   }
 
@@ -422,6 +424,8 @@ class Server {
   }
 
   setupAfterFeature() {
+    // Todo rename onAfterSetupMiddleware in next major release
+    // Todo pass only `this` argument
     this.options.after(this.app, this, this.compiler);
   }
 
@@ -721,6 +725,10 @@ class Server {
 
       if (fn) {
         fn.call(this.listeningApp, err);
+      }
+
+      if (typeof this.options.onListening === 'function') {
+        this.options.onListening(this);
       }
     });
   }
--- a/lib/options.json
+++ b/lib/options.json
@@ -195,6 +195,9 @@
     "noInfo": {
       "type": "boolean"
     },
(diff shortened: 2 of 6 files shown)


hiroppy:
Thanks

PR #1933, opened 2019-05-28
Insert client implementation from within Dev Server API

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Not yet, they will be necessary though

### Motivation / Use-Case

The goal here is to make it possible for the Dev Server API to pass a local client implementation over to the client dynamically. From this, we can make the easy step of:

`clientMode: 'path/to/client-implementation.js'`

### Breaking Changes

- The client should work the same after this change. I'll need to confirm that applying the `ProvidePlugin` does not cause problems for the compiler.

### Additional Info

- I'm not sure this is the best solution to the problem. If there is a better way to pass in a script which the client source will rely on, please recommend it
- Possibly a bad choice to have the "global" variable `__webpack_dev_server_client__`, not sure there is a better way though. Optimally, we could restrict this variable only to the `client` bundle that the Dev Server inserts.
- Should we let the user implement exponential connect retries themselves, or keep it the way I have made it?

Edit:
- also: safe to use `class` on client side?


hiroppy:
We should add tests for clients before refactoring. If we don't, we don't know if it's really breaking changes.

Loonride:
@hiroppy I know, I can write some tests. I just wanted to put the solution out for review and discussion, in case someone has a better idea of how to pass along the client implementation from the API to the client src.

Loonride:
Is the only way of really testing the client to do `puppeteer` or other e2e tests?

hiroppy:
Currently, only Client.test.js.(so just e2e)😞

I've started to test about `client/default/index.js` as unit test using jest's jsdom since yesterday because I want to refactor this code and investigate bugs.

lib/utils/getClientPath.js
```diff
@@ -0,0 +1,13 @@
...
+  no-unused-vars
+*/
+
+function getClientPath(options) {
+  // this is where we can change the source of the client implementation
+  // eg. WebsocketClient or a user-provided implementation
+  return require.resolve('./../clients/SockJSClient');
+}
+
+module.exports = getClientPath;
```
evilebottnawi: It should be static method for SockJSClient, i.e.
BaseClient should have `static getClientPath(options) { throw new Error('Need implementation'); }` and
SockJSClient should have `static getClientPath(options) { return require.resolve('./../clients/SockJSClient'); }`
For other clients too

lib/utils/updateCompiler.js
```diff
@@ -48,6 +49,11 @@ function updateCompiler(compiler, options) {
     compilers.forEach((compiler) => {
       const config = compiler.options;
       compiler.hooks.entryOption.call(config.context, config.entry);
+
+      const providePlugin = new webpack.ProvidePlugin({
+        __webpack_dev_server_client__: getClientPath(),
+      });
+      providePlugin.apply(compiler);
```
evilebottnawi: As i said above `SockJSClient.getClientPath(options)`

client-src/default/socket.js
```diff
@@ -32,15 +36,14 @@ const socket = function initSocket(url, handlers) {
...
-  sock.onmessage = function onmessage(e) {
-    // This assumes that all data sent via the websocket is JSON.
-    const msg = JSON.parse(e.data);
+  client.onMessage((data) => {
+    const msg = JSON.parse(data);
     if (handlers[msg.type]) {
       handlers[msg.type](msg.data);
     }
-  };
+  });
```
evilebottnawi: Will be great standardize client API, but we can do this in future not in this PR

Loonride: So you want to revert this and have the new client files, but not use them yet?

evilebottnawi: Keep this as is in this PR, we discussion about this in future

evilebottnawi:
Let's update snapshots (we will rewrite this tests in future)

Loonride:
> Let's update snapshots (we will rewrite this tests in future)

Update snapshots in this PR?

evilebottnawi:
Yes

evilebottnawi:
`yarn test:only -- -u`

hiroppy:
Looks good to me, but on the other hand, I am afraid of bugs by this change(of course in the future too).
So we should add tests of the client code as soon as possible.

Loonride:
@hiroppy Agreed. Separate PR for new client tests? I will wait for addition of your jest jsdom tests and then work on adding more.

hiroppy:
@Loonride I think that it is good to separate PRs but at the next refactoring, I hope that a PR will have tests:) 
I'm Sorry but I don't have enough time for coding because I am on vacation now(from 24th to 14th), so maybe it will take some time to submit a PR about client test. Although I can review at night:) I think that it is fine if you add tests in the following directory hierarchy. In addition, I'm writing only `default`, so we won't conflict. Thanks.

https://github.com/webpack/webpack-dev-server/pull/1933#issuecomment-496800711

Loonride:
@hiroppy Do you mean I should do unit tests for the new components I've added (SockJSClient, etc.)? Because if you are doing tests for `default/index.js`, that is pretty much the only other client file that needs to be tested, expect like `default/socket.js`.

I'll also add more e2e tests in `Client.test.js`, as I think we should have tests that:

- Confirm reloading works in `inline` mode
- Confirm `hot`/`hotOnly` works
- Intercept the `sockjs` communications (and WS in the future) to check the messages being sent
- Test error handling in case client gets disconnected from server

Would you want such tests in `Client.test.js`, or in the new `client` directory?

hiroppy:
> Do you mean I should do unit tests for the new components I've added (SockJSClient, etc.)? 

if needed

> if you are doing tests for default/index.js, that is pretty much the only other client file that needs to be tested, except like default/socket.js

I was going to mock this line(`socket(socketUrl, onSocketMsg)`) because it can be divided completely. And this is an e2e test. (now, I'm testing as unit tests)
https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/index.js#L272

> Would you want such tests in Client.test.js, or in the new client directory?

I created a `client` directory, I think Client.test.js targets to check a URL path. (the file name might have to be changed) Also, Also, I thought that I needed to do a lot of testing, so I wanted to split the file.

Loonride:
I am currently writing a complex e2e test, I will just put it in `Client.test.js` and then consult you later where to move it

hiroppy:
Yeah, that's fine 👍

hiroppy:
And maybe, we should test using `ProvidePlugin` before merging.

> also: safe to use class on client side?

We use babel, so it's ok.

evilebottnawi:
Can you rebase too?

evilebottnawi:
Something strange with CI, let's try to rebase as i said above

Loonride:
> Can you rebase too?

When you say rebase does it matter if I do merge or rebase? Just curious.

evilebottnawi:
git rebase master

hiroppy:
Looks to fail to rebase.

```
$ git pull upstream master
$ git checkout inject-client
$ git rebase master inject-client
```

Loonride:
I think I mistakenly reset to `HEAD` while in the middle of the rebase which just messed everything up.

evilebottnawi:
Yes, you can open new PR if you don't know how fix this problem

Loonride:
> Yes, you can open new PR if you don't know how fix this problem

I think it is good now, just need to update `Routes` snapshot again.

PR #1937, opened 2019-05-30
feat(server): add serverMode option

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

The goal of this option is to allow the user to choose the current server implementation, or provide their own implementation as a path/class.

It is a GSoC Refactor goal: https://github.com/webpack/webpack-dev-server/issues/1860

### Breaking Changes

None


lib/Server.js
```diff
@@ -67,6 +66,45 @@ class Server {
...
+    }
+
+    if (!serverImplFound) {
+      throw new Error(
+        "serverMode must be a string denoting a default implementation (eg. 'sockjs'), a full path to " +
+          'a JS file which exports a class extending BaseServer (webpack-dev-server/lib/servers/BaseServer) ' +
+          'via require.resolve(...), or the class itself which extends BaseServer'
+      );
+    }
+    this.ServerImplementation = ServerImplementation;
```
evilebottnawi: 1. Let's move this to util `getSocketServerImplementation`
2. Don't think we need `function`
3. Let's rewrite this on `switch`

test/options.test.js
```diff
@@ -340,6 +341,15 @@ describe('options', () => {
         success: [true],
         failure: [''],
       },
+      serverMode: {
+        success: [
+          '',
```
evilebottnawi: '' should be failure

lib/Server.js
Loonride: @evilebottnawi `function` option is good if the user wants to supply an implementation, but their file is exporting multiple things, for example. So like if they want to do:

```
serverMode: require('./my-implementation').Server
```

evilebottnawi: I think better use `require.resolve('./my-implementation/Server')`
/cc @hiroppy

test/ServerMode.test.js
```diff
@@ -0,0 +1,133 @@
...
+    });
+  });
+
+  describe('custom sockjs implementation', () => {
+    it('uses supplied server implementation', (done) => {
+      server = testServer.start(
+        config,
+        {
+          sockPath: '/foo/test/bar/',
+          serverMode: class MySockJSServer extends BaseServer {
```
Loonride: @evilebottnawi being able to pass in class also makes testing easier

lib/Server.js
hiroppy: I'm fine either way:) Currently, it was fixed as `require.resolve`.

lib/utils/getSocketServerImplementation.js
```diff
@@ -0,0 +1,41 @@
...
+      // since errors could be difficult to understand
+      ServerImplementation = options.serverMode;
+      break;
+    default:
+      serverImplFound = false;
+  }
+
+  if (!serverImplFound) {
+    throw new Error(
+      "serverMode must be a string denoting a default implementation (eg. 'sockjs'), a full path to " +
```
hiroppy: nit

```diff
- eg
+ e.g.
```

lib/Server.js
```diff
@@ -67,6 +67,16 @@ class Server {
...
 
+    if (this.options.serverMode === undefined) {
+      this.options.serverMode = 'sockjs';
+    } else {
+      this.log.warn(
+        'serverMode is an experimental option, meaning its usage could potentially change without warning'
+      );
+    }
+
+    this.ServerImplementation = getSocketServerImplementation(this.options);
```
evilebottnawi: I think we should rename variable to `this.socketServerImplementation` to avoid misleading, we already store value in `this.socketServer` so let's use `socket` prefix too

Loonride: I'm going to capitalize first letter to clarify that it is a class.

lib/Server.js
```diff
@@ -67,6 +67,19 @@ class Server {
...
+    if (this.options.serverMode === undefined) {
+      this.options.serverMode = 'sockjs';
+    } else {
+      this.log.warn(
+        'serverMode is an experimental option, meaning its usage could potentially change without warning'
+      );
+    }
+
+    // this.SocketServerImplementation is a class, so it must be instantiated before use
+    this.SocketServerImplementation = getSocketServerImplementation(
```
evilebottnawi: `this.SocketServerImplementation` -> `this.socketServerImplementation`

PR #1940, opened 2019-05-31
test(client): add e2e reload client tests

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

N/A

### Motivation / Use-Case

https://github.com/webpack/webpack-dev-server/pull/1933

More tests are needed for the client. These tests confirm that reloading in the browser is working.

### Breaking Changes

Sometimes reload took long to occur, so I set a long wait time of 10000. I think this was fixed after I set `poll: 500`.

Also note that I added a `temp` directory in the testing directory so that my tests can write to files for reload simulation

### Additional Info

Things that can still be added onto this:
- Intercept the `sockjs` communications to confirm all the messages pass as expected
- Check the console output to make sure everything is working
- Test cases where there should be errors. For example, set `hotOnly: true` then do something that cannot be reloaded with `hot` and see the error in the console, or shut off the server early and make sure the console says `WDS Disconnected`

EDIT:
Another use case for this method
- Check that the client successfully gets `hot-update.json`


evilebottnawi:
Also need fix linting problem

Loonride:
@evilebottnawi What should I name the other e2e test if I move it into the same directory? Maybe `ClientOptions.test.js` since it confirms that the options work on the client?

evilebottnawi:
Sounds good

Loonride:
@evilebottnawi Why not `test/e2e/`? May be misleading if some tests are not in the test directory

Loonride:
I made the `test/e2e` change, I can switch it if you disagree.

evilebottnawi:
:+1: We can refactor tests in any time, just want keep e2e test separate from unit tests, same for integration tests we should do, but not in this PR, let's do it future

.gitignore
```diff
@@ -12,3 +12,5 @@ node_modules
 .vscode
 
 .eslintcache
+
+test/temp/*.css
```
hiroppy: Did you use editorconfig?

Loonride: What does that mean? I'm not sure what editorconfig does. Is there something wrong with doing this?

hiroppy: I'm not sure but only this file has EoF icon.

Loonride: Oh I see. I saw that too but overlooked it earlier. I think the file just needs another line at the end.

evilebottnawi:
What is wrong with azure :confused: :

evilebottnawi:
Good job!

PR #1943, opened 2019-05-31
refactor: utils and server

evilebottnawi:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

no need

### Motivation / Use-Case

move `opn` login to own util like `bonjour`

We should move `opn` logic from `status` util because it is misleading status should only output logs (based on name of util), so let's do it in next pr

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/lib/utils/runOpen.js
+++ b/lib/utils/runOpen.js
@@ -1,0 +1,21 @@
+'use strict';
+
+const open = require('opn');
+
+function runOpen(uri, options, log) {
+  let openOptions = {};
+  let openMessage = 'Unable to open browser';
+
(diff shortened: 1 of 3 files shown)


lib/utils/runOpen.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const open = require('opn');
+
+function runBonjour(uri, options, log) {
```
hiroppy: runBonjour should be renamed.

evilebottnawi: :+1:

PR #1948, opened 2019-06-01
refactor(client): separate into modules

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

yep

### Motivation / Use-Case

This is the first step to refactor client code.

### Breaking Changes

no


client-src/default/index.js
```diff
@@ -8,28 +8,32 @@ const stripAnsi = require('strip-ansi');
...
+};
+
+self.addEventListener('beforeunload', () => {
+  status.isUnloading = true;
+});
+
 if (typeof window !== 'undefined') {
   const qs = window.location.search.toLowerCase();
-  hotReload = qs.indexOf('hotreload=false') === -1;
+  options.hotReload = !qs.includes('hotreload=false');
```
evilebottnawi: Can't be used, `includes` is not supported old browsers, we can change this only in major release

client-src/default/index.js
```diff
@@ -8,28 +8,32 @@ const stripAnsi = require('strip-ansi');
...
-  const scriptElements = document.scripts || [];
-  const currentScript = scriptElements[scriptElements.length - 1];
-  if (currentScript) {
-    return currentScript.getAttribute('src');
-  }
-  // Fail as there was no script to use.
-  throw new Error('[WDS] Failed to get current script source.');
-}
+const sendMessage = require('./utils/sendMessage');
+const reloadApp = require('./utils/reloadApp');
```
evilebottnawi: Move this to `client-src/default/utils`, let's keep server and client in difference directories

client-src/default/index.js
evilebottnawi: You we need setup polyfill for this

test/client-src/default/utils/getCurrentScriptSource.test.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
```
evilebottnawi: Let's rename test directory to `client` not `client-src`. `src` postfix used because we build bundle

client-src/default/index.js
```diff
@@ -180,50 +158,49 @@ const onSocketMsg = {
...
+let { hostname, protocol } = urlParts;
 
 // check ipv4 and ipv6 `all hostname`
 if (hostname === '0.0.0.0' || hostname === '::') {
   // why do we need this check?
   // hostname n/a for file protocol (example, when using electron, ionic)
   // see: https://github.com/webpack/webpack-dev-server/pull/384
   // eslint-disable-next-line no-bitwise
-  if (self.location.hostname && !!~self.location.protocol.indexOf('http')) {
+  if (self.location.hostname && self.location.protocol.includes('http')) {
```
evilebottnawi: Also don't sue includes

client-src/default/index.js
hiroppy: ok, I'll add babel-polyfill

evilebottnawi: @hiroppy let's do this in other PR, using `babel-polyfill` directly can break many apps, because only one `babel-polyfill` can be defined, we should using `core-js` directly

evilebottnawi: @hiroppy let's leave this as is in this PR and change this in other PR (where we include polyfills)

client-src/default/index.js
evilebottnawi: Let's leave as is in this PR as i said above

hiroppy:
updated snapshots

PR #1950, opened 2019-06-01
feat(client): introduce core-js

hiroppy:
- [ ] This is a **bugfix**
- [x] This is a **feature**

modify only

### Motivation / Use-Case

https://github.com/webpack/webpack-dev-server/pull/1948#discussion_r289607053

### Breaking Changes

no


client-src/default/polyfill.js
```diff
@@ -0,0 +1,4 @@
+'use strict';
+
+require('core-js/features/string/includes');
+require('core-js/features/array/includes');
```
evilebottnawi: We pollution global namespace :disappointed: We should use name import like in docs https://github.com/zloirock/core-js

hiroppy: sorry, why?

evilebottnawi: Maybe we need setup support `import`/`export` in our babel config for client too

evilebottnawi: Using `require('core-js/features/array/includes');` modify `Array.prototype`, so if developer already  have modified `Array.prototype` we break can break their app. Using `import arrayIncludes from 'core-js/features/array/includes'` don't modify `Array.prototype` so it is safe to use

hiroppy: Makes sense. ok, I'll change

hiroppy:
@evilebottnawi PTAL

hiroppy:
I'll create other new pr.

PR #1951, opened 2019-06-01
fix(client): make use of self

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

no

### Motivation / Use-Case

just replacing `window` to `self` for worker context.

### Breaking Changes

no


client-src/default/index.js
```diff
@@ -31,10 +31,9 @@ self.addEventListener('beforeunload', () => {
   status.isUnloading = true;
 });
 
-if (typeof window !== 'undefined') {
-  const qs = window.location.search.toLowerCase();
-  options.hotReload = qs.indexOf('hotreload=false') === -1;
-}
+const qs = self.location.search.toLowerCase();
+options.hotReload = qs.indexOf('hotreload=false') === -1;
+
```
evilebottnawi: We can't remove `if (typeof window !== 'undefined') {` check because developers who use client code in `node` (example for tests) will get `undefined` error due `self.location.search`. I think better don't touch this

hiroppy: fm.. ok. I agreed

PR #1958, opened 2019-06-04
test: delete temp directory

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

modify only

### Motivation / Use-Case

Currently, main.css doesn't delete after running tests and it causes the errors of prettier.
In addition, we don't need `temp` directory. If you want it, you can use `fs.mkdtemp`.

### Breaking Changes

no

### DIFF
--- a/test/SockJSServer.test.js
+++ b/test/SockJSServer.test.js
@@ -40,11 +40,11 @@ describe('SockJSServer', () => {
 
       const client = new SockJS('http://localhost:8080/sockjs-node');
 
-      client.onmessage = function(e) {
+      client.onmessage = (e) => {
         data.push(e.data);
       };
 
-      client.onclose = function() {
+      client.onclose = () => {
         data.push('close');
       };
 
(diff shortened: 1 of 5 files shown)


test/SockJSServer.test.js
```diff
@@ -40,11 +40,11 @@ describe('SockJSServer', () => {
 
       const client = new SockJS('http://localhost:8080/sockjs-node');
 
-      client.onmessage = function(e) {
+      client.onmessage = (e) => {
```
hiroppy: warning  Unexpected unnamed function  func-names

test/SockJSServer.test.js
```diff
@@ -40,11 +40,11 @@ describe('SockJSServer', () => {
...
         data.push(e.data);
       };
 
-      client.onclose = function() {
+      client.onclose = () => {
```
hiroppy: warning  Unexpected unnamed function  func-names

Issue #1960, opened 2019-06-04
GSoC Refactor: CLI Refactor (CLI/API Uniformity Improvements, webpack-cli serve)


Loonride:
I decided to separate this from the `clientMode/serverMode` refactor effort. Here are my plans:

- [ ] make any "CLI Only" options available to API as well (e.g. [`devServer.progress`](https://webpack.js.org/configuration/dev-server/#devserverprogress))
- [ ] bring options handling in [`bin/webpack-dev-server`](https://github.com/webpack/webpack-dev-server/blob/master/bin/webpack-dev-server.js) to a bare minimum
- [ ] minimize [`createConfig`](https://github.com/webpack/webpack-dev-server/blob/master/lib/utils/createConfig.js) such that it only normalizes CLI arguments to API options, but does not handle defaults. Instead, the API should handle setting defaults.
- [ ] move the logic of [`bin/webpack-dev-server`](https://github.com/webpack/webpack-dev-server/blob/master/bin/webpack-dev-server.js) into `webpack-cli serve`, such that `webpack-cli serve` calls the Dev Server API.

Issue created at `webpack-cli`: https://github.com/webpack/webpack-cli/issues/900

EDIT:

- [ ] make the CLI throw an error (or possibly just a warning for backwards compatability?) if an argument is provided that does not exist

Parikshit-Hooda:
@hiroppy we can close this issue now.

evilebottnawi:
Still no

evilebottnawi:
Let's close, it is done, but with some bugs, but, we will fix them

PR #1961, opened 2019-06-04
feat(server/client): made progress option available to API

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

GSoC Goal: https://github.com/webpack/webpack-dev-server/issues/1960

The main goal of this is to remove functionality from `bin/webpack-dev-server` and into the API. This is the first step, moving `ProgressPlugin` usage over to the API.

As a side effect of this, I discovered that the CLI-specific part was using `argv.profile`. To not make this a breaking change, I added a `profile` option as well, so that it can get passed into the API.

### Breaking Changes

Despite not being documented, you could use the `progress` option on API before. However, it only outputted progress to the client console. Now, progress is also outputted to the command line.

### Additional Info

New documentation will be needed for `profile` option, along with changes to `progress` option


Loonride:
Need to change paths for new test after this is merged: https://github.com/webpack/webpack-dev-server/pull/1958

test/e2e/Progress.test.js
```diff
@@ -0,0 +1,62 @@
...
+        progress: true,
+        watchOptions: {
+          poll: 500,
+        },
+      };
+      testServer.startAwaitingCompilation(reloadConfig, options, done);
+    });
+
+    afterAll((done) => {
+      testServer.close(done);
```
hiroppy: fs.unlinkSync(cssFilePath);

lib/Server.js
```diff
@@ -91,6 +91,7 @@ class Server {
     // TODO this.<property> is deprecated (remove them in next major release.) in favor this.options.<property>
     this.hot = this.options.hot || this.options.hotOnly;
     this.headers = this.options.headers;
+    this.profile = !!this.options.profile;
```
evilebottnawi: `this.options.profile`, don't use `this.<option>` it is unnecessary and all `this.<option>` will be removed in next major release

lib/Server.js
```diff
@@ -147,6 +148,10 @@ class Server {
   }
 
   setupProgressPlugin() {
+    new webpack.ProgressPlugin({
+      profile: this.profile,
+    }).apply(this.compiler);
```
evilebottnawi: hm, we really need two plugins? I think we can use one instance of plugin

Loonride: I think 2 are needed since one supplies a custom handler but the other uses the default handler: https://github.com/webpack/webpack/blob/master/lib/ProgressPlugin.js#L103

evilebottnawi: hm, we need investigate this and tests this, looks like we setup two progress plugin and it can be invalid usage

test/e2e/Progress.test.js
```diff
@@ -0,0 +1,64 @@
...
+    });
+
+    describe('on browser client', () => {
+      jest.setTimeout(30000);
```
hiroppy: need to delete after https://github.com/webpack/webpack-dev-server/pull/1965 is merged

Loonride:
What is wrong with Windows CI?

lib/Server.js
Loonride: @evilebottnawi Wouldn't it be the equivalent of two plugins tapping into the exact same events? Assuming the plugin was made properly.

Loonride: @evilebottnawi Are my changes here causing the Windows CI problems? I'm unsure why progress plugin outputs stuff to `stderr`, but maybe that is causing a problem? https://github.com/webpack/webpack/blob/master/lib/ProgressPlugin.js#L49

evilebottnawi:
Something leak on windows due jest freeze, we need investigate this

Loonride:
I think the problem is that `ProgressPlugin` probably does not guarantee the same output each time, so I should do a more generalized test on e2e client to see that it is outputting some progress data. I think the hanging can be solved via the solution here: https://github.com/webpack/webpack-dev-server/pull/2018

test/e2e/Progress.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+describe('client progress', () => {
+  describe('using hot', () => {
+    beforeAll((done) => {
+      fs.writeFileSync(
+        cssFilePath,
+        'body { background-color: rgb(0, 0, 255); }'
+      );
+      const options = {
+        port: 9000,
```
hiroppy: this test should use a port variable

test/e2e/Progress.test.js
hiroppy: need to add `fs.unlinkSync(cssFilePath);`

Loonride: done

test/e2e/Progress.test.js
```diff
@@ -45,7 +45,12 @@ describe('client progress', () => {
               'body { background-color: rgb(255, 0, 0); }'
             );
             page.waitFor(10000).then(() => {
-              expect(res.includes('[WDS] 0% - compiling.')).toBeTruthy();
+              const testExp = /\[WDS\] [0-9]{1,3}% - compiling\./;
+              const match = res.find((line) => {
+                return testExp.test(line);
+              });
+              // eslint-disable-next-line no-undefined
+              expect(match).not.toEqual(undefined);
```
evilebottnawi: Maybe using snapshots for this?

Loonride: @evilebottnawi It is not outputting the same thing each time, like differences in numbers and such.

test/cli/cli.test.js
```diff
@@ -20,6 +20,23 @@ describe('CLI', () => {
...
+      .catch(done);
+  });
+
+  it('--progress --profile', (done) => {
+    testBin('--progress --profile')
+      .then((output) => {
+        expect(output.code).toEqual(0);
+        // should profile
+        expect(
+          output.stderr.includes('ms after chunk modules optimization')
```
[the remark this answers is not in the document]
Loonride: I will try, last time output did not always match though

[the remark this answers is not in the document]
evilebottnawi: We can implement helper for tests what remove unnecessary stuff from output to keep snapshot small and testable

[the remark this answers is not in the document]
Loonride: > We can implement helper for tests what remove unnecessary stuff from output to keep snapshot small and testable

I will try that.

> No need !! we already validate this only as boolean, so no need convert to boolean again

It could be undefined though.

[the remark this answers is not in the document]
evilebottnawi: Thanks, you are right :+1:

Loonride:
Testing with full snapshot again, last time there were some inconsistencies in output for a few CI builds though.

Loonride:
@evilebottnawi take a look at CI, there are minor differences in console output like one line getting printed twice or lines not getting printed. That's why I said snapshot is difficult.

lib/options.json
```diff
@@ -422,6 +425,7 @@
       "pfx": "should be {String|Buffer} (https://webpack.js.org/configuration/dev-server/#devserverpfx)",
       "pfxPassphrase": "should be {String} (https://webpack.js.org/configuration/dev-server/#devserverpfxpassphrase)",
       "port": "should be {Number|String|Null} (https://webpack.js.org/configuration/dev-server/#devserverport)",
+      "profile": "should be {Boolean}",
```
evilebottnawi: Please put link here to avoid changing in future

hiroppy:
CI fails https://dev.azure.com/webpack/webpack-dev-server/_build/results?buildId=3998&view=ms.vss-test-web.build-test-results-tab

Loonride:
Hmm still can't pass CI though. I tried to choose a pretty generic RegExp, surprised it is not getting outputted in some cases

Loonride:
Maybe two ProgressPlugins together does not always work

Loonride:
This is the first time CI is passing consistently, and I did not really change anything. Previously, it was regularly failing on some of the Windows machines. This tells me it is probably a problem with `puppeteer`.

Maybe I should skip the e2e test I added for Windows, then investigate e2e problems later? Could it be causing problems if two e2e tests are running concurrently?

evilebottnawi:
@Loonride maybe, we need investigate this, sometimes puppeteer just crashes on windows due unknown reasons, i think it is internal bug in `puppeteer`.

PR was completed?

Loonride:
@evilebottnawi This is complete.

I will investigate Windows `puppeteer` problems later.

Loonride:
@evilebottnawi I added 'progress' and 'profile' option tests

hiroppy:
need to rebase

PR #1967, opened 2019-06-04
refactor(client): make use of ESM instead of CJS

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

no

### Motivation / Use-Case

related #1950

### Breaking Changes

no

### DIFF
--- a/client-src/default/index.js
+++ b/client-src/default/index.js
@@ -1,14 +1,14 @@
-'use strict';
-
-/* global __resourceQuery WorkerGlobalScope self */
 /* eslint prefer-destructuring: off */
-const stripAnsi = require('strip-ansi');
-const socket = require('./socket');
-const overlay = require('./overlay');
-const { log, setLogLevel } = require('./utils/log');
-const sendMessage = require('./utils/sendMessage');
-const reloadApp = require('./utils/reloadApp');
-const createSocketUrl = require('./utils/createSocketUrl');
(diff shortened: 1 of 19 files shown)


client-src/default/index.js
```diff
@@ -1,14 +1,14 @@
...
-const { log, setLogLevel } = require('./utils/log');
-const sendMessage = require('./utils/sendMessage');
-const reloadApp = require('./utils/reloadApp');
-const createSocketUrl = require('./utils/createSocketUrl');
+import stripAnsi from 'strip-ansi';
+import socket from './socket';
+import {
+  clear as clearOverlay,
+  showMessage as showMessageOverlay,
+} from './overlay';
```
hiroppy: `* as overlay` is prohibited by eslint.

hiroppy:
I checked using sample and there was no problem. Let's see how it goes.

hiroppy:
I think `exports["default"] = _default` will break userland.

### CJS

```ts
// ts
import * as socket from 'webpack-dev-server/client/socket'
```

### ESM

_depends on settings of tsconfig_

```ts
// ts
import socket from 'webpack-dev-server/client/socket'
```

So, we will either change to the corresponding both style or change to the next branch. Anyway, I attached the label to block merging.

hiroppy:
@evilebottnawi please approve if possible

PR #1970, opened 2019-06-05
fix(server): revert #1794

### LINKED ISSUE #1934: webpack-dev-server@3.4.1 ignores changes in node_modules
- Operating System: Windows (Server 2019)
- Node Version: 10.15.0
- NPM Version: 6.4.1
- webpack Version: 4.32.2
- webpack-dev-server Version: 3.4.1

- [x] This is a **bug**

### Code

```js
// main/webpack.config.js
const HtmlWebpackPlugin = require('html-webpack-plugin');
const baseUrl = "/";

/** @type {() => import("webpack").Configuration} */
module.exports = () => ({
    plugins: [
        new HtmlWebpackPlugin({
            template: "index.ejs",
            metadata: { baseUrl }
        })
    ]
})
```

```js
// main/src/index.js
const {Service} = require("service"); // a dependency from a local node module

const div = document.createElement("div");
div.innerText = new Service().method();

document.querySelector("body").appendChild(div);
```

### Expected Behavior
When any changes occur in the `node_modules`, `webpack-dev-server` should trigger a new build.

### Actual Behavior
It is not happening after I upgraded to the latest `webpack-dev-server@3.4.1`. With the earler version of `3.3.1` it was working perfectly. If the fix is non-trivial, are there better workarounds rather than pinning the version?

I have created the repo https://github.com/Sayan751/webpack-reloading-issue to demonstrate this problem. The steps to reproduce the problem is detailed there.

hiroppy:
- [x] This is a **bugfix**
- [ ] This is a **feature**

yes

### Motivation / Use-Case

fixes #1934

### Breaking Changes

yes


test/ContentBase.test.js
```diff
@@ -63,76 +63,6 @@ describe('ContentBase', () => {
...
-      // check if node_modules folder is in watched list
-      expect(
-        watchedPaths[path.join(contentBasePublic, 'node_modules')]
-      ).toEqual(['index.html']);
-      // check if bar folder is not in watched list
-      const folderWatched = !!watchedPaths[path.join(contentBasePublic, 'bar')];
-      expect(folderWatched).toEqual(false);
-      done();
-    });
-  });
```
evilebottnawi: I this we should rewrite test to check what we watch `node_modules` (to avoid future regressions)

PR #1973, opened 2019-06-05
fix(server): Change clientLogLevel order to be called first

hiroppy:
- [x] This is a **bugfix**
- [ ] This is a **feature**

yes

### Motivation / Use-Case

maybe related #1972.

## before

Screen Shot 2019-06-05 at 16 18 45

## after

Screen Shot 2019-06-05 at 16 21 26

### Breaking Changes

no


lib/Server.js
```diff
@@ -694,11 +694,15 @@ class Server {
...
+      if (this.clientLogLevel) {
+        this.sockWrite([connection], 'log-level', this.clientLogLevel);
+      }
+
       if (this.hot) {
         this.sockWrite([connection], 'hot');
       }
 
-      if (this.options.liveReload !== false) {
+      if (this.options.liveReload) {
```
evilebottnawi: Can be breaking change for non boolean value, let's avoid this in this PR

hiroppy: ok, I'll change it at next major version.

evilebottnawi: Yes

hiroppy:
Sorry, I contained another commit to put .gitkeep.

PR #1976, opened 2019-06-05
chore: don't support Node6

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

no

### Motivation / Use-Case

Next major version doesn't support Node6.

### Breaking Changes

yes

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "client"
   ],
   "engines": {
-    "node": ">= 6.11.5"
+    "node": ">= 8.9.0"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "client"
   ],
   "engines": {
-    "node": ">= 6.11.5"
+    "node": ">= 8.16.0"
```
evilebottnawi: Why `8.16.0`? We should support with `8.9.0` (low LTS verion)

hiroppy: updated

PR #1977, opened 2019-06-06
feat(client): add clientMode option

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

This option makes it possible to provide a custom client socket implementation.

This mimics the `serverMode` implementation with one exception: `clientMode` does not accept a function (specifically, a class, as `serverMode` does). The reason for this is that it is not possible (or just pointlessly complex and impractical) to make a class within an executing server-side script, then pass it along to the client. There must be a path to the file that `module.exports` the client implementation.

Since the only thing we really need to pass to the client is the path to the implementation, I made the helper `getSocketClientPath`. There is no reason for a helper to be returning the whole `ClientImplementation` class, since it is not actually used server-side.

### Breaking Changes

None

### Additional Info

One consideration:
Should more error handling be added to confirm that the `serverMode/clientMode` are implemented correctly? For example, there could be some functionality in the Dev Server that tests the contents of `ClientImplementation.protoype`, to see if all the required functions are implemented.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -54,8 +54,6 @@ class Server {
 
     validateOptions(schema, options, 'webpack Dev Server');
 
-    updateCompiler(compiler, options);
-
     this.compiler = compiler;
     this.options = options;
 
@@ -67,6 +65,7 @@ class Server {
 
     this.log = _log || createLogger(options);
 
+    // set serverMode default
     if (this.options.serverMode === undefined) {
       this.options.serverMode = 'sockjs';
     } else {
@@ -74,6 +73,16 @@ class Server {
         'serverMode is an experimental option, meaning its usage could potentially change without warning'
       );
     }
+    // set clientMode default
+    if (this.options.clientMode === undefined) {
+      this.options.clientMode = 'sockjs';
+    } else {
+      this.log.warn(
+        'clientMode is an experimental option, meaning its usage could potentially change without warning'
+      );
+    }
+
+    updateCompiler(this.compiler, this.options);
 
     // this.SocketServerImplementation is a class, so it must be instantiated before use
     this.socketServerImplementation = getSocketServerImplementation(
--- a/lib/options.json
+++ b/lib/options.json
@@ -434,7 +434,7 @@
       "reporter": "should be {Function} (https://github.com/webpack/webpack-dev-middleware#reporter)",
       "requestCert": "should be {Boolean}",
       "serveIndex": "should be {Boolean} (https://webpack.js.org/configuration/dev-server/#devserverserveindex)",
-      "serverMode": "should be {String|Function} (https://webpack.js.org/configuration/dev-server/#devserverservermode-)",
+      "serverMode": "should be {String|Function} (https://webpack.js.org/configuration/dev-server/#devserverservermode)",
       "serverSideRender": "should be {Boolean} (https://github.com/webpack/webpack-dev-middleware#serversiderender)",
       "setup": "should be {Function} (https://webpack.js.org/configuration/dev-server/#devserversetup)",
       "sockHost": "should be {String|Null} (https://webpack.js.org/configuration/dev-server/#devserversockhost)",
--- a/lib/options.json
+++ b/lib/options.json
@@ -48,6 +48,9 @@
         "warning"
       ]
     },
+    "clientMode": {
+      "type": "string"
+    },
     "compress": {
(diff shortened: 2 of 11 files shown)


lib/options.json
```diff
@@ -390,6 +393,7 @@
       "ca": "should be {String|Buffer}",
       "cert": "should be {String|Buffer}",
       "clientLogLevel": "should be {String} and equal to one of the allowed values\n\n [ 'info', 'warn', 'error', 'debug', 'trace', 'silent' ]\n\n (https://webpack.js.org/configuration/dev-server/#devserverclientloglevel)",
+      "clientMode": "should be {String} (https://webpack.js.org/configuration/dev-server/#devserverclientmode-)",
```
evilebottnawi: extra `-` at end of line

hiroppy:
CI fails. 

options validation clientMode

```
Error: expect(received).toBeInstanceOf(expected)

Expected constructor: ValidationError
Received constructor: JestAssertionError
```

Loonride:
`getSocketClientPath(options)` is only called if `inline` is enabled, so the `clientMode` option is only validated by it when `inline` is enabled. Should I keep it as is, move it outside the `inline` check so that the `clientMode` option gets validated anyway, or throw an error when `clientMode` is provided without `inline` mode enabled?

Loonride:
I'm now wondering why this empty string is passing this test: https://github.com/webpack/webpack-dev-server/blob/master/test/options.test.js#L350. Because that should not be throwing a `ValidationError`, it should be throwing an error later on. Problem with `options.test.js`?

lib/Server.js
```diff
@@ -67,13 +65,24 @@ class Server {
...
+    // set clientMode default
+    if (this.options.clientMode === undefined) {
+      this.options.clientMode = 'sockjs';
+    } else {
+      this.log.warn(
+        'clientMode is an experimental option, meaning its usage could potentially change without warning'
+      );
+    }
+
+    updateCompiler(this.compiler, this.options);
```
evilebottnawi: Why we move this?

Loonride: @evilebottnawi I reorganized it to have the default setting for `serverMode` and `clientMode` together above that. `updateCompiler` uses `this.options.clientMode`, so the default setting for it must be set before `updateCompiler` is called. It can be switched back, I just thought it would be nice to have default setting for `serverMode` and `clientMode` right next to one another

test/e2e/ClientMode.test.js
```diff
@@ -0,0 +1,71 @@
...
+
+const testServer = require('../helpers/test-server');
+const config = require('../fixtures/client-config/webpack.config');
+const runBrowser = require('../helpers/run-browser');
+
+describe('clientMode', () => {
+  describe('sockjs', () => {
+    beforeAll((done) => {
+      const options = {
+        port: 9000,
```
hiroppy: Sorry but I change tests system. See https://github.com/webpack/webpack-dev-server/pull/2005.
And I'll merge https://github.com/webpack/webpack-dev-server/pull/2007 asap. Please change 🙏

Loonride: @hiroppy done

test/e2e/__snapshots__/ClientMode.test.js.snap
```diff
@@ -0,0 +1,19 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`clientMode custom client on browser client logs additional messages to console 1`] = `
+Array [
+  "Hey.",
+  "open",
+  "liveReload",
+  "[WDS] Live Reloading enabled.",
+  "hash",
+  "ok",
```
hiroppy: Why isn't `onClose` called?

Loonride: @hiroppy `testServer.close` happens after the snapshot is tested, so the client is not closed before that. I could maybe rearrange that to test `onClose` as well.

Loonride:
@evilebottnawi @hiroppy Can I get a review for this?

lib/options.json
hiroppy: Do we need to modify this? L437 was fixed it.

Loonride: > Do we need to modify this? L437 was fixed it.

Yes, I mistakenly did not remove that `-`, but they should be removed in both places.

PR #1984, opened 2019-06-06
refactor: scripts

evilebottnawi:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

no required

### Motivation / Use-Case

refactor

### Breaking Changes

no

### Additional Info

no


package.json
```diff
@@ -24,12 +24,13 @@
...
+    "prepare": "rimraf ./ssl/*.pem && npm run build:client",
     "transpile:index": "babel client-src/default --out-dir client --ignore \"./client-src/default/*.config.js\"",
     "transpile:clients": "babel client-src/clients --out-dir client/clients",
-    "build:index": "webpack ./client-src/default/index.js -o client/index.bundle.js --color --config client-src/default/webpack.config.js",
-    "build:live": "webpack ./client-src/live/index.js -o client/live.bundle.js --color --config client-src/live/webpack.config.js",
-    "build:sockjs": "webpack ./client-src/sockjs/index.js -o client/sockjs.bundle.js --color --config client-src/sockjs/webpack.config.js",
+    "build:client:index": "webpack ./client-src/default/index.js -o client/index.bundle.js --color --config client-src/default/webpack.config.js",
+    "build:client:live": "webpack ./client-src/live/index.js -o client/live.bundle.js --color --config client-src/live/webpack.config.js",
+    "build:client:sockjs": "webpack ./client-src/sockjs/index.js -o client/sockjs.bundle.js --color --config client-src/sockjs/webpack.config.js",
+    "build:client": "rimraf ./client/* && npm-run-all -s -l -p \"build:client:**\"",
```
evilebottnawi: - We don't extract `rm` scripts because it doesn't have sense, nobody will be used their separate like `npm run rm:ssl` 
- We use `build:client` because we should migrate on `babel` in next major release for server, so we have these command in `prepare`:
```
npm run build:client - build client
npm run build:server - build server (in next major release when we migrate on babel)
npm run build - build client and server
```

hiroppy: Why do you use babel at the server?

evilebottnawi: Using all the features ES like async/await (already supported in node@8), private fields, `import`/`export` and etc.

evilebottnawi: It is not high priority but will be great to have consistent ecma 2018/2019 code in client and server

hiroppy: ok, I want to use typescript.😉 I need types..lol

Loonride: I thought Node 6 was being dropped so we could use async/await, etc. without problems?

hiroppy: ESM is not supported.

evilebottnawi: Yes ESM is not supported, it doesn't matter if we only use babel or typescript we will build server so i rename this command, approve?

hiroppy: Yes, approved

evilebottnawi:
Sometimes our tests are freeze we need investigate this

PR #1986, opened 2019-06-06
refactor: tests

evilebottnawi:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

not required

### Motivation / Use-Case

Refactor test part #1

### Breaking Changes

No

### Additional Info

Ideally each option **should have**:
1. Unit test (own)
2. Integration test (Server.test.js)
3. E2E test (can be avoided for non client options)
4. CLI test (can be avoided right now)

We should concentrate on `1` and `2` points right now.

Other changes what we should do:
- avoid using `--runInBand` it is decrease time on tests
- union option.test.js and validation.test.js (they do same)
- rewrite some test in more effective way
- better naming of tests


hiroppy:
maybe we can not avoid using `--runInBand` because many tests have same port.

evilebottnawi:
@hiroppy We can setup for tests using difference port :smile:

hiroppy:
Yes, but we should consider managing ports because it is easy to conflict ports.

evilebottnawi:
I think we can use `getPort` utils in test helper, it is require refactor, but it is possible and tests always use difference port

evilebottnawi:
Anyway let's do it in other PRs, not here :smile:

evilebottnawi:
Missing snapshot (updated)

PR #1988, opened 2019-06-06
refactor: server

evilebottnawi:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

not required

### Motivation / Use-Case

Refactor

### Breaking Changes

No

### Additional Info

See below


lib/Server.js
```diff
@@ -506,7 +507,7 @@ class Server {
         this.setupMagicHtmlFeature();
       },
       setup: () => {
-        if (typeof this.options.setup === 'function') {
+        if (this.options.setup) {
```
evilebottnawi: No need check on `typeof` because we validate options and it is always will be Function https://github.com/webpack/webpack-dev-server/blob/master/lib/options.json#L313. Same for `before`/`after`/`onListening`

hiroppy:
Why did you combine server and status?

lib/Server.js
```diff
@@ -724,39 +725,81 @@ class Server {
...
+    }
+
+    if (this.options.historyApiFallback) {
+      this.log.info(
+        `404s will fallback to ${colors.info(
+          useColors,
+          this.options.historyApiFallback.index || '/index.html'
+        )}`
+      );
+    }
```
evilebottnawi: No need util what contains only `logs`, util should do something more than logs

lib/Server.js
```diff
@@ -724,39 +725,81 @@ class Server {
...
-      this.showStatus();
+        this.log.info(
+          'Broadcasting "http" with subtype of "webpack" via ZeroConf DNS (Bonjour)'
+        );
+      }
 
-      if (fn) {
-        fn.call(this.listeningApp, err);
+      if (this.options.open) {
+        runOpen(this.url, this.options, this.log);
```
evilebottnawi: Move `runOpen` from `status` utils because it is misleading

lib/Server.js
```diff
@@ -724,39 +725,81 @@ class Server {
...
+        ? '/'
+        : '/webpack-dev-server/';
+
     this.hostname = hostname;
+    this.url = `${createDomain(this.options, this.listeningApp)}${suffix}`;
 
     return this.listeningApp.listen(port, hostname, (err) => {
       this.createSocketServer();
 
+      this.showStatus();
```
evilebottnawi: We show status and when run bonjure and/or opn

lib/Server.js
```diff
@@ -724,39 +725,81 @@ class Server {
...
       }
 
-      if (typeof this.options.onListening === 'function') {
+      if (this.options.onListening) {
         this.options.onListening(this);
       }
+
+      if (fn) {
+        fn.call(this.listeningApp, err);
+      }
```
evilebottnawi: We should run `onListening` before callback

hiroppy:
What is the status of this?

evilebottnawi:
@hiroppy i think we can merge this, here we avoid misleading in `status` util

lib/Server.js
hiroppy: Also, we can modify L528(`if (this.options.contentBase !== false) {`) ?

evilebottnawi:
I will do it for next

PR #1996, opened 2019-06-08
test: make use of async/await

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

modify only

### Motivation / Use-Case

We don't support Node6, so we can use async/await.

### Breaking Changes

no

### Additional Info

Next step, I'll refactor test/run-server and delete TODO.


hiroppy:
I deleted test/Server.test.js because I failed to rebase from master to next.

.babelrc
```diff
@@ -1,3 +1,8 @@
 {
-  "presets": ["@babel/preset-env"]
+  "presets": ["@babel/preset-env"],
+  "env": {
+    "test": {
+      "plugins": ["@babel/plugin-transform-runtime"]
```
hiroppy: I'm not sure but jest uses babel as the default, so we need `regeneratorruntime` to use async/await. If we want to use the native async/await, we have to set `current` using `@babel/env`.

client-src/clients/BaseClient.js
```diff
@@ -1,5 +1,3 @@
-'use strict';
```
hiroppy: Affected by `"sourceType": "module"` in `./client-src/.eslintrc`

hiroppy:
Merge #1997 first.

.babelrc
evilebottnawi: We don't need this plugin for async/await:
1. It is inside `@babel/preset-env`
2. We use next branch, in node@8 it is built-in

evilebottnawi: Jest don't use babel as the default, we need `babel-jest` package for this

hiroppy: no, if we have `.babelrc`, jest uses `babel-jest`. see the documentation

hiroppy:
I want to wait for #2005 merged

.babelrc
evilebottnawi: :+1:

Issue #1990, opened 2019-06-07
server.close no longer works as of 3.4.0


Aghassi:
- Operating System: macOS 10.14.4
- Node Version: 10.16.0
- NPM Version: 6.9.0
- webpack Version: 4.33.0
- webpack-dev-server Version: 3.6.0

- [ x ] This is a **bug**
- [ ] This is a **modification** request

### Code

We have a block of code where we run webpack-dev-server and then kick off Cypress and Lighthouse. In each of these blocks of code our webpack builds fine, and the server runs. However, we can no longer kill the server with `server.close`. It does nothing (seemingly) and just leaves the process open and running.

```js
// webpack.config.js
```

```js
      new Promise((resolve, reject) => {
        server.listen(port, '0.0.0.0', async () => {
          try {
            const results = await (open ? cypress.open(cypressConfig) : cypress.run(cypressConfig));
            // Fail if there are tests that failed
            if (results.totalFailed > 0) {
              throw new CLIError(
                `Some tests failed. See results for more details ${emoji.get('point_up')}`,
              );
            }
            resolve(results);
          } catch (err) {
            // Reject on any other failures
            reject(err);
          } finally {
            console.log('server :', server);
            // Always close the server
            server.close();
          }
        });
      });
```

### Expected Behavior

The server should close when I call `server.close`

### Actual Behavior

The process remains open (we know this because our integration tests are failing because they hang on CI)

### For Bugs; How can we reproduce the behavior?

Use the node api to start the server and then call `close` on it. I can't provide a good example right now since this is code internal to the company I work for. From what I can tell though in debugging:

https://github.com/webpack/webpack-dev-server/blob/master/lib/servers/SockJSServer.js#L57

The above line doesn't work because the instance it is calling `close` on doesn't expose a close function. Below is the object I get back if I log out what `connection` is.

```js
[ SockJSConnection {
    _session:
     Session {
       session_id: undefined,
       heartbeat_delay: 25000,
       disconnect_delay: 5000,
       prefix: '/sockjs-node',
       send_buffer: [],
       is_closing: false,
       readyState: 1,
       timeout_cb: [Function],
       to_tref:
        Timeout {
          _called: false,
          _idleTimeout: 25000,
          _idlePrev: [TimersList],
          _idleNext: [TimersList],
          _idleStart: 37401,
          _onTimeout: [Function],
          _timerArgs: undefined,
          _repeat: null,
          _destroyed: false,
[...]

PR #2001, opened 2019-06-08
fix(open): set `wait: false` to run server.close successfully

hiroppy:
fixes: https://github.com/webpack/webpack-dev-server/issues/1990

- [x] This is a **bugfix**
- [ ] This is a **feature**

yes

### Motivation / Use-Case

see https://github.com/webpack/webpack-dev-server/issues/1990

commit: https://github.com/sindresorhus/open/commit/da2d663d163ad29bffea95540b1b9ea302e17871#diff-168726dbe96b3ce427e7fedce31bb0bcR23

Our open(opn)'s version is old, and `wait` seems to be `true` by default in this version.

### Breaking Changes

no

### DIFF
--- a/lib/utils/runOpen.js
+++ b/lib/utils/runOpen.js
@@ -3,11 +3,14 @@
 const open = require('opn');
 
 function runOpen(uri, options, log) {
-  let openOptions = {};
+  // https://github.com/webpack/webpack-dev-server/issues/1990
+  const openOptions = { wait: false };
   let openMessage = 'Unable to open browser';
 
   if (typeof options.open === 'string') {
-    openOptions = { app: options.open };
+    Object.assign(openOptions, {
+      app: options.open,
+    });
     openMessage += `: ${options.open}`;
   }
 
--- a/lib/utils/runOpen.js
+++ b/lib/utils/runOpen.js
@@ -4,13 +4,11 @@ const open = require('opn');
 
 function runOpen(uri, options, log) {
   // https://github.com/webpack/webpack-dev-server/issues/1990
-  const openOptions = { wait: false };
+  let openOptions = { wait: false };
   let openMessage = 'Unable to open browser';
 
   if (typeof options.open === 'string') {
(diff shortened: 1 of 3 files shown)


hiroppy:
@evilebottnawi I added one commit(https://github.com/webpack/webpack-dev-server/pull/2001/commits/fdc6d31744ced7e3f76ceb62ccc13ffe80d65226). PTAL

lib/utils/runOpen.js
```diff
@@ -3,11 +3,14 @@
...
-  let openOptions = {};
+  // https://github.com/webpack/webpack-dev-server/issues/1990
+  const openOptions = { wait: false };
   let openMessage = 'Unable to open browser';
 
   if (typeof options.open === 'string') {
-    openOptions = { app: options.open };
+    Object.assign(openOptions, {
+      app: options.open,
+    });
```
evilebottnawi:
```js
let openOptions = { wait: false };
// ...
if (typeof options.open === 'string') {
  openOptions = Object.assign({} openOptions, { app: options.open });
}
```

No need modify `openOptions` variable, it is very bad practice

hiroppy: `openOptions = Object.assign({} openOptions, { app: options.open });` equals `Object.assign(openOptions, { app: options.open })`.

I don't think this is a very bad practice.

evilebottnawi: Yes, but it is not good, we modify original object and it is bad practice because you can forget about this in future and it is lead to some problems when you refactor code. 

When you pass options to third party package(s) better always cloning object. Other package can modify object too so it can lead to very strange errors and situations.

hiroppy: > When you pass options to third party package(s) better always cloning object. Other package can modify object too so it can lead to very strange errors and situations.

If you say, we should not assign to the same variable.

I don't have any strong opinions, so I'll change.

hiroppy:
@evilebottnawi PTAL

PR #2004, opened 2019-06-09
feat(client): attach [WDS] prefix automatically

hiroppy:
- [ ] This is a **bugfix**
- [x] This is a **feature**

yes

### Motivation / Use-Case

Attach `[WDS]` prefix automatically if we use `log.xxxx`.

### Breaking Changes

no

### DIFF
--- a/client-src/default/utils/log.js
+++ b/client-src/default/utils/log.js
@@ -1,6 +1,6 @@
-'use strict';
+import { getLogger } from 'loglevel';
 
-const log = require('loglevel').getLogger('webpack-dev-server');
+export const log = getLogger('webpack-dev-server');
 
 const INFO = 'info';
 const WARN = 'warn';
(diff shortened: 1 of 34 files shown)


client-src/default/utils/log.js
```diff
@@ -2,6 +2,16 @@ import { getLogger } from 'loglevel';
...
 
+const originalFactory = log.methodFactory;
+
+log.methodFactory = (methodName, logLevel, loggerName) => {
+  const rawMethod = originalFactory(methodName, logLevel, loggerName);
+
+  return (message) => {
+    rawMethod(`[WDS] ${message}`);
+  };
+};
```
evilebottnawi: No need this:
1. It is potentially break interface (it is bad practice)
2. We should migrate on `webpack-log`
3. It should be supported as factory method (not overriding)

Issue #2006, opened 2019-06-10
Uncaught ReferenceError: __webpack_dev_server_client__ is not defined


bureson:
- Operating System: Windows 10
- Node Version: 11.12.0
- NPM Version: 6.9.0
- webpack Version: 4.33.0
- webpack-dev-server Version: 3.7.0

- [x] This is a **bug**
- [ ] This is a **modification** request

### Expected Behavior

Webpack should compile a working bundle.

### Actual Behavior

When trying the app in the webpack-dev-server mode (having the webpack-dev-server as part of the URL enabling hot reloading), browser reports a following error:

```
Uncaught ReferenceError: __webpack_dev_server_client__ is not defined
```

I believe it is coming from the following line introduced in version 3.7.0: https://github.com/webpack/webpack-dev-server/blob/3d9288a34b715428c521cd22c6f70b6b2d6a04f4/client-src/default/socket.js#L7

Where `__webpack_dev_server_client__` doesn't seem to be defined.

evilebottnawi:
It is expected, how you setup webpack-dev-server. This line respect for variable https://github.com/webpack/webpack-dev-server/blob/master/lib/utils/updateCompiler.js#L54 looks you have invalid configuration

bureson:
Thank you for the reply @evilebottnawi 

Can you please take a look at the config we are using for setting up the devServer?

```
const webpackDevServerOptions = Object.assign({}, configList[0].devServer, {
      publicPath: configList[0].output.publicPath
    });
    const server = new WebpackDevServer(compiler, webpackDevServerOptions);
    return new Promise(resolve => server.listen(8080, 'localhost', resolve));
```

where the devServer object is as follows:
```
    devServer: {
      contentBase: './output/build',
      public: 'triggerz.eu.ngrok.io'
    }
```

and publicPath is `'/static/dashboard/'`

I would also like to mention that the app loads just fine without using the `webpack-dev-server` token in the url.

evilebottnawi:
@bureson config looks good, very strange, because we automatically mount provide plugin, maybe you can create minimum reproducible test repo?

bureson:
@evilebottnawi it's a little tricky to extract the whole project into a reproducible repo. 

However when I open either `index.bundle.js` or `live.bundle.js` which I'm assuming are compiled package files, the variable `__webpack_dev_server_client__` is there available only once and it comes from the assignment which I'm mentioning in the first post.

roy1910:
[...]

PR #2015, opened 2019-06-11
fix(client): add default fallback for client

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Not yet, probably should resolve asap

### Motivation / Use-Case

Fixes https://github.com/webpack/webpack-dev-server/issues/2006 by providing a default fallback in case `__webpack_dev_server_client__` is undefined.

I moved the `default` client sources to `client/default` so that the file path structure is the same, allowing `socket.js` to `require('../clients/SockJSClient')` in both `client` and `client-src`

### Breaking Changes

None

### Additional Info

We still need to figure out how to insert `__webpack_dev_server_client__` variable into `live` mode.


client-src/default/socket.js
```diff
@@ -4,7 +4,20 @@
 /* eslint-disable
   camelcase
 */
-const Client = __webpack_dev_server_client__;
+
+// this SockJSClient is here as a default fallback, in case inline mode
+// is off or the client is not injected. This will be switched to
+// WebsocketClient when it becomes the default
+const SockJSClient = require('../clients/SockJSClient');
```
hiroppy: Move to L19

hiroppy:
We should release as a patch version asap. Thanks.

Loonride:
@evilebottnawi Looks good?

client-src/default/socket.js
```diff
@@ -4,7 +4,21 @@
...
+  Client = __webpack_dev_server_client__;
+} catch (e) {
+  // this SockJSClient is here as a default fallback, in case inline mode
+  // is off or the client is not injected. This will be switched to
+  // WebsocketClient when it becomes the default
+
+  // eslint-disable-next-line global-require
+  const SockJSClient = require('../clients/SockJSClient');
+  Client = SockJSClient;
+}
```
evilebottnawi: Maybe more simple? Like:
```
const Client = typeof __webpack_dev_server_client__ !== 'undefined' ? __webpack_dev_server_client__ : require('../clients/SockJSClient');
```

lib/utils/addEntries.js
```diff
@@ -20,7 +20,7 @@ function addEntries(config, options, server) {
     const sockPath = options.sockPath ? `&sockPath=${options.sockPath}` : '';
     const sockPort = options.sockPort ? `&sockPort=${options.sockPort}` : '';
     const clientEntry = `${require.resolve(
-      '../../client/'
+      '../../client/default/'
```
evilebottnawi: Breaking change, some packages can use this directly, so changing path can break these apps

package.json
```diff
@@ -25,7 +25,7 @@
     "test": "npm run test:coverage",
     "pretest": "npm run lint",
     "prepare": "rimraf ./ssl/*.pem && npm run build:client",
-    "build:client:default": "babel client-src/default --out-dir client --ignore \"./client-src/default/*.config.js\"",
+    "build:client:default": "babel client-src/default --out-dir client/default --ignore \"./client-src/default/*.config.js\"",
```
evilebottnawi: Breaking change, as i written above

Loonride:
@evilebottnawi any ideas how to make `require('../clients/SockJSClient')` resolve to the right place in both `client` and `client-src`?

evilebottnawi:
@Loonride hm, we already keep clients in difference directory, can't see problem

Loonride:
> @Loonride hm, we already keep clients in difference directory, can't see problem

The problem is the directory organization of `client` and `client-src` are not the same. Therefore, when `socket.js` in `client-src` uses `require('../clients/SockJSClient')` it will work

When `socket.js` in `client` uses `require('../clients/SockJSClient')`, it does not work, the correct path would be `./clients/SockJSClient`.

evilebottnawi:
Good catch, i think we can change location and create `index.js` with content like:
```
require('./default/index')

```

What do you think?

Loonride:
> Good catch, i think we can change location and create `index.js` with content like:
> 
> ```
> require('./default/index')
> ```
> 
> What do you think?

That is a good idea, let me push an alternative solution that I just made though which may be simpler and we can compare.

Loonride:
@evilebottnawi my idea is to use `webpack.NormalModuleReplacementPlugin` to direct the paths to the right place. The only problem I have with this is that it makes the files in `client-src` not work unless accompanied by the proper webpack config. But users should really not be using stuff in `client-src` anyway, right?

Loonride:
@evilebottnawi thoughts?

Loonride:
@evilebottnawi @hiroppy Could you take a look so we can resolve this issue?

client-src/default/webpack.config.js
```diff
@@ -15,4 +17,12 @@ module.exports = {
       },
     ],
   },
+  plugins: [
+    new webpack.NormalModuleReplacementPlugin(/\/clients\//, (resource) => {
```
hiroppy: memo: We will delete this plugin at the next major version because we can change the directory structure.

client-src/live/webpack.config.js
```diff
@@ -33,5 +34,11 @@ module.exports = {
         to: path.resolve(__dirname, '../../client/live.html'),
       },
     ]),
+    new webpack.NormalModuleReplacementPlugin(/\/clients\//, (resource) => {
```
hiroppy: memo: We will delete this plugin at the next major version because we can change the directory structure.

hiroppy:
I'll release today. thanks!

evilebottnawi:
Don't do this in future without my approve, here bug, other module(s) can have clients name and we break this apps
/cc @hiroppy @Loonride

client-src/default/webpack.config.js
evilebottnawi: We don't need this plugin what potentially break other applications, just create is file and re-require logic

evilebottnawi:
We should do regexp more strictly now all clients directories in requests rewrites in all modules, i.e. if package use clients directory we rewrite their import/require too, it is very very very big problem, I have many issues about this in DM gutter and other channels and I don't access to fix it right now, please guys take care about it asap, it is incredibly high priority

hiroppy:
@evilebottnawi Would you want to revert?

Loonride:
@evilebottnawi Sorry about that, I did not think it would impact other modules.

@hiroppy I say revert, then we follow the other method that @evilebottnawi suggested

Issue #1719, opened 2019-03-15
Reload occurs on every entry even only one entry changed.


orange4glace:
- Operating System:
- Node Version: 10.15.3
- NPM Version: 6.4.1
- webpack Version: 4.29.6
- webpack-dev-server Version: 3.2.1

- [ ] This is a **bug**
- [x] This is a **modification** request

### Code

```js
module.exports = [{
    entry: {
      a: './a.js'
    },
    target: 'web',
    resolve: {
      modules: ['node_modules', 'app'],
      extensions: ['.js', '.jsx']
    },
    output: {
      path: __dirname + '/dist',
      publicPath: '/',
      filename: '[name].js'
    }
  },{
    entry: {
      a: './b.js'
    },
    target: 'web',
    resolve: {
      modules: ['node_modules', 'app'],
      extensions: ['.js', '.jsx']
    },
    output: {
      path: __dirname + '/dist',
      publicPath: '/',
      filename: '[name].js'
    }
  },

]
```

**a.js**
```js
console.log('a.js')
```
**a.html**
```js
<html>
  <body>
    <script src='./a.js'></script>
  </body>
</html>
```

**b.js**
```js
console.log('b.js')
```
**b.html**
```js
<html>
  <body>
    <script src='./b.js'></script>
  </body>
</html>
```

### Expected Behavior

When either `a.js` or `b.js` is changed, like, let's say if `a.js` is changed. Then only `a.html` should be reloaded since `b.html` does not make use of `a.js`.

### Actual Behavior

`a.html` and `b.html` are reloaded together..

### For Features; What is the motivation and/or use-case for the feature?

I'm making a web app and there's a separation between data layer and presentation layer. data layer is executed on main page and presentation layer is executed on its pop-up page. When only presentation layer script is changed, only presentation pop-up page is reloaded so the underlying data can be preserved.

evilebottnawi:
Duplicate https://github.com/webpack/webpack-dev-server/issues/1503, thanks for issue, feel free to send a PR with fix

orange4glace:
@evilebottnawi Thanks for the comment. It seems time to code XD

evilebottnawi:
@orange4glace we are happy when you send PRs

orange4glace:
[...]

Issue #1503, opened 2018-09-12
webpack-dev-server recompiles all files when using webpack array of configuration


KenilDomadia:
`webpack-dev-server` when used with array of webpack configuration - it recompiles all the file instead of file that was affected. 

For example, if I have webpack config as  below:

``` javascript
var path = require("path");
var webpack = require("../../");
module.exports = [
	{
		name: "mobile",
		// mode: "development || "production",
		entry: "./src/mobile",
		output: {
			path: path.join(__dirname, "dist"),
			filename: "mobile.js"
		},
		plugins: [
			new webpack.DefinePlugin({
				ENV: JSON.stringify("mobile")
			})
		]
	},

	{
		name: "desktop",
		// mode: "development || "production",
		entry: "./src/desktop",
		output: {
			path: path.join(__dirname, "dist"),
			filename: "desktop.js"
		},
		plugins: [
			new webpack.DefinePlugin({
				ENV: JSON.stringify("desktop")
			})
		]
	}
];
```

From above example -  Any change in `src/mobile/index.js`, will result in recompilation/rebuild of `desktop.js` as well. I expect the webpack-dev-server should recompile just the file that has changed (`mobile.js`). I am using array of configuration with 15 - 20 output files, which really slows down the incremental build process.

I tried using `lazy` and `cache` mode option of `webpack-dev-server` but it doesn't seem to help. Am I missing any configuration option here?

dakom:
On a similar note - I'm not getting live-reloading with multiple configs. I get the rebuild when looking at the terminal, but the browser doesn't refresh (if I manually refresh, I do see the changes).

I'm happy to file a separate issue but I'm wondering if this is specifically due to multiple configs or some other bug... does live-reload work for you in this case?

dakom:
Nevermind - my issue is fixed by _not_ having the manual entry point added as suggested at https://stackoverflow.com/a/41223040/784519

Removing that fixed things - though it could be that I had it added to all the configs whereas it would have worked if only added to the first _shrug_

evilebottnawi:
Duplicate https://github.com/webpack/webpack-dev-middleware/issues/338
WIP

KenilDomadia:
Are we working on resolving the issue [webpack/webpack-dev-middleware#338](https://github.com/webpack/webpack-dev-middleware/issues/338)?

evilebottnawi:
@KenilDomadia First we should find where this happens and why

KenilDomadia:
[...]

PR #2017, opened 2019-06-11
fix: #1719

orange4glace:
- [x] This is a **bugfix**
- [x] This is a **feature**

fix for #1719 

When sending a 'ok' and 'invalid' message, compiler name is passed as a data.
compiler name can be defined in config options. If user does not provided it, the script generates a random guid as a compiler name.


client-src/default/index.js
```diff
@@ -9,6 +9,32 @@ const { log, setLogLevel } = require('./utils/log');
...
+    urlParts = url.parse(scriptHost || '/', false, true);
+  }
+  const { auth, path } = urlParts;
+
+  // eslint-disable-next-line no-undefined
+  if (path !== null && path !== undefined && path !== '/') {
+    const parsedQuery = querystring.parse(path);
+    compilerName = parsedQuery.compilerName;
+  }
+})(__resourceQuery)
```
evilebottnawi: Move this to `util`

client-src/live/index.js
```diff
@@ -60,6 +60,7 @@ $(() => {
       }
     },
     ok() {
+      console.log('OK!!');
```
evilebottnawi: Why you need `What is OK!!`? It is misleading need better text

lib/Server.js
```diff
@@ -719,7 +719,7 @@ class Server {
         return;
       }
 
-      this._sendStats([connection], this.getStats(this._stats), true);
+      this._sendStats(null, [connection], this.getStats(this._stats), true);
```
evilebottnawi: null?

lib/utils/updateCompiler.js
```diff
@@ -7,6 +7,13 @@
 const webpack = require('webpack');
 const addEntries = require('./addEntries');
 
+function guid() {
+  function s4() {
+    return ((1 + Math.random()) * 0x10000 | 0).toString(16).substring(1);
+  }
+  return s4() + s4() + '-' + s4() + '-' + s4() + '-' + s4() + '-' + s4() + s4() + s4();
+}
```
evilebottnawi: Better use package for this

orange4glace:
Agreed. Rather than modifying both server and client code, separating websocket connection like using channel by modifying only server code (and minimum client code modification as possible) seems to be a better implementation. I was kind of in a hurry so the implementation is actually dirty. I'll improve it with further commits :)

orange4glace:
The new commit remains client code unchanged (so the code is completely same with the upstream master code) and only server code is changed. Please see the #2017 (even it contains tiny errors and corrected by the next commits)

I am trying to write and run tests, but running test keeps failing. It seems like Windows related error. I'm using VSCode in Windows 10 and if any code change happens with Windows environment editor, ESLint test goes fail, with unknown error.

Could you check if error happens on test with my repository source?

evilebottnawi:
What is error(s) you got?

orange4glace:
Actually the error is much same with what [Azure pipeline did](https://github.com/webpack/webpack-dev-server/pull/2017/checks?check_run_id=147933407). It seems error happens on **Lint:prettier**

Error log is [here](https://pastebin.com/kYrRkUjZ), and the complete log file is [here](https://pastebin.com/55y7WteG)

evilebottnawi:
You can run tests using `npm run test:only` and avoid fixing linting issue right now, we can do this at last step of PR

alexander-akait:
https://github.com/webpack/webpack-dev-server/issues/1503

PR #2021, opened 2019-06-12
fix: error message of clientLogLevel

sottar:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

The error message of `clientLogLevel` is not correct, so fixed it.

ref. https://webpack.js.org/configuration/dev-server/#devserverclientloglevel

### Breaking Changes

not breaking change

### DIFF
--- a/lib/options.json
+++ b/lib/options.json
@@ -389,7 +389,7 @@
       "bonjour": "should be {Boolean} (https://webpack.js.org/configuration/dev-server/#devserverbonjour)",
       "ca": "should be {String|Buffer}",
       "cert": "should be {String|Buffer}",
-      "clientLogLevel": "should be {String} and equal to one of the allowed values\n\n [ 'info', 'warn', 'error', 'debug', 'trace', 'silent' ]\n\n (https://webpack.js.org/configuration/dev-server/#devserverclientloglevel)",
+      "clientLogLevel": "should be {String} and equal to one of the allowed values\n\n [ 'none', 'info', 'error', 'warning' ]\n\n (https://webpack.js.org/configuration/dev-server/#devserverclientloglevel)",
       "compress": "should be {Boolean} (https://webpack.js.org/configuration/dev-server/#devservercompress)",
       "contentBase": "should be {Number|String|Array} (https://webpack.js.org/configuration/dev-server/#devservercontentbase)",
       "disableHostCheck": "should be {Boolean} (https://webpack.js.org/configuration/dev-server/#devserverdisablehostcheck)",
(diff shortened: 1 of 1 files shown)


hiroppy:
It has the reason(https://github.com/webpack/webpack-dev-server/pull/1901#issuecomment-493494912), but we've already had `next` branch, so I'll accept this change.

lib/options.json
```diff
@@ -389,7 +389,7 @@
       "bonjour": "should be {Boolean} (https://webpack.js.org/configuration/dev-server/#devserverbonjour)",
       "ca": "should be {String|Buffer}",
       "cert": "should be {String|Buffer}",
-      "clientLogLevel": "should be {String} and equal to one of the allowed values\n\n [ 'info', 'warn', 'error', 'debug', 'trace', 'silent' ]\n\n (https://webpack.js.org/configuration/dev-server/#devserverclientloglevel)",
+      "clientLogLevel": "should be {String} and equal to one of the allowed values\n\n [ 'none', 'info', 'error', 'warning' ]\n\n (https://webpack.js.org/configuration/dev-server/#devserverclientloglevel)",
```
hiroppy: The description should have `warn` and `silent`, and we should update documentation.

```diff
- [ 'none', 'info', 'error', 'warning' ]
+ [ 'none', 'silent', 'info', 'error', 'warning', 'warn' ]
```

sottar: Thank you.
I tried `warn` and `silent`, but it doesn't work as well 🤔 Is it intended?

sottar: Ah~~ it's my bad. The version of `webpack-dev-server` in my local is old. I'll fix it.
Then can I handle to update documentation?

hiroppy: Oops, yep. I confused. `none` and `warning` are going to be deprecated at the next major version, instead of `silent` and `warn`.

hiroppy: @sottar We've already had the [updated documentation](https://github.com/webpack/webpack.js.org/pull/2976
). So, I'll merge this documentation after merging this pr.

hiroppy:
related documentation https://github.com/webpack/webpack.js.org/pull/2976

lib/options.json
sottar: Oh, I see. So the documentation is the latest right? 
It seems different between you suggested above.

hiroppy: @sottar sorry, I forgot `trace`...

https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/utils/log.js#L5-L15

please update 🙇

sottar:
@hiroppy Thank you~~ enjoy staying in Europe 😁

PR #2028, opened 2019-06-14
feat/refactor(server): move CLI server start into API listen method

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Not yet, would like feedback on the idea first.

### Motivation / Use-Case

I have added a `start` method in the API, which the CLI calls, rather than calling the API's `listen` method.

- Firstly, the CLI will do significantly less configuration/error handling, as the API will do it instead
- API users can find more use in `start` method, because it does not make sense that they have to provide `host` and `port` via `listen(port, hostname, fn)`, when they probably already pass those into `new Server(options)` as dev server options.
- API users can make use of `socket` option and the automatic free port assignment via `start`, or opt out of those options via `listen`.

### Breaking Changes

Likely not, need to add tests to be sure though

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -4,8 +4,6 @@
 
 /* eslint-disable no-shadow, no-console */
 
-const fs = require('fs');
-const net = require('net');
 const debug = require('debug')('webpack-dev-server');
 const importLocal = require('import-local');
 const yargs = require('yargs');
@@ -15,7 +13,6 @@ const setupExitSignals = require('../lib/utils/setupExitSignals');
 const colors = require('../lib/utils/colors');
 const processOptions = require('../lib/utils/processOptions');
 const createLogger = require('../lib/utils/createLogger');
-const findPort = require('../lib/utils/findPort');
 const getVersions = require('../lib/utils/getVersions');
 const options = require('./options');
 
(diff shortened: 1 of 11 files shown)


bin/webpack-dev-server.js
```diff
@@ -115,58 +112,7 @@ function startDevServer(config, options) {
...
-              }
-            });
-          }
-        });
-
-        clientSocket.connect({ path: options.socket }, () => {
-          throw new Error('This socket is already used');
-        });
-      }
-    });
```
evilebottnawi: I think we can solve this without new method, it is very inconsistently with node api for servers, we can move this code without new method, just listen `error` event and handle this inside server class

Loonride: Do you mean handle it inside the `listen(port, hostname, fn)` method, rather than making a new method?

Loonride:
Tests that need to be added for this:

`socket` option:
- confirm that it works on CLI and API
- simulate `chmod` failure to ensure that handler works
- simulate EADDRINUSE for server start, then consequent ECONNREFUSED to see if the server can be created after initial failure

`port` option:
- should still work as string and number on API/CLI, if the string parses to a number
- on API, port should now go to a free port if an undefined one is provided

Loonride:
> Let's separate this PR in two:
> 
> 1. Move socket check
> 2. Move findPort
> 
> Anyway looks good, better handle this in two PRs because:
> 
> 1. Easy review
> 2. Less chance of making a mistake
> 3. Easy create changelog
> 
> Anyway good job

Closing in favor of https://github.com/webpack/webpack-dev-server/pull/2061

Issue #1843, opened 2019-05-05
Let "open" and "openPage" open a different url then bound one


damianobarbati:
This is a **modification** request.

Scenario: you are developing a webapp on your desktop but you need to access it from from your smartphone.

This imply having the webpack-dev-server to bind to 0.0.0.0 (otherwise dev-server will not serve dev.mydomain.com resolved to desktop from the smartphone with custom DNS).

When firing dev-server with open=true, the url opened is "https://0.0.0.0" which ends up in a "certificate invalid page".

I currently see no option to open a custom url (i.e: https://dev.mydomain.com) which results in the correct opened page.

Suggestion: maybe `openPage` could stop prefixing `host` if given url is an absolute one.

i.e: 
```
openPage: 'https://dev.mydomain.com', => opens https://dev.mydomain.com
openPage: '/home', => opens HOST + /home
```

evilebottnawi:
hm, can you provide minimum reproducible test repo with problem?

msegers:
I was actually also hoping for some more options with the openPage prop.

We serve our build through a separate proxy (outside webpack). say dev server servers on port 8080, but we actually browse our application on 8081. with openPage I cannot specify: `https:/localhost:8081/mypage` as it will open `https://localhost:8080/httpss:/localhost:8081/mypage`

evilebottnawi:
@msegers Can you create minimum reproducible test repo with? Looks like it is bug

PR #2048, opened 2019-06-17
fix: allow openPage to open absolute urls

EslamHiko:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case
fixes: https://github.com/webpack/webpack-dev-server/issues/1843

### Breaking Changes
no


lib/utils/runOpen.js
```diff
@@ -12,7 +12,11 @@ function runOpen(uri, options, log) {
     openMessage += `: ${options.open}`;
   }
 
-  return open(`${uri}${options.openPage || ''}`, openOptions).catch(() => {
+  const pageUrl = /^(https?:)?\/\//.test(options.openPage)
```
mistic: I think you should use something like https://github.com/ogt/valid-url instead of that simple regex in order to validate if the url is valid.

lib/utils/runOpen.js
```diff
@@ -12,7 +12,11 @@ function runOpen(uri, options, log) {
...
+    ? options.openPage
+    : `${uri}${options.openPage || ''}`;
+
+  return open(pageUrl, openOptions).catch(() => {
     log.warn(
       `${openMessage}. If you are running in a headless environment, please do not use the --open flag`
```
mistic: I think that error message should be updated to also reflect hints of failures when using a non valid absolute url

test/server/utils/runOpen.test.js
```diff
@@ -80,6 +80,24 @@ describe('runOpen util', () => {
         `);
       });
     });
+
+    it('on specify URL with page in Google Chrome ', () => {
```
mistic: I think that description doesn't make sense when running in the test runner

lib/utils/runOpen.js
EslamHiko: I don't think we need a full separate package to validate the url. I think Regex is enough.

lib/utils/runOpen.js
EslamHiko: no failures when using non valid absolute url it'll append it to the server.
ex: openPage is ht:/nonvalid and the devserver is http://localhost:8080 
the result will be  http://localhost:8080/ht:/nonvalid

test/server/utils/runOpen.test.js
EslamHiko: I did specify http/https in the description, and I added http test, Thanks.

mistic: 👍

lib/utils/runOpen.js
mistic: 👍

lib/utils/runOpen.js
mistic: I would still prefer to use a better method, but I would left this consideration for others

evilebottnawi: I think better use this package https://github.com/sindresorhus/is-absolute-url for check

hiroppy:
Thanks

Aghassi:
@hiroppy @evilebottnawi Nice job, this looks like an awesome PR. Will it be coming in a patch version or in a major? It's unclear from discussion here.

evilebottnawi:
@Aghassi minor release in near future

PR #2049, opened 2019-06-18
fix(client): default client fallback with redirection to all default directory files

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

Fix https://github.com/webpack/webpack-dev-server/issues/2006,

Fix problems introduced by https://github.com/webpack/webpack-dev-server/pull/2015

I made a redirect to each file in the `default` directory, because it's better to be safe I think.

### Breaking Changes

None


Loonride:
@evilebottnawi @hiroppy Here is a fix.

If you revert the previous fix, I will just rebase and fix any problems that result from that.

Loonride:
Also I should add tests to make sure all the redirection are correct.

Would it make more sense if [`test/client`](https://github.com/webpack/webpack-dev-server/tree/master/test/client) tested files in the `client` directory, rather than the `client-src` directory, since `client` directory is what users will actually use? If we switched to this method, I could simply point those tests towards the redirection files that I create in this PR, and it would confirm that they work.

Loonride:
The change I just made makes the `client` tests use `client` directory, rather than `client-src`. Those are the files that the dev server and users use, so they should be tested. The unit tests pass through the redirection files that I added to each individual client file, confirming that they work. 

This fix spans more than I expected, but probably is safer than other solutions.

Loonride:
Closing in favor of https://github.com/webpack/webpack-dev-server/pull/2069

PR #2051, opened 2019-06-20
feat(server): check that serverMode implementation is correct

knagaitsev:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

This feature gives useful error messages when the user provides an incorrect `serverMode` implementation.

### Breaking Changes

None

### DIFF
--- a/lib/schemas/utils/socketServerImplementationPrototype.json
+++ b/lib/schemas/utils/socketServerImplementationPrototype.json
@@ -1,0 +1,27 @@
+{
+  "type": "object",
+  "properties": {
+    "constructor": {
+      "instanceof": "Function"
+    },
+    "send": {
+      "instanceof": "Function"
+    },
+    "close": {
+      "instanceof": "Function"
+    },
+    "onConnection": {
+      "instanceof": "Function"
+    }
+  },
+  "errorMessage": {
+    "properties": {
+      "constructor": "- serverMode must have a constructor that takes a single server argument and calls super(server) on the superclass BaseServer, found via require('webpack-dev-server/lib/servers/BaseServer')",
+      "send": "- serverMode must have a send(connection, message) method that sends the message string to the provided client connection object",
+      "close": "- serverMode must have a close(connection) method that closes the provided client connection object",
+      "onConnection": "- serverMode must have a onConnection(f) method that calls f(connection) whenever a new client connection is made"
+    }
+  },
+  "required": ["constructor", "send", "close", "onConnection"],
+  "additionalProperties": false
+}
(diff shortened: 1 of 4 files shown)


alexander-akait (changes requested):
I think it is extra checking and should be solved using `schema-util` instead adding checks in source code, we should search way how we can integrate this in schema util

Loonride:
> I think it is extra checking and should be solved using `schema-util` instead adding checks in source code, we should search way how we can integrate this in schema util

Switched to use of `schema-util`. I don't think it has the ability out of the box to check the parameter count of a function, but also that is not a check we really need.

hiroppy (approved):

lib/schemas/utils/socketServerImplementationPrototype.json
```diff
@@ -0,0 +1,27 @@
...
+    "properties": {
+      "constructor": "- serverMode must have a constructor that takes a single server argument and calls super(server) on the superclass BaseServer, found via require('webpack-dev-server/lib/servers/BaseServer')",
+      "send": "- serverMode must have a send(connection, message) method that sends the message string to the provided client connection object",
+      "close": "- serverMode must have a close(connection) method that closes the provided client connection object",
+      "onConnection": "- serverMode must have a onConnection(f) method that calls f(connection) whenever a new client connection is made"
+    }
+  },
+  "required": ["constructor", "send", "close", "onConnection"],
+  "additionalProperties": false
+}
```
evilebottnawi: I think we can have one schema (put this inside original schema) for better acquaintance

alexander-akait (changes requested):
Good job, one note

lib/schemas/utils/socketServerImplementationPrototype.json
Loonride: @evilebottnawi The problem is that I couldn't find a way to test the prototype of a function with `schema-utils` when it matched `"instanceof": "Function"`. I think it only looks at `"properties"` if the thing in question matches `"type": "object"`

Loonride: Unless you just mean make `options.json` into something like:
```
{
"defaultSchema": { ... },
"socketServerImplementationPrototypeSchema": { ... }
}
```

But that could be a breaking change if something else is using this schema.

evilebottnawi: hm we can rewrite schema-utils and allow to define own validator (where we can implement any logic), i will take care about this, let's keep this PR open, it is not high priority because we can improve documentation about this

ullashb (approved):

ullashb (approved):

ullashb (approved):

PR #2055, opened 2019-06-22
fix(client): Make public path on client match dev server publicPath

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

Fixes: https://github.com/webpack/webpack-dev-server/issues/1385

This creates an absolute URL based on provided `devServer.publicPath`, host, port, etc., then it sets `__webpack_public_path__` on the client to this absolute URL. This forces the client to request `hot-update.json` from the correct location.

This works because `hot-update.json` is requested from the webpack public path here: https://github.com/webpack/webpack/blob/master/lib/web/JsonpMainTemplate.runtime.js#L31 (`$require$.p` is public path).

The client must be informed of `publicPath`, so I pass it to the client using resource query along with `sockHost`, `sockPort`, `sockPath`. I did some work to improve tests and how these options are encoded/decoded through the resource query.

### Breaking Changes

Now, the `__webpack_public_path__` will be set to an absolute URL.


client-src/default/utils/updatePublicPath.js
```diff
@@ -0,0 +1,29 @@
...
+const getUrlParts = require('./getUrlParts');
+
+function updatePublicPath(resourceQuery) {
+  const urlParts = getUrlParts(resourceQuery);
+
+  let publicPath = urlParts.publicPath;
+
+  // this tests if publicPath is relative, and makes it an absolute url if it is
+  // will match something of the form '/.../', or simply '/'
+  if (/^\/.*\/$|^\/$/.test(publicPath)) {
```
evilebottnawi: Can we encapsulate this in `isRelativeUrl` (maybe using other repository for this check)?

client-src/default/index.js
```diff
@@ -9,6 +9,9 @@ const { log, setLogLevel } = require('./utils/log');
 const sendMessage = require('./utils/sendMessage');
 const reloadApp = require('./utils/reloadApp');
 const createSocketUrl = require('./utils/createSocketUrl');
+const updatePublicPath = require('./utils/updatePublicPath');
+
+updatePublicPath(__resourceQuery);
```
evilebottnawi: I am afraid what we change global variable and it is can break code in some rare case, maybe we can rewrite this on using own variable like:
```
const resourceQuery = getPublicPath(__resourceQuery);
```

Loonride: Yes good point

jony89:
Hey, thanks for the PR this is really needed. as the functionality for publicPath is broken now.

evilebottnawi:
Only in next major release, sorry it is breaking change

jony89:
> Only in next major release, sorry it is breaking change

sure but can this at least be merged for some of us that want to start work with the `next` branch?

evilebottnawi:
@jony89 yep, we will do this in near future (now we are working on next release webpack-dev-middleware, so it will be updated too)

jony89:
Hey, 21 days pass, any progress? can I help with anything?

PR #2061, opened 2019-06-24
feat/refactor(server): move socket handling into server.listen

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

This is the `socket` option handling portion of: https://github.com/webpack/webpack-dev-server/pull/2028. The goal is to move socket handling out of the CLI and into the API `listen` method. 

### Breaking Changes

API users will now get different behavior when they put a socket path in place of `port` in the `server.listen(port, hostname, fn)` method.

### Additional Info

This should be merged before I start work on the `findPort` portion of https://github.com/webpack/webpack-dev-server/pull/2028

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -755,20 +755,16 @@ class Server {
       }
     };
 
-    const listenCallback = () => {
+    const onListeningCallback = () => {
       if (typeof this.options.onListening === 'function') {
         this.options.onListening(this);
       }
     };
 
-    const setupAndListenCallback = () => {
-      setupCallback();
-      listenCallback();
-    };
-
     const fullCallback = (err) => {
-      setupAndListenCallback();
+      setupCallback();
       userCallback(err);
+      onListeningCallback();
     };
 
     // try to follow the Node standard in terms of deciding
@@ -780,12 +776,7 @@ class Server {
       // set this so that status helper can identify how the project is being run correctly
       this.options.socket = socket;
 
-      startUnixSocket(
-        this.listeningApp,
-        socket,
-        setupAndListenCallback,
-        userCallback
-      );
+      startUnixSocket(this.listeningApp, socket, fullCallback);
     } else {
       this.listeningApp.listen(port, hostname, fullCallback);
     }
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -6,7 +6,6 @@
   func-names
 */
 const fs = require('fs');
-const net = require('net');
 const path = require('path');
 const tls = require('tls');
 const url = require('url');
@@ -36,6 +35,7 @@ const routes = require('./utils/routes');
 const getSocketServerImplementation = require('./utils/getSocketServerImplementation');
 const handleStdin = require('./utils/handleStdin');
 const tryParseInt = require('./utils/tryParseInt');
+const startUnixSocket = require('./utils/startUnixSocket');
 const schema = require('./options.json');
 
 // Workaround for node ^8.6.0, ^9.0.0
@@ -776,55 +776,7 @@ class Server {
       // set this so that status helper can identify how the project is being run correctly
       this.options.socket = socket;
 
-      const chmodSocket = (done) => {
-        // chmod 666 (rw rw rw) - octal
-        const READ_WRITE = 438;
-        fs.chmod(socket, READ_WRITE, done);
-      };
-
-      const startSocket = () => {
-        this.listeningApp.on('error', (err) => {
-          userCallback(err);
-        });
-
-        this.listeningApp.listen(socket, hostname, (err) => {
-          if (err) {
-            fullCallback(err);
-          } else {
-            chmodSocket(fullCallback);
-          }
-        });
-      };
-
-      fs.access(socket, fs.constants.F_OK, (err) => {
-        if (err) {
-          // file does not exist
-          startSocket();
-        } else {
-          // file exists
-
-          const clientSocket = new net.Socket();
-
-          clientSocket.on('error', (err) => {
-            if (err.code === 'ECONNREFUSED' || err.code === 'ENOTSOCK') {
-              // No other server listening on this socket so it can be safely removed
-              fs.unlinkSync(socket);
-
-              startSocket();
-            }
-          });
-
-          clientSocket.connect({ path: socket }, () => {
-            // if a client socket successfully connects to the given socket path,
-            // it means that the socket is in use
-            const err = new Error('This socket is already used');
-            clientSocket.destroy();
-            // do not call onListening or the setup method, since the server
-            // cannot start listening on a used socket
-            userCallback(err);
-          });
-        }
-      });
+      startUnixSocket(this.listeningApp, socket, fullCallback, userCallback);
     } else {
       this.listeningApp.listen(port, hostname, fullCallback);
     }
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -6,6 +6,7 @@
   func-names
 */
 const fs = require('fs');
+const net = require('net');
 const path = require('path');
 const tls = require('tls');
 const url = require('url');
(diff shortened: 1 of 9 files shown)


lib/Server.js
```diff
@@ -741,23 +743,102 @@ class Server {
...
+            // if a client socket successfully connects to the given socket path,
+            // it means that the socket is in use
+            const err = new Error('This socket is already used');
+            clientSocket.destroy();
+            // do not call onListening or the setup method, since the server
+            // cannot start listening on a used socket
+            userCallback(err);
+          });
+        }
+      });
```
evilebottnawi: Can we refactor this on utils? Maybe multiple utils

Loonride: I'll try

lib/Server.js
```diff
@@ -785,7 +789,12 @@ class Server {
       // set this so that status helper can identify how the project is being run correctly
       this.options.socket = socket;
 
-      startUnixSocket(this.listeningApp, socket, fullCallback, userCallback);
+      startUnixSocket(
+        this.listeningApp,
+        socket,
+        setupAndListenCallback,
+        userCallback
+      );
```
evilebottnawi: Let rewrite this in style (pseudo code):
```
startUnixSocket(
  this.listeningApp,
  socket,
  (error) => { 
    // Here logic for callback
  }
);
```

Idea is not pass a lot of argument in util, just use closure for this

Loonride: @evilebottnawi I was passing in 2 callbacks because I didn't want `options.onListening` to be called if an error occurred before the server starts listening. But maybe that is not important. If you think it is important, maybe we could have:

```
startUnixSocket(
  this.listeningApp,
  socket,
  (error, startedListening) => { 
    if (startedListening) {
    }
    // other functions that always get called, regardless of if server started listening
  }
);
```

hiroppy:
This pr has breaking changes so I think we should change from master to next.

Loonride:
Should I switch it over to `next` branch?

evilebottnawi:
Yes

hiroppy:
@Loonride Could you rebase?

hiroppy:
@Loonride Do you need to commits of the latest master before merging? If you need, I'll rebase the next branch from master.

Loonride:
@hiroppy For this PR the latest master commits should not affect it

hiroppy:
After all, I'll rebase from the master to the next branch every time. https://github.com/webpack/webpack-dev-server/compare/master...next 
I just want to ask if you want to check if these changes can run on the latest master commits branch before merging.

lib/utils/startUnixSocket.js
```diff
@@ -0,0 +1,58 @@
...
+    listeningApp.listen(socket, 511, (err) => {
+      if (err) {
+        cb(err);
+      } else {
+        chmodSocket(cb);
+      }
+    });
+  };
+
+  fs.access(socket, fs.constants.F_OK, (e) => {
```
hiroppy: We don't need to use a callback. Just use `promisify`.

Loonride: @hiroppy is this what you meant?:

https://github.com/webpack/webpack-dev-server/pull/2061/files#diff-2e3826007762dc4188ae8303ccd8b036R32-R38

test/server/utils/startUnixSocket.test.js
```diff
@@ -0,0 +1,124 @@
...
+    // this test is significant because previously the callback was called
+    // twice if a file at the given socket path already existed, but
+    // could be removed
+    it('should only call server.listen callback once', (done) => {
+      const userCallback = jest.fn();
+      startUnixSocket(testUnixSocket.server, socketPath, userCallback);
+      setTimeout(() => {
+        expect(userCallback).toBeCalledTimes(1);
+        done();
+      }, 10000);
```
hiroppy: I think this value is big but do we need 10000?

lib/utils/startUnixSocket.js
hiroppy: Yes, thanks!

test/server/utils/startUnixSocket.test.js
Loonride: I think 3000 should be enough.

hiroppy:
Need to rebase.

lib/Server.js
```diff
@@ -732,23 +734,54 @@ class Server {
...
         this.options.onListening(this);
       }
-    });
+    };
+
+    const fullCallback = (err) => {
+      setupCallback();
+      userCallback(err);
+      onListeningCallback();
+    };
```
evilebottnawi: I think we should refactor this in future, looks very misleading

Loonride: Really those 3 functions can all be in `fullCallback`, since they are only called by `fullCallback`. Originally I was doing some other things with these individual functions to avoid a few breaking changes and stay organized, but do you think just putting all their functionality in that one callback will be cleaner?

evilebottnawi: @Loonride i think we should allow `userCallback` to be async, also ehre good questions should `userCallback` will be called before or after `onListeningCallback`, potentially user callback should be called after all operation, it is allow do something with server (or not with server) when webpack-dev-server starts, for example for other web server, (for example if you use php project, you should use this callback to run php dev server)

Loonride: > i think we should allow userCallback to be async

Maybe we should do this in a separate PR?

> potentially user callback should be called after all operation

So this is a good order the way I have above, because `userCallback` is after `setupCallback`, right?

evilebottnawi: > Maybe we should do this in a separate PR?

Yes, just thoughts

> So this is a good order the way I have above, because userCallback is after setupCallback, right?

oh, it is interesting, i think `userCallback` should be after all callbacks, because i want to get socket information after starts server

Loonride: > oh, it is interesting, i think userCallback should be after all callbacks, because i want to get socket information after starts server

With the current ordering, `userCallback` should be able to access socket information (because `this.createSocketServer()` is in `setupCallback`.)

And I think the ordering of these two:
```
userCallback(err);
onListeningCallback();
```

should not matter, as user should probably only use one or the other (the `userCallback` calls the callback for the `Server.listen` method, and the `onListeningCallback` calls the user provided `options.onListening` callback).

There is, however, a question of whether these two callbacks should always be called in the same places. For consistency, I would say yes. However, `onListening` currently gets called in these two places:

https://github.com/webpack/webpack-dev-server/pull/2061/files#diff-2e3826007762dc4188ae8303ccd8b036R25

https://github.com/webpack/webpack-dev-server/pull/2061/files#diff-2e3826007762dc4188ae8303ccd8b036R25

But these places are technically not the server listening, they are just errors. Could we possibly change the `onListening` option to accept both an error object and the server object? Optimally, it would be `onListening(err, server)`, but since the current api is `onListening(server)`, `onListening(server, err)` would reduce breaking changes. Thoughts?

evilebottnawi: /cc @hiroppy what do you think about this too

hiroppy: I want to change `onListening(err, server)`. In fact, this format is correct as Node.js theory.

evilebottnawi: :+1: agree

Loonride: So can we get this merged, and then I'll do 2 PRs:

1. async `userCallback`
2. `onListening(err, server)`

But @evilebottnawi can you clarify the async one. A user can already do this without problems:

```
server.listen(port, host, async (err) => {
  await something();
  // ...
});
```

And I think there are no things before `userCallback` that are asynchronous. Maybe you mean we should do:

```
setupCallback();
await userCallback(err);
onListeningCallback();
```

But the only thing this would do is delay `onListeningCallback` until a potentially async `userCallback` is complete. 

Is this what you meant?

evilebottnawi: > But @evilebottnawi can you clarify the async one? A user can already do this without problems:
```
server.listen(port, host, async (err) => {
  await something();
  // ...
});
```

Yep, but it is require to create own util for running webpack-dev-server, in some cases it is unnecessary, because you want run one simple task when  webpack-dev-server started

> But the only thing this would do is delay onListeningCallback until a potentially async userCallback is complete.

> Is this what you meant?

It is expected and can be used for some features

evilebottnawi:
Good job!

PR #2069, opened 2019-06-25
fix(client): stricter reg exp to redirect sockjs client path

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

Fixes problem introduced by https://github.com/webpack/webpack-dev-server/pull/2015 using a stricter reg exp. 

I added an iframe e2e test because the issue came up from users using `iframe` mode: https://github.com/webpack/webpack-dev-server/issues/2006

### Breaking Changes

None


client-src/default/webpack.config.js
```diff
@@ -18,11 +18,14 @@ module.exports = {
...
   },
   plugins: [
-    new webpack.NormalModuleReplacementPlugin(/\/clients\//, (resource) => {
-      resource.request = resource.request.replace(
-        /\/clients\//,
-        '/../clients/'
-      );
-    }),
+    new webpack.NormalModuleReplacementPlugin(
+      /^\.\/clients\/SockJSClient$/,
```
evilebottnawi: We need search way how we can check what module was loaded from our `client-src` - it is right way don't break other code

Loonride: You mean to test that it is loading the correct module in the webpack compilation? Could we run this webpack config in a test, look at the webpack CLI output, and see if it contains the right path?

evilebottnawi: Yes, but it can solved using `NormalModuleReplacementPlugin` as you use, just need more checks (need check context contains 'webpack-dev-server/client-src' and rewrite only in this case)

Loonride: @evilebottnawi I did something like this, except instead I matched `'/client-src/default'`, because it is possible the dev server is not in a `webpack-dev-server` directory if someone clones it into a different named directory. If we want to be very strict about it, we could match against `path.resolve` to get the exact context we want, but I'm afraid that could result in some problem.

Also, I included backwards slash (`\\`) in my matching string because windows seems to use that for the context paths

Loonride:
> Just check `if (resource.context.startsWith(process.cwd())) { }`, we need rewrite `clients` only for `webpack-dev-server`, for other package it should be not rewritting

Done. Good solution!

PR #2077, opened 2019-06-27
fix(server): fix header check for socket server

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Not yet

### Motivation / Use-Case

Headers are not passed in the same way on connection for `ws` and `sockjs`, so the socket server implementation needs to reflect that, then the server can check the headers.

### Breaking Changes

None


lib/servers/SockJSServer.js
```diff
@@ -57,8 +57,14 @@ module.exports = class SockJSServer extends BaseServer {
...
-  // f should return the resulting connection
+  // f should return the resulting connection and, optionally, the connection headers
   onConnection(f) {
-    this.socket.on('connection', f);
+    this.socket.on('connection', (connection) => {
+      f(connection, connection.headers);
+    });
+  }
+
+  onConnectionClose(connection, f) {
```
evilebottnawi: Why we need this?

lib/servers/SockJSServer.js
```diff
@@ -57,8 +57,14 @@ module.exports = class SockJSServer extends BaseServer {
     connection.close();
   }
 
-  // f should return the resulting connection
+  // f should return the resulting connection and, optionally, the connection headers
   onConnection(f) {
-    this.socket.on('connection', f);
+    this.socket.on('connection', (connection) => {
+      f(connection, connection.headers);
```
evilebottnawi: Why not use only `connection`?

lib/Server.js
```diff
@@ -677,25 +677,22 @@ class Server {
     const SocketServerImplementation = this.socketServerImplementation;
     this.socketServer = new SocketServerImplementation(this);
 
-    this.socketServer.onConnection((connection) => {
+    this.socketServer.onConnection((connection, headers) => {
```
evilebottnawi: connection already has headers, i think we don't need extra argument

lib/servers/SockJSServer.js
Loonride: @evilebottnawi Because `ws` connection object does not work the same, it does not have a `headers` property (https://github.com/webpack/webpack-dev-server/pull/2077/files#diff-b111fb9e1bc06a00edb8d0416cabf86dR33). The goal is to not assume anything about the `connection` object, since it could differ between socket server types.

lib/servers/SockJSServer.js
Loonride: @evilebottnawi As I said above, I want to not assume anything about the `connection` object. Probably most `connection` objects will work like `connection.on('close', f)`, which is the case with both `sockjs` and `ws`. But maybe some other library implements it like `connection.onclose = f`.

lib/Server.js
```diff
@@ -677,25 +677,22 @@ class Server {
...
       if (!connection) {
         return;
       }
 
-      if (
-        !this.checkHost(connection.headers) ||
-        !this.checkOrigin(connection.headers)
-      ) {
+      if (headers && (!this.checkHost(headers) || !this.checkOrigin(headers))) {
```
sokra: This looks like it's too easy to accidentally skip the security check.

Maybe changing it to `if (!headers || ...`

Loonride: @sokra Good point. I think if users do not implement `onConnection` correctly though, it will be hard to figure out what is wrong. Maybe before that I'll log a warning if headers are missing.

evilebottnawi: @Loonride it should be no warning, we should drop connection if headers are not present

Loonride: @evilebottnawi If `!headers` is true it means that the user made a socket server implementation that did not call the `onConnection(f)` callback correctly, so I think that could cause confusion if there is little explanation for failed connection. Let me send a PR and I'll show you what I mean

PR #2090, opened 2019-07-02
feat(client): 'ws' clientMode

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

Adds `clientMode: 'ws'` with tests to confirm that it works, including full e2e tests.

### Breaking Changes

None


Loonride:
> /cc @Loonride something wrong with node@6

I think this was just a very unsuccessful e2e test run, not particular to node 6. Unless there is something in the output that I missed

PR #2095, opened 2019-07-03
test(client): socket helper tests

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

Added tests to confirm that the client `socket` helper is using client implementations correctly.

### Breaking Changes

None


Loonride:
@evilebottnawi What do you think I should name it? It follows conventions of the `client` test directory, but you are right that maybe it is too generic.

evilebottnawi:
Yes, just other name, maybe more complex, because name is misleading, any ideas?

Loonride:
> Yes, just other name, maybe more complex, because name is misleading, any ideas?

How about `socket-util.test.js`?

hiroppy:
hm, this file has already included in `utils/`, so I think `util` is unnecessary.

PR #2098, opened 2019-07-04
refactor(server): move publicPath configuration from CLI to API

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Not yet

### Motivation / Use-Case

This is part of the effort to move configuration changes out of the CLI and into the API. The goal here is to set the `publicPath` option in the API if it is not already set.

### Breaking Changes

No breaking changes on CLI.

On API, if a user was not specifying a `publicPath` in the past, it would not be set. Now, it will be set based on `compiler.options.output.publicPath`, if it exists. Should this PR be on `next` for this reason?

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -32,6 +32,7 @@ const createDomain = require('./utils/createDomain');
 const runBonjour = require('./utils/runBonjour');
 const routes = require('./utils/routes');
 const getSocketServerImplementation = require('./utils/getSocketServerImplementation');
+const setPublicPath = require('./utils/setPublicPath');
 const schema = require('./options.json');
 
 // Workaround for node ^8.6.0, ^9.0.0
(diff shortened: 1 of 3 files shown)


lib/Server.js
```diff
@@ -58,6 +59,8 @@ class Server {
     this.compiler = compiler;
     this.options = options;
 
+    setPublicPath(compiler, options);
```
evilebottnawi: Let's move this in `updateCompiler`

hiroppy:
This pr should be changed to the next.

lib/Server.js
Loonride: @evilebottnawi this is not altering the compiler at all though, only using the compiler's option data to set `devServer.publicPath`. But if you still think it should be there I have no strong opinion.

Loonride:
I think I will just open new PR on next

Loonride:
Closing in favor of https://github.com/webpack/webpack-dev-server/pull/2099

PR #2100, opened 2019-07-04
chore(template): added browser option in issue template and contribution guide refactoring

anikethsaha:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

NA

### Motivation / Use-Case
As this project is client-side heavy too. A proper details are required for browser too when resolving an issue ! So along with other options, added browser too

### Breaking Changes
No

### Additional Info
NA

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -11,6 +11,7 @@
 - NPM Version:
 - webpack Version:
 - webpack-dev-server Version:
+- Browser:
 
 <!--
   Please place an x, no spaces, in all [ ] that apply


package.json
```diff
@@ -31,7 +31,6 @@
     "build:client:live": "webpack ./client-src/live/index.js -o client/live.bundle.js --color --config client-src/live/webpack.config.js",
     "build:client:sockjs": "webpack ./client-src/sockjs/index.js -o client/sockjs.bundle.js --color --config client-src/sockjs/webpack.config.js",
     "build:client": "rimraf ./client/* && npm-run-all -s -l -p \"build:client:**\"",
-    "webpack-dev-server": "cd $INIT_CWD && node ../../../bin/webpack-dev-server.js",
```
hiroppy: Don't delete this. This is for samples.

evilebottnawi: @hiroppy it doesn't worj on windows, also we need this line when we don't have tests (for manual testing), but right now we have many tests and good helpers, so i think it is legacy trash and we can remove this safely

anikethsaha: > node ../../../bin/webpack-dev-server.js",

also is this path working? 
if we are keeping the script then I guess we need to consider `cross-env`

Thoughts ?

anikethsaha: cc @evilebottnawi @hiroppy

CONTRIBUTING.md
```diff
@@ -52,10 +52,61 @@ Pull requests often need some real-world testing.
...
+-   misc
+-   tests
+
+The **header** is mandatory.
+
+Any line of the commit message cannot be longer 100 characters. This allows the message to be easier
+to read on GitHub as well as in several git tools.
+
+For more information about what each part of the template mean, head up to the documentation in the
+[angular repo](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format)
```
evilebottnawi: let's remove this from this issue and send other PR with this update, also we use convetial-commits, so type is invalid

anikethsaha: type in the commit format ?

anikethsaha:
@evilebottnawi  please review again 👍

package.json
hiroppy: > it doesn't work on windows, also we need this line when we don't have tests (for manual testing), but right now we have many tests and good helpers, so i think it is legacy trash and we can remove this safely

How to build the examples? We need this line.

evilebottnawi: @hiroppy hm, do you use this command? I just use `node bin/webpack-dev-server` in command line

hiroppy: @evilebottnawi Yes, because README says to use it.😅 e.g. https://github.com/webpack/webpack-dev-server/tree/master/examples/cli/bonjour

anikethsaha: >  We need this line. If this line doesn't work on Windows, we should change this line.

the `cd $INIT_CWD` should be changed with `cd %INIT_CWD%` if it's in windows! but then it won't work in other OS 😬 . 
what can be done? `cross-env` ? but I don't think its worth using and increasing dependency just for this !

hiroppy: > but I don't think its worth using and increasing dependency just for this !

I don't think so, because we have already had many examples. If you delete this line, we should change all examples build style. I recommend adding `cross-env` as devDependencies if we want to run on Windows.

anikethsaha: okay ! it seems we can add it then !

anikethsaha: Will add a seperate PR for that, we can discuss clearly there I guess

anikethsaha: @hiroppy done ! 👍

Loonride: @anikethsaha I'm not sure if you were working on this but I couldn't get `cross-env` to do this successfully on Windows so I wrote my own solution here: https://github.com/webpack/webpack-dev-server/pull/2112

anikethsaha: Yeah cross env was not the working for me too.

PR #2103, opened 2019-07-05
refactor: setupExitSignals handle server close

noscripter:
setupExitSignals should handle server close event while it is always undefined

- [ ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

setupExitSignals should handle server close event while it is always undefined


evilebottnawi:
Already fixed

PR #2106, opened 2019-07-06
feat(server): add stdin for API

### LINKED ISSUE #1712: [feature request] --stdin as option in config
System:
    OS: macOS 10.14.3
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
  Binaries:
    Node: 8.10.0 - ~/.nvm/versions/node/v8.10.0/bin/node
    Yarn: 1.9.4 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v8.10.0/bin/npm
  Browsers:
    Chrome: 72.0.3626.121
    Firefox: 65.0.1
    Safari: 12.0.3

- [ ] This is a **bug**

### Code

Would be really useful to be able to do:
```
module.exports = {
  //...
  devServer: {
    stdin: true
  }
};
```

instead of having to run webpack-dev-server with --stdin on CLI only [as specified in the docs](https://webpack.js.org/configuration/dev-server/#devserverstdin---cli-only)

### For Features; What is the motivation and/or use-case for the feature?

Some popular dev tools (like vue-cli) have opinionated wrappers around webpack-dev-server that make it only possible to configure webpack via their own config files (where they pass along options to webpack-dev-server's `devServer` configuration object).

Unfortunately, this obscures really useful flags like stdin

Thank you for the amazing software!

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

fixes #1712

This makes `stdin` an option for both CLI and API, and helps in moving configuration changes out of the CLI for CLI refactor.

### Breaking Changes

None


lib/Server.js
```diff
@@ -58,6 +59,8 @@ class Server {
     this.compiler = compiler;
     this.options = options;
 
+    handleStdin(options);
```
evilebottnawi: hm, why we need it here?

Loonride: @evilebottnawi Do you think no helper would be better? I just thought having helper would be cleaner

evilebottnawi: i think we need this inside server file, it is logic for CLI only

lib/Server.js
```diff
@@ -59,7 +58,16 @@ class Server {
...
+    if (options.stdin) {
+      // listening for this event only once makes testing easier,
+      // since it prevents event listeners from hanging open
+      process.stdin.once('end', () => {
+        // eslint-disable-next-line no-process-exit
+        process.exit(0);
+      });
+
+      process.stdin.resume();
+    }
```
evilebottnawi: No, i don't mean we don't need help, i just answer why we use this here?

Loonride: @evilebottnawi The goal is to make this option not CLI-only. As mentioned in https://github.com/webpack/webpack-dev-server/issues/1712 it is desired for this to be possible.

Also this is beneficial because I want to reduce what the CLI does as much as possible for the refactor.

hiroppy:
I think tests on Mac will be broken these changes.

```
FAIL test/cli/cli.test.js (102.875s)
  ● CLI › --stdin without stdin end should time out

    Timeout - Async callback was not invoked within the 60000ms timeout specified by jest.setTimeout.Error: 

      119 |   });
      120 | 
    > 121 |   it('--stdin without stdin end should time out', (done) => {
          |   ^
      122 |     const configPath = resolve(
      123 |       __dirname,
      124 |       '../fixtures/simple-config/webpack.config.js'

      at Spec (node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.it (test/cli/cli.test.js:121:3)
      at Object.describe (test/cli/cli.test.js:17:1)
```

EslamHiko:
@Loonride since you are testing after calling the helper, that ensures that `process.stdin.resume()` has been called I think you can check for this flag `process.stdin._readableState.paused` it's false when stdin is enabled. instead of waiting for it to time out.

Loonride:
@EslamHiko I'm not seeing the `process.stdin._readableState.paused` on Windows 7, Node v8.10.0, so I'm not sure if that is the right flag to look at. But good suggestion, I will see if something else from `_readableState` can be used to check this

Loonride:
I stuck with a solution of waiting for a timeout, because working with child process `stdin` seemed unreliable when I tried to use `stdin.isPaused()`.

Loonride:
Oddly inconsistent behavior with `execa` where `timedOut` is undefined on MacOS. Seeing if checking for `failed` property will work on both. Possibly an `execa` issue?

evilebottnawi:
@Loonride looks like problem on execa side? Can you try next version of execa, maybe they fixed this, if it is works we can minimum required nodejs version for devDependcies to `8.9` and describe this in Contributing guide

Loonride:
> @Loonride looks like problem on execa side? Can you try next version of execa, maybe they fixed this, if it is works we can minimum required nodejs version for devDependcies to `8.9` and describe this in Contributing guide

@evilebottnawi why node `8.9`? It looks like tests are passing in node `6`, and the upgrade seems to have solved it.

evilebottnawi:
don't worry, CircleCi has ability to start CI worker and connect them using ssh, maybe azure has some feature and you can debug problem on CI machine  (usisng CLI), need investigate

EslamHiko:
@Loonride that's strange I test on windows 10 and it reads `process.stdin._readableState.paused` without problems. this is a screenshot I hope it helps.
Capture-stdin

hiroppy:
@Loonride Thanks. Yep, so don't worry.

Loonride:
Closing in favor of https://github.com/webpack/webpack-dev-server/pull/2186

PR #2107, opened 2019-07-07
chore: added type checking using comments

anikethsaha:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Added script to do type linting in build

### Motivation / Use-Case
As this package as quite a more number of utility methods and logics spread. Type checking is quite needful here. Migrating whole to typescript, for now, is quite a work. 
So as discussed with @evilebottnawi , it's better to do the type checking in comments as `webpack` does too. 

### Breaking Changes
No
Just an addition of `tsconfig.json` in root and `lint:type` in CI

### Additional Info
This type-checking is done with reference to [`jsDoc`](https://jsdoc.app/index.html)

**Addition dependency** : 
- `typescript` as devDependency

As of now, I have refactored `addEntries.js`, and for time being, only included `lib/utils/addEntries.js` in the tsconfig. as the completing of the changing to dir/files, the tsconfig will be changed.

These changes have no effect on the main logic or the core of the codebase.


hiroppy:
In my option, maintenance of jsdoc is difficult and the possibility of information becoming old becomes high. If you want to introduce types, I think you should move to typescript. If the migration is hard, I will do it. Anyway, I'm -1 on this changes.

anikethsaha:
I am not aiming to create a doc with this, just as type checking in lint! that too not for all directories, just those utilities!
Yeah, typescript migration is more efficient for sure but this can fill the time gap and also it will be 100% with typescript too!

hiroppy:
We've already had `next` branch, so we can move to typescript in this branch. If you want a type, I will be able to build the base. I think that comments are incomplete and do not make sense.

hiroppy:
@evilebottnawi  I wanna know what you think. Honestly, I wanna migrate to typescript, so if you agree with me, I'll migrate to typescript on next branch.

hiroppy:
sorry but I can't agree with these changes. First, if we want to migrate to typescript, please show me the roadmap. Secand, jsdoc is a bad idea, typescript has allowJS option, so we can migrate to typescript slowly. Also, learning jsdoc costs as same as learning typescript. Sorry but I’m sure this is technical debt. If you use allowJS option, I’ll agree with you. Anyway, first, we must decide we want to migrate to typescript or not. Really sorry.

evilebottnawi:
@hiroppy 
> we must decide we want to migrate to typescript or not

I think yes, sometimes i look on code in some places and can't remember possible types and in some places it is help to maintain, i don't want use pure typescript because it can be hard for other developers to send PRs with fixes so i think using types through JSDocs is good idea. It is not hard priority for us bug it is help to us.

> If you use allowJS option, I’ll agree with you.

Can you clarify?

anikethsaha:
I think if we migrate with removing the `allowjs` options, it will be a big task!.
Keeping `allowjs` and migrating step by step can be done!

For me, JSDocs is an option which we need even after migrating to TS, not just for now, plus it's giving us the advantage of having a type checking in build too. and without any major change!

Having this and then migrating to TS will help a lot with the migration as we would be having the types in the comment
JSDocs is having good docs so it's not quite a bit of load for us too.


*I think webpack core package might have faced the same problem of too much load with migration that's why they have this in the codebase*

hiroppy:
**If we use JSdoc, when will we move to TypeScript?** If you do not make it clear, you can not make a migration forever. 

## why I think we should avoid Jsdoc?

- comments are basically stale and often forgotten
- types in jsdoc are very weak
- types should be inferred as far as possible

## typescript is difficult or not?

Many users have already used typescript, and I think typescript is not difficult because we can use `any` type. Also, it is easy to move to typescript when we can use any type in the first step because TypeScript owns only types, namespaces(not recommended), interfaces and enums(not recommended). Compatible with es2015 and not complicated.

https://www.npmjs.com/package/typescript

## my idea

If you want to migrate to typescript step by step, you can change only that file as typescript using allowJS.

## finally

I have experience of moving from flow to typescript in a large-scale product. There is no need to make the type strong from the beginning.

evilebottnawi:
> If we use JSdoc, when will we move to TypeScript? If you do not make it clear, you can not make a migration forever.

Should we migrate on typescript fully? webpack use JSDoc for type and all fine so we can do this also, also you can find other popular packages what written on js  and use ts only as linting, for example https://github.com/sindresorhus/execa

> comments are basically stale and often forgotten

It is not hard to track, also we ccan configure typescript report about this, types should help to us and don't increase work time on issues/features so JSDos is ok solution

> types in jsdoc are very weak
> types should be inferred as far as possible

look at it from the side that this is just another linter for our code, it is not new language or something also, just linter as eslint

> typescript is difficult or not?

yep, typescript is not difficult, rewriting can take us time and for novice users who very often uses webpack, it can be difficult, so will be great keep code in js and use types as part of lint process

anikethsaha:
> If you want to migrate to typescript step by step, you can change only that file as typescript using allowJS.

agreed 👍 
 
> typescript is difficult or not?

yeah, we can use `any`, but I think a stable release should come with `strict` on in `tsconfig` which will throw with any, so again we need to revisit that any to resolve it. 

**But for time being sure we can use `any` and `allowJs` 👍**

### Then as per my view if the whole need to migrate to TS (even with step by step using `allowJS`) , it should not be included in this release at least. 

For me, we can do both and both seem fair to me, I will give more preference to **JSDoc -> TS(with JSDoc)**

sokra:
Here my 2 cents.

* Choose whatever approach you feel comfortable with the most, but agree on one
  * I don't care that much, both approaches are already used in the webpack orgs
* Both approaches have pro and cons
* jsdoc can't get stale as it is checked by tsc on CI
* jsdoc can be forgotten when not using strict mode, similar to types in TS
* types in jsdoc are as strong as in TS, both using the typescript compiler under the hood. Sometimes you need a `@typedef` to create valid JSdoc, where you could inline the code in TS
* Both TS and JSdoc uses the same type inference: You need types at the same places in TS or JSDoc
* JSDoc is a bit more verbose when not adding descriptions and a bit less verbose when adding descriptions (You need JSDoc in TS to add description comments)
* JSDoc is a bit more verbose when using generic types: `let x = new Set<Node>()` vs `/** @type {Set<Node>} */ let x = new Set();`
* You can't use `import` with JSDoc, while when using TS you could compile it to `require` easily
* With JSDoc you can publish without compile step
  * github and npm code are identical
  * Users could consume with github dependency, branches and pull requests can be consumed easily
  * `yarn link` is simpler

A technical clarification: JSDoc means "Using `tsc --noEmit --allowJs` and add type info via JSDoc where needed"

hiroppy:
@sokra Thank you for your answering.

As I said, 

> If we use JSdoc, when will we move to TypeScript? If you do not make it clear, you can not make a migration forever.

First, I hope we must decide on this choice. If we don't migrate to typescript, I agree to jsdoc, but I think it is easy to migrate to typescript using `any`. If we migrate to typescript, I'll be able to pay the cost of migrating. fortunately,  our organization has jsdoc(webpack) and typescript(webpack-cli), so we have both pieces of knowledge. From now on, our code is bigger than now. I think this choice is important. Thanks.

evilebottnawi:
I prefer jsdoc, because it is easy for other developers contribute package, also it is allow to developers solve all on js and when if it is accepted, add types without additional time wasting. Also we can add types step by step without rewriting on typescript and do not spend a lot of time on this and the money of our sponsors

anikethsaha:
either of both looks good !...
just got to make a choice here!

Loonride:
I don't have much experience using either but it seems:

- JSDoc would be a positive change since it can't get stale via CI, and usually PRs are not introducing many new functions
- I don't really have enough experience with typescript to argue for one side or another (I will say that I like not having to recompile anything when working on the server here, but that shouldn't be a deciding factor)

hiroppy:
As I said above I prefer migrating to typescript to using JSdoc. But I respect @evilebottnawi opinion. Agree with you.

> I prefer jsdoc, because it is easy for other developers contribute package, also it is allow to developers solve all on js and when if it is accepted, add types without additional time wasting. Also we can add types step by step without rewriting on typescript and do not spend a lot of time on this and the money of our sponsors

hiroppy:
@evilebottnawi Yes, I wanna create a tracking issue like [this](https://github.com/nodejs/node/issues/11273) for contributors and us.

anikethsaha:
Can we have a review on this (on those types)??

cc @evilebottnawi  @hiroppy

Loonride:
@evilebottnawi I agree with what you said here:

> I prefer jsdoc, because it is easy for other developers contribute package, also it is allow to developers solve all on js and when if it is accepted, add types without additional time wasting. Also we can add types step by step without rewriting on typescript and do not spend a lot of time on this and the money of our sponsors

JSDoc seems easy enough to add in, and I agree that typescript can be an additional barrier for contributors.

PR #2112, opened 2019-07-09
fix(examples): make examples webpack-dev-server npm script work on Windows

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

As described here: https://github.com/webpack/webpack-dev-server/pull/2100#discussion_r300542619, `$INIT_CWD` does not work on Windows. I tried solving with `cross-env` but it did not work, so here is an alternative.

### Breaking Changes

None

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,6 @@
     "chokidar": "^2.1.6",
     "compression": "^1.7.4",
     "connect-history-api-fallback": "^1.6.0",
-    "cross-env": "^5.2.0",
     "debug": "^4.1.1",
     "del": "^4.1.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -40,6 +40,7 @@
     "chokidar": "^2.1.6",
     "compression": "^1.7.4",
     "connect-history-api-fallback": "^1.6.0",
+    "cross-env": "^5.2.0",
```
evilebottnawi: Can't find usage

Loonride: Yes just a mistake.

PR #2099, opened 2019-07-04
refactor(server): move publicPath changes from CLI to API

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**
- [x] This is a **code refactor**
- [ ] This is a **test update**
- [ ] This is a **docs update**
- [ ] This is a **metadata update**

### For Bugs and Features; did you add new tests?

Not yet

### Motivation / Use-Case

Move https://github.com/webpack/webpack-dev-server/pull/2098 to `next` branch.

### Breaking Changes

Yes

### Additional Info

The block I deleted from `createConfig` was changed since the last `next` rebase, but it should be easy to deal with.

PR #2113, opened 2019-07-10
refactor(server): move filename option from createConfig to API

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

Move default fallback of `options.filename` out of `createConfig`, since `createConfig` is only used on the CLI.

### Breaking Changes

If `filename` is not set, it will now default to the compiler's `options.output.filename`. However, having `filename` set will have no effect unless `lazy` is enabled, so this should not break anything. Does this warrant moving this to `next`?


hiroppy:
I think we solve this issue just by moving to `lib/utils/updateCompiler.js`. Is there any advantage to dividing it into utils other than making it easier to write tests?

Loonride:
> I think we solve this issue just by moving to `lib/utils/updateCompiler.js`. Is there any advantage to dividing it into utils other than making it easier to write tests? I think it is a good thing to standardize processing.

The only reason for the util is to make testing easier.

I was following the style that I did here: https://github.com/webpack/webpack-dev-server/pull/2099. However, I'm not convinced these things really belong in `updateCompiler`, since they are only using the compiler's data, not changing the compiler in any way.

### A more uniform idea that I have is:

I make a helper `updateOptions(compiler, options)`. It will be called in `Server.js` before anything else, and `updateCompiler` will be called after it. Its only purpose will be to change properties of the `this.options` object, like setting defaults and doing essentially what `createConfig` currently does. Then we will not have to branch out with many different helpers like I was doing in this PR.

Loonride:
/cc @evilebottnawi @hiroppy I switched to using `normalizeOptions` here.

Do you think I should proceed by working through these options one PR at a time, or by moving many in 1 PR?

hiroppy:
I'm fine either way. How much is the amount of changes?

Loonride:
> I'm fine either way. How much is the amount of changes?

@hiroppy Sorry for late response. I'm essentially just moving everything from `createConfig` to the new `normalizeOptions`. I will just put a quick draft together. I might close this PR and combine it with that draft, as this is potential small breaking changes.

Loonride:
closing in favor of https://github.com/webpack/webpack-dev-server/pull/2174

PR #2116, opened 2019-07-11
feat(server): add transportMode

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**
- [ ] This is a **code refactor**
- [ ] This is a **test update**
- [ ] This is a **docs update**
- [ ] This is a **metadata update**

### For Bugs and Features; did you add new tests?

Yes, based on moving around and changing existing `clientMode` and `serverMode` tests

### Motivation / Use-Case

https://github.com/webpack/webpack-dev-server/issues/1860

This replaces `clientMode` and `serverMode` options with the usage:

```
transportMode: {
  server: 'ws',
  client: 'ws'
}
```

### Breaking Changes

Experimental options `clientMode` and `serverMode` removed

### Additional Info

- `transportMode-option.test.js` is bulky and maybe should be broken down
- the configuration I added into `Server.js` could definitely be moved into some unified `updateOptions` helper to be tested more easily.

PR #2117, opened 2019-07-12
refactor(server): add updateOptions API helper

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

I am working on going through `createConfig` and moving any CLI-specific behavior to the API. Since there is only one `createConfig` file, I think it makes sense to parallel it with one API helper to update options and set defaults.

I think it will get out of hand to have a helper for each option that needs to have a default setting, such as what I am doing in:

- https://github.com/webpack/webpack-dev-server/pull/2099
- https://github.com/webpack/webpack-dev-server/pull/2116

So the above PRs need to be updated to use this if we agree on this solution.

### Breaking Changes

None

### Additional Info

I will add tests for `transportMode` later, since `serverMode` and `clientMode` will be removed.


hiroppy:
Sorry for the late reply.

> maybe we can move normalize all options inside this util

Agree 👌

Loonride:
> Good idea, maybe we can move normalize all options inside this util

I agree, but I don't think we should move them in this PR because some of them will cause breaking changes (since the normalizing done in `createConfig` is currently CLI only.)

hiroppy:
We might be better to change the base branch to the `next` for safety.

evilebottnawi:
@hiroppy i think don't need use next branch for this

Loonride:
@hiroppy Currently there are no breaking changes, I just moved stuff from the start of the `Server.js` file into the new helper. Then I plan to move other options from `createConfig` into here. When it is a breaking change I will add it in on `next`.

test/server/utils/updateOptions.test.js
```diff
@@ -0,0 +1,75 @@
...
+    },
+  ];
+
+  cases.forEach((data) => {
+    describe(data.title, () => {
+      let compiler;
+      beforeAll(() => {
+        let webpackConfig;
+        if (data.multiCompiler) {
+          // eslint-disable-next-line global-require
```
hiroppy: You can delete this comment

test/server/utils/updateOptions.test.js
```diff
@@ -0,0 +1,75 @@
...
+          webpackConfig = require('../../fixtures/multi-compiler-config/webpack.config');
+        } else {
+          // eslint-disable-next-line global-require
```
hiroppy: ditto

lib/utils/updateOptions.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+/* eslint-disable
+  no-undefined
+*/
+
+function updateOptions(compiler, options) {
```
evilebottnawi: @Loonride maybe `normalizeOptions` better name here? Anyway code looks good

PR #2123, opened 2019-07-15
chore(deps): update package-lock.json

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**
- [ ] This is a **code refactor**
- [ ] This is a **test update**
- [ ] This is a **docs update**
- [x] This is a **metadata update**

### For Bugs and Features; did you add new tests?

no

### Motivation / Use-Case

run `audit fix`.

```

>= 4.17.12

├─┬ @commitlint/cli@8.0.0
│ ├─┬ @commitlint/lint@8.0.0
│ │ ├─┬ @commitlint/parse@8.0.0
│ │ │ ├─┬ conventional-commits-parser@2.1.7
│ │ │ │ └── lodash@4.17.14  deduped
│ │ │ └── lodash@4.17.14  deduped
│ │ ├─┬ @commitlint/rules@8.0.0
│ │ │ └─┬ @commitlint/ensure@8.0.0
│ │ │   └── lodash@4.17.11 
│ │ └── lodash@4.17.11 
│ ├─┬ @commitlint/load@8.0.0
│ │ ├─┬ @commitlint/resolve-extends@8.0.0
│ │ │ └── lodash@4.17.11 
│ │ └── lodash@4.17.11 
│ └── lodash@4.17.11 
```

These PRs have merged but they haven't published new version yet, so we cannot fix.(because they lock lodash's version)
https://github.com/conventional-changelog/commitlint/commit/4cb979df81328684467fd0dad130440ce725831d
https://github.com/conventional-changelog/commitlint/pull/724

### Breaking Changes

no

PR #2124, opened 2019-07-15
chore(deps): update husky and lint-staged

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

no

### Motivation / Use-Case

Wait for #2123.

closes: #2085
closes: #2087

### Breaking Changes

no


.prettierignore
```diff
@@ -4,3 +4,4 @@ coverage
 client-src/live/web_modules/
 node_modules
 CHANGELOG.md
+package-lock.json
```
evilebottnawi: We can remove this from here, because prettier doesn't change format of package-lock.json

wizardofhogwarts: > We can remove this from here, because prettier doesn't change format of package-lock.json

I agree.

hiroppy:
@evilebottnawi PTAL

PR #2125, opened 2019-07-15
refactor(routes): refactor utils/routes.js

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

yes

### Motivation / Use-Case

just refactoring.

### Breaking Changes

no

### DIFF
--- a/lib/utils/routes.js
+++ b/lib/utils/routes.js
@@ -3,37 +3,31 @@
 const { createReadStream } = require('fs');
 const { join } = require('path');
 
+const clientBasePath = join(__dirname, '..', '..', 'client');
+
 function routes(app, middleware, options) {
   app.get('/__webpack_dev_server__/live.bundle.js', (req, res) => {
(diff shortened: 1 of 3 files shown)


evilebottnawi:
@hiroppy What do you think about routes for next version? I think we should remove some of them

hiroppy:
@evilebottnawi You mean some endpoints are unnecessary?

evilebottnawi:
@hiroppy yes, i think we don't need this because it is legacy stuff

hiroppy:
All endpoints?

evilebottnawi:
@hiroppy need investigate using `index.bundle.js` directly is invalid, because it is require some options for detect what transportMode we use, also socker options can be invalid, i think we should remove usage using `script` tag with this route

hiroppy:
@evilebottnawi ok, understood. I'll investigate it. May I merge this pr?

Issue #1831, opened 2019-04-29
Remove Sock injection when options.hot is false?


plauclair:
- Operating System: macOS 10.14.4
- Node Version: 10.15.0
- NPM Version: 6.9.0
- webpack Version: 4.30.0
- webpack-dev-server Version: 3.3.1

- [ ] This is a **bug**
- [x] This is a **modification** request

### Code

```js
const serverOptions = { hot: false };
```

### Expected Behavior

When setting hot to false, socket is still added as a dependency. Maybe I understand wrong, but shouldn't dev server _only_ serve when `options.hot` is false?

### Actual Behavior

Current code always injects Socket dependency even if hot is disabled. Currently this can be achieved through `options.inline = false` but that seems to be something completely different.

https://github.com/webpack/webpack-dev-server/blob/master/lib/utils/addEntries.js#L21

### For Features; What is the motivation and/or use-case for the feature?

We use dev server to serve files in an environment that does not support web sockets. Having the socks module crashes the process even when hot is false.

evilebottnawi:
hot !== sockjs, we use sockjs for getting new assets, maybe we can add option `sockjs: true | false`, PR welcome

plauclair:
@evilebottnawi would gladly submit a PR. Anything special that comes to mind that would need to be taken into account?

evilebottnawi:
I think no

yordis:
> hot !== sockjs, we use sockjs for getting new assets,

@evilebottnawi would you mind to mind to explain what do you mean?

Also, the issue I am facing is that I am the socket client is using `window.location.{hostname, port}` I believe and I am including my JavaScript in a second server.

Rails Server: `http://localhost:3000`
Webpack Dev Server: `http://localhost:5000`

Because of this, I keep getting errors

```
sockjs.js:1793 GET http://192.168.141.56:3000/sockjs-node/info?t=1556570647605 net::ERR_CONNECTION_REFUSED
```

Because my dev-server is not running in `3000`.

I think we should take some extra configs to be able to change the `hostname` and the `port`.

evilebottnawi:
We should not add client when hot is false, because it is doesn't have sense

evilebottnawi:
@plauclair i was wrong, we should not include client when `hot` and `liveReload` is false

plauclair:
Makes sense, we have an issue open here, I'll try to squeeze in something this week. This makes it much easier to go about it.

Simon-Tang:
[...]

PR #2133, opened 2019-07-18
fix: disable socket injection when hot & liveReload are disabled

EslamHiko:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case
Solves : https://github.com/webpack/webpack-dev-server/issues/1831

### Breaking Changes
No


test/e2e/__snapshots__/Iframe.test.js.snap
```diff
@@ -24,7 +24,8 @@ Array [
 
 exports[`Client iframe console.log liveReload disabled 1`] = `
 Array [
-  "Hey.",
+  "Failed to load resource: the server responded with a status of 404 (Not Found)",
+  "Failed to load resource: the server responded with a status of 404 (Not Found)",
```
evilebottnawi: Looks something wrong

EslamHiko: yes liveReload is disabled and hot is disabled too that's why the WebSocket is not working

evilebottnawi: :+1:

Loonride: I think it is possible that this could be a variable number of 404 error messages, since the client keeps retrying to connect. Maybe this should be handled some way for this particular test?

test/client/clients/WebsocketClient.test.js
```diff
@@ -60,3 +66,99 @@ describe('WebsocketClient', () => {
     });
   });
 });
+
+describe('websocket client injection', () => {
```
Loonride: This should be moved to the `test/server` directory. Either in `Server.test.js`, or possibly a new test file of its own. @evilebottnawi thoughts on where it should go exactly?

evilebottnawi: @EslamHiko yep, let's move this test to `test/server`
@Loonride maybe `test/server/hot-and-liveReload-integration`, no idea how better naming this :smile:

EslamHiko: @Loonride @evilebottnawi okay, will move it to `test/server/socket-injection` what do you think about the name? 😅

evilebottnawi: `test/server/socket-injection.test.js` :+1:

test/server/socket-injection.test.js
```diff
@@ -0,0 +1,106 @@
...
+    it('should be injected', (done) => {
+      req
+        .get('/sockjs-node/info')
+        .expect(200)
+        .then(({ res }) => {
+          expect(JSON.parse(res.text).websocket).toBe(true);
+          done();
+        });
+    });
+  });
```
evilebottnawi: Let's add test where `hot: false, liveReload: true` for explicit check

test/e2e/__snapshots__/Iframe.test.js.snap
EslamHiko: @Loonride any idea to handle these 404 errors?

test/e2e/__snapshots__/ClientOptions.test.js.snap
```diff
@@ -25,6 +25,9 @@ Array [
 exports[`Client console.log liveReload disabled 1`] = `
 Array [
   "Hey.",
+  "Failed to load resource: the server responded with a status of 404 (Not Found)",
+  "[WDS] Disconnected!",
```
EslamHiko: should I add a check to disable `"[WDS] Disconnected!"` message when `hot` &  `liveReload` are disabled ? @Loonride @evilebottnawi

evilebottnawi: I think yes, better output message like `[WDS] hot and live reload was disables`

client-src/default/index.js
```diff
@@ -136,7 +136,9 @@ const onSocketMessage = {
     log.error(error);
   },
   close() {
-    log.error('[WDS] Disconnected!');
+    if (options.hot || options.liveReload) log.error('[WDS] Disconnected!');
+    else
+      {log.error('[WDS] Hot Module Replacement & Live Reloading are disabled!');}
```
evilebottnawi:
```
if (condition) { 
 // Logic
} else {
  // Logic
}
```

lib/Server.js
```diff
@@ -737,7 +737,9 @@ class Server {
     this.hostname = hostname;
 
     return this.listeningApp.listen(port, hostname, (err) => {
-      this.createSocketServer();
+      if (this.options.hot || this.options.liveReload !== false) {
```
hiroppy:
```diff
- if (this.options.hot || this.options.liveReload !== false) {
+ if (this.options.hot || this.options.liveReload) {
```

EslamHiko: @hiroppy sorry for my late reply but I think if I changed it, it won't create the socket server if options.liveReload was `undefined` the default value. I suggest making little refactor and write `this.options.liveReload = this.options.liveReload !== false` at the beginning of the constructor and replace all `this.options.liveReload !== false` with `this.options.liveReload`.

evilebottnawi: /cc @hiroppy

hiroppy: ok, we will fix this in the next vesion.

test/e2e/__snapshots__/Iframe.test.js.snap
Loonride: > @Loonride any idea to handle these 404 errors?

@EslamHiko Sorry I missed your reply, it seems tests generally pass as the timing is consistent to only output the error message twice, but just in case:

You could check that every element in that array is the same. Then make the snapshot only test the first element of the output array. This will allow for a variable number of the same error message.

Loonride:
@evilebottnawi Is it possible this should be on `next`? It is a good idea, but it is conceivable that users might still want the socket server running but do not have the right server settings. For example, someone could have the server running and are expecting to connect to the socket with a custom client not served by webpack-dev-server.

evilebottnawi:
@Loonride hm, what is use case for this action?

Loonride:
> @Loonride hm, what is use case for this action?

I can't think of a good use case. I was thinking hypothetically someone could connect to the socket server without using the served client, but there is likely no use for this.

Anyway, LGTM.

EslamHiko:
I failed to rebase & created a fresh PR here: https://github.com/webpack/webpack-dev-server/pull/2601

PR #2143, opened 2019-07-22
fix(server): set port before instantiating server

### LINKED ISSUE #2142: Websocket and /info requests go to wrong domain when using default port
- Operating System: osx
- Node Version: 12.6.0
- NPM Version: 6.9.0
- webpack Version: 4.36.1
- webpack-dev-server Version: 3.7.2
- Browser: Firefox and Chrome (both)

- [x] This is a **bug**

### Code
(See repo that demonstrates the problem: https://github.com/joeldenning/webpack-dev-server-ws-bug)

Run `webpack-dev-server` without a port specified (not through the CLI nor webpack config). The server will be automatically assigned a default port, usually 8080.

Now start a static http server on a different port (let's say port 5000). Create an html file that loads the webpack bundle: `<script src="http://localhost:8080/main.js">`.

Make sure that webpack-dev-server is configured to reload the page whenever the bundle is recompiled. The webpack-dev-server client will attempt to open a web socket to the wrong host. When the web socket fails, it will start making XHR requests to `/info` in an infinite loop, once every few seconds. Each of the requests will fail because they are to the wrong host.

### Expected Behavior

The websocket and `/info` requests should go to `http://localhost:8080`. For example the XHR request for `/info` should go to `http://localhost:8080/info`.

### Actual Behavior

The web socket and XHR requests go to http://localhost, without the `:8080` port on it. For example, the `/info` request goes to `http://localhost/info`.

Note that if you specify `--port 8080` when you run webpack-dev-server, that the behavior is corrected and the `/info` request goes to `http://locahost:8080/info`.

webpack-dev-server should work without having to specify a port.

https://github.com/joeldenning/webpack-dev-server-ws-bug

joeldenning:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

This resolves #2142.

### Breaking Changes

No breaking changes.

### Additional Info

The reason this fixes the bug is that we are setting `options.port` **before** calling `new Server()` instead of afterwards. The constructor in Server calls function that need the port to already be present.


lib/utils/processOptions.js
```diff
@@ -23,7 +24,21 @@ function processOptions(config, argv, callback) {
...
+    findPort(options.port)
+      .then((port) => {
+        options.port = port;
+        callback(config, options);
+      })
+      .catch((err) => {
+        // eslint-disable-next-line no-console
+        console.error(err.stack || err);
+        // eslint-disable-next-line no-process-exit
+        process.exit(1);
```
Loonride: I think it should be `throw err;` here as it was before

Loonride:
This solution looks good to me, just one note.

@evilebottnawi Unfortunately, this problem will make it difficult, or maybe impossible, to move `findPort` into the API for the CLI refactor.

test/cli/cli.test.js
```diff
@@ -78,13 +78,22 @@ describe('CLI', () => {
     testBin('--sockPath /mysockPath')
       .then((output) => {
         expect(
-          output.stdout.includes('http://localhost&sockPath=/mysockPath')
+          output.stdout.includes('http://localhost:8080&sockPath=/mysockPath')
```
hiroppy: Don't use the port number directly. You can use [this file](https://github.com/webpack/webpack-dev-server/blob/master/test/ports-map.js).

Loonride: @hiroppy One problem with using the `ports-map` is that this is intentionally testing the case where the CLI defaults to a port when not specified. In such test cases, maybe `findPort` should be mocked to return a specific port.

lib/utils/processOptions.js
joeldenning: Doing so would cause an unhandled promise rejection, which might go unnoticed by the user of webpack-dev-server. Right? Why is a rejected promise better than exiting the process?

I chose to follow the pattern [here](https://github.com/webpack/webpack-dev-server/pull/2143/files#diff-08aa20d4f0b23d049d0ea9d0377a03d8R16) since it seemed more likely to be helpful to the user. Am very happy to be corrected, though, if I'm missing something :)

joeldenning: To add to this, my understanding is that `.catch(err => {throw err})` is essentially a no-op that just makes an already rejected promise remain rejected with the same err. So I went looking for something better and found the process.exit() code from a few lines above my code.

Loonride: You are right, this looks better

test/cli/cli.test.js
joeldenning: I just tried to mock `findPort` and `defaultPort` with `jest.mock()`. However, they do not work because `testBin` spawns a child process that does not have jest's mocked modules in it.

I think that for the purposes of this test, it's not important that findPort finds any particular port as much as it's important that it finds a port at all. I think I'll change the tests to just verify that a port was found at all, by matching the output to a regex `/localhost:[0-9+]/`

Thoughts?

joeldenning: I just pushed code that does what I suggested ^

Let me know if you know of a better way to do this.

joeldenning:
The failed tests are related to a "page crashed" error that seems to be unrelated to my code changes.

test/cli/cli.test.js
Loonride: Seems fine to me. One thing I have been thinking to make sure tests like this do not attach to a port that later creates a collision is to move the hard [port mappings](https://github.com/webpack/webpack-dev-server/blob/05bdb0ce0738c1d014c949575d15e77b6132ed25/test/ports-map.js#L49) up by 10, so that 8080, 8081, 8082, etc. are guaranteed available for these variable port tests. Not needed in this PR though, just an extra precaution.

joeldenning: 👍 I won't add that now but yeah I agree that would be cool in a future PR

joeldenning:
@Loonride @hiroppy anything else before this is good to merge?

test/cli/cli.test.js
hiroppy: Yes, we need to reserve some ports for CLI tests.

Loonride:
Looks good to me.

joeldenning:
Anybody know why one of the azure checks is still pending? When I go to Azure it says that everything passed. Is it because a commit on master isn't merged into this branch?

I'm trying out merging master into this branch again.

Loonride:
@joeldenning It looks like a Page crashed issue. Not your fault, there is still some instability with the e2e tests here. Your PR can still be merged if a couple CI builds are failing/timing out for this particular reason since your changes are not causing it.

joeldenning:
All checks are now passing, after merging master in to retrigger the build.

Time to merge?

mistic:
@evilebottnawi could you merge this?

PR #2148, opened 2019-07-23
test(server): increase port mapping base number

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

N/A

### Motivation / Use-Case

As mentioned in https://github.com/webpack/webpack-dev-server/pull/2143, if we need to test CLI default port selection, which will use ports 8080, 8081, 8082... we should make sure the hard port mapping does not use these ports near port 8080. We just need to provide enough room for this default port assignment so that there will never be collisions.

### Breaking Changes

None

### DIFF
--- a/test/ports-map.js
+++ b/test/ports-map.js
@@ -46,7 +46,7 @@ const portsList = {
   Iframe: 1,
 };
 
-let startPort = 8079;
+let startPort = 8089;
 const ports = {};
 
 Object.keys(portsList).forEach((key) => {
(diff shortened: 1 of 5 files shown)


test/ports-map.js
```diff
@@ -46,7 +46,7 @@ const portsList = {
   Iframe: 1,
 };
 
-let startPort = 8079;
+let startPort = 8089;
```
hiroppy: Why did you change this? I thought it was only necessary to change the number of cli from 2 to 10.

Loonride: @hiroppy There are other tests that use the default ports near 8080, such as: https://github.com/webpack/webpack-dev-server/blob/master/test/server/utils/findPort.test.js. So I think we should provide some extra room to make sure those ports are available.

PR #2163, opened 2019-07-30
test(e2e): More e2e test improvements

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

- There was some problematic logic in `ClientOptions.test.js`
- `Client.test.js` and `Progress.test.js` could potentially write to the same `main.css`, so I split their tests to use two different configs

### Breaking Changes

None

### DIFF
--- a/client-src/clients/SockJSClient.js
+++ b/client-src/clients/SockJSClient.js
@@ -12,7 +12,8 @@ module.exports = class SockJSClient extends BaseClient {
     this.sock = new SockJS(url);
 
     this.sock.onerror = (err) => {
-      console.error(err);
+      // TODO: use logger to log the error event once client and client-src
+      // are reorganized to have the same directory structure
     };
   }
 
--- a/client-src/clients/SockJSClient.js
+++ b/client-src/clients/SockJSClient.js
@@ -10,6 +10,8 @@ module.exports = class SockJSClient extends BaseClient {
   constructor(url) {
     super();
     this.sock = new SockJS(url);
+
+    this.sock.onerror = () => {};
   }
 
   static getClientPath(options) {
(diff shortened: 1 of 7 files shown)


Loonride:
/cc @evilebottnawi I don't think just adding an empty `client.onerror` handler is a breaking change, right? Or is it preventing something from being thrown that would normally get thrown?

I got the idea that this might solve the problem from: https://github.com/websockets/ws/issues/246, https://github.com/websockets/ws/issues/656, but not entirely sure why it would change anything

Loonride:
Page crashes still happening. I've thought about making some mechanism to retry the test if a page crashes, but that might just be sweeping an issue under the rug.

Maybe we should just wait until there is further progress on this https://github.com/GoogleChrome/puppeteer/issues/1454

client-src/clients/SockJSClient.js
```diff
@@ -11,7 +11,9 @@ module.exports = class SockJSClient extends BaseClient {
     super();
     this.sock = new SockJS(url);
 
-    this.sock.onerror = () => {};
+    this.sock.onerror = (err) => {
+      console.error(err);
```
evilebottnawi: We should use logger, instead `console`

Loonride: @evilebottnawi This is another case where the path from `SockJSClient` to `utils/log` is not the same in the `client-src` directory compared to the `client` directory. I think we should wait and add this after reorganizing `client` directories on `next`, rather than do more of this redirection: https://github.com/webpack/webpack-dev-server/blob/c2da5322b078dbe9c3f6580ca2f301ba3e282452/client-src/default/webpack.config.js#L21.

Loonride: Let me remove the `console.error` for now, since it did not exist before and we won't introduce anything new in console output.

PR #2174, opened 2019-08-03
refactor(options): Move options from createConfig to normalizeOptions

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Not yet

### Motivation / Use-Case

Currently conceptualizing how options handling should be moved into the API. There will be many minor breaking changes, but I think they will all ultimately be for the better.

### Breaking Changes

Read the comments in `normalizeOptions`.


lib/utils/createConfig.js
```diff
@@ -202,16 +154,12 @@ function createConfig(config, argv, { port }) {
...
     options.openPage = argv.openPage;
   }
 
-  if (typeof argv.open !== 'undefined') {
-    options.open = argv.open !== '' ? argv.open : true;
-  }
-
-  if (options.open && !options.openPage) {
-    options.openPage = '';
+  // eslint-disable-next-line no-undefined
```
hiroppy: It seems unnecessary. Or if we need it, We'll be able move this rule to .eslintrc.

lib/utils/normalizeOptions.js
```diff
@@ -4,7 +4,119 @@
...
+  // not using Object.assign because we want the current options object to remain,
+  // but populate it with data from devServer if it does not already exist
+  Object.keys(devServerOptions).forEach((key) => {
+    if (options[key] === undefined) {
+      options[key] = devServerOptions[key];
+    }
+  });
+
+  if (!options.publicPath) {
+    // eslint-disable-next-line
```
hiroppy: why?

lib/utils/normalizeOptions.js
```diff
@@ -4,7 +4,119 @@
...
+
+  // options that default to true:
+  if (options.liveReload === undefined) {
+    options.liveReload = true;
+  }
+  if (options.inline === undefined) {
+    options.inline = true;
+  }
+
+  // ---- The options below were already here, they are not breaking changes ---- //
```
hiroppy: This comment is needed?

test/server/utils/createConfig.test.js
```diff
@@ -210,131 +209,11 @@ describe('createConfig', () => {
     expect(config).toMatchSnapshot();
   });
 
-  it('publicPath option (not specify)', () => {
-    const config = createConfig(webpackConfig, argv, { port: 8080 });
+  // removed: publicPath tests
```
hiroppy: ditto

test/server/utils/createConfig.test.js
```diff
@@ -210,131 +209,11 @@ describe('createConfig', () => {
...
-      Object.assign({}, webpackConfig, {
-        devServer: { watchOptions: { poll: true } },
-      }),
-      argv,
-      { port: 8080 }
-    );
-
-    expect(config).toMatchSnapshot();
-  });
+  // removed: watchOptions tests
```
hiroppy: ditto

test/server/utils/createConfig.test.js
```diff
@@ -402,55 +281,7 @@ describe('createConfig', () => {
...
-      }),
-      argv,
-      { port: 8080 }
-    );
-
-    config.contentBase = path.relative(process.cwd(), config.contentBase);
-
-    expect(config).toMatchSnapshot();
-  });
+  // removed: contentBase tests
```
hiroppy: ditto

lib/utils/normalizeOptions.js
Loonride: My mistake, these comments were just helping me stay organized while working on this

lib/utils/createConfig.js
Loonride: It is necessary, but I moved to the top of the `createConfig` file for convenience in this file. Let's remove the `no-undefined` requirement in another PR.

lib/utils/normalizeOptions.js
Loonride: This was there originally, but I don't know why. eslint is fine without it.

Loonride:
/cc @evilebottnawi @hiroppy Can you review this? I've moved most options from `createConfig` to `normalizeOptions`, skipping a few that will need some additional changes to be moved.

I've realized that one difficulty with the approach of merging the compiler's `options.devServer` with CLI options is default values for CLI. This is because you can't tell the difference between a CLI-set default value and a value that the user set explicitly. I'm hoping that maybe I can completely remove default CLI values on the CLI-side (with the exceptions probably just being host and port, as we need those for the `listen` method) then I can set default values in `normalizeOptions` if they are not explicitly set by the user.

hiroppy:
I wanna wait for https://github.com/webpack/webpack-dev-server/pull/2202 because this pr is related is-absolute-url.

hiroppy:
> I'm hoping that maybe I can completely remove default CLI values on the CLI-side

Agree. I hope so.

lib/utils/createConfig.js
```diff
@@ -1,21 +1,20 @@
 'use strict';
 
-const path = require('path');
-const isAbsoluteUrl = require('is-absolute-url');
+/* eslint-disable
+  no-undefined
+*/
```
evilebottnawi: Let's avoid this in favor `eslint-disable-next-line`

lib/utils/createConfig.js
```diff
@@ -1,21 +1,20 @@
...
 
-  // This updates both config and firstWpOpt
-  firstWpOpt.mode = defaultTo(firstWpOpt.mode, 'development');
-
   if (argv.bonjour) {
     options.bonjour = true;
   }
 
-  if (argv.host && (argv.host !== 'localhost' || !options.host)) {
+  if (argv.host) {
```
evilebottnawi: Why we remove `&& (argv.host !== 'localhost' || !options.host)`?

Loonride: We do not have to make this change now, but we need to carefully rethink how we handle the options: `host`, `port`, and `socket`, since these are relevant for the `listen` method.

The logic in `&& (argv.host !== 'localhost' || !options.host)` is made difficult by the fact that `localhost` is the CLI default, so we can't tell if the user explicitly asked for `localhost` or if it was set by default:

https://github.com/webpack/webpack-dev-server/blob/eab6acd2972e39c8b2d83f7a6614d40b8e14bf63/bin/options.js#L175

My proposal is:

**Don't set default values for the CLI options host or port**. I have done this in `webpack-cli serve`:
https://github.com/webpack/webpack-cli/blob/44d0fb01f47852b4dfc7423b8e1dfa6c51872e55/packages/serve/flags.js

**Have a clear hierarchy for which value gets used**. The order of precedence should be:
1. CLI value
2. webpackConfig.devServer value
3. default value

You can see I have implemented this order of precedence in these lines:
https://github.com/webpack/webpack-cli/pull/1011/files#diff-24921f08fe12bbbb2c25f5ed41d757b4R19-R23

evilebottnawi: Oh, i think we should keep https://github.com/webpack/webpack-cli/blob/44d0fb01f47852b4dfc7423b8e1dfa6c51872e55/packages/serve/flags.js on our side, because when we want add cli option it is require change something on webpack-cli side, it is not good

evilebottnawi: i.e. webpack-cli will work with `webpack-dev-server@3`?

evilebottnawi: @Loonride `webpack-cli` should not have `webpack-dev-server` in deps, they should use `peerDependenciesMeta`

Issue #1479, opened 2018-08-29
can't close webpack-dev-server --progress with Ctrl + C


danburzo:
* Operating System: macOS High Sierra (10.13.6)
* Node Version: v10.9.0
* NPM Version: 6.2.0
* webpack Version: 4.16.5
* webpack-dev-server Version: 3.1.5

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

Running the dev server with: `yarn webpack-dev-server --progress`. I'm not including webpack configs since seems to be reproducing on a variety of projects, so I assume it's something that happens regardless of the config? (Happy to provide more info otherwise).

### Expected Behavior

Ctrl + C stops the dev server.

### Actual Behavior

Ctrl + C does not do anything, dev server continues to compile modules.

_May be a regression of #860, #1360?_

michael-ciniawsky:
Could you try with `v3.1.6` ? https://github.com/webpack/webpack-dev-server/releases/tag/v3.1.6 (#1432)

danburzo:
Woo-hoo! Missed that there's 3.1.6 available, works like a charm with it. Thank you! :-)

qkdreyer:
Still an issue with [`v3.1.9`](https://github.com/webpack/webpack-dev-server/releases/tag/v3.1.9)

SebT:
Having the issue on Linux and Mac with webpack-dev-server `v3.1.8`.

evilebottnawi:
Please create minimum reproducible test repo with steps

superhawk610:
@SebT I've also had this issue, specifically when using the `-theme react` preset from the ReasonReact starter. I'm working on getting reproducible steps - it doesn't seem to happen in a dry environment, so I imagine it's some conflict with another file watcher that occurs after some period of use.

I'll report back if I get anything useful.

superhawk610:
The error (in this case, at least) appears to occur when the dev server gets stuck compiling. `Ctrl+C` at this point doesn't actually end any processes - all 3 processes listed above remain active after sending `SIGINT` to the yarn process.

After further attempts, this still doesn't reproduce the error consistently :/ FWIW, I'm using VSCode with the Reason/OCaml extension.

dcorb:
One developer in our team is having the same issue (the rest of the team not). So it's hard to offer a reproducible test.

evilebottnawi:
If somebody can reproducible test please post how here, we can fix it asap

Krzychek:
[...]

PR #2181, opened 2019-08-07
fix(server): fix setupExitSignals usage

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No, should I add tests for `setupExitSignals`?

### Motivation / Use-Case

Fixes problem with `setupExitSignals` mentioned in: https://github.com/webpack/webpack-dev-server/issues/1479

An alternative to this approach is simply to get rid of `setupExitSignals`, but I think it's better to keep the helper and just change the function a bit.

### Breaking Changes

Altered `setupExitSignals` function API.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -19,8 +19,11 @@ const getVersions = require('../lib/utils/getVersions');
 const options = require('./options');
 
 let server;
+const serverData = {
+  server: null,
+};
 
-setupExitSignals(server);
+setupExitSignals(serverData);
 
 // Prefer the local installation of webpack-dev-server
 if (importLocal(__filename)) {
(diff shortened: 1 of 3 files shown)


Loonride:
> Not sure, how it is solve problem

Before, we were just passing undefined into the function. The function could not see when the variable was set to something else outside of the function.

Here, I am passing an object into the function. The `setupExitSignals` helper will be able to see when the properties of this object are updated from outside of the function.

evilebottnawi:
@Loonride can we add tests fir this case?

Loonride:
> @Loonride can we add tests fir this case?

Yes I will do that soon.

Loonride:
@evilebottnawi It looks like husky is not working for me. Usually works for me on Windows 7, but I'm on Ubuntu 16.04 right now. Possibly just a bad `npm install` or something. Anyway, that means `commitlint` was not run here. Could this cause a problem with the commit messages?

evilebottnawi:
@Loonride 
> Could this cause a problem with the commit messages?

don't worry, i will fix commit message before merge (I often correct them for a better changelog), anyway using good commit message help me to do it :smile:

hiroppy:
Thanks

PR #2183, opened 2019-08-09
fix(tests): minor fix for transportMode.client test case

jamesgeorge007:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

N/A

### Motivation / Use-Case
There was a slight typo in #2116 that got unnoticed.

### Breaking Changes
N/A

### Additional Info
 N/A

### DIFF
--- a/test/e2e/TransportMode.test.js
+++ b/test/e2e/TransportMode.test.js
@@ -14,7 +14,9 @@ describe('transportMode client', () => {
     {
       title: 'sockjs',
       options: {
-        transportMode: 'sockjs',
+        transportMode: {
+          client: 'sockjs',
+        },
       },
     },
     {


jamesgeorge007:
cc @evilebottnawi

jamesgeorge007:
@hiroppy I guess you mean that it should be refactored to be:-
```js
options.transportMode: {
  client: 'sockjs'
}
```
:thinking:

jamesgeorge007:
@hiroppy didn't get you :thinking:

hiroppy:
https://github.com/webpack/webpack-dev-server/blob/master/lib/utils/normalizeOptions.js#L20-L25

jamesgeorge007:
@hiroppy so how should the respective case be instead :question:

jamesgeorge007:
```js
options: {
  transportMode.client: 'sockjs'
}
```
Is this the required way :thinking:

hiroppy:
Why do you introduce this change? 

We must add these cases.

string: transportMode: 'sockjs', 'ws'
object: transportMode.server, transportMode.client

And we've alerady had enough test cases.

jamesgeorge007:
I assumed this from the diff (#2116) such that the change was specific to `clientMode` (transportMode.client)

jamesgeorge007:
As seen from the test-case it is something specific to `transportMode client`:-
https://github.com/Loonride/webpack-dev-server/blob/aa17f7ca58eb3bd97a097f4901b58295acbe7334/test/e2e/TransportMode.test.js#L12-18

jamesgeorge007:
@hiroppy So this PR doesn't make any sense at all :thinking:

jamesgeorge007:
@evilebottnawi thoughts :thinking:

evilebottnawi:
@jamesblight what is problem here?

We support, `transportMode: 'ws` and `transportMode: { client: 'ws' , `server`: 'ws'}`?

jamesgeorge007:
@evilebottnawi as per what I see, the test was specific for `transportMode client` [here](https://github.com/Loonride/webpack-dev-server/blob/aa17f7ca58eb3bd97a097f4901b58295acbe7334/test/e2e/TransportMode.test.js#L12-19) 
As evident from the diff, the respective test case was specific to `options.clientMode` which should be set to `transportMode.client` right :thinking: 
Screenshot from 2019-08-09 15-47-41
Instead, the entire `transportMode` object was being set.

jamesgeorge007:
> @jamesblight what is problem here?
wrong user mentioned :sweat_smile:

jamesgeorge007:
> @jamesblight what is problem here?

wrong user mentioned :sweat_smile:

evilebottnawi:
I think test is right here

jamesgeorge007:
@evilebottnawi @hiroppy 
```js
options: {
  clientMode: 'sockjs'
}
```
was updated to be:-
```js
options: {
  transportMode: 'sockjs'
}
```
As per what I infer, instead of having `clientMode` and `serverMode` #2116 introduced `transportMode` as:-
```js
transportMode: {
  client: '',
  'server: ''
}
```

Hence, `options.transportMode.client` should be an equivalent of `options.clientMode` right :thinking:
Instead the respective case made use of just `options.transportMode` which made me confused :confused:

jamesgeorge007:
Apologies for the trouble here. Kindly correct if I'm wrong :pray:

jamesgeorge007:
Apologies for the trouble here. Kindly correct me if I'm wrong :pray:

evilebottnawi:
`options.clientMode` and `options.serverMode` were removed, now only `transportMode` option

jamesgeorge007:
Even `transportMode` comprises of `client` and `server` entries right?

evilebottnawi:
@jamesgeorge007 yes, you can set `transportMode: { client: myImplementation, server: 'ws}` for example

Loonride:
@jamesgeorge007 

When we just had:

```
options: {
  clientMode: 'sockjs'
}
```

We were relying on the fact that the default serverMode was also `sockjs`, so that server and client could communicate. This is still the case, but in the future, the default mode will be changed to `ws`.

This is better, because it specifies both the client and server modes explicitly:

```
options: {
  transportMode: 'sockjs'
}
```

since it is equivalent to:

```
options: {
  transportMode: {
    client: 'sockjs',
    server: 'sockjs'
  }
}
```

PR #2188, opened 2019-08-10
chore(deps): add webpack-cli to peers

eps1lon:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Can publish a package with this PR and see if it actually works but these kind of fixes have been done multiple times to make yarn v2 happy. Considering `webpack-dev-server` throws at runtime anyway if `webpack-cli` is missing this shouldn't bother anyone

### Motivation / Use-Case

yarn v2 throws currently with 

```

```

### Breaking Changes

I guess this new warning might be annoying if you don't run the `webpack-dev-server` cli. If this is a blocker I can try and see if https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md works for yarn v2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,8 @@
     "webpack-cli": "^3.3.6"
   },
   "peerDependencies": {
-    "webpack": "^4.0.0"
+    "webpack": "^4.0.0",
+    "webpack-cli": "^3.0.0"
   },
   "author": "Tobias Koppers @sokra",
   "bugs": "https://github.com/webpack/webpack-dev-server/issues",


evilebottnawi:
Thanks for PR, we can't do this, it is break CFA, vue-cli, angular-cli and etc, you should put webpack-cli in your deps, also next major release was without bin, and you need webpack-cli to run webpack-dev-server using CLI

eps1lon:
> Thanks for PR, we can't do this, it is break CFA, vue-cli, angular-cli and etc, you should put webpack-cli in your deps, also next major release was without bin, and you need webpack-cli to run webpack-dev-server using CLI

@evilebottnawi  Please read the PR description. I already have webpack-cli as a dependency but yarn v2 will throw errors if a package does not declare its dependencies.

Maybe @arcanis could elaborate why this is important.

arcanis:
I'll copy what I said on [this issue](https://github.com/facebook/create-react-app/pull/6859#issuecomment-519463498):

> Fwiw I don't think using an unsafe pattern known to break under some circumstances (omitting peer dependencies from the manifest) just to remove warnings is a good tradeoff, especially when those warning can already disappear for half the users by using the right feature (npm is [also going to implement optional peer dependencies](https://github.com/npm/cli/pull/224), btw) ...

evilebottnawi:
Feel free to send a PR with `peerDependenciesMeta` (looks it is good idea)

PR #2191, opened 2019-08-12
chore(deps): add webpack-cli to peers

eps1lon:
Follow-up on #2188 according to https://github.com/webpack/webpack-dev-server/pull/2188#issuecomment-520373076

- [x] This is a **bugfix**
- [ ] This is a **feature**

Can publish a package with this PR and see if it actually works but these kind of fixes have been done multiple times to make yarn v2 happy. Considering `webpack-dev-server` throws at runtime anyway if `webpack-cli` is missing this shouldn't bother anyone

### Motivation / Use-Case

yarn v2 throws currently with 

```

```

### Breaking Changes

I guess this new warning might be annoying if you don't run the `webpack-dev-server` cli. If this is a blocker I can try and see if https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md works for yarn v2.

### Additional Info
https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md


package.json
```diff
@@ -114,7 +114,13 @@
     "webpack-cli": "^3.3.6"
   },
   "peerDependencies": {
-    "webpack": "^4.0.0"
+    "webpack": "^4.0.0",
+    "webpack-cli": "^3.0.0"
```
evilebottnawi: Please remove `webpack-cli` from here, this line break CFA, vue-cli, angular-cli and etc big boilerplates

arcanis:
> /cc @arcanis it is require what `webpack-cli` was in `peerDependencies`?

Yes. The `peerDependenciesMeta` field affect how package managers will treat the entries for the existing `peerDependencies` field - it doesn't extend it.

> Because we can't do this right now

Again, if you can't only because it would cause warnings to appear for npm's users then I'd say it's too bad because omitting the peer dependencies will cause both npm and Yarn to silently generate invalid dependency trees, whereas warnings will only cause a slight temporary discomfort at best.

Does it *actually* break something? As in, a consumer stops working altogether?

evilebottnawi:
@arcanis do you know other developers, they starts create a lot of issues and send PRs with removing this, we want remove bin for webpack-dev-server@4 so i think better avoid this right now for webpack-dev-server@3

arcanis:
> If this is a blocker I can try and see if https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md works for yarn v2.

Optional peer dependencies work for Yarn 1, Yarn 2, pnpm 3.2+, and (planned for) npm 6.11.

> you know other developers, they starts create a lot of issues and send PRs with removing this

I sympathise, however note that the same is also true for us - we get issues telling us that Yarn is installing dependencies the wrong way etc, and after digging we notice it's caused by undeclared dependencies like this. Plus I guess you still get PRs to add it, in the end 😄

evilebottnawi:
@arcanis but here not `Optional peer dependencies`

arcanis:
They are; that's how optional peer dependencies work.

The `peerDependenciesMeta` lists settings that modify the default behaviour for the dependencies field of the given name (`optional` means that the peer dependency being unmet shouldn't be reported as a warning, for example), but they still have to be declared in `peerDependencies` (otherwise we don't have the range hint for the peer dependency, for example).

evilebottnawi:
@arcanis thansk for clarify, maybe we should wait https://github.com/npm/cli/pull/224 before merge and release this

eps1lon:
@evilebottnawi `npm@6.11` was released which includes https://github.com/npm/cli/pull/224

I don't know how to make the checks pass though. Failures seem unrelated.

hiroppy:
We need to wait for npm to be updated in Node.js.

salper:
> We need to wait for npm to be updated in Node.js.

It seems to be the case. Any news on this?

arcanis:
@evilebottnawi @hiroppy I've opened another PR (https://github.com/webpack/webpack-dev-server/pull/2396) that uses a smoother pattern and doesn't print any warnings even on older package managers. Can one of you give it a look and tell me how it looks? We plan to release Yarn 2 next week, it would be nice to sort this out before then if possible 😀

jacklaurencegaray:
https://github.com/yarnpkg/berry/issues/815

evilebottnawi:
Fixed for yarn v2, feel free to feedback

PR #2196, opened 2019-08-14
 feat(util): accept port as object type

hiroppy:
- [ ] This is a **bugfix**
- [x] This is a **feature**

yes

### Motivation / Use-Case

Accept port as object type.

```js
{
  port: 8081
  port: null
  port: {
    basePort: 8081,
	highestPort: 8090,
    retry: 5
  }
}

```

### Breaking Changes

no


lib/options.json
```diff
@@ -246,6 +246,20 @@
...
+            "basePort": {
+              "type": "number"
+            },
+            "highestPort": {
+              "type": "number"
+            },
+            "retry": {
+              "type": "number"
+            }
+          }
```
evilebottnawi: Let's add `required: ['basePort']`, because empty object is invalid setup

lib/utils/findPort.js
```diff
@@ -1,39 +1,34 @@
...
-  // Try to find unused port and listen on it for 3 times,
-  // if port is not specified in options.
-  // Because NaN == null is false, defaultTo fails if parseInt returns NaN
-  // so the tryParseInt function is introduced to handle NaN
-  const defaultPortRetry = defaultTo(
-    tryParseInt(process.env.DEFAULT_PORT_RETRY),
-    3
-  );
+  portfinder.basePort = defaultPort;
+  portfinder.highestPort = portfinder.basePort + 3;
```
evilebottnawi: Magic number, we should use contant for this, anyway i think keeping `process.env.DEFAULT_PORT_RETRY` is good idea

hiroppy:
Updated:)

lib/options.json
```diff
@@ -246,6 +246,21 @@
...
+        {
+          "type": "object",
+          "properties": {
+            "basePort": {
+              "type": "number"
+            },
+            "highestPort": {
+              "type": "number"
+            },
+            "retry": {
```
evilebottnawi: After testing and investigating, i think `retry` is unnecessary here, we have `process.env.DEFAULT_PORT_RETRY` constant so we can avoid option here, it is unnecessary. Also we should respect `process.env.DEFAULT_PORT_RETRY` when port is number/string/null.

lib/utils/findPort.js
```diff
@@ -1,39 +1,36 @@
...
-    });
-  });
-}
-
-function findPort(port) {
-  if (port) {
-    return Promise.resolve(port);
+async function findPort(options) {
+  if (tryParseInt(options)) {
+    return tryParseInt(options);
```
evilebottnawi:
```js
var maybeInt = tryParseInt(options);

if (maybeInt) { 
  return maybeInt;
}
```

lib/utils/findPort.js
```diff
@@ -1,39 +1,36 @@
...
-  const defaultPortRetry = defaultTo(
-    tryParseInt(process.env.DEFAULT_PORT_RETRY),
-    3
-  );
+  const defaultRetry = tryParseInt(process.env.DEFAULT_PORT_RETRY) || 3;
+
+  portfinder.basePort = defaultPort;
+  portfinder.highestPort = portfinder.basePort + defaultRetry;
+
+  if (Object.prototype.toString.call(options) === '[object Object]') {
```
evilebottnawi: Move this to util `isObj`

lib/utils/findPort.js
```diff
@@ -1,39 +1,36 @@
...
+
+    if (retry) {
+      portfinder.highestPort = portfinder.basePort + retry;
+    } else if (highestPort) {
+      portfinder.highestPort = options.highestPort;
+    }
+  }
 
-  return pRetry(runPortFinder, { retries: defaultPortRetry });
+  return portfinder.getPortPromise();
```
evilebottnawi: I think here breaking change, when i use `8080` port and when port is busy next port should be `8081`, then `8082` and continue, look `basePort` and `highestPort` just get random port from range.

hiroppy: Not breaking change. See https://github.com/http-party/node-portfinder/blob/master/lib/portfinder.js#L357-L359

evilebottnawi: We need avoid this breaking, because it is good logic when port is unavailable we get next port instead random, just need test when we use `port` as number/string/null and port unavailable we get next port instead random from range

hiroppy: How to resolve this situation?

```
{
  basePort: 8080,
  highestPort: 9000
}
```

and default retry is 3.

1. 8080 <- used
2. 8087 <- used
3. 9000 <- used

in this case, we'll finish searching even if we can't find an empty port?
Random retry is very unstable.

I think if highestPort is set we might be better to ignore retry.

evilebottnawi: We should avoid `retry` option for this case, we already say range of ports, retry option is needed only when we use `port` as number/null or string

hiroppy: if `port` is set, it will go to https://github.com/webpack/webpack-dev-server/pull/2196/files#diff-487ca80b65e47f5165c8419f1664af7cR8 and finish. And if `port` is `null` or `undefined`, this code has already set the highest port. https://github.com/webpack/webpack-dev-server/pull/2196/files#diff-487ca80b65e47f5165c8419f1664af7cR15

evilebottnawi: I want to say what we really do not need `retry` option for `object` notation, because it doesn't have sense, if i write:
```
{
  basePort: 8080,
  highestPort: 9000
}
```

I want any port from this range and the retry option in this case is misleading, what `retry` means in this case? Try to get busy multiple times busy port (for example 8080 is busy and we try to get 8080 again and again) or try to get other free port (for example 8080 is busy and we try to get 8081 then 8082)?

To be honesty retry env variable is very misleading also, maybe we need rename this option here too

thoughts?

hiroppy: > I want any port from this range and the retry option in this case is misleading,  

Yes, I think so too. Actually, I confused when I was implementing this feature.

In the first place we offer the highestPort option, so is this retry option still necessary?(just providing the default value(3) or rename `DEFAULT_PORT_RETRY` ...)

evilebottnawi: Give me time to think about it, I'm a little busy today, so I'll write my thoughts tomorrow :+1:

hiroppy: Yes, sure. Thank you for your review!

PR #2200, opened 2019-08-16
fix: add null check for connection.headers

### LINKED ISSUE #2199: TypeError: Cannot read property 'headers' of null at Server.socket.on (***/node_modules/webpack-dev-server/lib/servers/SockJSServer.js:68:32)
- Operating System: Windows 10
- Node Version: 12.8.1
- NPM Version: 6.10.3
- webpack Version: 4.39.2
- webpack-dev-server Version: since 3.5.0
- Browser: Chrome 76

- [X] This is a **bug**

### Expected Behavior

`connection` should not be null

### Actual Behavior

`TypeError: Cannot read property 'headers' of null at Server.socket.on (***/node_modules/webpack-dev-server/lib/servers/SockJSServer.js:68:32)`

This bug is difficult to reproduce. It happens occasionally.

https://github.com/webpack/webpack-dev-server/blob/2d44ef8088db2312516274cc2d5edda348064712/lib/servers/SockJSServer.js#L65-L69

line 67 should become

```js
if (connection) {
  f(connection, connection.headers)
}
```

wood1986:
fix #2199

- [X] This is a **bugfix**
- [ ] This is a **feature**


Loonride:
If the connection object is null, there is no use in passing it along to the callback since we can't do anything with it. I think it is better off being:

```
if (connection) {
  f(connection, connection.headers);
}
```

wood1986:
> If the connection object is null, there is no use in passing it along to the callback since we can't do anything with it. I think it is better off being:
> 
> ```
> if (connection) {
>   f(connection, connection.headers);
> }
> ```

I thought about this. As you have a check in

https://github.com/webpack/webpack-dev-server/blob/a3558944b900f5f8695421522eeb102711b81ce7/lib/Server.js#L662-L672

I want to reuse.

wood1986:
> need tests

I added

wood1986:
I need some help to re-queue the build

wood1986:
Any updates?

test/server/servers/SockJSServer.test.js
```diff
@@ -84,9 +84,20 @@ describe('SockJSServer', () => {
...
       }, 3000);
     });
+
+    it('should not throw an exception when connection is null', (done) => {
+      expect(() => {
+        socketServer.onConnection(() => {});
+        socketServer.socket.emit('connection', null);
+      }).not.toThrow();
+
+      setTimeout(() => {
```
Loonride: This timeout is not needed, remove this and remove the `done` parameter from the function.

test/server/servers/SockJSServer.test.js
```diff
@@ -84,9 +84,20 @@ describe('SockJSServer', () => {
         done();
       }, 3000);
     });
+
+    it('should not throw an exception when connection is null', (done) => {
+      expect(() => {
+        socketServer.onConnection(() => {});
```
Loonride: Before this `expect` test, could you make a function `const cb = jest.fn()`, then pass it into the onConnection like: `onConnection(cb)`. Then after this confirm that the callback is actually being called with the expected values like: `expect(cb.mock.calls[0]).toEqual([null, null])`.

test/server/servers/SockJSServer.test.js
```diff
@@ -10,7 +10,7 @@ describe('SockJSServer', () => {
   let socketServer;
   let listeningApp;
 
-  beforeAll((done) => {
+  beforeEach((done) => {
```
Loonride: These can stay as `beforeAll` and `afterAll`. It is reusing the same server for all tests. Switching to `beforeEach/afterEach` would not change anything because there is only that one `describe` call on the same level as these calls.

Loonride:
> I want to reuse.

Understood. I have added some notes on the tests.

test/server/servers/SockJSServer.test.js
wood1986: * Reason for using `beforeEach`:
As my test needs to test null scenario and we are mutating the same environment, it shows
image

* Reason for the `timeout`
It won't pass in the build machine only for Windows node-8 and Windows node-6.
image
See https://dev.azure.com/webpack/webpack-dev-server/_build/results?buildId=5531

test/server/servers/SockJSServer.test.js
wood1986: * Reason for the `timeout`
It won't pass in the build machine only for Windows node-8 and Windows node-6.
image
See https://dev.azure.com/webpack/webpack-dev-server/_build/results?buildId=5531

test/server/servers/SockJSServer.test.js
Loonride: Yes that makes sense. My understanding of `beforeEach/afterEach` in describe blocks was incorrect.

test/server/servers/SockJSServer.test.js
Loonride: That is caused by a stability issue with our e2e tests. You can ignore it, your changes are not causing it (you can see that your test ran and passed after the problematic test), and your changes can still be merged if CI is failing for this reason. The `timeout` can still be removed.

wood1986: Thanks. I am going to make the change

wood1986: Yes, it was the stability issue. Right now, all are passed. Thanks

test/server/servers/SockJSServer.test.js
```diff
@@ -84,9 +84,18 @@ describe('SockJSServer', () => {
         done();
       }, 3000);
     });
+
+    it('should not throw an exception when connection is null', () => {
+      expect(() => {
+        const cb = jest.fn();
```
Loonride: Sorry one more nitpick but could you move this `cb` definition above the `expect` line, then move `expect(cb.mock.calls[0]).toEqual([null, null]);` below the `not.toThrow`? Just to isolate the behavior we want to "not throw".

wood1986: nvm and I love it

Loonride:
Looks good to me /cc @evilebottnawi @hiroppy

wood1986:
When will you guys do the npm publish?

Issue #2190, opened 2019-08-12
"Error: Watching remote files is not supported" on windows with latest


docteurzoidberg:
- Operating System: Windows 10
- Node Version: 10.16.1
- NPM Version: 6.10.3
- webpack Version: 4+
- webpack-dev-server Version: 3.8.0
- Browser: none

- [x] This is a **bug**
- [ ] This is a **modification** request

### Expected Behavior

'npm run serve' should work like always

### Actual Behavior

'npm run serve' give me an "Error: Watching remote files is not supported.". 
-> I cannot build / run anything with the dev-server

### For Bugs; How can we reproduce the behavior?

Running 'npm run serve' on any vue-cli boilerplate webapp

evilebottnawi:
Looks you try to watch remove files

> Running 'npm run serve' on any vue-cli boilerplate webapp

Sorry, i can't reproduce nothing, please create **minimum** reproducible test repo or provide your configuration

docteurzoidberg:
https://github.com/vuejs/vue-cli/issues/4424

evilebottnawi:
@docteurzoidberg still no configuration

docteurzoidberg:
@evilebottnawi i haven't any, it's the default vue-cli project !

docteurzoidberg:
@evilebottnawi > I found the bug; it's because of the pwd of my project: "c:\dev\test-broken" 
It doesn't work on windows when parent dir is named DEV xD
I moved it elsewhere and now it works flawlessly, in another folder....

evilebottnawi:
need investigate, maybe wrong regexp

evilebottnawi:
Thanks for more infromation which can help us

docteurzoidberg:
https://i.imgur.com/etToqUI.png not working in c:\dev\test-broken
https://i.imgur.com/w4sHePq.png working in c:\users\drzoid\test-broken

Same project, working in any other folder with no "dev" parent folder

evilebottnawi:
@docteurzoidberg can you provide you project, or minimum reproducible test repo, it can help fix problem fastly, thanks

docteurzoidberg:
sorry, i don't have git to publish on windows setting, i'm at camping right now. i can do it next week.

The project is done with vue-cli ("npm install vue-cli -g"), then "vue init test-broken"

evilebottnawi:
@docteurzoidberg no additional configurations?

docteurzoidberg:
> 
[...]

PR #2202, opened 2019-08-16
fix(server): use regex instead of isAbsoluteUrl

EslamHiko:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case
fixes: https://github.com/webpack/webpack-dev-server/issues/2190

### Breaking Changes
no

### Additional Info
If we want to still `isAbsoluteUrl` use can add an extra check for the second character in the contentBase and replace all `isAbsoluteUrl(String(contetBase))` with `isAbsoluteUrl(String(contetBase)) && String(contetBase)[1] !== ':'` assuming that always local paths will have `:` as a second character ex:`c: , e: , ...` but I think reverting the changes to the original regex will be better, what do you think ?

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -23,7 +23,6 @@ const serveIndex = require('serve-index');
 const webpack = require('webpack');
 const webpackDevMiddleware = require('webpack-dev-middleware');
 const validateOptions = require('schema-utils');
-const isAbsoluteUrl = require('is-absolute-url');
 const normalizeOptions = require('./utils/normalizeOptions');
 const updateCompiler = require('./utils/updateCompiler');
 const createLogger = require('./utils/createLogger');
@@ -339,7 +338,7 @@ class Server {
       contentBase.forEach((item) => {
         this.app.get('*', express.static(item));
       });
-    } else if (isAbsoluteUrl(String(contentBase))) {
+    } else if (/^(https?:)?\/\//.test(contentBase)) {
       this.log.warn(
         'Using a URL as contentBase is deprecated and will be removed in the next major version. Please use the proxy option instead.'
       );
@@ -392,7 +391,7 @@ class Server {
       });
     } else if (
       typeof contentBase !== 'number' &&
-      !isAbsoluteUrl(String(contentBase))
+      !/^(https?:)?\/\//.test(contentBase)
     ) {
       this.app.get('*', serveIndex(contentBase));
     }
@@ -401,11 +400,14 @@ class Server {
   setupWatchStaticFeature() {
     const contentBase = this.options.contentBase;
 
-    if (isAbsoluteUrl(String(contentBase)) || typeof contentBase === 'number') {
+    if (
+      /^(https?:)?\/\//.test(contentBase) ||
+      typeof contentBase === 'number'
+    ) {
       throw new Error('Watching remote files is not supported.');
     } else if (Array.isArray(contentBase)) {
       contentBase.forEach((item) => {
-        if (isAbsoluteUrl(String(item))) {
+        if (/^(https?:)?\/\//.test(item) || typeof item === 'number') {
           throw new Error('Watching remote files is not supported.');
         }
         this._watch(item);
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -23,11 +23,11 @@ const serveIndex = require('serve-index');
 const webpack = require('webpack');
 const webpackDevMiddleware = require('webpack-dev-middleware');
 const validateOptions = require('schema-utils');
+const isAbsoluteUrl = require('is-absolute-url');
 const normalizeOptions = require('./utils/normalizeOptions');
 const updateCompiler = require('./utils/updateCompiler');
 const createLogger = require('./utils/createLogger');
 const getCertificate = require('./utils/getCertificate');
-const checkUrl = require('./utils/checkUrl');
 const status = require('./utils/status');
 const createDomain = require('./utils/createDomain');
 const runBonjour = require('./utils/runBonjour');
@@ -339,7 +339,7 @@ class Server {
       contentBase.forEach((item) => {
         this.app.get('*', express.static(item));
       });
-    } else if (checkUrl(contentBase)) {
+    } else if (isAbsoluteUrl(String(contentBase))) {
       this.log.warn(
         'Using a URL as contentBase is deprecated and will be removed in the next major version. Please use the proxy option instead.'
       );
@@ -390,7 +390,10 @@ class Server {
       contentBase.forEach((item) => {
         this.app.get('*', serveIndex(item));
       });
-    } else if (typeof contentBase !== 'number' && !checkUrl(contentBase)) {
+    } else if (
+      typeof contentBase !== 'number' &&
+      !isAbsoluteUrl(String(contentBase))
+    ) {
       this.app.get('*', serveIndex(contentBase));
     }
   }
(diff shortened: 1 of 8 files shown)


evilebottnawi:
I think we need ti wait answer https://github.com/sindresorhus/is-absolute-url/issues/7 before doing something here

EslamHiko:
@hiroppy done
@evilebottnawi sorry I didn't see the comment, we have now a util file we can use any regex we want.we can modify is-absolute-url regex and use it in `checkUrl` util. what do you think?

EslamHiko:
@evilebottnawi yes, I kept only the tests.

lib/utils/runOpen.js
```diff
@@ -14,7 +14,7 @@ function runOpen(uri, options, log) {
   }
 
   const pageUrl =
-    options.openPage && isAbsoluteUrl(options.openPage)
+    options.openPage && isAbsoluteUrl(String(options.openPage))
```
evilebottnawi: No need this here, if you have problem place let's solve this in other PR

test/server/contentBase-option.test.js
```diff
@@ -301,6 +301,29 @@ describe('contentBase option', () => {
         done();
       }
     });
+    it('Should not throw exception (local path with lower case first character)', (done) => {
+      try {
+        // eslint-disable-next-line no-unused-vars
+        server = new Promise((resolve, reject) => {
+          testServer.start(config, {
+            contentBase:
```
evilebottnawi: Let's add two tests here `c:\absolute\path\to\content-base` and `C:\absolute\path\to\content-base`

test/server/contentBase-option.test.js
```diff
@@ -324,6 +324,49 @@ describe('contentBase option', () => {
         expect(true).toBe(false);
       }
     });
+    it("Should not throw exception (local path with lower case first character & has '-')", (done) => {
+      try {
+        // eslint-disable-next-line no-unused-vars
+        server = new Promise((resolve, reject) => {
+          testServer.start(config, {
+            contentBase: 'c:absolutepath\tocontent-base',
```
evilebottnawi: Typo :smile:

test/server/contentBase-option.test.js
```diff
@@ -301,6 +301,72 @@ describe('contentBase option', () => {
...
+    it('Should not throw exception (local path with lower case first character)', (done) => {
+      try {
+        // eslint-disable-next-line no-unused-vars
+        server = new Promise((resolve, reject) => {
+          testServer.start(config, {
+            contentBase:
+              contentBasePublic.charAt(0).toLowerCase() +
+              contentBasePublic.substring(1),
+            watchContentBase: true,
+            port: 2222,
```
hiroppy: Don't use the port number directly.

EslamHiko: if I use `port` it'll cause `listen EADDRINUSE: address already in use 127.0.0.1:8107` error and cause the tests to hang, can I use `port+1`,`port+2` or add new values in `ports-map` ?

hiroppy: Don't add the port number.

<details>

```diff
diff --git a/test/server/contentBase-option.test.js b/test/server/contentBase-option.test.js
index fe120e5..bcde1e6 100644
--- a/test/server/contentBase-option.test.js
+++ b/test/server/contentBase-option.test.js
@@ -269,10 +269,10 @@ describe('contentBase option', () => {
 
   describe('testing single & multiple external paths', () => {
     afterEach((done) => {
-      testServer.close(() => {
-        done();
-      });
+      console.log('aa');
+      testServer.close(done);
     });
+
     it('Should throw exception (string)', (done) => {
       try {
         // eslint-disable-next-line no-unused-vars
@@ -302,71 +302,40 @@ describe('contentBase option', () => {
       }
     });
     it('Should not throw exception (local path with lower case first character)', (done) => {
-      try {
-        // eslint-disable-next-line no-unused-vars
-        server = new Promise((resolve, reject) => {
-          testServer.start(config, {
-            contentBase:
-              contentBasePublic.charAt(0).toLowerCase() +
-              contentBasePublic.substring(1),
-            watchContentBase: true,
-            port: 2222,
-          });
-          resolve(testServer);
-        });
-
-        server.then((testServer) => {
-          testServer.close(() => {
-            done();
-          });
-        });
-      } catch (e) {
-        expect(true).toBe(false);
-      }
+      testServer.start(
+        config,
+        {
+          contentBase:
+            contentBasePublic.charAt(0).toLowerCase() +
+            contentBasePublic.substring(1),
+          watchContentBase: true,
+          port,
+        },
+        done
+      );
     });
     it("Should not throw exception (local path with lower case first character & has '-')", (done) => {
-      try {
-        // eslint-disable-next-line no-unused-vars
-        server = new Promise((resolve, reject) => {
-          testServer.start(config, {
-            contentBase: 'c:\\absolute\\path\\to\\content-base',
-            watchContentBase: true,
-            port: 2223,
-          });
-          resolve(testServer);
-        });
-
-        server.then((testServer) => {
-          testServer.close(() => {
-            done();
-          });
-        });
-      } catch (e) {
-        expect(true).toBe(false);
-      }
+      testServer.start(
+        config,
+        {
+          contentBase: 'c:\\absolute\\path\\to\\content-base',
+          watchContentBase: true,
+          port,
+        },
+        done
+      );
     });
     it("Should not throw exception (local path with upper case first character & has '-')", (done) => {
-      try {
-        // eslint-disable-next-line no-unused-vars
-        server = new Promise((resolve, reject) => {
-          testServer.start(config, {
-            contentBase: 'C:\\absolute\\path\\to\\content-base',
-            watchContentBase: true,
-            port: 2224,
-          });
-          resolve(testServer);
-        });
-
-        server.then((testServer) => {
-          testServer.close(() => {
-            done();
-          });
-        });
-      } catch (e) {
-        expect(true).toBe(false);
-      }
+      testServer.start(
+        config,
+        {
+          contentBase: 'C:\\absolute\\path\\to\\content-base',
+          watchContentBase: true,
+          port,
+        },
+        done
+      );
     });
-
     it('Should throw exception (array)', (done) => {
       try {
         // eslint-disable-next-line no-unused-vars
```

</details>

You can patch this code.

EslamHiko:
/cc @evilebottnawi @hiroppy 
CI is green 😄

lib/Server.js
```diff
@@ -405,7 +405,7 @@ class Server {
       throw new Error('Watching remote files is not supported.');
     } else if (Array.isArray(contentBase)) {
       contentBase.forEach((item) => {
-        if (isAbsoluteUrl(String(item))) {
+        if (isAbsoluteUrl(String(item)) || typeof item === 'number') {
```
evilebottnawi: Why we add `typeof item === 'number'`?

EslamHiko: @evilebottnawi it's already blocking number values in options.js accepting only strings, in case in future refactors if we change the type of array items to accept any type we still need it to throw an exception if the array item is number type.

EslamHiko: @evilebottnawi we can remove number type from options.json 
```
  "contentBase": {
      "anyOf": [
        {
          "enum": [false]
        },
        {
          "type": "number"
        },
        {
          "type": "string"
        },
        {
          "type": "array",
          "items": {
            "type": "string"
          },
          "minItems": 1
        }
      ]
    },
```
and remove ` || typeof item === 'number'` checks and update tests to check for the exception ```"should be {String|Array} (https://webpack.js.org/configuration/dev-server/#devservercontentbase)"```
what do you think ?

evilebottnawi: Let's do it next version

PR #2215, opened 2019-08-22
fix(server): remove number type from contentBase (next)

EslamHiko:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case
ContentBase accepts array with only string values but can accept a single number value. removing number type from options.json will fix this bug. 

### Breaking Changes
No

### Additional Info
Had to remove 'to port' test case because it won't accept number values anymore as contentBase.


evilebottnawi:
Done in next, anyway thanks for the PR

PR #2229, opened 2019-08-27
fix: fatal error when current JS is not inside script tag

misak113:
* for example when html-webpack-inline-source-plugin is used, the JS is inlined in HTML in script tag. So in this case document.currentScript.getAttribute('src') is undefined
* undefined cannot be used as string with method replace. This cause a fatal error and crash app

- [ ] This is a **bugfix**
- [ ] This is a **feature**

### DIFF
--- a/client-src/default/utils/createSocketUrl.js
+++ b/client-src/default/utils/createSocketUrl.js
@@ -16,8 +16,10 @@ function createSocketUrl(resourceQuery) {
     // Else, get the url from the <script> this file was called with.
     let scriptHost = getCurrentScriptSource();
 
-    // eslint-disable-next-line no-useless-escape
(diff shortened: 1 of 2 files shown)


misak113:
Hi, I'm sorry for failing PR, but I added a special case to tests, where getCurrentScriptSource returns undefined. So it could be enough for test coverage. It's exactly the value, which would have been failing tests when my change is not applied.
https://github.com/webpack/webpack-dev-server/pull/2229/commits/b48830490e29ef97b21375d3f5e6e9a3240c8483#diff-2ee53e81db4e6170cf44cc607e849f1aR17

evilebottnawi:
We need fix error

Issue #2233, opened 2019-08-28
Option `silent` does not show initial startup information


aryzing:
- Operating System: Ubuntu 18.04 LTS
- Node Version: v11.12.0
- NPM Version: 6.7.0
- webpack Version: 4.39.2
- webpack-dev-server Version: 3.8.0
- Browser: Chrome 76

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

```js
  devServer: {
    contentBase: false,
    hot: true,
    quiet: true,
  },
```

### Expected Behavior

To see the initial startup information in the console, and nothing else

### Actual Behavior

There is no output at all to the console.

### For Bugs; How can we reproduce the behavior?

Check out this repro repo: https://github.com/aryzing/wds-repro-quiet-flag-not-working

rishabh3112:
@aryzing I wasn't able to reproduce the issue at the repo shared above and is working as it should be.
Here is the output with `quiet: true`:
![image](https://user-images.githubusercontent.com/15812317/64019641-56c23800-cb4d-11e9-83e7-af47bafe1a9c.png)

aryzing:
Where's the initial startup information, like the URL of the server? The docs say that the initial startup info should be printed in quiet mode.

aryzing:
For reference, [docs here](https://github.com/webpack/webpack.js.org/blob/a7b1fcf9935460cdaf96b9ff8e015ce07fbae74c/src/content/configuration/dev-server.md#devserverquiet-)

rishabh3112:
My bad. Will look into it soon.

PR #1486, opened 2018-09-02
fix(utils/createLogger): ensure `quiet` always takes precedence (`options.quiet`)

lukeed:
- [x] This is a **bugfix**
- [ ] This is a **code refactor**
- [ ] This is a **test update**
- [ ] This is a **typo fix**
- [ ] This is a **metadata update**

### For Bugs and Features; did you add new tests?

Doesn't look like `logLevel` is currently tested, aside from validation. Also not sure how to test this~

### Motivation / Use-Case

Setting `noInfo` and `quiet` to `true` still reported errors (and warnings), despite the fact that `quiet` should silence everything.

### Breaking Changes

N/A

### Additional Info

Screenshot includes `wdm` error message. Also displaying the full `devServer` config above it. 

screen shot 2018-09-02 at 12 43 38 pm

PR #2235, opened 2019-08-31
feat: add status in quiet log level

rishabh3112:
- [x] This is a **bugfix**
- [x] This is a **feature**

### Motivation / Use-Case
#2233 

### Breaking Changes
May break some tests

### DIFF
--- a/lib/utils/status.js
+++ b/lib/utils/status.js
@@ -7,7 +7,7 @@ const runOpen = require('./runOpen');
 // TODO: don't emit logs when webpack-dev-server is used via Node.js API
 function status(uri, options, log, useColor) {
   if (options.quiet === true) {
-    log = log({
+    log = logger({
       name: 'wds',
       logger,
       timestamp: options.logTime,
--- a/lib/utils/status.js
+++ b/lib/utils/status.js
@@ -7,6 +7,7 @@ const runOpen = require('./runOpen');
 // TODO: don't emit logs when webpack-dev-server is used via Node.js API
 function status(uri, options, log, useColor) {
   if (options.quiet === true) {
+    // Add temporary logger to output just the status of the dev server
     log = logger({
       name: 'wds',
       logger,
--- a/lib/utils/status.js
+++ b/lib/utils/status.js
@@ -1,10 +1,19 @@
 'use strict';
 
+const logger = require("webpack-log");
 const colors = require('./colors');
 const runOpen = require('./runOpen');
 
 // TODO: don't emit logs when webpack-dev-server is used via Node.js API
 function status(uri, options, log, useColor) {
+  if (options.quiet === true) {
+    log = log({
+      name: 'wds',
+      logger,
+      timestamp: options.logTime,
+    });
+  }
+
   const contentBase = Array.isArray(options.contentBase)
     ? options.contentBase.join(', ')
     : options.contentBase;


Loonride:
Looks like it should work, but needs tests.

rishabh3112:
Added tests for quiet flag.

lib/utils/status.js
```diff
@@ -1,10 +1,20 @@
...
 
+const logger = require('webpack-log');
 const colors = require('./colors');
 const runOpen = require('./runOpen');
 
 // TODO: don't emit logs when webpack-dev-server is used via Node.js API
 function status(uri, options, log, useColor) {
+  if (options.quiet === true) {
+    // Add temporary logger to output just the status of the dev server
+    log = logger({
```
evilebottnawi: Why don't use `log`?

rishabh3112: Assuming you want `logger` to be `log`, because `log` is also the function argument for status function.

evilebottnawi: Can you clarify?

rishabh3112: When `--quiet` is passed, The `log` received by the status function would have log level as `silent`, due to which `log.info` would not print anything to `stdout`. That's why overriding the passed log so that only status is printed + original log remain unaffected for rest of the modules. 

Hope that explains the solution.

evilebottnawi: hm, look we can fix it in master, only for next

rishabh3112: I am afraid but I didn't get what you meant above?

evilebottnawi: @rishabh3112 all fine, latest webpack has logger and use other logic we need switch on this logic and rewrite some places

rishabh3112: I can help if you can point me where changes has to be done

evilebottnawi: @rishabh3112 we will refactor this in next version, can't be node in master right now

rishabh3112: node?

evilebottnawi: @rishabh3112 no, new version of webpack has logger, we need migrate on this logger instead using `webpack-log`

rishabh3112: Got it. Thanks for clarifying.
Is master fine with webpack-log or should I close this PR ?

evilebottnawi: @rishabh3112 All good here

test/cli/cli.test.js
```diff
@@ -29,6 +29,22 @@ describe('CLI', () => {
       .catch(done);
   });
 
+  it('--quiet', async (done) => {
+    const output = await testBin('--quiet');
+    expect(output.code).toEqual(0);
+    expect(output.stdout.split('\n').length === 3).toBe(true);
+    expect(
+      output.stdout.includes('Project is running at http://localhost:8080/')
```
Loonride: Please use `ports-map` to get a port that is not 8080, then specify it with `--port`. It looks like there are some unused `cli` ports here that can be used: https://github.com/webpack/webpack-dev-server/blob/master/test/ports-map.js#L6

rishabh3112: Gotcha, Will look into it soon.

lib/utils/status.js
```diff
@@ -1,10 +1,20 @@
 'use strict';
 
+const logger = require('webpack-log');
 const colors = require('./colors');
 const runOpen = require('./runOpen');
 
 // TODO: don't emit logs when webpack-dev-server is used via Node.js API
 function status(uri, options, log, useColor) {
+  if (options.quiet === true) {
```
kelly-tock: does this mean that there is now no way to silence these extra logs when using in node api?

evilebottnawi: yes, this log should be outputted always, in future we will use webpack logger and you can filter this output

kelly-tock: that would be great. this seems like a regression IMO. we now have an extra set of console output. if possible when the webpack logger option is available would be great to link that to this issue.

kelly-tock: link to documentation? not sure what you mean.

evilebottnawi: No docs right now, because it is hidden feature, look here https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L52

endiliey: @evilebottnawi

I am facing this as well. I use webpackbar and this extra console log noise actually make it weird because webpackbar tries to clear screen when showing the fancy bar

After this PR
sad

Before
good

Can we add an option to hide this output ? I'm sure many packages that use webpackbar is affected by this unknowingly :(

cc my team @yangshun

kelly-tock: pretty pretty please?  this output decreases dev experience IMO.

evilebottnawi: Feel free to send a PR with new option

lukeed: This reverts #1486 and I believe it to _now_ be a breaking change.

> this log should be outputted always

Sorry, strongly disagree. What's the point of the option if it doesn't do what it says? I understand the custom logger is just a documentation issue right now, but it still points to `quiet` (and `noInfo`) being pointless/improperly named

lukeed: You actually can't fix this with a custom logger, because this setup noise is printed by `utils/status.js` directly, which is where this `log` override is happening. There's no intermediary step

kelly-tock:
it seems like not a lot of people want this change. how can we get it back to be what it was?

aprilandjan:
It is really frustrating finding out that  currently these no way to silence these extra logs when using in node api 😭

Issue #2273, opened 2019-10-09
Self-signed SSL certs not valid on MacOS 10.15 (Catalina)


tarrall:
(PR inbound)

- Operating System:  MacOS 10.15
- Node Version: 12.9.0
- NPM Version: 6.10.3
- webpack Version: 4.39.3
- webpack-dev-server Version: 3.8.2
- Browser: Chrome

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

(Usual webpack setup; no specific code needed to repro)

### Expected Behavior
Autogenerated SSL certificate should be accepted by common browsers.

### Actual Behavior
As of MacOS 10.15, the SSL certificate validation logic is more strict and requires some additional certificate extensions.  Ref: https://support.apple.com/en-us/HT210176

### For Bugs; How can we reproduce the behavior?
Start up webpack-dev-server with SSL enabled on a MacOS 10.15 system; attempt to hit the server in Chrome (probably Safari as well).

stefanovualto:
@evilebottnawi , I am still having a `NET::ERR_CERT_INVALID` with version `3.8.2` which should  includes #2274

evilebottnawi:
Release will be today

evilebottnawi:
Done https://github.com/webpack/webpack-dev-server/pull/2274

evilebottnawi:
Release https://github.com/webpack/webpack-dev-server/releases/tag/v3.9.0

PR #2274, opened 2019-10-09
fix(server): Add extKeyUsage to self-signed cert

tarrall:
For issue #2273 

- [x] This is a **bugfix**
- [ ] This is a **feature**

Sorry, I did not.  I am not sure if the codebase has tests to ensure that the generated cert works on MacOS?

I did manually test to confirm the new cert works on both MacOS 10.14 and MacOS 10.15; it should not impact users on other platforms.

### Motivation / Use-Case

Self-signed cert does not work on MacOS 10.15 Catalina.

### Breaking Changes

None.

### Additional Info

MacOS 10.15 Catalina has additional requirements for self-signed
certs: https://support.apple.com/en-us/HT210176

Chrome (and, I assume, Safari) will not let you clickthrough the
"invalid cert" warning (error is NET::ERR_CERT_INVALID) if the
ExtendedKeyUsage extension is not present with at least the
id-kp-serverAuth OID. 

Firefox does not seem to be affected; I think they may have their
own SSL stack.


evilebottnawi:
@tarrall hm, so with this PR, it is allow to clickthrough the
"invalid cert" warning, right? Can you provide screen shots before and after? Anyway thanks for the PR

tarrall:
That is correct; apparently clickthrough not allowed with `NET::ERR_CERT_INVALID` but is allowed with `NET::ERR_CERT_AUTHORITY_INVALID`.

I should be able to get some screenshots tomorrow AM (USA time).

evilebottnawi:
@tarrall before you also have `NET::ERR_CERT_INVALID`? Maybe we should report to repo which we use for generating certs

tarrall:
Before and after screenshots attached -- courtesy of a teammate.

Re reporting upstream... I considered it, but I think the upstream modules are functioning as intended.  You're using the `selfsigned` module, which in turn uses `forge`.  They're just doing what the call tells them to do; since webpack-dev-server doesn't ask for the extended key usage extensions (before this PR) it's correct for them to not add those.

The cert isn't outright a "bad cert"; it just doesn't include all the stuff that MacOS 10.15 demands of a cert **used for a website**.

The [forge example](https://github.com/digitalbazaar/forge/blob/master/examples/create-cert.js#L47-L52) includes these extensions.  (I skipped "emailProtection" in this PR because I can't think of any case where the self-signed cert we're creating would be used for that purpose.)

Before:

before_screenshot

After:

after_screenshot

evilebottnawi:
@tarrall thanks for feedback, we can merge this PR for better development experience, but we should ivnestigate why cert doesn't work on MacOS 10.15 Catalina. Maybe you can do it and help us?

tarrall:
MacOS 10.15 Catalina has additional requirements for self-signed
certs.  Please see: https://support.apple.com/en-us/HT210176 for documentation.

While Apple does not say in that doc exactly **why** they did it, I think "improved security" is the general goal; they are ensuring that only certificates which explicitly say they're intended to be used for server authentication.

Per https://tools.ietf.org/html/rfc5280#page-44 clients are allowed to require specific extensions:

>    [...] Certificate using
>    applications MAY require that the extended key usage extension be
>    present and that a particular purpose be indicated in order for the
>    certificate to be acceptable to that application.

Since the certificate you are generating is for server authentication, you should always include that extension.

taiken:
@evilebottnawi what about merging this hotfix? we cant run development with https on updated MACos ;(

> @tarrall thanks for feedback, we can merge this PR for better development experience, but we should ivnestigate why cert doesn't work on MacOS 10.15 Catalina. Maybe you can do it and help us?

olivandesign:
Same here. Really looking forward for this hotfix
> @evilebottnawi what about merging this hotfix? we cant run development with https on updated MACos ;(
> 
> > @tarrall thanks for feedback, we can merge this PR for better development experience, but we should ivnestigate why cert doesn't work on MacOS 10.15 Catalina. Maybe you can do it and help us?

tarrall:
OK I **think** the one failing lint that's left is from the commit message... in theory should pass if run again?  Not positive, not sure how to test that locally and can't find anything in the repo that defines the required spec for commit messages.  Please feel free to reword that commit message if necessary!

evilebottnawi:
@tarrall don't worry we fix it before merge

l-hendriks:
In the meantime I found out you can visit the local webpage in safari but not in chrome. So until this fix is released you can at least get the dev environment running in safari.

k1sul1:
Yeah let me just fire up Safari on my linux laptop... 

Thanks for fucking all chromium based browsers. Luckily 3.8.2 still works.

evilebottnawi:
@k1sul1 what is problem?

k1sul1:
#2313

akx:
This _does_ break Chrome on Linux, unfortunately. See my discussion and investigation in https://github.com/webpack/webpack-dev-server/issues/2313#issuecomment-553046535.

PR #2276, opened 2019-10-10
fix: add hostname and port to bonjour name to prevent name collisions

roblan:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

It was impossible to run mutliple instances of webpack-dev-server if they all had `bonjour` because of 'name already in use' error.

### Breaking Changes

No

### DIFF
--- a/lib/utils/runBonjour.js
+++ b/lib/utils/runBonjour.js
@@ -2,9 +2,10 @@
 
 function runBonjour({ port }) {
   const bonjour = require('bonjour')();
+  const os = require('os');
 
   bonjour.publish({
-    name: 'Webpack Dev Server',
+    name: `Webpack Dev Server ${os.hostname()}:${port}`,
     port,
     type: 'http',
     subtypes: ['webpack'],
--- a/test/server/utils/__snapshots__/runBonjour.test.js.snap
+++ b/test/server/utils/__snapshots__/runBonjour.test.js.snap
@@ -2,7 +2,7 @@
 
 exports[`runBonjour should call bonjour.publish 1`] = `
 Object {
-  "name": "Webpack Dev Server",
+  "name": "Webpack Dev Server hostname:1111",
   "port": 1111,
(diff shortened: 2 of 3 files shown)


roblan:
@evilebottnawi test added

lib/utils/runBonjour.js
```diff
@@ -2,9 +2,10 @@
 
 function runBonjour({ port }) {
   const bonjour = require('bonjour')();
+  const os = require('os');
 
   bonjour.publish({
-    name: 'Webpack Dev Server',
+    name: `Webpack Dev Server ${os.hostname()}:${port}`,
```
evilebottnawi: Just questions - this names is used somewhere?

Sorry, i am not familiar with `bonjour`,

roblan: bonjour is used to broadcast services in current network, and it's used as service name

evilebottnawi: maybe better use `hostname` from here https://github.com/webpack/webpack-dev-server/blob/61b949e076b887734d318889e8a6420eccd99404/lib/Server.js#L743

roblan: I've check it, and it's set to 'localhost' (and it will be in 9/10 cases) so it could still colide with some other webpack-dev-server in local network.

evilebottnawi: @roblan port should be difference in this case, you can't run multiple webpack-dev-server using the same port

roblan: @evilebottnawi not on the same machine, but this name should be unique across local network.

evilebottnawi: :+1:

evilebottnawi:
Need fix lint problem and we can merge

roblan:
Should be ok now

evilebottnawi:
Thanks!

PR #2303, opened 2019-11-05
fix(client): don't override protocol for socket connection to 127.0.0.1

### LINKED ISSUE #2302: Allow http: connection to 127.0.0.1
- [ ] This is a **bug**
- [X] This is a **modification** request

### Code

```js
// webpack.config.js

...
devServer: {
  disableHostCheck: true,
  public: "http://127.0.0.1:3333",
  port: 3333,
  host: "127.0.0.1"
}
```

### Expected Behavior

Site is loaded via `https://`, dev-server is running on `http://127.0.0.1:3333`. I expect I can connect with Chrome and Firefox because `http://` should be allowed for `127.0.0.1` without errors.

### Actual Behavior

Webpack overrides the protocol with `https:`: https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/utils/createSocketUrl.js#L47-L56

Which results in wrong protocol error.

### For Features; What is the motivation and/or use-case for the feature?

Chrome and Firefox accept `http://` connection to `127.0.0.1` even when the actual website is loaded via `https://`:

```
> window.location
... some site with https (and without CORS, otherwise 127.0.0.1 will get CORS errors)
> fetch('http://example.org');
... will fail because mixed content is not allowed
> fetch('http://127.0.0.1');
... works
```

But webpack-dev-server forces `https`: https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/utils/createSocketUrl.js#L47-L56 

I assume the check should be:

```
  if (
    hostname &&
    hostname !== '127.0.0.1' && ## line added
    (self.location.protocol === 'https:' || urlParts.hostname === '0.0.0.0')
  ) {
    protocol = self.location.protocol;
  }
```

slightly related:
* If `transportMode` with `sockjs` is used this will block the request: https://github.com/sockjs/sockjs-client/issues/486
* If `transportMode` with `ws` is used it works in Chrome but FF request at least version 70: https://bugzilla.mozilla.org/show_bug.cgi?id=1376309

nougad:
Chrome and Firefox accept `http://`/`ws://` mixed-content connection to
`127.0.0.1` even when the actual website is loaded via `https://`.

Fixes #2302 

- [X] This is a **bugfix**
- [X] This is a **feature**

I'm not sure if the [current tests](https://github.com/webpack/webpack-dev-server/blob/master/test/client/utils/__snapshots__/createSocketUrl.test.js.snap) actually contain correct assertions.

### Motivation / Use-Case

described in https://github.com/webpack/webpack-dev-server/issues/2302

### Breaking Changes

nope

### DIFF
--- a/client-src/default/utils/createSocketUrl.js
+++ b/client-src/default/utils/createSocketUrl.js
@@ -46,6 +46,7 @@ function createSocketUrl(resourceQuery) {
   // because the browser doesn't accept non-secure websockets.
   if (
     hostname &&
+    hostname !== '127.0.0.1' &&
     (self.location.protocol === 'https:' || urlParts.hostname === '0.0.0.0')
   ) {
     protocol = self.location.protocol;
--- a/test/client/utils/__snapshots__/createSocketUrl.test.js.snap
+++ b/test/client/utils/__snapshots__/createSocketUrl.test.js.snap
@@ -4,6 +4,8 @@ exports[`createSocketUrl should return the url when __resourceQuery is ?test 1`]
 
 exports[`createSocketUrl should return the url when __resourceQuery is http://0.0.0.0 1`] = `"http://localhost/sockjs-node"`;
 
+exports[`createSocketUrl should return the url when __resourceQuery is http://127.0.0.1 1`] = `"http://127.0.0.1/sockjs-node"`;
+
 exports[`createSocketUrl should return the url when __resourceQuery is http://user:pass@[::]:8080 1`] = `"http://user:pass@localhost:8080/sockjs-node"`;
 
 exports[`createSocketUrl should return the url when __resourceQuery is http://user:password@localhost/ 1`] = `"http://user:password@localhost/sockjs-node"`;
@@ -21,6 +23,8 @@ exports[`createSocketUrl should return the url when __resourceQuery is undefined
 exports[`createSocketUrl should return the url when the current script source is ?test 1`] = `"/sockjs-node"`;
 
 exports[`createSocketUrl should return the url when the current script source is http://0.0.0.0 1`] = `"http://localhost/sockjs-node"`;
+
+exports[`createSocketUrl should return the url when the current script source is http://127.0.0.1 1`] = `"http://127.0.0.1/sockjs-node"`;
 
 exports[`createSocketUrl should return the url when the current script source is http://user:pass@[::]:8080 1`] = `"http://user:pass@localhost:8080/sockjs-node"`;
 
--- a/test/client/utils/createSocketUrl.test.js
+++ b/test/client/utils/createSocketUrl.test.js
@@ -10,6 +10,7 @@ describe('createSocketUrl', () => {
     'http://0.0.0.0',
     'https://localhost:123',
     'http://user:pass@[::]:8080',
+    'http://127.0.0.1',
     // TODO: comment out after the major release
     // https://github.com/webpack/webpack-dev-server/pull/1954#issuecomment-498043376
     // 'file://filename',


nougad:
Thank you @hiroppy for your fast comment. Looking at the current tests it seams like the serialization of the `URL` object generated wrong snapshots: 

https://github.com/webpack/webpack-dev-server/blob/master/test/client/utils/__snapshots__/createSocketUrl.test.js.snap#L13

That makes adding new tests hard since the assertions are arbitrary. Is my assumption correct? Is there a ticket for replacing the snapshot with actual tests?

hiroppy:
I think this line is incorrect. https://github.com/webpack/webpack-dev-server/blob/00903f6b798fd9d6f42728c090ae9a8b9018b8d5/client-src/default/utils/createSocketUrl.js#L14

> Is there a ticket for replacing the snapshot with actual tests?

Currently no. I'll check this. thanks!

nougad:
Awesome! let me know when you found the bug. I'm happy to add tests for this PR :smile:

odbayar:
Will it work for "localhost" too?

nougad:
@odbayar AFAIK it only works with `127.0.0.1`.

@hiroppy That line is actually correct. The method expects only a query like `?http://0.0.0.0&asdf=32` as parameter but the tests provide a full URL. If I remove the `substr` other tests fail.

I updated the tests to pass only the querystring as parameter but that lines are fishy as well:

https://github.com/webpack/webpack-dev-server/blob/00903f6b798fd9d6f42728c090ae9a8b9018b8d5/client-src/default/utils/createSocketUrl.js#L19-L22

which leads to wrong tests like that one:

https://github.com/webpack/webpack-dev-server/blob/master/test/client/utils/__snapshots__/createSocketUrl.test.js.snap#L23-L25

I removed the block since it looks unused and I could not make any sense out of it.

Lastly I cleaned some unrelated unit tests to use the correct assertions and refactored the `createSocketUrl` file to make it easier testable. I made everything a separate commit. Let me know if you want separate PRs instead.

nougad:
@hiroppy could you please have a second look if the improvements for unit testing make sense? Thanks

nougad:
@evilebottnawi I understand this change was growing a bit more than expected over time. Please have a look at the individual commits:

* https://github.com/webpack/webpack-dev-server/pull/2303/commits/75e620a9c9874316b236323e5ebd8c18d747bb9f - thats the actual change. But as @hiroppy pointed out above, the unit tests for that file are incorrect
* https://github.com/webpack/webpack-dev-server/pull/2303/commits/0df1f0c0cf67063b0ea26daa79fbe00084527971 fixes then the unit tests for the file
* https://github.com/webpack/webpack-dev-server/pull/2303/commits/d1a86d742eb4d7346d1fdeefbe05a38efe0de528 is a general improvement of unit test assertions and I'm happy to discard/move into a different PR
* https://github.com/webpack/webpack-dev-server/pull/2303/commits/8b0ef4a28acb4ae7922d0820740b48d5428ceccf is the actual big diff because it introduces some refactoring into the file to make it easier to test. Again, I'm happy of not adding those additional tests/refactoring or move it into a second PR to unblock the actual change.

Please let me know which change you want in a separate PR or if it's ok to review it as part of this since everything is kind of related.

evilebottnawi:
@nougad let's do refactor and fix tests in other PRs for better git history and good changelog

nougad:
I created two separate PRs with fixing tests and improving assertions:

* https://github.com/webpack/webpack-dev-server/pull/2323 - fix(client): fix tests for createSocketUrl 
* https://github.com/webpack/webpack-dev-server/pull/2324 - fix(test): replace include.toBeTruthy with toContain

I will update this PR here one the test fix is merged since it's a prerequisite to have working unit tests.

nougad:
@evilebottnawi rebased and updated to only include the one relevant commit.

Once merged I will create another PR with the refactoring on top.

nougad:
Is there anything left for this PR? I promise even more extensive testing with the next PR once this one is merged :) https://github.com/nougad/webpack-dev-server/commit/acd46c872abe64fad0f92904e154a5e7a88be55f

hiroppy:
thanks

BinarySpike:
Could this be expanded to `localhost` as well?

Issue #2310, opened 2019-11-09
transportMode: 'ws' interferes with proxy { ws: true }


joonas-lahtinen:
- Operating System: any
- Node Version: any
- NPM Version: any
- webpack Version: any
- webpack-dev-server Version: any
- Browser: any

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

```js
// webpack.config.js
...
devServer: {
  transportMode: 'ws',
  proxy: {
    '/api': {
      target: 'http://localhost:8888',
      ws: true,
    },
  },
},
...
```

### Expected Behavior

Proxying Websockets works similar to `transportMode: 'sockjs'`

### Actual Behavior

Websocket connections get closed with `400 Bad Request` error.

### For Bugs; How can we reproduce the behavior?

Simultaneously enable `transportMode: 'ws'` and `proxy: { ws: true }`

hiroppy:
fixed by #2311, thank!

PR #2311, opened 2019-11-09
fix(server): respect sockPath on transportMode: 'ws' (#2310)

joonas-lahtinen:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Added tests to verify that sockPath is respected.

### Motivation / Use-Case

Fixes bug #2310.


evilebottnawi:
Please accept CLA

lib/servers/WebsocketServer.js
```diff
@@ -3,22 +3,38 @@
...
+      });
+    });
+
     this.wsServer.on('error', (err) => {
       this.server.log.error(err.message);
     });
   }
 
+  shouldUpgrade(req) {
+    const pathname = url.parse(req.url).pathname;
```
hiroppy: memo: we'll use URLSearchParams after major release.

joonas-lahtinen:
Thanks for the quick reviews. I noticed a bug in the tests based on the code coverage difference report, should be fixed now and using the more forward compatible ws.Server.shouldHandle()

lib/servers/WebsocketServer.js
joonas-lahtinen: I fixed the code to use WebSocket.Server.shouldHandle()
https://github.com/websockets/ws/blob/master/doc/ws.md#servershouldhandlerequest
Should be more futureproof?

joonas-lahtinen:
This also fixes the bug that whole `webpack-dev-server` process exits due to an unhandled error if the proxy target is not open:

`[HPM] Rewriting path from "/api" to ""
[HPM] GET /api ~> http://localhost:8888
[HPM] Upgrading to WebSocket
events.js:174
      throw er; // Unhandled 'error' event
      ^

Error [ERR_STREAM_DESTROYED]: Cannot call write after a stream was destroyed
`

Loonride:
@joonas-lahtinen I ran your new tests without the changes that you made, and they passed. How are your changes different from the existing behavior? The `ws` docs say for the path option - "Accept only connections matching this path" (https://github.com/websockets/ws/blob/master/doc/ws.md#new-websocketserveroptions-callback), so paths that do not match the one specified are already rejected.

My suggestion is simply to remove the changes you made but keep the tests for us to merge, because the tests are useful for articulating this.

joonas-lahtinen:
Thanks for the patience. After some digging I found the spot to inject testing the different transportMode options with proxy.

Loonride:
LGTM /cc @evilebottnawi @hiroppy

hiroppy:
thanks

PR #2336, opened 2019-11-29
feat(client): refactor createSocketUrl to make it better unit tested

nougad:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

Follow up from https://github.com/webpack/webpack-dev-server/pull/2303 to add more tests


client-src/default/utils/createSocketUrl.js
```diff
@@ -3,41 +3,60 @@
...
     // If this bundle is inlined, use the resource query to get the correct url.
-    // strip leading `?` from query string
-    urlParts = url.parse(resourceQuery.substr(1));
+    // format is like `?http://0.0.0.0:8096&sockPort=8097&sockHost=localhost`
+    urlParts = url.parse(
+      resourceQuery
+        // strip leading `?` from query string to get a valid URL
+        .substr(1)
+        // replace first `&` with `?` to have a valid query string
+        .replace('&', '?'),
```
evilebottnawi: Can you clarify here?

client-src/default/utils/createSocketUrl.js
```diff
@@ -3,41 +3,60 @@
...
+      true
+    );
   } else {
     // Else, get the url from the <script> this file was called with.
     const scriptHost = getCurrentScriptSource();
-    urlParts = url.parse(scriptHost || '/', false, true);
+    urlParts = url.parse(scriptHost || '/', true, true);
```
evilebottnawi: Why we need parse query string here?

client-src/default/utils/createSocketUrl.js
nougad: The full URL is something like `http://example.com/?http://0.0.0.0:8096&sockPort=8097&sockHost=localhost` while this method only gets the path: `?http://0.0.0.0:8096&sockPort=8097&sockHost=localhost`

This will convert the path into a full url by 1) removing the leading `?` and 2) replace the first `&` with a `?`. The result is a valid URL which can be parsed: `url.parse('http://0.0.0.0:8096?sockPort=8097&sockHost=localhost')`

The removal of the leading `?` was already done before. The difference is the replace of `&` to get a valid URL. That allows to use `urlParts.query` instead of before `querystring.parse(urlParts.path)`.

I agree the parsing and replacement is a bit funky but I also don't get why the parameter is a invalid url instead of just passing the host as a normal query parameter.

client-src/default/utils/createSocketUrl.js
nougad: This is related to the comment above. Having a full URL including parsed query string saves from another `querystring.parse(urlParts.path)` later on.

client-src/default/utils/createSocketUrl.js
evilebottnawi: @nougad what about(rewrite) fix it in other PRs?

evilebottnawi:
Thanks!

client-src/default/utils/createSocketUrl.js
nougad: Not sure how to fix this because the actual problem comes from a problematic protocol to signal the sock endpoint. I don't think I have enough knowledge about the details to change the protocol.

PR #1838, opened 2019-05-01
fix(client-src): don't use self.location.port

hiroppy:
ISSUE: https://github.com/webpack/webpack-dev-server/issues/1777
REF: https://github.com/webpack/webpack-dev-server/pull/1792
REF: https://github.com/webpack/webpack-dev-server/pull/1664

- [x] This is a **bugfix**
- [ ] This is a **feature**
- [ ] This is a **code refactor**
- [ ] This is a **test update**
- [ ] This is a **docs update**
- [ ] This is a **metadata update**

### For Bugs and Features; did you add new tests?

Revert from #1664.

### Motivation / Use-Case

this pr is just a revert, but it is WIP status.(to validate)

### Breaking Changes

nope

### Additional Info

fixes #1777

PR #1792, opened 2019-04-11
feat: add sockPort options in queryString

zhangyuang:
- [ ] This is a **bugfix**
- [x] This is a **feature**
- [ ] This is a **code refactor**
- [ ] This is a **test update**
- [ ] This is a **docs update**
- [ ] This is a **metadata update**

### For Bugs and Features; did you add new tests?
in tests/client.test.js

### Motivation / Use-Case

i find webpack-dev-server start socketjs/client and the port use location.port
if only client side rendering it's ok
but i use server side rendering, so i have two ports with server: 7001, client:8080
when i open localhost:7001 sockjs-node send websocket request to ws://localhost:7001
but in fact it should  send websocket request to ws://localhost:8080
so i feel the port will be webpack-dev-server start port and not location.port
so i add a sockPort option which we can custom sockJs port
image

### Breaking Changes
No

PR #2341, opened 2019-12-06
feat(client): allow sock port to use browser location's port

jkrehm:
Add the ability to set `sockPort: 'location'` to make the sock port dynamic.

This change would allow the port to be specified by an individual proxy, e.g. if different team members have their environments configured differently.

- [ ] This is a **bugfix**
- [x] This is a **feature**

I added the scenario to an existing test.

### Motivation / Use-Case

Related to #1664 which was semi-addressed by #1792 and then reverted by #1838.

This change provides a way for the sock port to use the browser location's port which may or may not be known at build time. For example,
* one team member hits webpack-dev-server directly at http://localhost:3000
* another uses a Docker container that proxies it via http://localhost.example.com
* a third uses Nginx to proxy through SSL with https://dev.example.com

For all three to be able to use it the way they want they would have to modify their local copy of webpack.config.js to set the `sockPort` to either 3000, 80, or 443 and then remember to never check in their change. Alternatively a "magic" value (in this case `"location"`) would be set and the client JS would swap in the browser's port number and all three developers would happily use their current setup.

### Breaking Changes

No breaking changes (additive only).

### Additional Info

N/A

### DIFF
--- a/client-src/default/utils/createSocketUrl.js
+++ b/client-src/default/utils/createSocketUrl.js
@@ -76,7 +76,11 @@ function getSocketUrl(urlParts, loc) {
   // they are not provided
   const sockHost = query.sockHost || hostname;
   const sockPath = query.sockPath || '/sockjs-node';
-  const sockPort = query.sockPort || port;
+  let sockPort = query.sockPort || port;
+
+  if (sockPort === 'location') {
+    sockPort = loc.port;
+  }
 
   return url.format({
     protocol,
--- a/test/client/utils/createSocketUrl.test.js
+++ b/test/client/utils/createSocketUrl.test.js
@@ -141,6 +141,16 @@ describe('createSocketUrl', () => {
       'http://localhost',
       'http://localhost:8097/sockjs-node',
     ],
+    [
+      '?http://example.com:8096&sockPort=location',
+      'http://something.com',
+      'http://example.com/sockjs-node',
+    ],
+    [
+      '?http://0.0.0.0:8096&sockPort=location',
+      'http://localhost:3000',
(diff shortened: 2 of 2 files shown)


jkrehm:
I committed using my work email address and I did not have that attached to my Github account. I do now. I've signed the CLA. Is there a way to kick the bot and tell it to try again?

rishabh3112:
@jkrehm closing and reopening the PR would help here.

jkrehm:
Closing and re-opening fixed the CLA issue. I'm not sure what that latest Windows node-12 failure is about. I don't think it's related to my changes (all the rest succeeded and its complaint is that a port is already in use).

What would you like me to do? Are you all right with ignoring the failure as a false-negative?

rishabh3112:
@jkrehm I don't think so we could ignore. Tagging @hiroppy for more info on this.

jkrehm:
I closed the PR and re-opened again and now the Windows node-10 pipeline failed with the same "address already in use" error. Maybe the pipelines are running on the same agent? If so and if two of the tests run at the same time and both try to bind to the same port it would generate that error.

rishabh3112:
I got the problem. I have tagged one of the code maintainer of the dev server. He will look into it soon.

hiroppy:
CI was re-executed and CI became green. CI is sometimes flaky.

evilebottnawi:
Thanks for the PR

jkrehm:
Thank you!

saiqulhaq:
this feature should be written on documentation

evilebottnawi:
PR welcome to docs

PR #2150, opened 2019-07-24
feat(server): add `contentBasePublicPath` option

iamandrewluca:
- [x] This is a **feature**
- [x] Test
- [x] Docs https://github.com/webpack/webpack.js.org/pull/3291

Tell the server at what URL to serve `devServer.contentBase`. If there was a file `assets/manifest.json`, it would be served at `/serve-content-base-at-this-url/manifest.json`

__webpack.config.js__

```javascript
module.exports = {
  //...
  devServer: {
    contentBase: path.join(__dirname, 'assets'),
    contentBasePublicPath: '/serve-content-base-at-this-url'
  }
};
```

### Motivation / Use-Case

Now `webpack-dev-server` serves static files (`contentBase`) from root ignoring `publicPath`
Using added option we can serve static files from `publicPath`
This use case is needed in create-react-app, when want to serve all files from `publicPath`
doesn't matter in development or production

~~https://github.com/iamandrewluca/plsdelete-wds-use-case~~
Above repo will be deleted because does not reflect current changes after review.
From start the option was intended to be `contentBasePrependPublic: boolean`

### Breaking Changes

No breaking changes

### Related

https://github.com/facebook/create-react-app/pull/7259
https://github.com/facebook/create-react-app/issues/6135
https://github.com/facebook/create-react-app/pull/6280

PR #2374, opened 2019-12-26
fix(server): fallthrough non `GET` and `HEAD` request to routes defined in after option

### LINKED ISSUE #2370: The [before] and [after] option hook not allow http POST request?
Operating System: Mac OS 10.15.2
Node Version: 13.5.0
NPM Version: 6.13.4
webpack Version: 4.41.4
webpack-dev-server Version: 3.10.1
Browser: Chrome

! This is a bug

I was writing an express middleware, try to send request dev server, but it not allowed!
** But webpack-dev-server 3.9.0 work fine!!!**

Screen Shot 2019-12-25 at 11 22 09 AM

jvasseur:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

Before #2150, non-GET request where forwarded to any middleware defined in the after option, since #2150, they are handled by `serve-index` instead that always reply with a 405 error.

This fix issue #2370.

### Breaking Changes

No

### Additional Info

Closes #2370

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -390,13 +390,27 @@ class Server {
 
     if (Array.isArray(contentBase)) {
       contentBase.forEach((item) => {
-        this.app.use(contentBasePublicPath, serveIndex(item));
+        this.app.use(contentBasePublicPath, (req, res, next) => {
+          // serve-index doesn't fallthrough non-get/head request to next middleware
+          if (req.method !== 'GET' && req.method !== 'HEAD') {
+            return next();
+          }
+
+          serveIndex(item)(req, res, next);
+        });
       });
     } else if (
       typeof contentBase !== 'number' &&
       !isAbsoluteUrl(String(contentBase))
     ) {
-      this.app.use(contentBasePublicPath, serveIndex(contentBase));
+      this.app.use(contentBasePublicPath, (req, res, next) => {
+        // serve-index doesn't fallthrough non-get/head request to next middleware
+        if (req.method !== 'GET' && req.method !== 'HEAD') {
+          return next();
+        }
+
+        serveIndex(contentBase)(req, res, next);
+      });
     }
   }
 
--- a/test/server/after-option.test.js
+++ b/test/server/after-option.test.js
@@ -29,6 +29,10 @@ describe('after option', () => {
           appArg.get('/after/some/path', (_, response) => {
             response.send('after');
           });
+
+          appArg.post('/after/some/path', (_, response) => {
+            response.send('after POST');
+          });
         },
         port,
       },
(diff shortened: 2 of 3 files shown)


lib/Server.js
```diff
@@ -390,13 +390,25 @@ class Server {
 
     if (Array.isArray(contentBase)) {
       contentBase.forEach((item) => {
-        this.app.use(contentBasePublicPath, serveIndex(item));
+        this.app.use(contentBasePublicPath, (req, res, next) => {
+          if (req.method !== 'GET') {
```
evilebottnawi: Not sure it is good solution, developers can want to handle `GET` in own middleware too

jvasseur: AFAIK, `serve-index` let `GET` request that it can't handle fall-back to the next middleware, it's just non-GET request that are blindly replied as `405`.

This is basically a revert of the change from `app.get` to `app.use` done in #2150 while still keeping the `contentBasePublicPath` feature.

jvasseur:
@evilebottnawi the `serve-static` doesn't exhibit this problem because it already forward non-GET request to the next middleware : https://github.com/expressjs/serve-static/blob/master/index.js#L73-L76

iamandrewluca:
I'm taking a look into `serve-index` implementation

lib/Server.js
iamandrewluca:
```suggestion
          if (req.method !== 'GET' && req.method !== 'HEAD') {
```

lib/Server.js
```diff
@@ -390,13 +390,25 @@ class Server {
...
+          serveIndex(item)(req, res, next);
+        });
       });
     } else if (
       typeof contentBase !== 'number' &&
       !isAbsoluteUrl(String(contentBase))
     ) {
-      this.app.use(contentBasePublicPath, serveIndex(contentBase));
+      this.app.use(contentBasePublicPath, (req, res, next) => {
+        if (req.method !== 'GET') {
```
iamandrewluca:
```suggestion
        if (req.method !== 'GET' && req.method !== 'HEAD') {
```

evilebottnawi: Can you clarify?

iamandrewluca:
`serve-static` has `fallthrough`, but we don't use it in our case, it must be added.
`serve-index` [does not have ](https://github.com/expressjs/serve-index/pull/61) `fallthrough`, and I agree with the workaround in this PR.

@jvasseur also make sure you check also for `HEAD` method
```js
if (req.method !== 'GET' && req.method !== 'HEAD') {
// ...
```

jvasseur:
@iamandrewluca `true` is already the [default value](https://github.com/expressjs/serve-static/blob/master/index.js#L51) for the fallthrough parameter of `serve-static`.

I've added the check for `HEAD` in the code.

iamandrewluca:
recommend to rename PR into
```
fix(server): fallthrough non `GET` and `HEAD` request forward to routes defined in after option
```

iamandrewluca:
@jvasseur pls add a comment in code, that `serve-index` does not fallthrough like `serve-static`

jvasseur:
@iamandrewluca done

iamandrewluca:
@jvasseur add at the end of PR description this. It will automatically close the issue when PR is merged.
```
Closes #2370
```

jvasseur:
Done

hiroppy:
Thanks!

iamandrewluca:
Released in https://github.com/webpack/webpack-dev-server/releases/tag/v3.10.2

PR #2385, opened 2020-01-08
Chore upgrade http proxy middleware

lPadier:

### Motivation / Use-Case
This upgrades `http-proxy-middleware` to v0.20.0 and updates `micromatch` (a transitive dependency) to v4.x.x

### Breaking Changes
`micromatch` 4 drops support for node < 8, but the `next` branch has already dropped support for node < 8.


evilebottnawi:
Something wrong with jest on macos with node@13 :confused:

evilebottnawi:
Done for next

PR #2386, opened 2020-01-08
feat(server): forward logger to dev-middleware

nicolasbrugneaux:
- [ ] This is a **bugfix**
- [x] This is a **feature**

### Motivation / Use-Case

This is to help reading the output of the webpack-dev-middleware when using a single compiler and multiple servers. Currently, the logs are not very useful, since they just prefix everything with `[wdm]` because the logger isn't forwarded via options, and we can't pass it via `options` since it's not part of the schema.

I attached a screenshot that explains why it's a bit hard to read the logs.

### Breaking Changes

I don't believe this can break anything

### Additional Info

Current logs:
Screen Shot 2020-01-08 at 17 41 26

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -114,7 +114,7 @@ class Server {
     this.setupHooks();
     this.setupApp();
     this.setupCheckHostRoute();
-    this.setupDevMiddleware();
+    this.setupDevMiddleware(_log);
 
     // set express routes
     routes(this.app, this.middleware, this.options);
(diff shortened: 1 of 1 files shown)


nicolasbrugneaux:
cool, if you use webpack-logger instead then the issue _should_ resolve itself! I'll keep an eye out :) Thanks for answering!

evilebottnawi:
Closing, anyway, thanks for the PR

PR #2396, opened 2020-01-18
Adds an optional peer dependency on webpack-cli

arcanis:
- [x] This is a **bugfix**

Can publish a package with this PR and see if it actually works but these kind of fixes have been done multiple times to make Yarn 2 happy. Considering `webpack-dev-server` throws at runtime anyway if `webpack-cli` is missing this shouldn't bother anyone

### Motivation / Use-Case

Yarn 2 currently throws when using `webpack-dev-server` due to it trying to require `webpack-cli` without declaring it anywhere in its dependencies:

```

```

### Breaking Changes

There is no breaking change for anyone - not even a warning for older package managers. This is a completely safe bugfix that will only impact Yarn users.

### Additional Info

Supercedes: https://github.com/webpack/webpack-dev-server/pull/2191

Documentation: https://next.yarnpkg.com/configuration/manifest#peerDependenciesMeta.optional

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,6 +116,11 @@
   "peerDependencies": {
     "webpack": "^4.0.0 || ^5.0.0"
   },
+  "peerDependenciesMeta": {
+    "webpack-cli": {
+      "optional": true
+    }
+  },
   "author": "Tobias Koppers @sokra",
   "bugs": "https://github.com/webpack/webpack-dev-server/issues",
   "homepage": "https://github.com/webpack/webpack-dev-server#readme",


arcanis:
Ping @evilebottnawi?

evilebottnawi:
@arcanis npm still not fixed bug about warning optional peer deps?

arcanis:
They have fixed in November, so if you want to merge the other PR it'll be fine. If you're still a bit wary that's ok, that's why I opened this alternate PR: it only adds the `peerDependenciesMeta` field (without explicit `peerDependency`, so it won't affect npm at all. It's a good in-between.

evilebottnawi:
@hiroppy @Loonride feel free to feedback

Issue #2403, opened 2020-01-27
WebSocket transportMode is missing heartbeat


ianschmitz:
- Operating System:
- Node Version: 10.18.1
- NPM Version: 	6.13.4
- webpack Version: 4.41.5
- webpack-dev-server Version: 3.9.0
- Browser: Chrome 79

- [x] This is a **bug**
- [ ] This is a **modification** request

### Expected Behavior
webpack-dev-server sends keep alive heartbeat when using `transportMode: 'ws'`, [consistent with sockjs-node](https://github.com/sockjs/sockjs-node/blob/4c9f3009be13962b0785138ccdf527c5abe925f6/lib/transport/websocket.js#L74-L82) to keep the connection open while developing.

### Actual Behavior
WebSocket connection times out after a period of inactivity. A page refresh is needed to reconnect.

Create React App related issue: https://github.com/facebook/create-react-app/issues/8091

### For Bugs; How can we reproduce the behavior?
Use `transportMode: 'ws'`, connect to running webpack-dev-server instance, and wait for socket to timeout.

Here's the RFC for ping/pong in the WebSocket spec: https://tools.ietf.org/html/rfc6455#section-5.5.2

evilebottnawi:
Easy to implement - https://github.com/websockets/ws#how-to-detect-and-close-broken-connections, somebody want to be a champion?

Loonride:
> Easy to implement - https://github.com/websockets/ws#how-to-detect-and-close-broken-connections, somebody want to be a champion?

I can do this

evilebottnawi:
Client doesn't require additional code (browser automatically sends ping)

PR #2404, opened 2020-01-29
feat(server): add ws heartbeat

Loonride:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes, and also restructured the test file for WebsocketServer slightly

### Motivation / Use-Case

Add ws heartbeat

#2403 (need investigate client code)

### Breaking Changes

None


Loonride:
> Good job, what about client? Second part of https://github.com/websockets/ws#how-to-detect-and-close-broken-connections

@evilebottnawi Good point. However, we are using a native browser WebSocket (https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API), not a `ws` client WebSocket. The browser WebSocket seems to not have a `ping` method, nor can it listen for pings from the server (https://stackoverflow.com/questions/10585355/sending-websocket-ping-pong-frame-from-browser). Do you think I should just implement a "ping" message which is not a real ping but accomplishes the same task?

evilebottnawi:
Hm, we need investigate that, i will look at that in near future

evilebottnawi:
I looked and tested the solution, it works well, I’m not sure about the client code, but the current code can be merged, a little later I will investigate the client in more detail

gamedev8:
@Loonride This is great, but I think you will still eventually get a timeout if you are stepping through your code in the browser dev tools. The browser is holding up the whole application so the client code  cannot communicate with the server. You can easily test this by setting a breakpoint and wait until it times out.

If that is true, then maybe the client code needs to be hardened as well. It should be able to reconnect to the server if it detects that the connection is broken.

Loonride:
> @Loonride This is great, but I think you will still eventually get a timeout if you are stepping through your code in the browser dev tools. The browser is holding up the whole application so the client code cannot communicate with the server. You can easily test this by setting a breakpoint and wait until it times out.
> 
> If that is true, then maybe the client code needs to be hardened as well. It should be able to reconnect to the server if it detects that the connection is broken.

@gamedev8 I tested this and it seems that the WebSocket client can still be pinged by the server even while stepping through your code in dev tools. Could you test this yourself and confirm?

gamedev8:
@Loonride I tested it a few days ago and produced the issue, but it is possible that I did not set up properly. I will try again.

PR #2410, opened 2020-01-29
refactor: remove the `features` option

evilebottnawi:
BREAKING CHANGE: the `features` option was remove without alternatives

- [ ] This is a **bugfix**
- [ ] This is a **feature**

existing

### Motivation / Use-Case

Remove unnecessary option

### Breaking Changes

Yes

### Additional Info

No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -557,7 +557,7 @@ class Server {
       runnableFeatures.push('after');
     }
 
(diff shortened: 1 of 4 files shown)


hiroppy:
`test/server/socket-option.test.js` fails...

evilebottnawi:
Yep, it is bug in jest, we ignore it right now, but we should resolve that before stable release

Issue #2380, opened 2020-01-02
Some errors are not handled correctly for proxied requests


OliverJAsh:
- Operating System: macOS 10.15.2
- Node Version: 12.8.1
- NPM Version: 6.10.2
- webpack Version: see `package.json` below
- webpack-dev-server Version: see `package.json` below
- Browser: N/A

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

```js
// webpack.config.js
module.exports = {
    devServer: {
        port: 4000,
        proxy: {
            '/api': 'http://localhost:3000'
        }
    }
}
```

```js
// index.js
const express = require("express");

const app = express();

app.get("*", (req, res) => {
  res.send("foo");
});

// error handling middleware
app.use((error, req, res, next) => {
  res.send("my custom error page");
});

app.listen(3000);
```

```jsonc
// package.json
{
  "dependencies": {
    "express": "^4.17.1",
    "webpack": "^4.41.5",
    "webpack-cli": "^3.3.10",
    "webpack-dev-server": "^3.10.1"
  }
}
```

### Expected Behavior

1. Run `node index.js`
2. Run `webpack-dev-server`
3. `curl "localhost:4000/api/%"`

The response should be `"my custom error page"`, indicating that the Express error handling middleware for the API server was used.

### Actual Behavior

The response is `URIError: Failed to decode param …`, indicating that the Express error handling middleware for the API server was _not_ used.

I would not expect `webpack-dev-server` to change how errors are handled for proxied requests.

#### Technical notes

I played around with using `http-proxy-middleware` on its own to see if the root issue was inside there, but using the following code, I could not reproduce this issue:

```ts
const httpProxyMiddleware = require("http-proxy-middleware");
const express = require("express");

{
  const app = express();

  app.get("*", (req, res) => {
    res.send("foo");
  });

  app.use((error, req, res, next) => {
    res.send("my custom error page");
  });

  app.listen(3000);
}

{
  const app = express();

  app.use(httpProxyMiddleware({ target: "http://localhost:3000" }));

  app.listen(4000);
}
```

### For Bugs; How can we reproduce the behavior?

See above

### For Features; What is the motivation and/or use-case for the feature?

N/A

evilebottnawi:
> I would not expect webpack-dev-server to change how errors are handled for proxied requests.

We don't do it, can you create minimum reproducible test repo?

OliverJAsh:
[...]

PR #2425, opened 2020-02-03
Also forward error requests to the proxy

OliverJAsh:
Fixes https://github.com/webpack/webpack-dev-server/issues/2380

- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

See https://github.com/webpack/webpack-dev-server/issues/2380 and https://github.com/webpack/webpack-dev-server/issues/2380#issuecomment-581378628.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -288,7 +288,7 @@ class Server {
         this.websocketProxies.push(proxyMiddleware);
       }
 
-      this.app.use((req, res, next) => {
+      const handle = (req, res, next) => {
         if (typeof proxyConfigOrCallback === 'function') {
           const newProxyConfig = proxyConfigOrCallback();
 
@@ -319,7 +319,11 @@ class Server {
         } else {
           next();
         }
-      });
+      };
+
+      this.app.use(handle);
+      // Also forward error requests to the proxy so it can handle them.
+      this.app.use((error, req, res, next) => handle(req, res, next));
     });
   }
 
--- a/test/server/proxy-option.test.js
+++ b/test/server/proxy-option.test.js
@@ -326,6 +326,19 @@ describe('proxy option', () => {
       // Parse application/json
       proxy.use(bodyParser.json());
 
+      // This forces Express to try to decode URLs, which is needed for the test
+      // associated with the middleware below.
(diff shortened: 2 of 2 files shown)


lib/Server.js
```diff
@@ -319,7 +319,11 @@ class Server {
         } else {
           next();
         }
-      });
+      };
+
+      this.app.use(handle);
+      // Also forward error requests to the proxy so it can handle them.
+      this.app.use((error, req, res, next) => handle(req, res, next));
```
evilebottnawi: It is create two instances of proxy that was reduce performance

OliverJAsh: I don't see how that will hurt performance.

The proxy middleware is created once, outside of `handle`:

```ts
      proxyMiddleware = getProxyMiddleware(proxyConfig);
```

… and the result is used in `handle` on each request, be it a non-error or error request.

evilebottnawi: Each request now run `handle` (https://github.com/webpack/webpack-dev-server/blob/29a822ea8426820617745b3f355c344e84f7c01b/lib/Server.js#L291)  twice so we potentially send require to middleware twice https://github.com/webpack/webpack-dev-server/blob/29a822ea8426820617745b3f355c344e84f7c01b/lib/Server.js#L318

OliverJAsh: I don't believe that's true. `handle` will only run once:

- If the request has not errored, `handle` will be called via `app.use(handle)`.
- If the request has errored, `handle` will be called via `app.use((error, req, res, next) => handle(req, res, next))`.

You can see this for yourself if you add some logging to `handle`.

evilebottnawi: Hm, looks you are right https://expressjs.com/en/guide/error-handling.html

evilebottnawi:
Something broken on CI, I rerun tests

evilebottnawi:
Bit thanks, sometimes I get too much error flow, and do not have time to investigate them in detail, I apologize for the situation and am very glad that you yourself could find the problem and fix it

OliverJAsh:
@evilebottnawi No problem, thanks for reviewing my PR! 😄

PR #2489, opened 2020-03-31
feat: allow multiple paths in contentBasePublicPath

jleifeld:
- [ ] This is a **bugfix**
- [x] This is a **feature**

I add tests to verify that multiple contentBasePublicPath's are working.

### Motivation / Use-Case

I need multiple static directories with different paths. The problem there is that there is no global root path for the assets.

### Breaking Changes

This PR is free of breaking changes and provide downwards compatibility.


evilebottnawi:
> I need multiple static directories with different paths. The problem there is that there is no global root path for the assets.

Can you provide real use case?

jleifeld:
When you have several micro applications which all have custom static folders:

Application 1: `/application-one/static/`
Application 2: `/application-two/static/`
Application 3: `/application-three/static/`

evilebottnawi:
@jleifeld just for information - you have one big app with own content static and multiple micro apps which used in on big app with own statics?

jleifeld:
Yes, that is the case. The main application is extendable with plugins. These plugins have all their own static folder. In our case we have these static folders:

```
for the main application: '/administration/static',
for the plugins: '/pluginName/static'
```

This allows each plugin to have a scoped folder which does not get in conflict with other files from other plugins

hiroppy:
I'll change the CI code of lint asap.

hiroppy:
Need to update documentation.

snitin315:
> Need to update documentation.
> https://github.com/webpack/webpack.js.org

I will do it if you agree.

jleifeld:
@snitin315 Of course 👍

evilebottnawi:
@hiroppy Feel free to merge after CI green

hiroppy:
Thanks

PR #1546, opened 2018-10-25
feat(Server): add `/invalidate` endpoint

camertron:
- [ ] This is a **bugfix**
- [x] This is a **feature**
- [ ] This is a **code refactor**
- [ ] This is a **test update**
- [ ] This is a **docs update**
- [ ] This is a **metadata update**

### For Bugs and Features; did you add new tests?

Sadly no. How should I go about testing this? Happy to work something up given some direction.

### Motivation / Use-Case

We're trying to use docker-compose for local development of a giant monolithic rails app that uses webpack for several pages written with react. For performance reasons, we're mounting a volume inside the rails container using NFS, which doesn't support file system events. Consequently, making changes to our react components doesn't trigger an automatic webpack refresh. This PR is an attempt to expose and endpoint that, when hit, will trigger a recompile. The idea is to run a separate container with a regular 'ol non-NFS volume that will listen for file system events and ping the /invalidate endpoint. Actually, while I have you here, I've been trying to figure out an easy way to get the list of files webpack-dev-server is watching, but that appears to be pretty well concealed. Any tips?

### Additional Info

I added this change by creating a branch from master, but, if this change is accepted, I would love it if it could be released as both a 2.x and a 3.x version bump, since we are still stuck on 2.x at the moment.

Issue #1720, opened 2019-03-16
Page not reloading in iframe mode (inline: false)


Loonride:
- Operating System: Windows 7
- Node Version: 8.10.0
- NPM Version: 6.4.1
- webpack Version: 4.29.6
- webpack-dev-server Version: 3.2.1

- [x] This is a **bug**
- [ ] This is a **modification** request

### Expected Behavior
Page should reload in iframe mode on changes.

### Actual Behavior
Page does not reload on changes (console output does not update). 

### For Bugs; How can we reproduce the behavior?
reproducible test repo: https://github.com/Loonride/webpack-dev-server-iframe
Do npm start, go to http://localhost:8080/webpack-dev-server/main.
Change something in src/index.js, the console output will not change and the message "App hot update." does not go away. Upon manual page refresh, the changes appear.

evilebottnawi:
@Loonride I think we should remove `inline` for next major release and remove `inline` client

Loonride:
> @Loonride I think we should remove `inline` for next major release and remove `inline` client

Yes I agree

evilebottnawi:
So i think we can close the issue, it was always buggy and looks nobody uses it

PR #2493, opened 2020-04-01
feat: add invalidate endpoint

EslamHiko:
- [ ] This is a **bugfix**
- [x] This is a **feature**

yes

complete: https://github.com/webpack/webpack-dev-server/pull/1546
### Breaking Changes
no


test/server/utils/routes.test.js
```diff
@@ -64,6 +64,10 @@ describe('routes util', () => {
     });
   });
 
+  it('GET request to invalidate endpoint', (done) => {
+    req.get('/invalidate').expect(200, done);
+  });
```
evilebottnawi, line 69: Will be great to test what we really do invalidate - get stats before do request and do get stats again

EslamHiko: I can't detect any changes in `server.middleware`

EslamHiko: and `server.middleware.context.stats` is always `undefined`

evilebottnawi: Strange, should be no

EslamHiko: I've thought about an Idea maybe we can create an option `invalidateCallback` and we create a function that creates a file then delete after confirming that the file exists.

evilebottnawi: Bad idea, maybe we need fix something in tests, need debug

lib/Server.js
```diff
@@ -118,7 +118,7 @@ class Server {
     this.setupDevMiddleware();
 
     // set express routes
-    routes(this.app, this.middleware, this.options);
```
hiroppy, line 121: Why did you change this?

EslamHiko: @hiroppy To pass the server, so I can call server.invalidate()

hiroppy: I'm -1 on this change because if you pass `this`, we can't understand what variables are used clearly.

EslamHiko: @hiroppy I tried `(this.app, this.middleware, this.options, this.invalidate)` and called `invalidate()` only but this resulted in `500 "Internal Server Error"` so what about changing `this` into `server` for better reading ex: 
```
const server = this;
routes(server);
```

EslamHiko: one thing that worked was moving 
```javascript
invalidate(callback) {
    if (this.middleware) {
      this.middleware.invalidate(callback);
    }
  }
``` 
to `routes.js` and make it only accessable via the api call, but that would require adding an option `invalidateCallBack` so we can pass it ex: 
```javascript
    this.invalidateCallBack = this.options.invalidateCallBack;
    // set express routes
    routes(this.app, this.middleware, this.options, this.invalidateCallBack);
```
and in `routes.js` 
```javascript
function routes(app, middleware, options, invalidateCallBack) {
  app.get('/invalidate', (_req, res) => {
    if(middleware)
      middleware.invalidate(invalidateCallBack)
    res.end();
  });
}
```

evilebottnawi:
@hiroppy I think it is normal to pass `this` in some cases, like webpack pass `compiler` and `compilation` to hooks and utils functions

hiroppy:
I got it but it's hard to read for me like a black box.

evilebottnawi:
@hiroppy I think we can do refactor for utils and code for next branch, we have many bad places that we can improve, maybe using destructuring is good idea in many places

hiroppy:
@evilebottnawi Totally agree with you. I'll improve these codes on the next branch.

evilebottnawi:
@hiroppy So i think we can merge it, we should fix the problem https://github.com/webpack/webpack-dev-server/issues/1720 (we can't remove inline while not fix it + it will good for next) and do release

lib/utils/routes.js
```diff
@@ -30,6 +34,11 @@ function routes(app, middleware, options) {
     createReadStream(join(clientBasePath, 'live.html')).pipe(res);
   });
 
+  app.get('/invalidate', (_req, res) => {
```
sokra, line 37: This should better be prefix with `webpack-dev-server` maybe `/webpack-dev-server/invalidate`. Otherwise `/invalidate` is unusable for the user and apps using `/invalidate` in their router break.

PR #2501, opened 2020-04-03
fix: check if serverData object exist

jeffin143:
- [x] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case
https://github.com/webpack/webpack-dev-server/issues/1479#issuecomment-549831416

This is just a minor patch

### Breaking Changes
None

### Additional Info
 Nothing


hiroppy:
This is an internal function and this argument always exists.

PR #2531, opened 2020-04-22
BREAKING CHANGE: switch default transportMode to ws

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

switching to ws as default

this may require getting rid of iframe to pass tests since I think iframe mode relies on sockjs

Should the warning that `transportMode` is an experimental option stay for the moment, or should we get rid of it?

### Breaking Changes

ws will be default instead of sockjs

Edit: also changed default sockPath from `sockjs-node` to `ws`


evilebottnawi:
@Loonride hm, CI broken

Loonride:
> /cc @Loonride friendly ping

Yes it made sense to get rid of iframe/live mode first, since it relied on sockjs I believe

patrickszmucer:
@Loonride @hiroppy @alexander-akait I've recently upgraded my WDS to 4.x and I stated seeing issues with the websocket randomly dying when the server is running. Switching back to sockjs works, but I'm just curious what was the reason for changing the default to websockets?

snitin315:
>Switching back to sockjs works, but I'm just curious what was the reason for changing the default to websockets?

Are you using any 3rd party plugin like the `error-overlay-webpack-plugin`?

patrickszmucer:
I'm not - I have error overlays disabled and I wrote some custom code that listens to [these events](https://github.com/webpack/webpack-dev-server/blob/master/client-src/utils/sendMessage.js#L8) and changes the color of my banner to show when the server is ready/reloading/stopped. I noticed that the banner now occasionally flickers between stopped/ready because I'm getting "webscket close" events (with a 1006 status).

snitin315:
Please create an issue with reproducible example.

PR #2533, opened 2020-04-22
chore(deps): update chokidar to v3

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

no

### Motivation / Use-Case

This PR is separated from other updated deps PRs because I want to know how chokidar goes.

### Breaking Changes

yes, chokidar doesn't support Node@6.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -43,8 +43,8 @@ jobs:
     strategy:
       maxParallel: 6
       matrix:
-        node-13:
-          node_version: ^13.0.0
+        node-14:
+          node_version: ^14.0.0
           webpack_version: latest
         node-12:
           node_version: ^12.0.0
@@ -54,9 +54,6 @@ jobs:
           webpack_version: latest
         node-8:
           node_version: ^8.9.0
-          webpack_version: latest
-        node-6:
-          node_version: ^6.9.0
           webpack_version: latest
         node-10-canary:
           node_version: ^10.13.0
@@ -102,8 +99,8 @@ jobs:
     strategy:
       maxParallel: 6
       matrix:
-        node-13:
-          node_version: ^13.0.0
+        node-14:
+          node_version: ^14.0.0
           webpack_version: latest
         node-12:
           node_version: ^12.0.0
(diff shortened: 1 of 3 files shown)


paulmillr:
Do you know why there are two chokidar versions in package-lock.json? I see 2.1 and 3.3.

hiroppy:
I think another package depends on Chokidar. I don't have a laptop now but I'll check it.(maybe dev-middleware)

paulmillr:
All good. That was `@babel/cli` and it's devDependency, which means end-users won't install it.

paulmillr:
There is zero need to make the close method async. Don't you know a promise is just an abstraction over a callback function?

```js
  close(cb) {
    this.sockets.forEach((socket) => {
      this.socketServer.close(socket);
    });
    this.sockets = [];
    let prom = Promise.all(this.contentBaseWatchers.map(watcher => watcher.close()));
    this.listeningApp.kill(() => {
      this.middleware.close(() => {
        prom.then(cb);
      });
    });
  }
```

evilebottnawi:
@paulmillr bad idea, tests will be unstable + jest will be warning about unclosed processes + other code can use `close` as method that guarantees close, so multiple listen/close could potentially cause a failing due to memory consumption

evilebottnawi:
@paulmillr makes sense, but we still have to fix it because our code now closes everything synchronously

paulmillr:
That is not true. Your close method specifically expects a callback. It then passes the callback below, to `this.middleware.close()`. If `middleware.close` expects a callback, then obviously your `close` method **is not** syncronous.

evilebottnawi:
@paulmillr 

> That is not true

What is not true ? :confused: 

We close chokidar watchers synchronously here https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L804 in master. What are you talking about?

paulmillr:
Yean, you can just apply my patch above and nothing from those are going to break. The external api for them will remain the same. 

>yep we can potential will break many boilerplates like vue-cli, CRA

evilebottnawi:
@paulmillr I already said that this is a good idea/makes sense, what is the problem?

Issue #2541, opened 2020-04-27
Upgrade "del" dependency


madeken:
https://github.com/webpack/webpack-dev-server/blob/375ab2332706ead66335e8f54391ef37f718e129/package.json#L45

Is a bit out of date, and has a bug in which it ends it ends up dragging in `@types/node@13.9.3` which can be problematic for typescript users.

hiroppy:
This will be fixed by https://github.com/webpack/webpack-dev-server/pull/2542.

evilebottnawi:
Done for `v4`

PR #2542, opened 2020-04-28
chore(deps): upgrade some deps

hiroppy:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

N/A

### Motivation / Use-Case

Update all deps.

Jest has many breaking changes so I modified mocks and some event loop such as setTimeout.

### Breaking Changes

yes

### Additional Info

fixes: https://github.com/webpack/webpack-dev-server/issues/2541


hiroppy:
internal-ip doesn't support Node@8. I'll delete it.

```
    /home/vsts/work/1/s/node_modules/internal-ip/index.js:35
      } catch {}

```

hiroppy:
puppeteer is breaking when using Node@14 so I downgraded from v3 to v2.
FYI: https://github.com/puppeteer/puppeteer/issues/5719

hiroppy:
CI is green.

package.json
```diff
@@ -36,39 +36,40 @@
...
+    "strip-ansi": "^6.0.0",
+    "supports-color": "^7.1.0",
     "url": "^0.11.0",
     "webpack-dev-middleware": "^3.7.2",
-    "webpack-log": "^2.0.0",
-    "ws": "^6.2.1",
-    "yargs": "12.0.5"
+    "webpack-log": "^3.0.1",
+    "ws": "^7.2.5",
+    "yargs": "^15.3.1"
```
evilebottnawi, line 72: Please do no touch yargs, it can potential break CLI

package.json
```diff
@@ -36,39 +36,40 @@
...
     "selfsigned": "^1.10.7",
-    "semver": "^6.3.0",
+    "semver": "^7.3.2",
     "serve-index": "^1.9.1",
     "sockjs": "0.3.20",
     "sockjs-client": "1.4.0",
     "spdy": "^4.0.2",
-    "strip-ansi": "^3.0.1",
-    "supports-color": "^6.1.0",
+    "strip-ansi": "^6.0.0",
```
evilebottnawi, line 66: We can't do it, we will break client on old browsers, we need setup babel on our client and then do update (let's do backport it from previously next)

hiroppy:
@evilebottnawi PTAL

package.json
```diff
@@ -36,39 +36,40 @@
     "release": "standard-version"
   },
   "dependencies": {
+    "@jest/test-sequencer": "^25.4.0",
```
evilebottnawi, line 39: I think we should move it to `devDependencies`

hiroppy:
@evilebottnawi updated

package.json
```diff
@@ -42,33 +42,33 @@
...
     "strip-ansi": "^3.0.1",
-    "supports-color": "^6.1.0",
+    "supports-color": "^7.1.0",
     "url": "^0.11.0",
     "webpack-dev-middleware": "^3.7.2",
-    "webpack-log": "^2.0.0",
-    "ws": "^6.2.1",
-    "yargs": "12.0.5"
+    "webpack-log": "^3.0.1",
+    "ws": "^7.2.5",
```
evilebottnawi, line 70: /cc @Loonride hope we don't breaking anything, can you look?

evilebottnawi:
@hiroppy Why do you use `feature/upgrade-deps` branch for `hot: true`?

evilebottnawi:
No we can't squash this PR

hiroppy:
@evilebottnawi I said you yesterday to merge this branch early as it will be the basis for all our work.

package.json
hiroppy: This module supports only Node.js so not breaking changes.

Loonride: based on ws changelog it should be fine

lib/options.json
```diff
@@ -134,10 +134,14 @@
...
-      "type": "boolean"
-    },
-    "hotOnly": {
-      "type": "boolean"
+      "anyOf": [
+        {
+          "type": "boolean"
+        },
+        {
+          "enum": ["only"]
```
Loonride, line 142: I'm not sure I like this change to the hot option because webpack core CLI also has `--hot` boolean flag (https://webpack.js.org/api/cli/) and I don't think it's great for usage of a flag of the same name to diverge.

evilebottnawi: @Loonride we should implement `hot: only` on webpack-cli side too

evilebottnawi: The main problem is that we already have a lot of options and setting up everything is terribly complicated and redundant, we must simplify everything as much as possible

evilebottnawi: And in some edge cases it is normal what package have same arguments but with many values

hiroppy:
NOTE: Must select `rebase and merge` when merging.

evilebottnawi:
@hiroppy good job, let's remove `inline` and live client

/cc @Loonride I remember we wanted to change the structure of client directory, can you take care, thanks!

hiroppy:
@evilebottnawi This PR commit was put together...😞 why did you select squash and merge...? we can't generate the true changelog..

evilebottnawi:
@hiroppy Oh, my mistake, anyway I want to put more information to CHANGELOG for new major release, don't worry, I will write it before release

Issue #2540, opened 2020-04-27
devServer.index param has no effect


T1mL3arn:
- Operating System: Win 8.1 x64
- Node Version: 10.11.0
- NPM Version: 6.4.1
- webpack Version: 4.43.0
- webpack-dev-server Version: 3.10.3
- Browser: Firefox 75.0

- [x] This is a **bug**
- [ ] This is a **modification** request

### Description

`devServer.index` param does not work as expected. [Documentation says](https://webpack.js.org/configuration/dev-server/#devserverindex) it is:

>The filename that is considered the index file

but the actual file being used is always `index.html`. 

### Code

```js
// webpack.config.js

module.exports = {
    ...
    devServer: {
        index: "foo.html", // <- this does nothing
    },
};
```

Full project to demonstrate the bug is [here](https://github.com/t1ml3arn-bug-reports/webpack-dev-server-index-param-bug). 

### Expected Behavior

Given `foo.html` file is used as an index file

### Actual Behavior

Default `index.html` is used as an index file.

### How can we reproduce the behavior?

1. Clone [demo-project](https://github.com/t1ml3arn-bug-reports/webpack-dev-server-index-param-bug)
2. `npm install`
3. `npm start`
4. open [localhost:3000](http://localhost:3000)
5. see `index.html` instead of `foo.html`

hiroppy:
Thanks. I'll investigate this.

evilebottnawi:
We should improve docs, right setup:
```js
const path = require("path");

console.log(path.resolve("public/foo.html"))

module.exports = {
    entry: "./src/main.js",
    mode: "development",
    output: {
        path: path.resolve(__dirname, "dist/"),
    },
    devServer: {
        contentBase: path.join(__dirname, "public/"),
        staticOptions: { index: 'foo.html' },
        index: path.resolve("public/foo.html"),
        port: 3000
    },
};
```

evilebottnawi:
@hiroppy I think we should move all options for webpack-dev-middleware under `devServer: { devMiddleware }` to avoid misleading, let's discus about it when we will update webpack-dev-middleware

PR #2543, opened 2020-04-28
fix(index): modify index-option

hiroppy:
- [x] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

This option should be enabled.

### Breaking Changes

no

### Additional Info

fixes: https://github.com/webpack/webpack-dev-server/issues/2540


evilebottnawi:
It is not a bug, no need to fix

PR #2546, opened 2020-04-28
fix(hot): enable hot option as default

hiroppy:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Updated.

### Motivation / Use-Case

The hot option is as default from now on.

### Breaking Changes

yes


hiroppy:
My local doesn't fail tests...

lib/Server.js
```diff
@@ -69,6 +69,17 @@ class Server {
...
+    // don't move this position because addEntries called by updateCompiler checks this.options.hot|hotOnly
+    this.hot =
+      typeof this.options.hot === 'boolean' ||
+      typeof this.options.hotOnly === 'boolean'
+        ? this.options.hot || this.options.hotOnly
+        : true;
+
+    if (this.hot) {
+      this.options.hot = true;
+    }
```
evilebottnawi, line 81: Why do not use `this.options.hot` only (avoid using `this.hot`)?

hiroppy: addEntries checks this.options.hot so we need to add true when this.hot is true.  So you want to delete hotOnly or hotOnly should be added true as well?

evilebottnawi: Can we rewrite code on using `this.options.hot` only. What do you think about - `this.options.hot: boolean | "only"`?

hiroppy: ok, understood. so we need to rewrite this line(https://github.com/webpack/webpack-dev-server/blob/89ffb86cd26425c59e3937ca06a2c804a01b8f1d/lib/Server.js#L734). Can we delete hotOnly...?

evilebottnawi: Ideally - avoid using `this.[propertyFromOptions]` in favor `this.options.[property]` in all code, but we can do it in the separate PR

hiroppy: Yes, I'll work on this on Next branch.

lib/Server.js
```diff
@@ -86,7 +97,6 @@ class Server {
     this.contentBaseWatchers = [];
 
     // TODO this.<property> is deprecated (remove them in next major release.) in favor this.options.<property>
```
evilebottnawi, line 99: Remove TODO

lib/Server.js
evilebottnawi: Please remove `this.hot` in thus PR in favor `this.options.hot`

lib/Server.js
hiroppy: Why? this task hasn't been completed.

evilebottnawi: Oh, yes, let's keep it

hiroppy:
I've deleted hotOnly option but CI will fail because e2e/client.js is broking now. I need to investigate it.

client-src/default/index.js
```diff
@@ -15,7 +15,7 @@ const status = {
   currentHash: '',
 };
 const options = {
-  hot: false,
+  hot: true,
```
evilebottnawi, line 18: I think we can change it to false, because it is breaking `hot: false` logic, we run `.hot()` when hot enable, but if we use `hot: false` we doesn't run `.hot()` and default value now is `true`, so it can break logic for `hot: false`

hiroppy: Good catch, I fixed it. The CI's error was because of this.

test/e2e/TransportMode.test.js
```diff
@@ -77,7 +78,16 @@ describe('transportMode client', () => {
                         res.splice(i, 1);
                       }
                     }
+
+                    // TODO: delete this code when getting rid of webpack5
+                    if (isWebpack5) {
+                      const swap = res[0];
+                      res[0] = res[1];
+                      res[1] = swap;
+                    }
```
evilebottnawi, line 87: Why we need it?

hiroppy: It was deleted.

hiroppy:
TransportMode.test.js is really flaky. The order of snapshot often is not fine between v5 and v4.

test/e2e/TransportMode.test.js
```diff
@@ -14,18 +14,21 @@ describe('transportMode client', () => {
     {
       title: 'sockjs',
       options: {
+        hot: false,
```
hiroppy, line 17: This test has been set the value as `false` from before.

hiroppy:
@evilebottnawi CI is green, please review again.

evilebottnawi:
Oh, we merge it into `feature/upgrade-deps` :disappointed:

PR #2550, opened 2020-04-29
BREAKING CHANGE: remove inline and iframe/live mode

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Removing tests no longer in use

### Motivation / Use-Case

Remove inline option, still WIP

### Breaking Changes

Yes, removing inline, iframe/live mode


test/e2e/__snapshots__/ClientOptions.test.js.snap
```diff
@@ -3,6 +3,7 @@
 exports[`Client console.log clientLogLevel is silent 1`] = `
 Array [
   "Hey.",
+  "[HMR] Waiting for update signal from WDS...",
```
Loonride, line 6: This is an issue we should look at separately, clientLogLevel silent is not correctly setting the log level of HMR. We simply were not picking up on it before in this test because it was not inlining HMR before

Loonride:
@evilebottnawi @hiroppy Please review. Lots of files had to change because the inline option was so deeply rooted into everything

PR #2552, opened 2020-04-30
BREAKING CHANGE(server): get rid of conditions of Node version

hiroppy:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Updated.

### Motivation / Use-Case

Drop conditions of Node version.

### Breaking Changes

Yes

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -7,12 +7,10 @@
 */
 const fs = require('fs');
 const path = require('path');
-const tls = require('tls');
 const url = require('url');
 const http = require('http');
 const https = require('https');
 const ip = require('ip');
-const semver = require('semver');
 const killable = require('killable');
 const chokidar = require('chokidar');
 const express = require('express');
@@ -34,15 +32,6 @@ const runBonjour = require('./utils/runBonjour');
 const routes = require('./utils/routes');
 const getSocketServerImplementation = require('./utils/getSocketServerImplementation');
 const schema = require('./options.json');
-
-// Workaround for node ^8.6.0, ^9.0.0
-// DEFAULT_ECDH_CURVE is default to prime256v1 in these version
-// breaking connection when certificate is not signed with prime256v1
-// change it to auto allows OpenSSL to select the curve automatically
-// See https://github.com/nodejs/node/issues/16196 for more information
-if (semver.satisfies(process.version, '8.6.0 - 9')) {
-  tls.DEFAULT_ECDH_CURVE = 'auto';
-}
 
 if (!process.env.WEBPACK_DEV_SERVER) {
   process.env.WEBPACK_DEV_SERVER = true;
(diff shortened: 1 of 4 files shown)


lib/Server.js
```diff
@@ -656,35 +645,14 @@ class Server {
...
-            'HTTP/2 is currently unsupported for Node 10.0.0 and above, but will be supported once Express supports it'
-          );
-        }
-        this.listeningApp = https.createServer(this.options.https, this.app);
-      } else {
-        // The relevant issues are:
-        // https://github.com/spdy-http2/node-spdy/issues/350
-        // https://github.com/webpack/webpack-dev-server/issues/1592
+      if (this.options.http2) {
+        // TODO: we need to replace spdy with http2 which is an internal module
```
hiroppy, line 649: Ideally, we wanna use http2 which is an internal module but we don't have enough time to implement because the interface isn't the same, so we use spdy for all Node versions because spdy supports 10, 12, and 14.

hiroppy:
commitlint doesn't accept BREAKING CHANGE prefix lol

hiroppy:
https://github.com/webpack/webpack-dev-server/pull/2553 must be merged first.

evilebottnawi:
@hiroppy need rebase :+1:

azure-pipelines.yml
```diff
@@ -50,10 +50,10 @@ jobs:
           node_version: ^12.0.0
           webpack_version: latest
         node-10:
-          node_version: ^10.13.0
+          node_version: ^10.20.1
           webpack_version: latest
         node-10-canary:
-          node_version: ^10.13.0
+          node_version: ^10.20.1
```
evilebottnawi, line 56: Why we update version?

hiroppy: This is the latest version but you want to lock the v10.13.0?  Anyway, http2 is supported over v10.10.0 so I'm fine either way.

https://nodejs.org/api/http2.html

evilebottnawi: @hiroppy azure and github actions by default install latest version if you use `^10.13`

package.json
```diff
@@ -11,7 +11,7 @@
     "client"
   ],
   "engines": {
-    "node": ">= 6.11.5"
+    "node": ">= 10.20.1"
```
evilebottnawi, line 14: Return this to `10.13`

azure-pipelines.yml
hiroppy: yes I know but I don't know why you are sticking to v10.13.0.

package.json
hiroppy: I think 10.10 is better than 10.13 because this version supports http2 as a stable.

azure-pipelines.yml
evilebottnawi: @hiroppy by default we support minimum LTS

package.json
evilebottnawi: 10.13 is LTS and we support LTS

evilebottnawi: Also webpack@5 support 10.13 as minimum, so we should support this range too

azure-pipelines.yml
hiroppy: No, LTS is just v10, not v10.13.

package.json
hiroppy: > webpack@5 support 10.13 as minimum

ok, your answer is good for me. but 10.13 isn't minimum LTS.

evilebottnawi: @hiroppy 

10.12.0 - is not LTS https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md#2018-10-10-version-10120-current-targos

10.13.0 - is LTS https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md#2018-11-27-version-10140-dubnium-lts-rvagg

So we should support minimum supported LTS and now it is 10.13

hiroppy: thank you for sharing this info.

hiroppy:
I've changed the commit message from BREAKING CHANGE to fix because CI is going to fail when using BREAKING CHANGE.

PR #2563, opened 2020-05-02
migrate to built-in logger

hiroppy:
**DON'T MERGE but request maintainer's comments**

- [ ] This is a **bugfix**
- [x] This is a **feature**

Yep

### Motivation / Use-Case

We need to drop webpack-log because this package isn't under our management.
see https://www.npmjs.com/package/webpack-log

And webpack@4 supports `infrastructureLogging` system so I decided to migrate to it.  See https://webpack.js.org/configuration/other-options/#infrastructurelogging.

#### Output

```console
<i> [webpack-dev-server] Project is running at http://localhost:8080/
<i> [webpack-dev-server] webpack output is served from /
<i> [webpack-dev-server] Content not from webpack is served from /Users/hiroppy/webpack/webpack-dev-server
```

#### Dropped keys (CLI as well)

- logLevel
- quiet
- info

#### Modified keys
- clientLogLevel
  - deleted `debug`, `trace`, `silent`, and `warning` instead of `none`, `warn` and `verbose`

### Breaking Changes

Yep

### Additional Info
Currently, I don't know why but the CLI test of multi ports fails so I'm investigating it.


test/cli/cli.test.js
```diff
@@ -199,75 +197,9 @@ describe('CLI', () => {
...
-      cp: {
-        port: null,
-        done: false,
-      },
-      cp2: {
-        port: null,
-        done: false,
-      },
-    };
+  it('should use different random port when multiple instances are started on different processes', async () => {
```
hiroppy, line 200: This test fails so I need to fix it.

hiroppy: Oh, this test in the master branch is broken as well...

> cp1:null cp2: 8080

Yes this test passes because null !== 8080 but we don't expect it.

hiroppy: I've fixed this test.

bin/options.js
```diff
@@ -74,17 +74,12 @@ const options = {
...
-    group: DISPLAY_GROUP,
-    describe: 'Quiet',
-  },
   'client-log-level': {
     type: 'string',
     group: DISPLAY_GROUP,
     default: 'info',
     describe:
-      'Log level in the browser (trace, debug, info, warn, error or silent)',
+      'Log level in the browser (none, error, warn, info, log, verbose)',
```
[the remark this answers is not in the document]
hiroppy: hm, in fact, client-log-level should respect infrastructurelogging.level? What do you think?

lib/Server.js
```diff
@@ -38,13 +38,11 @@ if (!process.env.WEBPACK_DEV_SERVER) {
 }
 
 class Server {
-  constructor(compiler, options = {}, _log) {
+  constructor(compiler, options = {}, log) {
```
hiroppy, line 41: I remain `log` for users who want to inject a custom logger.

PR #2570, opened 2020-05-05
BREAKING CHANGE(server): remove setup and rename before and after

hiroppy:
- [ ] This is a **bugfix**
- [x] This is a **feature**

yes

### Motivation / Use-Case

Remove `setup` option and Rename `before` and `after`.

- setup -> N/A
- before -> onBeforeSetupMiddleware
- after -> onAfterSetupMiddleware

### Breaking Changes

yes


hiroppy:
I think it's hard to understand for users. Users want to rewrite only before or after so we need to explicit **before** and **after** as options name.

evilebottnawi:
@hiroppy :+1: I think we can think about it even better in the next version.

PR #2586, opened 2020-05-08
Refactor avoid this

evilebottnawi:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No need

### Motivation / Use-Case

Resolve todo

### Breaking Changes

Yes, if you use property from server

### Additional Info

Most likely we can do even more refactoring, but I suggest doing it in the next PRs to avoid regressions and problems with review

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -52,14 +52,7 @@ class Server {
       );
     }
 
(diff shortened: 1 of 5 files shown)


evilebottnawi:
Done for v4

PR #2601, opened 2020-05-13
fix: disable socket injection when hot & liveReload are disabled (V4)

EslamHiko:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case
fix : #1831
Alternative to : https://github.com/webpack/webpack-dev-server/pull/2133

### Breaking Changes
The socket server will be disabled when `hot` and `liveReload` are disabled.

### Additional Info
I updated the values of `liveReload: false` tests since `hot` became enabled by default.
I added tests for `ws` and moved `socket-injection.test.js` to `e2e` folder since e2e testing is the only way to test `ws` as it became the default websocket.
Also I've updated `console.log` test in `test/e2e/ClientOptions.test.js` to check for all WebSocket options.

Not sure if logging tests in `test/e2e/ClientOptions.test.js` are enough so I don't know whether I should delete `test/e2e/socket-injection.test.js` or not.

### DIFF
--- a/client-src/default/index.js
+++ b/client-src/default/index.js
@@ -136,7 +136,11 @@ const onSocketMessage = {
     log.error(error);
   },
   close() {
-    log.error('[WDS] Disconnected!');
+    if (options.hot || options.liveReload) {
+      log.error('[WDS] Disconnected!');
+    } else {
+      log.error('[WDS] Hot Module Replacement & Live Reloading are disabled!');
+    }
     sendMessage('Close');
   },
 };
--- a/client-src/default/index.js
+++ b/client-src/default/index.js
@@ -136,11 +136,8 @@ const onSocketMessage = {
     log.error(error);
   },
   close() {
-    if (options.hot || options.liveReload) {
-      log.error('[WDS] Disconnected!');
-    } else {
-      log.error('[WDS] Hot Module Replacement & Live Reloading are disabled!');
-    }
+    log.error('[WDS] Disconnected!');
+
     sendMessage('Close');
   },
 };
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -678,7 +678,9 @@ class Server {
     this.hostname = hostname;
 
     return this.listeningApp.listen(port, hostname, (err) => {
-      this.createSocketServer();
+      if (this.options.hot || this.options.liveReload !== false) {
+        this.createSocketServer();
+      }
 
       if (this.options.bonjour) {
         runBonjour(this.options);
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -678,7 +678,7 @@ class Server {
     this.hostname = hostname;
 
     return this.listeningApp.listen(port, hostname, (err) => {
-      if (this.options.hot || this.options.liveReload !== false) {
+      if (this.options.hot || this.options.liveReload) {
         this.createSocketServer();
       }
 
@@ -925,7 +925,7 @@ class Server {
 
     const watcher = chokidar.watch(watchPath, watchOptions);
     // disabling refreshing on changing the content
-    if (this.options.liveReload !== false) {
+    if (this.options.liveReload) {
       watcher.on('change', () => {
         this.sockWrite(this.sockets, 'content-changed');
       });
--- a/test/e2e/ClientOptions.test.js
+++ b/test/e2e/ClientOptions.test.js
@@ -412,53 +412,28 @@ describe('Client console.log', () => {
     },
   ];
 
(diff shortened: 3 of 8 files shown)


client-src/default/index.js
```diff
@@ -136,7 +136,11 @@ const onSocketMessage = {
     log.error(error);
   },
   close() {
-    log.error('[WDS] Disconnected!');
+    if (options.hot || options.liveReload) {
+      log.error('[WDS] Disconnected!');
+    } else {
+      log.error('[WDS] Hot Module Replacement & Live Reloading are disabled!');
+    }
```
evilebottnawi, line 143: I think it is unnecessary here to write `log.error('[WDS] Hot Module Replacement & Live Reloading are disabled!');`

/cc @Loonride @hiroppy

lib/Server.js
```diff
@@ -659,7 +659,9 @@ class Server {
     this.hostname = hostname;
 
     return this.listeningApp.listen(port, hostname, (err) => {
-      this.createSocketServer();
+      if (this.options.hot || this.options.liveReload !== false) {
```
evilebottnawi, line 662: Put `this.options.liveReload` in `normalizeOptions` with code:
```js
options.liveReload = typeof options.liveReload !== "undefined" ? options.liveReload : true;
```

client-src/default/index.js
hiroppy: Agree

test/e2e/ClientOptions.test.js
```diff
@@ -406,48 +419,54 @@ describe('Client console.log', () => {
...
+    cases.forEach(({ title, options }) => {
+      title += ` (${
+        Object.keys(mode).length ? mode.transportMode : 'default'
+      })`;
+      options = { ...mode, ...options };
+      it(title, (done) => {
+        const res = [];
+        const testOptions = Object.assign({}, baseOptions, options);
+
+        // TODO: use async/await when Node.js v6 support is dropped
```
hiroppy, line 432: You can use async/await already.

EslamHiko: @hiroppy I tried many times to use async/await but every time I get errors or wrong results, my last attempt was : 
```js
cases.forEach(async ({ title, options }) => {
      title += ` (${
        Object.keys(mode).length ? mode.transportMode : 'default'
      })`;
      options = { ...mode, ...options };
      testOptions = Object.assign({}, baseOptions, options);
      await testServer.startAwaitingCompilation(config, testOptions); // and without await
      it(title, (done) => {

        const res = [];
          runBrowser().then(({ page, browser }) => {
            page.goto(`http://localhost:${port2}/main`);
            page.on('console', ({ _text }) => {
              res.push(_text);
            });

            // wait for load before closing the browser
            page.waitForNavigation({ waitUntil: 'load' }).then(() => {
              page.waitFor(beforeBrowserCloseDelay).then(() => {
                browser.close().then(() => {
                  // Order doesn't matter, maybe we should improve that in future
                  expect(res.sort()).toMatchSnapshot();
                  testServer.close(done)
                });
              });
            });
          });
      });
    });
```
for now It works fine, I need help with rewriting this with async/await. 
@Loonride any thoughts for this ?

Loonride:
Looks like `test/client/index.test.js` snapshot needs updating

test/e2e/ClientOptions.test.js
Loonride: I believe the issue is forEach won't work with async, at the very top where you do `cases.forEach`. Try doing a normal for loop and see if that fixes it

EslamHiko: @Loonride thanks for the hint, what fixed the issue was adding `await` before `it` & adding the server starter inside `it`, not sure if converting the other promises helped ex : 
```
await it(title, async (done) => {
        await testServer.startAwaitingCompilation(config, testOptions);
        const res = [];
        const { page, browser } = await runBrowser();
        page.goto(`http://localhost:${port2}/main`);
        page.on('console', ({ _text }) => {
          res.push(_text);
        });
        // wait for load before closing the browser
        await page.waitForNavigation({ waitUntil: 'load' });
        await page.waitFor(beforeBrowserCloseDelay);
        await browser.close();
        // Order doesn't matter, maybe we should improve that in future
        await expect(res.sort()).toMatchSnapshot();
        await testServer.close(done);
      });
```
It's strange that using `for` for the inner loop didn't work & skipped many tests.

evilebottnawi:
@EslamHiko Please fix conflicts, sorry for the situation :smile:

PR #2617, opened 2020-05-19
Fix prefer hot from cli

evilebottnawi:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

Resolve TODO

### Breaking Changes

Yes

### Additional Info

No

### DIFF
--- a/lib/utils/createConfig.js
+++ b/lib/utils/createConfig.js
@@ -73,10 +73,11 @@ function createConfig(config, argv, { port }) {
     process.stdin.resume();
   }
 
-  // TODO https://github.com/webpack/webpack-dev-server/issues/616 (v4)
-  // We should prefer CLI arg under config, now we always prefer `hot` from `devServer`
-  if (!options.hot) {
-    options.hot = argv.hot;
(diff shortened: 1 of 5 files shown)


lib/utils/createConfig.js
```diff
@@ -73,10 +73,11 @@ function createConfig(config, argv, { port }) {
...
 
-  // TODO https://github.com/webpack/webpack-dev-server/issues/616 (v4)
-  // We should prefer CLI arg under config, now we always prefer `hot` from `devServer`
-  if (!options.hot) {
-    options.hot = argv.hot;
+  const hasHot = typeof argv.hot !== 'undefined';
+  const hasHotOnly = typeof argv.hotOnly !== 'undefined';
+
+  if (hasHot || hasHotOnly) {
+    options.hot = hasHotOnly ? 'only' : argv.hot;
```
evilebottnawi, line 80: argv-parser doesn't support multiple types for argument so for compatibility we have `--hot` and `--hot-only`, anyway we should deprecate our CLI in v4 in favor `webpack server` (`webpack-cli`), so we can do it safely

PR #2631, opened 2020-05-28
refactor(client): migrate client to built-in logger

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

Migrating to built-in logger in client

Followed slightly different approach for the actual log util than https://github.com/webpack/webpack-dev-server/pull/2563

### Breaking Changes

Yes, new client log levels

@hiroppy Do we need to worry about this for webpack@5 https://github.com/webpack/webpack-dev-server/pull/2563/commits/b6af4e96c59d8c3cecae534c2b9f7fb8ade8c4e9#diff-bd664d218d585ff7bd6ceced46bd877aR3


client-src/default/index.js
```diff
@@ -1,5 +1,11 @@
 'use strict';
 
+// webpack@5 doesn't inject node polyfill automatically
+window.global = window.global || window;
```
Loonride, line 4: this was needed to make the other polyfills work. Is this way to handle this okay, or is it a problem that this is ignoring webpack's node option https://webpack.js.org/configuration/node/ ?

Any suggestions?

lib/options.json
```diff
@@ -11,16 +11,7 @@
...
-        "info",
-        "warn",
-        "error",
-        "debug",
-        "trace",
-        "silent",
-        "none",
-        "warning"
-      ]
+      "enum": ["none", "error", "warn", "info", "log", "verbose"]
```
evilebottnawi, line 14: Let's move it to `clientOptions`, i.e.:
```
clientOptions: { 
  logging: {
    level: 'info',
    // Let's avoid implement debug in this PR, just for future, schema should not contain debug too
    debug: [
      'MyPlugin',
      /MyPlugin/,
      (name) => name.contains('MyPlugin')
    ]
  }
}
```

Like here https://webpack.js.org/configuration/other-options/#infrastructurelogging

client-src/default/index.js
```diff
@@ -55,17 +61,20 @@ const onSocketMessage = {
...
   },
   'still-ok': function stillOk() {
-    log.info('[WDS] Nothing changed.');
+    log.info('Nothing changed.');
     if (options.useWarningOverlay || options.useErrorOverlay) {
       overlay.clear();
     }
     sendMessage('StillOk');
   },
   'log-level': function logLevel(level) {
```
evilebottnawi, line 70: Let't rename it to `logging`

client-src/default/index.js
evilebottnawi: Hm, why we need `process` and `util`?

Loonride: > Hm, why we need process and util?

@evilebottnawi webpack5 tests were not passing, this made them work. Not sure if there is a better solution. @hiroppy Originally put that in here:
https://github.com/webpack/webpack-dev-server/commit/b6af4e96c59d8c3cecae534c2b9f7fb8ade8c4e9#diff-bd664d218d585ff7bd6ceced46bd877aR3

lib/options.json
Loonride: @evilebottnawi Should we rename the `client-log-level` option then? Maybe `client-logging`? I think just `logging` might be confusing.

evilebottnawi: @Loonride `webpack` has `infrastructureLogging`, so better to keep this same - https://webpack.js.org/configuration/other-options/

evilebottnawi: I think we should move all client options to `clientOptions`, maybe we should rename `clientOptions` to `client`, less characters in the name, what do you think?

client-src/default/index.js
evilebottnawi: We can use `alias` here when we will switch on `webpack@5`, we really need full `process/browser` and `util/util` packages? Maybe we can copy post some function to avoid using full packages?

lib/options.json
Loonride: Yes, I moved `logging` into `clientOptions`. I agree that it should be renamed to `client`, but that is an easy fix, I will do it later

client-src/default/index.js
Loonride: do you mean resolve.alias? https://webpack.js.org/configuration/resolve/#resolvealias

client-src/default/index.js
```diff
@@ -1,5 +1,10 @@
 'use strict';
 
+// webpack@5 doesn't inject node polyfill automatically
+window.global = window.global || window;
+window.process = window.process || {};
+window.process.env = window.process.env || {};
```
Loonride, line 6: @evilebottnawi I substituted `process/browser` out for this, I think it is all that is needed. I think we still need the `util/util` package as a dependency, since something in tapable needs it.

Loonride:
Proper handling of babel for the client should be merged before this: https://github.com/webpack/webpack-dev-server/pull/2652

Issue #2649, opened 2020-06-18
strip-ansi compatibility problem and general compatibility for old browsers


Loonride:
- [x] This is a **bug**
- [ ] This is a **modification** request

The issue is with the latest version of `strip-ansi` (6.0.0). It uses syntax that is not compatible with the oldest browsers, like `const`, which is reflected in the user's bundle (`localhost:8080/main.js`):

![strip-ansi-problem](https://user-images.githubusercontent.com/22580625/84986438-2e1f2500-b104-11ea-9d11-b52ab15225bd.png)

Latest version of webpack-dev-server is fine, because it uses `strip-ansi@3.0.1`.

To reproduce, simply use any config like https://github.com/Loonride/babel-repro , then navigate to `localhost:8080/main.js` after starting. You will not see `const` being used in the bundle with latest released webpack-dev-server version, but you will see it if you npm link the `v4` webpack-dev-server branch.

Usage of `strip-ansi` in the client is here: https://github.com/webpack/webpack-dev-server/blob/4ab1f21bc85cc1695255c739160ad00dc14375f1/client-src/default/index.js#L5

This is an issue we need to prevent, correct? If so, we generally need a better strategy so that we don't have to trust dependencies like `strip-ansi` to be compatible with old browsers.

### For Bugs; How can we reproduce the behavior?

Reproducing is explained above

alexander-akait:
Let's close, fixed, but we still uses not bundled version in some cases, but we have issue for this

PR #2652, opened 2020-06-18
fix(client): run babel on whole client dependency tree via webpack bundle

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Added bundle check test

### Motivation / Use-Case

We want the entire injected client code to be safe for older browsers: https://github.com/webpack/webpack-dev-server/issues/2649

The changes in here are mainly to create a client bundle that acts as a module, then pass the resourceQuery along to it so that everything behaves as it did before

### Breaking Changes

Yes, many breaking changes in client with the whole API, what do you think?

### DIFF
--- a/client-src/default/utils/reloadApp.js
+++ b/client-src/default/utils/reloadApp.js
@@ -1,6 +1,9 @@
 'use strict';
 
 /* global WorkerGlobalScope self */
+/* eslint-disable
+  camelcase
+*/
 
 const { log } = require('./log');
 
@@ -13,7 +16,12 @@ function reloadApp(
   }
   if (hot) {
     log.info('App hot update...');
-    const hotEmitter = require('webpack/hot/emitter');
+    const hotEmitter =
+      typeof __webpack_hot_emitter__ !== 'undefined'
+        ? // eslint-disable-next-line no-undef
+          __webpack_hot_emitter__
+        : // eslint-disable-next-line import/no-unresolved
+          require('webpack/hot/emitter');
     hotEmitter.emit('webpackHotUpdate', currentHash);
     if (typeof self !== 'undefined' && self.window) {
       // broadcast update to window
--- a/client-src/default/utils/reloadApp.js
+++ b/client-src/default/utils/reloadApp.js
@@ -1,27 +1,20 @@
 'use strict';
 
 /* global WorkerGlobalScope self */
-/* eslint-disable
-  camelcase
-*/
 
 const { log } = require('./log');
 
 function reloadApp(
   { hotReload, hot, liveReload },
-  { isUnloading, currentHash }
+  { isUnloading, currentHash },
+  hotEmitter
 ) {
   if (isUnloading || !hotReload) {
     return;
   }
   if (hot) {
     log.info('App hot update...');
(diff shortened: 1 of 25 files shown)


client-src/default/utils/reloadApp.js
```diff
@@ -13,7 +16,12 @@ function reloadApp(
   }
   if (hot) {
     log.info('[WDS] App hot update...');
-    const hotEmitter = require('webpack/hot/emitter');
+    const hotEmitter =
+      typeof __webpack_hot_emitter__ !== 'undefined'
+        ? // eslint-disable-next-line no-undef
+          __webpack_hot_emitter__
```
Loonride, line 22: I had to change slightly how the hot emitter is imported, because it is important that the right hot emitter module is used for hot reloading to work.

evilebottnawi:
> Yes, many breaking changes in client with the whole API, what do you think?

Can you clarify?

test/client/entry/entry.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+    afterAll(testServer.close);
+
+    it('should not include const', async () => {
+      const { text } = await req
+        .get('/main.js')
+        .expect('Content-Type', 'application/javascript; charset=UTF-8')
+        .expect(200);
+
+      expect(text).not.toContain('const ');
```
evilebottnawi, line 38: I think we need improve test here :smile: checking on `const` is not mean we have a problems :smile:

evilebottnawi: I think we can install `acorn` in dev deps and try to parse it

Loonride:
> > Yes, many breaking changes in client with the whole API, what do you think?
> 
> Can you clarify?

- the location where bundle is exported is moved from `client/default/index.bundle.js` to `client/entry/bundle.js` (we could change the name/location if needed)
- The `client-src/index.js` functionality changes. Everything there is now wrapped in a function that must be called, passing in the resourceQuery, so nothing will run unless the exported function is called
- In `client-src/entry`, a file is there called `bundle.js` that acts as the equivalent of the actual output bundle by simply exporting `client-src/default/index.js`
- The actual entry that is injected is `client/entry/index.js`, which passes `__resourceQuery` into the bundle

Loonride:
One other not great thing for debugging purposes is that all we will see for webpack-dev-server debugging is `bundle.js`. Not sure how to get the line numbers to work since it is a bundle in a bundle:

wds-screen-1

Previously it would be like this:

wds-screen-2

test/client/entry/entry.test.js
Loonride: Added an acorn test to check for ES5

Loonride:
I've realized that webpack@5 will use ES6 syntax no matter what, unless the user uses babel-loader. I don't think there's a workaround to that.

For that reason, I think this solution that I've created is overkill, and has too many breaking changes for the small fix that it does. I think it will be better to instead just make minor, case-specific fixes and hang on until we migrate to webpack@5.

Loonride:
Closing in favor of https://github.com/webpack/webpack-dev-server/pull/2658

Issue #1286, opened 2018-01-19
Add support for not having to pin strip-ansi


TheLarkInn:
Currently we've had to pin @sindresorhus's strip-ansi because of the use of arrow functions. We should essentially make whatever changes needed to allow us to use that code (transpilation etc.) so that it still works for strip-ansi@4+ but without breaking IE 11 compat. 

See https://github.com/webpack/webpack-dev-server/pull/1273

evilebottnawi:
@TheLarkInn need patch release after merge, otherwise `babel` would be a more rational solution

lydell:
See #1279 for a potential solution of running node_modules through Babel.

TheLarkInn:
I have an issue to add this but for now since it's breaking current
implementations I'd like to see pin and patched.

wayofthefuture:
Broke the entire build

kaiyoma:
This is causing issues with our project too. If I try to upgrade to the latest version, I start getting yarn errors:

```
error "nyc#string-width#strip-ansi" is wrong version: expected "4.0.0", got "3.0.1"
warning "nyc#string-width#strip-ansi#ansi-regex@^2.0.0" could be deduped from "2.1.1" to "ansi-regex@2.1.1"
error "nyc#strip-ansi#ansi-regex" is wrong version: expected "3.0.0", got "2.1.1"
error Found 2 errors.
```

We're currently on version 2.10.0 and would like to upgrade as the dev server is broken in IE11 in this version.

robvree:
This is also breaking our dev build in IE11. Running node_modules through babel is not a solution. Everything in node_modules should already be transpiled.

robvree:
Update: Updating from 2.11.0 to 2.11.1 (which uses v3 of strip-ansi) and moving my polyfill for es6-object-assign above the first import in my entry point (not sure if this is related) has fixed my IE11 build.

```
// polyFills for IE11 that need to run before the imports
require("es6-object-assign").polyfill();

// react-hot-loader must come before importing React
import "react-hot-loader";
import * as React from "react";
import { render } from "react-dom";
import App from "./staffhub/sh-application/components/App";
import DateUtils from "sh-application/utility/DateUtils";
import StringsStore from "sh-strings/store";
import { loadTheme } from "@uifabric/styling";
import { preinitApplication } from "sh-application";
import { initializeIcons } from "@uifabric/icons";

// polyFills for IE11
require("es6-promise").polyfill();
[...]

PR #2658, opened 2020-06-23
Fix: ES5 client

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

We need to make the client work with ES5 for webpack@4. This PR does some case-specific patches, rather than trying to fully solve the problem, since webpack@5 always has ES6 syntax in its bundles, unless the user uses babel-loader.

The `webpack/lib/logging/runtime` and `strip-ansi` modules use ES6, so they are placed in the `transpiled-modules` directory and have files that act as a proxy entry to them. Then, these entry files are turned into a bundle with babel-loader, so that the result is a module that works with ES5 and behaves the same way.

Explanation for the ES5 check test: Since webpack takes all the JavaScript and puts it into a string in an `eval` function, a bundle that will not work with ES5 can still pass a basic ES5 check, since all this bad ES6 syntax is inside of a string. The best course of action for a test I think is to locate these strings in a bundle that `eval` is being run on, then check that they parse with ES5.

https://github.com/webpack/webpack-dev-server/issues/1286 Actually was not fixed on v4 for the reason above, but should be fixed by this PR.

If  you look at the test run below only after the first 2 commits, you can see that the test does not pass because there are still modules being used that don't work with ES5: https://github.com/webpack/webpack-dev-server/pull/2658/files/d01ac6855c0bc20f23054c9ca8832eea2539588d

### Breaking Changes

None, client API is the same

### DIFF
--- a/client-src/transpiled-modules/webpack.config.js
+++ b/client-src/transpiled-modules/webpack.config.js
@@ -1,0 +1,39 @@
+'use strict';
+
+const path = require('path');
+const merge = require('webpack-merge');
+
+const base = {
+  mode: 'production',
+  output: {
+    path: path.resolve(__dirname, '../../client/transpiled-modules'),
+    libraryTarget: 'commonjs2',
+  },
+  module: {
+    rules: [
+      {
+        test: /\.js$/,
+        use: [
+          {
+            loader: 'babel-loader',
+          },
+        ],
+      },
+    ],
+  },
+};
+
+module.exports = [
+  merge(base, {
+    entry: path.join(__dirname, 'log.js'),
+    output: {
+      filename: 'log.js',
+    },
+  }),
+  merge(base, {
+    entry: path.join(__dirname, 'strip-ansi.js'),
+    output: {
+      filename: 'strip-ansi.js',
+    },
+  }),
+];
--- a/client-src/default/index.js
+++ b/client-src/default/index.js
@@ -6,7 +6,7 @@ window.process = window.process || {};
 window.process.env = window.process.env || {};
 
 /* global __resourceQuery WorkerGlobalScope self */
-const stripAnsi = require('strip-ansi');
+const stripAnsi = require('../transpiled-modules/strip-ansi');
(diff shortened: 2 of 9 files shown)


Loonride:
Looks like test instability again on Windows.

client-src/transpiled-modules/webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+      filename: 'log.js',
+    },
+  }),
+  merge(base, {
+    entry: path.join(__dirname, 'strip-ansi.js'),
+    output: {
+      filename: 'strip-ansi.js',
+    },
+  }),
+];
```
evilebottnawi, line 39: I am afraid, some modules can use new syntax (ES 6) and we can't catch it, for example in patch version

Loonride: @evilebottnawi If that happens when we are updating a dependency, it will be caught in CI here: https://github.com/webpack/webpack-dev-server/pull/2658/files#diff-24488ac612ba9666465a7a596040ac73R43

Loonride: But if we want full safety, we should do this: https://github.com/webpack/webpack-dev-server/pull/2652. I don't like how that PR ended up because it has so many breaking changes in the API and seems a little risky

Loonride:
Ah, issue with lodash is making lint fail: https://www.npmjs.com/advisories/1523

hiroppy:
@Loonride could you rebase this? This PR will be merged after merging https://github.com/webpack/webpack-dev-server/pull/2675

hiroppy (approved):

client-src/transpiled-modules/webpack.config.js
hiroppy: Sound good to me. CI is enough for me.

rishabh3112 (approved):

evilebottnawi (approved):

PR #2659, opened 2020-06-24
[WIP] Remove CLI default flag values for webpack-cli serve

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Not yet

### Motivation / Use-Case

We do not want default values for webpack-cli serve flags, because this makes it much easier to merge CLI flags with the user's devServer options in their webpack config

### Breaking Changes

Some changes in `normalizeOptions`, so instead of certain options being undefined, they will be set


Loonride:
I just realized that the default behavior I added in `normalizeOptions` was already added recently: https://github.com/webpack/webpack-dev-server/blob/eb9c58d32a2cf533d2195291bb37604e06acfaea/lib/utils/normalizeOptions.js#L5

evilebottnawi:
@Loonride Looks something broken on CI

Loonride:
I thought some slight refactoring on the current CLI flags (`options.js`) would be good to do here as well, since they are not necessary. There should be no breaking changes related to this besides slight util API changes like with `createConfig`. The 3 options for which I removed default CLI values are:

- `host` - the default of `localhost` is now set in `createConfig`
- `client-logging` - the default of `info` was already being set on the client side here: /home/kir/dev/node/webpack-dev-server/test/server/utils/createConfig.test.js
- `liveReload` - the default of `true` was already being set in `normalizeOptions`, so this default is already applied for both CLI and API

PR #2660, opened 2020-06-25
chore: upgrade dev middleware

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Not yet

### Motivation / Use-Case

webpack-dev-middleware needs to be updated

Take a look at the `options.json` file and tell me if these changes look good for the options.

### Breaking Changes

Many options being removed in favor of `devMiddleware` option.

Change where `publicPath` is handled, which is minor breaking change for API


lib/options.json
```diff
@@ -75,14 +75,11 @@
         }
       ]
     },
-    "disableHostCheck": {
-      "type": "boolean"
-    },
-    "fs": {
+    "devMiddleware": {
```
evilebottnawi, line 78: It is good place to do discussion, should we move all webpack-dev-middleware in the one place

/cc @hiroppy

Loonride: I think it makes sense because webpack-dev-middleware now does options validation that only allows for the properties specified. So we can just validate `devMiddleware` as an object, and if the user passes in bad webpack-dev-middleware options then webpack-dev-middleware validation will throw an error.

test/server/Server.test.js
```diff
@@ -140,12 +139,12 @@ describe('Server', () => {
...
   describe('Invalidate Callback', () => {
     describe('Testing callback functions on calling invalidate without callback', () => {
-      it('should be `noop` (the default callback function)', (done) => {
+      it('should use default `noop` callback', (done) => {
         const compiler = webpack(config);
         const server = new Server(compiler, baseDevConfig);
 
         server.invalidate();
-        expect(server.middleware.context.callbacks[0]).toBe(noop);
+        expect(server.middleware.context.callbacks.length).toEqual(1);
```
Loonride, line 147: webpack-dev-middleware no longer exports the noop function. Should we make it get exported from webpack-dev-middleware, or do you think this is sufficient for this test?

lib/options.json
hiroppy: I prefer this idea.

test/server/Server.test.js
hiroppy: I think it's ok.

evilebottnawi:
We have broken tests, maybe it is regressions, we need collect them here

lib/options.json
evilebottnawi: Let's rename it to `dev`, I think it is unnecessary to have `middleware` suffix, we don't have it for other middlewares

evilebottnawi: Also, it will be better for CLI - `webpack server --dev-something=value`

Loonride:
@evilebottnawi Yes, lots of broken tests because of the options changes that I still need to fix.

lib/utils/normalizeOptions.js
```diff
@@ -49,6 +55,22 @@ function normalizeOptions(compiler, options) {
   options.client.path = `/${
     options.client.path ? options.client.path.replace(/^\/|\/$/g, '') : 'ws'
   }`;
+
+  options.devMiddleware = options.devMiddleware || {};
+
+  // this was previously done in createConfig, but it should be done
+  // here for CLI/API uniformity
+  if (!options.devMiddleware.publicPath) {
```
Loonride, line 63: I thought this was a good time to move `publicPath` handling from `createConfig` into `normalizeOptions` so it is done both on the CLI and API. Let me know if you disagree, as it is a minor breaking change for API.

evilebottnawi:
@Loonride Agree, let's move options to the `dev` object

lib/utils/normalizeOptions.js
evilebottnawi: Yes, let's think about it in other PRs

evilebottnawi:
I think we should remove `stats` option from `webpack-dev-middleware` because new version take value from configuration, so you don't need it anymore, if you want to change it you can use `stats: WEBPACK_DEV_SERVER ? 'warnings-errors' : 'verbose'`

Loonride:
All tests should be passing once https://github.com/webpack/webpack-dev-middleware/pull/670 is released, then I update the webpack-dev-middleware dependency in this PR.

bin/options.js
```diff
@@ -62,17 +62,6 @@ const options = {
     describe: 'Open default browser with the specified page',
     requiresArg: true,
   },
-  color: {
```
Loonride, line 65: I removed the `color` CLI flag since it ends up setting `colors: true` in the stats option, which was also removed.

New behavior:
- default for `colors` is `false` (should it be `true`?)
- if a user wants to turn on color, they need to put in their webpack config: `stats: { colors: true }`.

Do you think we should do this approach, or should we keep the color CLI flag and then update the compiler stats option using this flag?

evilebottnawi: I think we don't need this option, default value is `stats.colors` (https://github.com/webpack/webpack/blob/master/lib/stats/DefaultStatsPresetPlugin.js#L168), `webpack-cli` should have `--stats-color` (or `--color`) flag to enable or disable it, it is out of scope webpack-dev-server

bin/options.js
```diff
@@ -62,17 +62,6 @@ const options = {
...
-    type: 'boolean',
-    alias: 'colors',
-    default: function supportsColor() {
-      // Use `require('supports-color').stdout` for supports-color >= 5.0.0.
-      // See https://github.com/webpack/webpack-dev-server/pull/1555.
-      return require('supports-color').stdout;
```
evilebottnawi, line 71: Do not forget to remove `supports-color`

Loonride: Done here: https://github.com/webpack/webpack-dev-server/pull/2660/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L62

lib/utils/createConfig.js
```diff
@@ -47,18 +47,18 @@ function createConfig(config, argv, { port }) {
...
   }
 
-  options.devMiddleware = options.devMiddleware || {};
+  options.dev = options.dev || {};
 
-  if (!options.devMiddleware.publicPath) {
+  if (!options.dev.publicPath) {
     // eslint-disable-next-line
-    options.devMiddleware.publicPath =
+    options.dev.publicPath =
```
evilebottnawi, line 54: I think it is really bad idea :smile: But let's resolve it on the other PR

lib/utils/createConfig.js
```diff
@@ -47,18 +47,18 @@ function createConfig(config, argv, { port }) {
...
       (firstWpOpt.output && firstWpOpt.output.publicPath) || '';
```
evilebottnawi, line 55: I think it is really bad idea smile But let's resolve it on the other PR

evilebottnawi:
@Loonride unstable CI?

Loonride:
After this PR we need to:
- Change how `publicPath` is handled
- Look deeper into test instability
- Investigate why webpack compilation calls the `done` hook many times if the compiler starts directly after an `fs.writeFileSync` of a CSS file in the compilation (it works fine if there is a 1 second delay between file writing and compilation start)

test/e2e/Socket-injection.test.js
```diff
@@ -8,7 +8,7 @@ const express = require('express');
 const request = require('supertest');
 const testServer = require('../helpers/test-server');
 const runBrowser = require('../helpers/run-browser');
-const port = require('../ports-map').WebsocketClient;
+const port = require('../ports-map').SocketInjection;
```
Loonride, line 11: This port was already being used which was causing some instability

test/server/contentBasePublicPath-option.test.js
```diff
@@ -4,7 +4,7 @@ const path = require('path');
 const request = require('supertest');
 const testServer = require('../helpers/test-server');
 const config = require('../fixtures/contentbase-config/webpack.config');
-const port = require('../ports-map')['contentBase-option'];
+const port = require('../ports-map')['contentBasePublicPath-option'];
```
Loonride, line 7: Same with this port

evilebottnawi:
@Loonride Great! But still failed on macOS node-10-canary :disappointed: Very strange

Loonride:
> @Loonride Great! But still failed on macOS node-10-canary  Very strange

@evilebottnawi The remaining instability seems to be from puppeteer. I will look closely at that later. Does everything else look good in this PR?

PR #2670, opened 2020-07-04
BREAKING CHANGES: create static option and remove replaced options

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Not yet

### Motivation / Use-Case

`static` option is replacing many existing options. See `options.json`.

### Breaking Changes

- Remove `contentBase`, `contentBasePublicPath`, `serveIndex`, `staticOptions`, `watchContentBase`, `watchOptions` in favor of `static`
- API changes related to `contentBase`

### DIFF
--- a/bin/cli-flags.js
+++ b/bin/cli-flags.js
@@ -81,7 +81,7 @@ module.exports = {
     {
       name: 'static',
       type: String,
-      describe: 'A directory or URL to serve HTML content from.',
+      describe: 'A directory to serve static content from.',
       group: RESPONSE_GROUP,
       multiple: true,
     },
--- a/bin/cli-flags.js
+++ b/bin/cli-flags.js
@@ -21,7 +21,7 @@ module.exports = {
       describe: 'Enables/Disables live reloading on changing files',
     },
     {
-      name: 'serveIndex',
+      name: 'static-serve-index',
       type: Boolean,
       describe: 'Enables/Disables serveIndex middleware',
       defaultValue: true,
@@ -85,15 +85,15 @@ module.exports = {
       describe: 'HTTP/2, must be used with HTTPS',
     },
     {
-      name: 'content-base',
+      name: 'static-directory',
       type: String,
       describe: 'A directory or URL to serve HTML content from.',
       group: RESPONSE_GROUP,
     },
     {
-      name: 'watch-content-base',
+      name: 'static-watch',
       type: Boolean,
-      describe: 'Enable live-reloading of the content-base.',
+      describe: 'Enable live-reloading of the static directory.',
       group: RESPONSE_GROUP,
     },
     {
--- a/bin/cli-flags.js
+++ b/bin/cli-flags.js
@@ -21,12 +21,6 @@ module.exports = {
       describe: 'Enables/Disables live reloading on changing files',
     },
     {
-      name: 'static-serve-index',
-      type: Boolean,
-      describe: 'Enables/Disables serveIndex middleware',
-      defaultValue: true,
-    },
-    {
       name: 'profile',
       type: Boolean,
       describe: 'Print compilation profile data for progress steps',
@@ -85,15 +79,9 @@ module.exports = {
       describe: 'HTTP/2, must be used with HTTPS',
     },
     {
-      name: 'static-directory',
+      name: 'static',
       type: String,
       describe: 'A directory or URL to serve HTML content from.',
-      group: RESPONSE_GROUP,
-    },
-    {
-      name: 'static-watch',
-      type: Boolean,
-      describe: 'Enable live-reloading of the static directory.',
       group: RESPONSE_GROUP,
     },
     {
--- a/bin/cli-flags.js
+++ b/bin/cli-flags.js
@@ -83,6 +83,7 @@ module.exports = {
       type: String,
       describe: 'A directory or URL to serve HTML content from.',
       group: RESPONSE_GROUP,
+      multiple: true,
     },
     {
       name: 'history-api-fallback',
--- a/bin/options.js
+++ b/bin/options.js
@@ -17,7 +17,7 @@ const options = {
     describe: 'Enables/Disables live reloading on changing files',
     default: true,
   },
-  serveIndex: {
+  'static-serve-index': {
     type: 'boolean',
     describe: 'Enables/Disables serveIndex middleware',
     default: true,
@@ -75,14 +75,14 @@ const options = {
     group: SSL_GROUP,
     describe: 'HTTP/2, must be used with HTTPS',
   },
-  'content-base': {
+  'static-directory': {
     type: 'string',
     describe: 'A directory or URL to serve HTML content from.',
     group: RESPONSE_GROUP,
   },
-  'watch-content-base': {
+  'static-watch': {
     type: 'boolean',
-    describe: 'Enable live-reloading of the content-base.',
+    describe: 'Enable live-reloading of the static directory.',
     group: RESPONSE_GROUP,
   },
   'history-api-fallback': {
--- a/bin/options.js
+++ b/bin/options.js
@@ -72,8 +72,7 @@ const options = {
   },
   static: {
     type: 'string',
-    describe:
-      'A comma-delimited string of directories to serve static content from.',
+    describe: 'A directory to serve static content from.',
     group: RESPONSE_GROUP,
   },
   'history-api-fallback': {
--- a/bin/options.js
+++ b/bin/options.js
@@ -17,11 +17,6 @@ const options = {
     describe: 'Enables/Disables live reloading on changing files',
     default: true,
   },
-  'static-serve-index': {
-    type: 'boolean',
-    describe: 'Enables/Disables serveIndex middleware',
-    default: true,
-  },
   profile: {
     type: 'boolean',
     describe: 'Print compilation profile data for progress steps',
@@ -75,14 +70,9 @@ const options = {
     group: SSL_GROUP,
     describe: 'HTTP/2, must be used with HTTPS',
   },
-  'static-directory': {
+  static: {
     type: 'string',
     describe: 'A directory or URL to serve HTML content from.',
-    group: RESPONSE_GROUP,
-  },
-  'static-watch': {
-    type: 'boolean',
-    describe: 'Enable live-reloading of the static directory.',
     group: RESPONSE_GROUP,
   },
   'history-api-fallback': {
--- a/bin/options.js
+++ b/bin/options.js
@@ -72,7 +72,8 @@ const options = {
   },
   static: {
     type: 'string',
-    describe: 'A directory or URL to serve HTML content from.',
+    describe:
+      'A comma-delimited string of directories to serve static content from.',
     group: RESPONSE_GROUP,
   },
   'history-api-fallback': {
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -309,26 +309,7 @@ class Server {
           publicPath,
           express.static(staticOption.directory, staticOption.staticOptions)
         );
-
-        if (staticOption.serveIndex) {
-          this.app.use(publicPath, (req, res, next) => {
-            // serve-index doesn't fallthrough non-get/head request to next middleware
-            if (req.method !== 'GET' && req.method !== 'HEAD') {
-              return next();
-            }
-
-            serveIndex(staticOption.directory, staticOption.serveIndex)(
-              req,
-              res,
-              next
-            );
-          });
-        }
       });
-
-      if (staticOption.watch) {
-        this.watchFiles(staticOption.directory, staticOption.watch);
-      }
     });
   }
 
@@ -350,10 +331,6 @@ class Server {
           });
         }
       });
-
-      if (staticOption.watch) {
-        this.watchFiles(staticOption.directory, staticOption.watch);
-      }
     });
   }
 
@@ -405,6 +382,12 @@ class Server {
       static: () => {
         this.setupStaticFeature();
       },
+      serveIndex: () => {
+        this.setupServeIndexFeature();
+      },
+      watchStatic: () => {
+        this.setupWatchStaticFeature();
+      },
       onBeforeSetupMiddleware: () => {
         if (typeof this.options.onBeforeSetupMiddleware === 'function') {
           this.setupOnBeforeSetupMiddlewareFeature();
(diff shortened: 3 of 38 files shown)


Loonride:
@evilebottnawi Do we need `directory` to be a required property for the `static` object option? There is a chance the user could want to set other static options and keep `directory` as the default current working directory.

lib/options.json
```diff
@@ -299,14 +265,77 @@
...
+        {
+          "type": "array",
+          "items": {
+            "anyOf": [
+              {
+                "type": "string",
+                "minLength": 1
+              },
+              {
+                "type": "object"
```
evilebottnawi, line 332: Let's create `definition` fro object notation and use here it

evilebottnawi: Example:
```
"anyOf": [
        {
          "type": "boolean"
        },
        {
          "$ref": "#/definitions/StaticString"
        },
        {
          "$ref": "#/definitions/StaticObject"
        },
        {
          "type": "array",
          "items": {
            "anyOf": [
              {
                "$ref": "#/definitions/StaticString"
              },
              {
                "$ref": "#/definitions/StaticObject"
              }
            ]
          },
          "minItems": 1
        }
      ]
```

lib/utils/normalizeOptions.js
```diff
@@ -2,14 +2,40 @@
...
+
+    options.static.publicPath = options.static.publicPath || '/';
+
+    if (
+      typeof options.static.watch === 'undefined' ||
+      options.static.watch === true
+    ) {
+      options.static.watch = {};
+    }
+  }
```
evilebottnawi, line 37: I think like this:
```
const defaultOptionsForStatic = {
    directory: process.cwd(),
    staticOptions: {},
    publicPath: ['/'],
    serveIndex: { icons: true },
    // Respect options from compiler.watchOptions
    watch: {
      // ...
    },
  };

if (options.static === 'undefined') {
    options.static = [{ defaultOptionsForStatic }];
  } else if (typeof options.static === 'boolean') {
    options.static = options.static ? [defaultOptionsForStatic] : false;
  } else if (typeof options.static === 'string') {
    options.static = [
      { ...defaultOptionsForStatic, directory: options.static },
    ];
  } else if (Array.isArray(options.static)) {
    options.static = options.static.map((item) => {
      if (typeof item === 'string') {
        return { ...defaultOptionsForStatic, directory: item };
      }

      return { ...defaultOptionsForStatic, ...item };
    });
  } else {
    options.static = [{ ...defaultOptionsForStatic, ...options.static }];
  }
```

Note - we should support `--no-static` for CLI

bin/cli-flags.js
```diff
@@ -85,15 +85,15 @@ module.exports = {
...
       type: String,
       describe: 'A directory or URL to serve HTML content from.',
       group: RESPONSE_GROUP,
     },
     {
-      name: 'watch-content-base',
+      name: 'static-watch',
       type: Boolean,
-      describe: 'Enable live-reloading of the content-base.',
+      describe: 'Enable live-reloading of the static directory.',
```
evilebottnawi, line 96: I think we should not provide here a lot of stuff, `--static=path/to/directory` and `--no-static` is enough, for other complex examples you need to use the configuration

lib/Server.js
```diff
@@ -308,112 +307,37 @@ class Server {
...
+            publicPath
+          );
         }
-
-        serveIndex(contentBase, { icons: true })(req, res, next);
       });
-    }
+
+      if (staticOption.watch) {
+        this._watch(staticOption.directory, staticOption.watch);
```
evilebottnawi, line 327: Let's rename `_watch` to `watchFiles`

lib/utils/normalizeOptions.js
```diff
@@ -1,15 +1,70 @@
...
-  options.watchOptions = options.watchOptions || {};
+  // TODO: improve this to not use .find for compiler watchOptions
+  const configArr = getCompilerConfigArray(compiler);
+  const watchOptionsConfig = configArr.find(
+    (config) => config.watch !== false && config.watchOptions
+  );
+  const watchOptions = watchOptionsConfig
+    ? watchOptionsConfig.watchOptions
+    : {};
+  console.log(compiler.options.watch);
```
evilebottnawi, line 15: `console.log`

test/server/static-directory-option.test.js
```diff
@@ -204,111 +208,37 @@ describe('contentBase option', () => {
...
-        });
-
-        expect(true).toBe(false);
-      } catch (e) {
-        expect(e.message).toBe('Watching remote files is not supported.');
-        done();
-      }
-    });
-    it('Should throw exception (number)', (done) => {
+    it('Should throw exception (external url)', (done) => {
```
Loonride, line 217: @evilebottnawi Is it correct that I should be throwing if `isAbsoluteUrl` for the static option is true?

lib/options.json
```diff
@@ -1,5 +1,54 @@
...
+            {
+              "type": "boolean"
+            },
+            {
+              "type": "object"
+            }
+          ]
+        }
+      },
+      "required": ["directory"]
```
Loonride, line 45: Do we need this to be required? It seems unnecessary and may be confusing if the user is just specifying watch options or something.

Loonride: I'm going to remove this and let it default to `process.cwd()` as usual if not provided, unless you think it is needed

bin/cli-flags.js
Loonride: So we will get rid of the `serve-index` CLI flag?

Loonride: Also, it seems our CLI system does not work well with a `no-static` flag, since `static` would have to be a string flag, not sure how to work around it without a hacky method

lib/options.json
evilebottnawi: Yep, you are right, we don't need it, because you can set only `watchOptions`

test/server/static-directory-option.test.js
evilebottnawi: No sure, I think we should improve it on `schema-utils` side, and support only paths, no URLs

bin/options.js
```diff
@@ -75,16 +70,11 @@ const options = {
     group: SSL_GROUP,
     describe: 'HTTP/2, must be used with HTTPS',
   },
-  'static-directory': {
+  static: {
```
evilebottnawi, line 73: Can be multiple

Loonride:
I'm getting `System limit for number of file watchers reached` error repeatedly if I do `npm run test`, though individual tests seem to be fine. Does it look like I'm doing anything wrong with opening or closing the watcher?

bin/options.js
Loonride: @evilebottnawi Like `--static=example/dir1,example/dir2`?

evilebottnawi:
We need to set https://github.com/paulmillr/chokidar/blob/master/index.js#L325 to true for tests, to avoid it

evilebottnawi:
I think here good place https://github.com/webpack/webpack-dev-server/blob/master/setupTest.js

lib/utils/createConfig.js
```diff
@@ -75,25 +70,11 @@ function createConfig(config, argv, { port }) {
...
-    options.contentBase = false;
-  }
-
-  if (argv.watchContentBase) {
-    options.watchContentBase = true;
+  if (argv.static === false) {
+    options.static = false;
+  } else if (argv.static) {
+    options.static = argv.static.split(',');
+    options.static = options.static.map((staticDir) => path.resolve(staticDir));
```
Loonride, line 77: Added multiple handling here

bin/options.js
evilebottnawi: Like `--static=example/dir1 --static=example/dir2`

lib/utils/createConfig.js
evilebottnawi: I think we should avoid this syntax, in favor multiple `--static=one --static=two`

test/helpers/test-server.js
```diff
@@ -8,6 +8,14 @@ let server;
 // start server, returning the full setup of the server
 // (both the server and the compiler)
 function startFullSetup(config, options, done) {
+  if (typeof options.static === 'undefined') {
+    options.static = {
+      watch: {
+        poll: true,
+      },
+    };
+  }
```
evilebottnawi, line 17: It still will create problems with to o many files opens when you try to tests the `watch` option

setupTest.js
```diff
@@ -1,3 +1,4 @@
 'use strict';
 
+process.env.CHOKIDAR_USEPOLLING = 'true';
```
Loonride, line 3: This seems to fix it locally, we will see what happens with CI.

evilebottnawi: Maybe we should use `true` instead `'true'`?

lib/utils/normalizeOptions.js
```diff
@@ -1,15 +1,70 @@
...
+    ? watchOptionsConfig.watchOptions
+    : {};
+
+  const defaultOptionsForStatic = {
+    directory: process.cwd(),
+    staticOptions: {},
+    publicPath: ['/'],
+    serveIndex: { icons: true },
+    // Respect options from compiler watchOptions
+    watch: watchOptions,
```
Loonride, line 22: Do we want this breaking change of `watch` being enabled by default? Previously `watchContentBase` was false by default

evilebottnawi: I think it is good what we watch it by default, most of developers enable it by default

lib/Server.js
```diff
@@ -501,24 +428,20 @@ class Server {
...
     }
 
-    if (this.options.contentBase && this.options.serveIndex) {
-      runnableFeatures.push('contentBaseIndex');
-    }
-
-    if (this.options.watchContentBase) {
-      runnableFeatures.push('watchContentBase');
+    if (this.options.static) {
+      runnableFeatures.push('serveIndex', 'watchStatic');
```
evilebottnawi, line 444: I think we can union `static`/`serveIndex`/`watchStatic` in one method and middleware, using one without the other doesn't make sense at all

Loonride: @evilebottnawi I tried that but it causes breaking changes when `historyApiFallback` is enabled. It seems that the order of middleware things that are enabled surrounding historyApiFallback is important

https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L554-L575

evilebottnawi: Why do nut put `static` feature after `historyApiFallback`, or maybe we should union `historyApiFallback` with `static` features?

Loonride: I think it makes sense to add it to `static` since it deals with serving static HTML content

evilebottnawi: Yep, i think we should add the `historyApiFallback` option to static

evilebottnawi: Good catch :+1: Thanks

Loonride: Although, is there any case where a user would want `static: false` with `historyApiFallback: true`? Or should we not worry about this edge case?

evilebottnawi: I think you are right, you can use `historyApiFallback` without `static`

evilebottnawi: I am afraid double `static` middleware reduce performance, but it is main for development, plus multiple `static` features in difference places look like hacky, can you describe the problem with `historyApiFallback` or maybe demonstrate?

Loonride: I'm looking into this now.

As a side note related to this, do you know why `middleware` is pushed to the features list 3 times? https://github.com/webpack/webpack-dev-server/blob/4ab1f21bc85cc1695255c739160ad00dc14375f1/lib/Server.js#L559

evilebottnawi: Don't know, but I hope we will have tests on this, maybe we can look at git history why it was done

Loonride: @evilebottnawi I'm not really sure why it was added: https://github.com/webpack/webpack-dev-server/pull/1891/files

Loonride: @evilebottnawi After some searching I figured out that `contentBaseFiles` (https://github.com/webpack/webpack-dev-server/blob/4ab1f21bc85cc1695255c739160ad00dc14375f1/lib/Server.js#L562) feature was placed before and after `historyApiFallback` for this reason:

https://github.com/bripkens/connect-history-api-fallback/blob/master/examples/static-files-and-index-rewrite/README.md#configuring-the-middleware

Is it possible that the `middleware` feature is added multiple times for a similar reason? Changing it so that there is only 1 `middleware` pushed does not seem to break any tests, but maybe tests were not made for it?

Loonride: I think I will not mess with the `middleware` runnable feature, but will investigate it later and see if I can add some tests for having it get pushed multiple times.

I have made the 3 features: `static`, `serveIndex`, and `watchStatic`. It would be ideal to have only the `static` feature, but it is complicated by the `historyApiFallback` feature.

evilebottnawi: It is very weird, because if we can't find something in `static` and `history` we should return `404`, not trying to do it again :confused: maybe I something don't understand

Loonride: @evilebottnawi It isn't a matter of finding something or returning a 404, but rather the priority of `static` vs `history` when they are both trying to serve something from the same path. I still don't know exactly why it works, but placing `static` middleware before and after seems to ensure that static files take priority over the history api fallback file.

evilebottnawi: Yep, I see, very weird, but we can't do something here :disappointed:

lib/Server.js
```diff
@@ -449,16 +379,13 @@ class Server {
...
         }
       },
-      // Todo rename to `static` in future major release
-      contentBaseFiles: () => {
+      static: () => {
         this.setupStaticFeature();
       },
-      // Todo rename to `serveIndex` in future major release
-      contentBaseIndex: () => {
+      serveIndex: () => {
```
evilebottnawi, line 385: `staticServeIndex`

lib/Server.js
```diff
@@ -449,16 +379,13 @@ class Server {
...
         this.setupServeIndexFeature();
       },
-      // Todo rename to `watchStatic` in future major release
-      watchContentBase: () => {
+      watchStatic: () => {
```
evilebottnawi, line 388: `staticWatch`

lib/Server.js
evilebottnawi: and `this.setupStaticServeIndexFeature`

lib/Server.js
evilebottnawi: and `this.setupStaticWatchFeature`

lib/Server.js
```diff
@@ -382,10 +382,10 @@ class Server {
       static: () => {
         this.setupStaticFeature();
       },
-      serveIndex: () => {
+      staticServeIndex: () => {
         this.setupServeIndexFeature();
```
evilebottnawi: And let's rename method names: `this.setupStaticServeIndexFeature`, as I written above :1st_place_medal:

Loonride: > And let's rename method names: `this.setupStaticServeIndexFeature`, as I written above

Just did that :+1:

PR #2672, opened 2020-07-07
chore: rename isLocalIp flag

rishabh3112:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No (would add if needed)

### Motivation / Use-Case
Keeping flags' syntax consistent overall

### Breaking Changes
Maybe?

### Additional Info
https://github.com/webpack/webpack-cli/pull/1640


Loonride:
Probably should be done for `liveReload` CLI flag as well. We can ignore `serveIndex` because it will be replaced in this PR: https://github.com/webpack/webpack-dev-server/pull/2670.

Yes, this a breaking change, as it is `--useLocalIp` vs `--use-local-ip`.

@evilebottnawi Do you think this should go on v4 due to breaking changes?

evilebottnawi:
@Loonride Yes, we should do for `liveReload`, I think it is more bug fixes, because webpack-cli also in rc, and do not use this

evilebottnawi:
But let's merge it for v4

Loonride:
We should also update in here then, correct? https://github.com/webpack/webpack-dev-server/blob/master/bin/options.js

rishabh3112:
@Loonride will update the PR with requested changes soon.
It's been long time since I dig into dev server's codebase, so wasn't well aware of implications of those changes 😅.

evilebottnawi:
Thanks!

PR #2680, opened 2020-07-17
test(client): work on test instability

Loonride:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

N/A

### Motivation / Use-Case

Testing stability. What we need to look at in client logs needs to be adjusted due to recent logger updates.

### Breaking Changes

[WDS] is switched to [webpack-dev-server] in 1 case


evilebottnawi:
Do you have idea why we have still broken mac os?

Loonride:
> Do you have idea why we have still broken mac os?

I'm not sure, it must be some other test instability causing it. Currently just trying to deal with the major instability problems, but I will keep investigating

PR #2691, opened 2020-08-02
feat(server): allow async bypass function for proxying

gromo:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

Currently result of bypass function is used immediately and there is no way to make it asynchronous. This update allows returning Promise from bypass function in proxy.

It can be useful for downloading content from original source and "patching" it, instead of storing the whole response.

### Breaking Changes

No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -292,7 +292,7 @@ class Server {
         this.websocketProxies.push(proxyMiddleware);
       }
 
-      const handle = (req, res, next) => {
+      const handle = async (req, res, next) => {
         if (typeof proxyConfigOrCallback === 'function') {
           const newProxyConfig = proxyConfigOrCallback();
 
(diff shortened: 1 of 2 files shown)


gromo:
@evilebottnawi I don't think my changes caused CI failure. Looks like the problem is in your azure pipeline or code it's running. Check other recent PRs - they all failed with the same error:
```
SyntaxError: /Users/runner/work/1/s/node_modules/watchpack/lib/chokidar.js: 'return' outside of function (4:1)

  2 | try {
  3 | 	module.exports = require("chokidar");
> 4 | 	return;
    | 	^
  5 | } catch(e) {
  6 | 	v3Err = e;
  7 | }
```

evilebottnawi:
Please send a fix to v4 branch, v3 is closed for new features

PR #2696, opened 2020-08-06
feat(server): allow async bypass function for proxying

gromo:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

Currently result of bypass function is used immediately and there is no way to make it asynchronous. This update allows returning Promise from bypass function in proxy.

It can be useful for downloading content from original source and "patching" it, instead of storing the whole response.

### Breaking Changes

No


hiroppy:
Thanks

Issue #2692, opened 2020-08-03
Webpack module federation breaks with multiple entrypoints


codepunkt:
This time it's @sokra who told me this is probably the place to open an issue.

- Operating System: Windows 10
- Node Version: 12.16.2
- NPM Version: 6.14.4
- webpack Version: 5.0.0-beta.22
- webpack-dev-server Version: 3.11.0
- Browser: Chrome 84.0.4147.105, Firefox 79.0, Edge 84.0.522.48

- [x] This is a **bug**

### Code

https://github.com/codepunkt/module-federation-examples/tree/dynamic-host-remote

### Expected Behavior

Using `webpack-dev-server` instead of `webpack` should still support module federation with additional content tacked onto the remoteEntry by defining it as an additional `entry`.

### Actual Behavior

Running "yarn start" to start `webpack-dev-server` breaks module federation and thus breaks the app in development mode.

### For Bugs; How can we reproduce the behavior?

1. clone repository
2. ensure you're on branch "dynamic-host-remote"
2. run `yarn` on repo root
3. go into "dynamic-host-remote" directory
4. run `yarn start` in "dynamic-host-remote" directory
5. open `localhost:3001` in the browser
6. encounter an error in the browser console that happens when executing app2/remoteEntry.js with the additional contents that were added to this entrypoint by `webpack-dev-server`
7. OPTIONAL: run `yarn build && yarn serve` and revisit `localhost:3001` to see production build working just fine.

evilebottnawi:
@webpack-bot move to webpack/webpack-dev-server

evilebottnawi:
Feel free to send a fix

codepunkt:
If I had a clue what the problem is, i'd send a PR right away. Spent a few hours with this, don't have the slightest idea 🤔

ScriptedAlchemy:
Okay, so the issue here is that WDS is likely appending its own entrypoint to the array / injecting a "require" to the hmr client. That typically serves as the entry module of an application. When ModuleFederation emits a remoteEnty with the same name as the entrypoint, they are merged. 

This lets us attach initialization code, to federated remotes. something like changing publicPath. 

__webpack_public_path__ = new URL(document.currentScript.src).origin + "/";


https://github.com/module-federation/module-federation-examples/pull/277

![image](https://user-images.githubusercontent.com/25274700/89253737-4bd01d00-d5d2-11ea-8f24-be5d4a284618.png)

[...]

PR #2699, opened 2020-08-10
fix: prepend entrypoints instead of append

ScriptedAlchemy:
- [x] This is a **bugfix**
- [ ] This is a **feature**

this resolves an issue with module federation

re #2692
### Motivation / Use-Case

Module Federation will not work with WDS in its current state

### DIFF
--- a/lib/utils/addEntries.js
+++ b/lib/utils/addEntries.js
@@ -84,8 +84,8 @@ function addEntries(config, options, server) {
       // in this case, entry is a string or an array.
       // make sure that we do not add duplicates.
       /** @type {Entry} */
-      const entriesClone = additionalEntries.slice(0);
-      [].concat(originalEntry).forEach((newEntry) => {
+      const entriesClone = [].concat(originalEntry);
+      additionalEntries.forEach((newEntry) => {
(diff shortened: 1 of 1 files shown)


ScriptedAlchemy:
Tests still need to be written

evilebottnawi:
Thanks!

evilebottnawi:
In my TODO

ScriptedAlchemy:
Thanks!

Loonride:
@ScriptedAlchemy Did you test this locally, and did it solve the issue? If so, was it with webpack@5?

From my tests, it appears that this does not solve the problem for webpack@5. Webpack 5 seems to not handle entries correctly if they are added on after the compiler is created, regardless of if they are appended or prepended. Basically, if any entry is added onto the entry array, the module that was supposed to be exported will be ignored, and just an empty object will be exported, no matter what has been added to the entry array.

I got the same result as described here: https://github.com/webpack/webpack-dev-server/issues/2719#issuecomment-679114178

This may need to be solved on the webpack side, so that entries can be injected after the compiler is created without causing problems.

alexander-akait:
Fixed in master, anyway thanks for the PR

PR #2707, opened 2020-08-17
fix(color): set value to stats.colors for InfrastructureLogger

hiroppy:
- [x] This is a **bugfix**
- [ ] This is a **feature**

yep

### Motivation / Use-Case

before

Screen Shot 2020-08-17 at 11 42 23

after

Screen Shot 2020-08-17 at 11 26 13

### Breaking Changes

no


hiroppy:
CI fails. I'll check it.

lib/utils/overwriteWebpackConfig.js
```diff
@@ -0,0 +1,15 @@
...
+    config[0].stats = {};
+    config[0].stats.colors = true;
+  } else if (typeof config[0].stats === 'object') {
+    if (config[0].stats.colors == null) {
+      config[0].stats.colors = true;
+    }
+  }
+}
+
+module.exports = overwriteWebpackConfig;
```
evilebottnawi, line 15: I don't think we should override `stats.colors`

evilebottnawi: We should respect original value

hiroppy: many products won't show colors because they don't set stats.colors.(default is enabled at v4) 
So I overwrite it to keep between v4 and v5.

https://github.com/webpack/webpack-dev-server/blob/75718b7e25a27da598340cb00e23628d9496cd1a/bin/options.js#L76

evilebottnawi: @hiroppy In webpack@5 it is `true` by default? Maybe we should use this logic inside webpack

hiroppy: no, so I overwrite colors to stats.colors.

evilebottnawi: I think we should not override `colors`, because it can break output in some terminals

hiroppy: ok so what's your solution?

evilebottnawi: Get this options from stats, maybe we should improve this options by default on webpack side using logic from `supported-color` package

alexander-akait:
Because we move our bin to webpack-cli, i.e. `webpack serve`, `webpack-cli` set colors https://github.com/webpack/webpack-cli/blob/master/packages/webpack-cli/lib/webpack-cli.js#L1969, for Node.js API developer should enable colors manually, I think we can close it

Issue #2484, opened 2020-03-26
Module exports are not preserved when using dev-server with Webpack 5 beta


csvn:
- Operating System: Windows 10
- Node Version: v13.10.0
- NPM Version: 6.13.7
- webpack Version: 5.0.0-beta.14
- webpack-dev-server Version: 3.10.13
- Browser: Chrome v82

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

Reproduction link: https://github.com/csvn/webpack-dev-server-export-issue-reproduction

```js
// webpack.config.js
module.exports = {
  mode: 'development',
  entry: './src/main.js',
  output: {
    ecmaVersion: 2015,
    library: 'main',
    libraryTarget: 'window'
  },
  devServer: {
    port: 3001,
    contentBase: require('path').join(__dirname, 'public'),
    open: false
  }
};
```

```js
// src/main.js
export const greeting = 'Hello World!';
```

#### Additional details

The entry point is messed up when when running `webpack-dev-server`. The exports will always be empty (`{}`) due to `webpack-dev-server` [prepending it's entry points](https://github.com/webpack/webpack-dev-server/blob/master/lib/utils/addEntries.js#L142), which does not seem to yield the correct results with Webpack 5 beta.

##### When using `webpack-dev-server`
```js
/******/ 	// Load entry module and return exports
/******/ 	__webpack_require__("./src/main.js");
/******/ 	return __webpack_require__("./node_modules/webpack-dev-server/client/index.js?http://localhost:3001");
```

##### When using `webpack` build
```js
/******/ 	// Load entry module and return exports
/******/ 	return __webpack_require__("./src/main.js");
```

### Expected Behavior
Values exported from `main.js` should be available on `window.main`.

### Actual Behavior
The exported object on `window.main` is `{}`

### How can we reproduce the behavior?
1. Clone the repo from the link above
2. Run `npm start` (or `npx webpack-dev-server`
3. Open http://localhost:3001
4. There are error in console due to export from `main.js` missing

evilebottnawi:
Pr welcome

hmnd:
A workaround for now at least is to set `config.devServer.injectClient: false`; this prevents dev-server from prepending its entry.

Loonride:
> /cc @Loonride Can you look at this?

I will take a look

kamilic:
this problem still existed on webpack@5.1.0.

sprilukin:
[...]

PR #2723, opened 2020-08-28
fix(utils): addEntries fix

Loonride:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

This is a PR for: https://github.com/webpack/webpack-dev-server/issues/2692#issuecomment-682223721

Basically, we should not be re-injecting entries that the user provided, we should only inject webpack-dev-server entries, as the webpack compiler should already handle the user's entries

### Breaking Changes

- The `addEntries` util API has changed to take a `compiler` argument


lib/utils/addEntries.js
```diff
@@ -159,6 +152,57 @@ function addEntries(config, options, server) {
...
+
+    if (webpack.version[0] === '5') {
+      // EntryPlugin is only available for webpack@5
+      // @ts-ignore
+      const EntryPlugin = require('webpack/lib/EntryPlugin'); // eslint-disable-line import/no-unresolved
+      // for webpack@5, we use a plugin to add global entries
+      compiler.hooks.make.tapAsync(
+        {
+          name: 'webpack-dev-server',
+          stage: -100,
```
Loonride, line 170: @evilebottnawi @sokra I've added the tapAsync method here, but the issue I've found now is that something like this where multiple Servers are attached to a compiler no longer works:

```
const server1 = new Server(compiler, options1);
const server2 = new Server(compiler, options2);
```

as it results in something like `Error: Conflicting entry option static = [object Object] vs [object Object]`

This worked before, as there were tests that did this which are now broken. (These tests would usually close the old server before a new one was attached)

Should we go about this by making multiple servers on one compiler discouraged, or maybe a hacky method of disabling the old hook taps like in https://github.com/webpack/tapable/issues/109 after the server is closed?

Loonride: Or I could also try to prevent duplicate entries by looking at the compilation globalEntry object.

Loonride: I added the hacky method for disabling old hook taps, but haven't added tests for it yet

evilebottnawi:
@Loonride ready for review?

Loonride:
@evilebottnawi Almost, but 2 things

1. Need advice on the above problem
2. This is appearing in CI for webpack@next:

```
<e> [webpack-dev-middleware] TypeError: done is not a function
<e>     at ready (/home/runner/work/webpack-dev-server/webpack-dev-server/test/helpers/test-server.js:51:7)
<e>     at Hook.eval [as callAsync] (eval at create (/home/runner/work/webpack-dev-server/webpack-dev-server/node_modules/webpack/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:27:1)
<e>     at Hook.CALL_ASYNC_DELEGATE [as _callAsync] (/home/runner/work/webpack-dev-server/webpack-dev-server/node_modules/webpack/node_modules/tapable/lib/Hook.js:18:14)
```

evilebottnawi:
@Loonride I think related https://github.com/webpack/webpack-dev-middleware/issues/682, maybe you can provide example of configuration?

Loonride:
It looks like the issue is here: https://github.com/webpack/webpack-dev-server/blob/c7e53b0d4006d76e73462ff67304f19dacecb6e7/test/e2e/ClientOptions.test.js#L404

lib/Server.js
```diff
@@ -50,7 +50,7 @@ class Server {
     this.wsHeartbeatInterval = 30000;
 
     normalizeOptions(this.compiler, this.options);
-    updateCompiler(this.compiler, this.options);
+    this._closeCallback = updateCompiler(this.compiler, this.options);
```
Loonride, line 53: do you agree with this naming since it is supposed to be an internal property?

evilebottnawi: Yes :+1:

Loonride:
@evilebottnawi Ready for review

lib/utils/addEntries.js
evilebottnawi: @Loonride I think we should prevent duplicate, in theory we can have property on `compiler` to achieve this, we use same logic here https://github.com/webpack/webpack-dev-middleware/blob/master/src/utils/setupWriteToDisk.js#L9 (but it only for webpack@4)

evilebottnawi (commented):
I think after release webpack-dev-server@4 we should focus on extracting hot and client stuff from webpack-dev-server to separate middleware (we already have webpack-hot-middlware and webpack-client-middleware, we need deprecate them too) and use hooks for modify compiler/compilation options, also after this we can refactor webpack-dev-server to plugin, what do you think?

lib/utils/addEntries.js
evilebottnawi: I think it is more clean hack :smile:

hiroppy:
> we should focus on extracting hot and client stuff from webpack-dev-server to separate middleware (we already have webpack-hot-middlware and webpack-client-middleware, we need deprecate them too) 

Extracting is sounds good to me, but why do you want to deprecate them? You mean creating new libs...?

evilebottnawi:
@hiroppy 

> Extracting is sounds good to me, but why do you want to deprecate them? 

Because they do the same, only in a dirty way, I think it will be easier for us to union all into the one project, it will be easier to maintain and fix

> You mean creating new libs...?

Yes, new, just maybe take name, webpack-hot-middleware sounds good

lib/utils/addEntries.js
Loonride: @evilebottnawi I agree it is more clean, but one issue I now see is that if the user does:

```
const server1 = new Server(compiler, options1);
server1.close(() => {
  const server2 = new Server(compiler, options2);
});
```

If the `options2` object results in a new set of entries, that will not be reflected in the compilation, since the old make callback would still be used.

lib/utils/addEntries.js
```diff
@@ -159,6 +152,79 @@ function addEntries(config, options, server) {
...
+      webpack.version[0] === '5' &&
+      !compiler.hasWebpackDevServerMakeCallback
+    ) {
+      // this counter allows the dev server to ensure only the most recent
+      // make callback is used, preventing duplicate entry injections
+      if (compiler.webpackDevServerMakeCallbackCount) {
+        // this invalidates old make callbacks
+        compiler.webpackDevServerMakeCallbackCount += 1;
+      } else {
+        compiler.webpackDevServerMakeCallbackCount = 1;
```
Loonride, line 172: My idea was to make a counter for callbacks, so that only the most recent callback will be used, fixing the problem I mentioned. Let me know if this seems too hacky.

ylemkimon:
This will fix #2484.

ylemkimon:
@Loonride Would you mind if I take this PR?

evilebottnawi:
@ylemkimon I think you can resent this PR, based on @Loonride activity, @Loonride  is not available now

PR #2728, opened 2020-09-02
fix(status): public path appearing as 'undefined'

cshawaus:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

Instead of:

```bash
webpack output is served from undefined
```

The correct public path for webpack 5 will be shown:

```bash
```

### Breaking Changes

No

### Additional Info

n/a

### DIFF
--- a/lib/utils/status.js
+++ b/lib/utils/status.js
@@ -26,7 +26,12 @@ function status(uri, options, log, useColor) {
   }
 
   log.info(
-    `webpack output is served from ${colors.info(useColor, options.publicPath)}`
+    `webpack output is served from ${colors.info(
+      useColor,
+      typeof options.contentBasePublicPath === 'string'
+        ? options.contentBasePublicPath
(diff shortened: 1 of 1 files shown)


evilebottnawi (changes requested):
Invalid fix, please provide reproducible test repo

cshawaus:
> Invalid fix, please provide reproducible test repo


https://github.com/cshawaus/vue-3-ts-demo/tree/webpack-5

evilebottnawi:
@cshawaus I will look at this in near future

ylemkimon:
Fixed in #2849.

alexander-akait:
@ylemkimon yep, anyway I think we should improve this, because `publicPath` can be `auto` or can be `Function`, maybe we should move it in plugin :thinking:

alexander-akait:
Fixed in master

PR #2729, opened 2020-09-02
fix: use find-cache-dir to temporarily store certificates

wojtekmaj:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No :( It's a path change, so no _new_ tests would apply here. There also didn't seem to be any tests I could modify.

### Motivation / Use-Case

This fixes HTTPS server not working properly in [Yarn PnP mode](https://yarnpkg.com/features/pnp) due to the way the packages are stored, unless the user manually [unplugs the package](https://yarnpkg.com/cli/unplug/). By using find-cache-dir, used widely (e.g. by babel-loader, copy-webpack-plugin), we ensure that Yarn PnP will correctly handle reading and writing these certificates.

See thread under one of Yarn package manager maintainer's tweet:
https://twitter.com/wojtekmaj91/status/1300849402001739776?s=20

### Breaking Changes

Theoretically, if someone used node_modules/webpack-dev-server/ssl as a place to store their certificates for webpack-dev-server to consume, they may have a problem. However, there are options to provide certificates to webpack-dev-server in a better way.


evilebottnawi:
Please send a PR to v4 branch, v3 accept only security fixes, I am with it, for webpack@5 we can use API, but we can improve it in future

lib/utils/getCertificate.js
```diff
@@ -34,6 +39,7 @@ function getCertificate(logger) {
     const attributes = [{ name: 'commonName', value: 'localhost' }];
     const pems = createCertificate(attributes);
 
+    mkdirp.sync(certificateDir);
```
evilebottnawi, line 42: https://nodejs.org/api/fs.html#fs_fs_mkdir_path_options_callback
```
fs.mkdirSync(path, { recursive: true })
```

minimum suppored version Node.js is `10.13` for the next major release

wojtekmaj: Ah, missed that. Wonderful! Will fix today.

wojtekmaj: Done!

evilebottnawi (approved):

hiroppy (approved):

hiroppy:
thanks

PR #2738, opened 2020-09-14
chore(readme): update 'create-react-app' url

rahul3v:
Correcting it to the original url rather then using the redirect one


hiroppy (approved):

evilebottnawi (approved):

hiroppy:
Thanks

PR #2760, opened 2020-10-11
Update overlay.js to display actual error message when works with webpack 5.0.0?

Xrumet:
- [x ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case
As a user I have updated to Webpack 5.0.0.
Not sure if this fix should be addressed here or in webpack/webpack repo.
Trying to use webpack-dev-server 3.11.0 in order with webpack 5.0.0 and using "overlay" feature I see data format is changed and errors are not displayed in browser's overlay.
This change should avoid the issue and be backward compatible.

### DIFF
--- a/client-src/default/overlay.js
+++ b/client-src/default/overlay.js
@@ -112,11 +112,17 @@ function clear() {
 // Compilation with errors (e.g. syntax error or missing modules).
 function showMessage(messages) {
   ensureOverlayDivExists((div) => {
+    let singleMessage = messages[0];
+
+    if (singleMessage && singleMessage.message) {
+      singleMessage = singleMessage.message;
+    }
+
     // Make it look similar to our terminal.
     div.innerHTML = `<span style="color: #${
       colors.red
     }">Failed to compile.</span><br><br>${ansiHTML(
-      entities.encode(messages[0].message || messages[0])
+      entities.encode(singleMessage)
     )}`;
   });
 }
--- a/client-src/default/overlay.js
+++ b/client-src/default/overlay.js
@@ -116,7 +116,7 @@ function showMessage(messages) {
     div.innerHTML = `<span style="color: #${
       colors.red
     }">Failed to compile.</span><br><br>${ansiHTML(
-      entities.encode(messages[0])
+      entities.encode(messages[0].message || messages[0])
     )}`;
   });
 }


evilebottnawi:
Can you provide example the problem? Maybe screenshot(s)?

Xrumet:
image

Xrumet:
Some why "message" object format is changed when I switch from webpack v4.44.2 to v5.0.0.
Instead of string Array now it is Array of objects with only "message" property.

Xrumet:
Hi, @evilebottnawi 
I am new on Github and newb with all this Nodejs and related stuff. 
I have no idea even how to run and pass all those tests/checks.
I do not know is this issue should be addressed in webpack-dev-server or in webpack or some webpack's dependency.
I hope the issue is pretty clear for you, and if you could assist/help with it - I will appreciate it...
Please let me know if you need some more information.

Thanks!

phara0hcom:
@Xrumet Did you also updated webpack-cli? I had an issue with the dev-server and webpack 5 but I downgraded webpack-cli back to 3.3.12 and it was all fine.

Xrumet:
@phara0hcom No, I am using Node API to start dev server, so I had removed webpack-cli at all.
Not sure, is it important, I put my dev-server configuration in webpack config like that:
devServer:{
   ....
		overlay: {
			errors: true,
			warnings: false 
		},
    ...
}

phara0hcom:
@Xrumet Ok, could you try to install webpack-cli `npm install -D webpack-cli` 
https://github.com/webpack/webpack-cli#about
If I read this then I think you need it.

Xrumet:
@phara0hcom sure, I have already tried it right after your previous comment.
I have installed webpack-cli v3.3.12 and also run my project with CLI instead of API.
Behavior is the same. I see error overlay itself. But, because error "data" format is changed - "entities.encode" or "ansiHTML" checks input string for "length" property (expecting a string but object is passed) and returns empty string.

So, I see "Build failed." message but do not see errors/warnings list.

Xrumet:
Hi @evilebottnawi 
Could you help with this fix?
I have try to run npm install on my windows and it failed some why.
So, I cannot even try to write and check unit tests to pass code coverage policy.

Seems like that's it for me :(

evilebottnawi:
@Xrumet Hi, can you rebase?

PR #2762, opened 2020-10-13
fix: add more stats presets

chenxsan:
- [x] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

I was trying to introduce a fix here https://github.com/webpack/webpack-dev-server/pull/2761, but only to find that tests are failing because webpack 5 with the default `stats` option produces different stats on the screen than webpack 4.

So I have to specify another preset for `stats`, e.g., [`'detailed'`](https://github.com/webpack/webpack/blob/60f7ce301df553d8ab0276ba2838ddb60ead1c94/declarations/WebpackOptions.d.ts#L589), unfortunately webpack-dev-server v3 doesn't support this value yet https://github.com/webpack/webpack-dev-server/blob/master/lib/options.json#L362, currently using `stats: 'detailed'` would throw errors:

```
✖ ｢wds｣: webpack Dev Server Invalid Options

options.stats should be {Object|Boolean} (https://webpack.js.org/configuration/dev-server/#devserverstats-)

error Command failed with exit code 1.
```


chenxsan:
This could be closed now as that fix was closed in favor of another fix.

PR #2764, opened 2020-10-14
fix: update CI

chenxsan:
- [x] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

After upgrading webpack from 4.43.0 to 4.44.2 https://github.com/webpack/webpack-dev-server/pull/2764/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519L12480, a lot of tests failed. Guess we should fix it first.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -45,19 +45,19 @@ jobs:
       matrix:
         node-14:
           node_version: ^14.0.0
-          webpack_version: 4
+          webpack_version: webpack-4
         node-12:
           node_version: ^12.0.0
-          webpack_version: 4
+          webpack_version: webpack-4
         node-10:
           node_version: ^10.13.0
-          webpack_version: 4
+          webpack_version: webpack-4
         node-8:
           node_version: ^8.9.0
-          webpack_version: 4
+          webpack_version: webpack-4
         node-6:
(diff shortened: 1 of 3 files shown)


evilebottnawi (commented):
We are working in v4 branch, no need to update this

chenxsan:
@evilebottnawi Do you mean pull request https://github.com/webpack/webpack-dev-server/pull/2761 will be merged into v4 not v3? I think you told webpack-dev-server v3 is going to support webpack v4 and v5? That's why I want to fix those CI problems first, or that pull request testing against webpack v4 will fail for webpack v4.

chenxsan:
Anyway, I think I've found the reason of the issue, I'll make a PR to [watchpack](https://github.com/webpack/watchpack) first. And if you're willing to accept this PR fixing the CI, I can continue it, otherwise let's drop it.

chenxsan:
`yarn jest` would pass while `yarn jest -- --coverage` would fail.

evilebottnawi:
Fixed for v4, I think we don't need it for v3, because v3 is very deprecated, we need to do release ASAP and focus on webpack-dev-server v4, I want to close it, anyway thanks for the PR and help

PR #2810, opened 2020-11-02
fix: remove default CLI values from flags

anshumanv:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Nope, do we have them for default values? if not then we can take a snapshot and test it

### Motivation / Use-Case

https://github.com/webpack/webpack-cli/issues/2018

These values are applied during arg parsing and config values are not able to override it.

### Breaking Changes

Not sure

### DIFF
--- a/bin/cli-flags.js
+++ b/bin/cli-flags.js
@@ -22,19 +22,16 @@ module.exports = {
     {
       name: 'liveReload',
       type: Boolean,
-      defaultValue: true,
       describe: 'Enables/Disables live reloading on changing files',
     },
     {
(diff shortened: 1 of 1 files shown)


snitin315 (commented):
Just found these are removed in `v4`. 
https://github.com/webpack/webpack-dev-server/blob/v4/bin/cli-flags.js

anshumanv:
Ah yeah, seems unlikely that we'll have a release before v4 so we'll need to wait till the issue is fixed, closing

/cc @evilebottnawi

Issue #2029, opened 2019-06-14
Error: Cannot find module 'webpack-cli/bin/config-yargs'


jamiek-acl:
- Operating System: Mac OS
- Node Version: 10.15.1
- NPM Version: 6.4.1
- webpack Version: 4.34.0
- webpack-dev-server Version: 3.2.1

- [x] This is a **bug**...?
- [ ] This is a **modification** request

### Code

```js
// webpack.config.js
const path = require('path');
const config = require('./webpack.dev.config');
const HtmlWebpackPlugin = require('html-webpack-plugin');

const newConfig = Object.assign({}, config);

newConfig.entry = ['@babel/polyfill', './app/features/Certifications/index', 'webpack/hot/only-dev-server'];

newConfig.output = {
  filename: 'webpack-certifications-bundle.js',
  path: path.resolve('../app/assets/webpack'),
  publicPath: '/',
};

newConfig.devServer = {
  historyApiFallback: true,
  hot: true,
  open: true,
  openPage: 'dashboard',
  port: 9000,
  proxy: {
    '/api': {
      target: 'http://localhost:9010',
      secure: false,
    },
  },
  publicPath: '/',
};

newConfig.plugins.push(new HtmlWebpackPlugin());

module.exports = newConfig;
module.exports.devtool = 'cheap-module-eval-source-map';

```

```js
// package.json
scripts: {
  "dev:cert:webpack": "NODE_ENV=development webpack-dev-server --config webpack.dev.certifications.config.js"
}
```

### Expected Behavior

Running `npm run dev:cert:webpack` starts webpack dev server.

### Actual Behavior
```
~/src/projects/client$ npm run dev:cert:webpack

> acl-project-manager@0.0.1 dev:cert:webpack /Users/jamie/src/projects/client
> NODE_ENV=development webpack-dev-server --config webpack.config.js

internal/modules/cjs/loader.js:583
    throw err;
    ^

Error: Cannot find module 'webpack-cli/bin/config-yargs'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
    at Function.Module._load (internal/modules/cjs/loader.js:507:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/Users/jamie/src/projects/client/node_modules/webpack-dev-server/bin/webpack-dev-server.js:77:1)
[...]

PR #2814, opened 2020-11-06
Usage With NPM Scripts modify "serve": "webpack serve" / Update README.md

blackraccoon000:
Beginners get confused by the guide when using webpack-dev-sercer.
This change can reduce the burden on the answer.

webpack/webpack-cli#1948
#2029 (comment)

- [ ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case
I looked at the docs and set it up, but when I changed it by looking at the link below, it worked.
I think the documentation should be changed.

https://github.com/webpack/webpack-dev-server/issues/2029#issuecomment-707034614

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,9 +67,26 @@ NPM package.json scripts are a convenient and useful means to run locally instal
 binaries without having to be concerned about their full paths. Simply define a
 script as such:
 
-```json
-"scripts": {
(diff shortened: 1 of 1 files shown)


alexander-akait:
Thanks for the PR, fixed in master

PR #2839, opened 2020-11-17
fix(utils): use plugin to add entries for webpack@5

ylemkimon:
- [x] This is a **bugfix**
- [ ] This is a **feature**
- [ ] This is a **code refactor**
- [x] This is a **test update**
- [ ] This is a **docs update**
- [x] This is a **metadata update**

### For Bugs and Features; did you add new tests?
Yes, added module federation test cherry-picked from #2723.

**TODO:** Add an unit test for `DevServerEntryPlugin`

### Motivation / Use-Case
Fixes #2484 and fixes #2692. Closes #2723.

Instead of modifying compiler configuration and calling `hooks.entryOption` after it has instantiated, this calls `compilation.addEntry` to add dev server entries into `globalEntry`, as suggested in https://github.com/webpack/webpack-dev-server/issues/2692#issuecomment-682439552.

As discussed in https://github.com/webpack/webpack-dev-server/pull/2723#pullrequestreview-483187667, the issue is with reusing the compiler instance with multiple dev servers, so I've decided to use a plugin, whose entries can be modified later.

I've considered using existing plugins:
* `EntryPlugin`: adds a single entry, so changing the number of entries is not possible because removing hooks is not supported
* `DynamicEntryPlugin`: adding a global entry, i.e., `name: undefined`, is not possible because name should be a string (https://git.io/JkleI)

So I've created `DevServerEntryPlugin` based on `DynamicEntryPlugin`. It would also allow easy detection.

### Breaking Changes
The signature of `webpack-dev-server/lib/utils/addEntries` has changed from `addEntries(config, options, server)` to `addEntries(compiler, options, server)`, but GitHub code search shows no code is using this directly.

### Additional Info
The first three commits are revert or refactor. Diff ignoring whitespace changes would be better for reviewing changes.

PR #2844, opened 2020-11-18
refactor(utils): addEntries to DevServerPlugin

ylemkimon:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case
https://github.com/webpack/webpack-dev-server/pull/2841#discussion_r526021624

Removed `DevServerEntryPlugin` and moved `addEntries` into `DevServerPlugin`.

### Breaking Changes
Yes. `lib/utils/addEntries` (exported as `Server.addDevServerEntrypoints`) function is now `lib/utils/DevServerPlugin`, a webpack plugin and no longer exported. It now takes a webpack-dev-server option as a constructor argument and deduplication should be handled by including only a single instance of the plugin.

```js
const webpackDevServer = require('webpack-dev-server');

webpackDevServer.addDevServerEntrypoints(compiler, devServerOptions);
```
should be changed to 
```js
const devServerPlugin = require('lib/utils/DevServerPlugin');

if (!config.plugins.find((p) => p.constructor === devServerPlugin)) {
  config.plugins.push(new devServerPlugin(devServerOptions));
}
```

Furthermore, using HotModuleReplacementPlugin from another webpack installation is not supported and should be avoided.

### Additional Info
All changes are refactoring, so ignoring whitespace would help reviewing the diff.


lib/utils/addEntries.js
```diff
@@ -2,123 +2,138 @@
```
ylemkimon, line 2: If I rename this file, GitHub doesn't seem to recognize it has moved. I'll rename the file name after review.

lib/utils/addEntries.js
```diff
@@ -143,43 +158,31 @@ function addEntries(compiler, options, server) {
...
     }
-
-    if (options.hot || options.hot === 'only') {
-      config.plugins = config.plugins || [];
-      if (
-        !config.plugins.find(
-          // Check for the name rather than the constructor reference in case
-          // there are multiple copies of webpack installed
-          (plugin) => plugin.constructor.name === 'HotModuleReplacementPlugin'
-        )
```
ylemkimon, line 177: This is changed to constructor check. I wonder the statement is valid, as multiple webpack usually leads to error.

lib/utils/addEntries.js
```diff
@@ -143,43 +158,31 @@ function addEntries(compiler, options, server) {
...
-      ) {
-        config.plugins.push(new webpack.HotModuleReplacementPlugin());
-      }
-    }
-  });
+  }
 }
 
-module.exports = addEntries;
+module.exports = DevServerPlugin;
```
evilebottnawi, line 188: Let's rename `addEntries.js` to `DevServerPlugin`

lib/utils/updateCompiler.js
```diff
@@ -1,64 +1,43 @@
...
-    compilersWithoutHMR.forEach((compiler) => {
-      // addDevServerEntrypoints above should have added the plugin
-      // to the compiler options
-      const plugin = findHMRPlugin(compiler.options);
-      if (plugin) {
+    if (!findPlugin(config, webpack.HotModuleReplacementPlugin)) {
+      // add and apply the HMR plugin, if it didn't exist before.
+      if (options.hot === true || options.hot === 'only') {
+        const plugin = new webpack.HotModuleReplacementPlugin();
+        config.plugins.push(plugin);
```
evilebottnawi, line 36: Why do not put this logic in the plugin too? Plugin can add other plugins :smile: It is normal

lib/utils/updateCompiler.js
```diff
@@ -1,64 +1,43 @@
...
         plugin.apply(compiler);
       }
-    });
-  }
+    }
+  });
 }
 
 module.exports = updateCompiler;
```
ylemkimon, line 43: I really like how this became clear and straight forward. I think we can separate these into another package (middleware or plugin) in the future.

evilebottnawi (commented):
Anyway good job and big thanks

lib/utils/addEntries.js
ylemkimon: If I rename this file, GitHub doesn't seem to recognize it has moved. I'll rename the file name after review.

lib/utils/updateCompiler.js
ylemkimon: I thought this is consistent, but since `DevServerPlugin` adds the hot entry it may make more sense to have it there. What do you think? Should I also move this too:
```
    const providePlugin = new webpack.ProvidePlugin({
      __webpack_dev_server_client__: getSocketClientPath(options),
    });
    providePlugin.apply(compiler);
```

lib/utils/addEntries.js
evilebottnawi: Yep, good catch

lib/utils/updateCompiler.js
evilebottnawi: Yes, we will implement `webpack-hmr-middleware`, we will start it after v4 release and stabilization

lib/utils/addEntries.js
```diff
@@ -2,123 +2,138 @@
...
+
+    // we're stubbing the app in this method as it's static and doesn't require
+    // a server to be supplied. createDomain requires an app with the
+    // address() signature.
+
+    this.app = server || {
+      address() {
+        return { port: options.port };
+      },
+    };
```
evilebottnawi, line 25: Can you clarify? I am afraid about it

lib/utils/updateCompiler.js
evilebottnawi: Yep, I think it make sense, ideally will be keeping all logic for modify compiler (add entries/plugins/change options and etc) should be happens in the plugin

lib/utils/addEntries.js
ylemkimon: I just copied from: https://github.com/webpack/webpack-dev-server/blob/6c2082fb8469af5025b024f7c2a7a5fa8fb03f53/lib/utils/addEntries.js#L20-L28 `git blame` shows it was added in https://github.com/webpack/webpack-dev-server/commit/e5dc236677e1865fce8afb903b9f833f7531333e.

evilebottnawi: I think we should remove this and also remove this https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L1034, no need to do it anymore, if somebody will need this, they can use plugin

lib/utils/addEntries.js
```diff
@@ -2,123 +2,139 @@
...
+    // eslint-disable-next-line no-shadow
+    const checkInject = (option, _config, defaultValue) => {
+      if (typeof option === 'boolean') return option;
+      if (typeof option === 'function') return option(_config);
+      return defaultValue;
+    };
 
-  // eslint-disable-next-line no-shadow
-  compilers.forEach((compiler) => {
     const config = compiler.options;
```
evilebottnawi, line 138: Let's rename it to `compilerOptions`, there is no `config` here

evilebottnawi (approved):
Looks good, let's rename file and I think we can merge it, good job and big thanks again, you are great :star:

evilebottnawi:
Forget about one - do we have multi compiler tests?

ylemkimon:
@evilebottnawi I cannot seem to fix type errors. If I `require('./getSocketClientPath')` in `addEntries`, it type-checks client source. Adding `"exclude": ["client"]` doesn't help as they are imported in `getSocketClientPath`.

ylemkimon:
> Forget about one - do we have multi compiler tests?

https://github.com/webpack/webpack-dev-server/blob/17b65002779e9802a24367688721b9be2e893f9c/test/server/utils/updateCompiler.test.js#L142-L151

ylemkimon:
If I add a module declaration file (`.d.ts`) for `getSocketClientPath`, it'd not look inside client sources. (Tried `jsdoc`, but it still looks inside client sources)

lib/utils/DevServerPlugin.js
```diff
@@ -0,0 +1,209 @@
...
+    /** @type {Boolean} */
+    const webTarget = [
+      'web',
+      'webworker',
+      'electron-renderer',
+      'node-webkit',
+      // eslint-disable-next-line no-undefined
+      undefined,
+      null,
+    ].includes(compilerOptions.target);
```
evilebottnawi, line 145: I found better way to do it in webpack@5, let's check:
```js
const isWebTarget = compiler.optons.externalsPresets 
  ? compiler.optons.externalsPresets.web
  : // Old logic for webpack@4
```

evilebottnawi (approved):
Very good job, only one improvement and we can merge it

lib/utils/DevServerPlugin.js
ylemkimon: Should we add client entries when `compiler.optons.externalsPresets.web === null`, i.e., targets are mixed, e.g., `['web', 'node']`?

evilebottnawi: No, only when it `web`, if it is not `web` and still in browser (I don't know this case :smile: ) you should use `injectHot: true`

evilebottnawi (approved):
Let's fix tests and merge it

ylemkimon:
I just realized I made a mistake, the plugin should be applied only when it didn't exist before and added.

lib/utils/updateCompiler.js
```diff
@@ -18,8 +18,8 @@ function updateCompiler(compiler, options) {
     } else {
       devServerPlugin = new DevServerPlugin(options);
       compilerOptions.plugins.push(devServerPlugin);
```
evilebottnawi, line 20: No need to adding this, just use `devServerPlugin.apply(compiler);`

lib/utils/DevServerPlugin.js
```diff
@@ -0,0 +1,211 @@
...
+
+    if (
+      hotEntry &&
+      !compilerOptions.plugins.find(
+        (p) => p.constructor === webpack.HotModuleReplacementPlugin
+      )
+    ) {
+      // add and apply the HMR plugin, if it didn't exist before.
+      const plugin = new webpack.HotModuleReplacementPlugin();
+      compilerOptions.plugins.push(plugin);
```
evilebottnawi, line 205: Just `apply` no need to push it in `plugins` array

lib/utils/updateCompiler.js
ylemkimon: I forgot the purpose of #2839, which was to support reusing a single compiler instance with multiple dev-servers, e.g.:
```js
const server1 = new Server(compiler, options1);
server1.close(() => {
  const server2 = new Server(compiler, options2);
});
```
Should this be supported or disallowed?

evilebottnawi: supported, but with only `apply` it should work fine

ylemkimon: But if options change, entires should be changed, right?

lib/utils/DevServerPlugin.js
ylemkimon: If it is not pushed to `plugins`, it'd add it again if the plugin is applied again.

lib/utils/updateCompiler.js
evilebottnawi: Yes, right, options should be updated

lib/utils/DevServerPlugin.js
evilebottnawi: Can you clarify? We should not touch `compilerOptions`, only use `plugin.apply(compiler)`

ylemkimon: If newly applied plugins are not added to `config.plugins`, then in
```js
const server1 = new Server(compiler, options1);
server1.close(() => {
  const server2 = new Server(compiler, options2);
});
```
each time the Server constructor is called, HMR plugin and this plugin will be applied. So they will be applied twice, causing duplicate hooks and entries.

ylemkimon: I think using a single compiler instance across multiple dev-server instances should be reconsidered, as it's impossible to remove or update hooks and it's not possible to persist state (whether plugins and hooks are added, etc.) without touching `compiler` or `compiler.options`.

evilebottnawi: I think here architecture problem, if we will be plugin we will not have this problem, so we need focus for v5 on refactor to plugin

evilebottnawi: We can mark it as limitation now

ylemkimon: Removed `compilerOptions.plugins.push(plugin)`. This PR is ready to merge.

ylemkimon: The plugin would have the same limitation, since the second dev-server (`server2`) cannot get or modify the plugin added by the first dev-server (`server1`).

evilebottnawi (approved):
Good job and big thanks, we need focus on CLI, it is the last issue (hope)

evilebottnawi:
We need remove `cli` logic, because it is deprecated and no longer work with webpack-cli, so we need remove this, but for better DX maybe we can keep CLI file, but what we will do:
- throw friendly error about migrate on `webpack serve`
- run `webpack serve` by default as `webpack` do https://github.com/webpack/webpack/blob/master/bin/webpack.js

What do you think?

ylemkimon:
How about we defer to `webpack`:
```js
// maybe throw a friendly error if webpack is not installed
process.argv.splice(2, 0, 'serve');
require('webpack/bin/webpack');
```
?

evilebottnawi:
Sounds good, but  am afraid webpack can change logic too in future and we will can broken again, so I think it would be better to use `webpack-cli` directly, yep copy/paste code, but it protected us from somethings unexpected

lib/utils/DevServerPlugin.js
```diff
@@ -0,0 +1,210 @@
...
+      path = `&path=${options.client.path}`;
+    }
+    /** @type {string} */
+    const port =
+      options.client && options.client.port
+        ? `&port=${options.client.port}`
+        : '';
+    /** @type {string} */
+    const clientEntry = `${require.resolve(
+      '../../client/default/'
```
shaodahong, line 62: Maybe `../../client-src/default/` ?

PR #2845, opened 2020-11-19
fix(cli): use webpack-cli@4

ylemkimon:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case
* Added `setupExitSignals` option. Takes a `boolean` and if `true` (default on CLI), the server will close and exit the process on `SIGINT` and `SIGTERM`
* `server.listen()` will find free port if `this.options.port` is not set and `port` argument is not passed
* Print a warning if `options.port` and `port` passed as an argument is different
* Added CLI calling `webpack serve`.

### Breaking Changes
* `server.listen()` now returns a Promise, which will resolve to `server.listeningApp`, after finding the free port.
* `webpack-dev-server` CLI, `utils/createConfig`, and `utils/processOptions` are removed.

### Additional Info
webpack-cli PR: https://github.com/webpack/webpack-cli/pull/2126


bin/webpack-dev-server.js
```diff
@@ -1,126 +1,151 @@
...
+/** @type {CliOption} */
+const cli = {
+  name: 'webpack-cli',
+  package: 'webpack-cli',
+  binName: 'webpack-cli',
+  installed: isInstalled('webpack-cli'),
+  url: 'https://github.com/webpack/webpack-cli',
+  preprocess() {
+    process.argv.splice(2, 0, 'serve');
+  },
```
ylemkimon, line 83: This file is copied from webpack/bin/webpack.js and added these lines.

lib/utils/setupExitSignals.js
```diff
@@ -1,21 +0,0 @@
...
-        });
-      } else {
-        // eslint-disable-next-line no-process-exit
-        process.exit();
-      }
-    });
-  });
-}
-
-module.exports = setupExitSignals;
```
evilebottnawi, line 21: We should keep it, it is very important, otherwise `CTRL+C` will be broken, maybe put it here https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L777?

lib/utils/findPort.js
```diff
@@ -1,39 +0,0 @@
...
-  // so the tryParseInt function is introduced to handle NaN
-  const defaultPortRetry = defaultTo(
-    tryParseInt(process.env.DEFAULT_PORT_RETRY),
-    3
-  );
-
-  return pRetry(runPortFinder, { retries: defaultPortRetry });
-}
-
-module.exports = findPort;
```
evilebottnawi, line 39: We should not remove this logic and it is very popular, let's move inside server class, when developers doesn't provide port, we try to use `8080` and if it is busy we try to get next (just reuse this logic)

bin/webpack-dev-server.js
```diff
@@ -1,126 +1,151 @@
...
 
-  try {
-    compiler = webpack(config);
-  } catch (err) {
-    if (err instanceof webpack.WebpackOptionsValidationError) {
-      console.error(colors.error(statsColors, err.message));
-      // eslint-disable-next-line no-process-exit
-      process.exit(1);
-    }
+  const isYarn = fs.existsSync(path.resolve(process.cwd(), 'yarn.lock'));
```
evilebottnawi, line 95: Sorry, my mistake, let's get logic from https://github.com/webpack/webpack/pull/11945/files, we will merge it, so we will support pnpm (no big different in code) also

evilebottnawi (commented):
We need improve this, but yes, it is very good job

lib/utils/setupExitSignals.js
ylemkimon: Shouldn't this be part of the CLI, `@webpack-cli/serve`?

evilebottnawi: hm, I think no, webpack-cli should don't know about internal logic + when we will refactor it to plugin it will be impossible, we call https://github.com/webpack/webpack-cli/blob/master/packages/serve/src/startDevServer.ts#L48 and we can put logic in `listen`, in think logic will work only when `options.port` is undefined

I will fix problem here https://github.com/webpack/webpack-cli/blob/master/packages/serve/src/startDevServer.ts#L36, after we do release v4

lib/utils/findPort.js
evilebottnawi: webpack-cli should don't know about internal logic + when we will refactor it to plugin it will be impossible, we call https://github.com/webpack/webpack-cli/blob/master/packages/serve/src/startDevServer.ts#L48 and we can put logic in listen, in think logic will work only when options.port is undefined

I will fix problem here https://github.com/webpack/webpack-cli/blob/master/packages/serve/src/startDevServer.ts#L36, after we do release v4

lib/utils/setupExitSignals.js
evilebottnawi: Wrong window please :smile: No, it is internal, we should keep internal logic for server inside our code

evilebottnawi: webpack-cli already do many logic right now, in the future it should only put only the plugin and nothing more

ylemkimon: This behavior might be undesirable when called directly via the API. Maybe add an option, `setupExitSignal`?

lib/utils/findPort.js
ylemkimon: Currently, the caller is responsible for setting the port:
```js
const server = new Server(compiler, {port});
server.listen(port, hostname);
```

lib/utils/setupExitSignals.js
evilebottnawi: Yep, source good for the new option for API

lib/utils/findPort.js
evilebottnawi: hm, maybe we should refactor API? and allow to set options only on `Server` constructor?

ylemkimon: I think it should be the other way around, as the only place the port option is used before `listen()` is creating the socket URL: https://github.com/webpack/webpack-dev-server/blob/1aa0d2db699d3d9edad7d354b4501d2773aa3675/lib/utils/DevServerPlugin.js#L35-L37

I think it should use the `location.port` by default if `sock.port` is not specified (the current `sockPort: 'location'` behavior). I wonder the following test case is valid: https://github.com/webpack/webpack-dev-server/blob/1aa0d2db699d3d9edad7d354b4501d2773aa3675/test/e2e/ClientOptions.test.js#L65-L79 as if one access the page via proxy on `http://localhost:${port2}/main`, should the socket URL also be via proxy on `http://localhost:${port2}/ws`?

evilebottnawi: hm, I think you are right, should we fix it in this PR? maybe we can do it separately?

bin/webpack-dev-server.js
evilebottnawi: https://github.com/webpack/webpack/pull/11945/files merged, we should have with the same, but with `serve` as you already done

lib/utils/findPort.js
ylemkimon: I've opened another PR #2850 since it's a breaking change.

ylemkimon: Now the problem is we have three sources of ports to use:
(1) `options.port` (not used currently)
(2) an argument passed to `listen`
(3) free port found by `findPort`

evilebottnawi: Any ideas how to better solve it? As I said before we will refactor it to plugin, so plugin will have only `options`, so maybe we should accept port only by options to `Server` class?

ylemkimon: I think we can remove or deprecate passing the port to `listen`.

evilebottnawi: Maybe we can keep compatibility with warning? if `port` are different, because in the next major release we refactor many places and I think there is no problem with this?

ylemkimon: https://github.com/webpack/webpack-dev-server/blob/0ab7237e9a2ddbc15db64ade9c0eb3be8c71b2b6/lib/Server.js#L600-L606

bin/webpack-dev-server.js
ylemkimon: Done :smile:

evilebottnawi:
I need to go home (tired :disappointed: ), let's finish tomorrow, do you have time? Just ping me when it will be ready for review

ylemkimon:
@evilebottnawi No worries, whenever you want. This PR is ready for review.

evilebottnawi:
Just note - let's add test on output using snapshot, we need to make sure your console output looks good, just start server via CLI do snapshot, change something and do snapshot again, I am afraid we can have some incompatibilities with webpack-cli

bin/cli-flags.js
```diff
@@ -37,6 +37,14 @@ module.exports = {
...
+      name: 'setup-exit-signals',
+      type: Boolean,
+      describe: 'Close and exit the process on SIGINT and SIGTERM',
+      group: ADVANCED_GROUP,
+      defaultValue: true,
+      negative: true,
+    },
     {
       name: 'stdin',
       type: Boolean,
```
ylemkimon, line 50: Another issue is `stdin`. Should it be handled by `webpack-cli` or added to `options` and handled in `dev-server` side like `setupExitSignals`?

package.json
```diff
@@ -108,11 +106,12 @@
     "typescript": "^3.9.7",
     "url-loader": "^4.1.0",
     "webpack": "^4.44.1",
```
shaodahong, line 108: Default webpack version should be `5`?

bin/cli-flags.js
evilebottnawi: @ylemkimon sorry for some delay, to be honest, I don't quite remember how this option works, ideally stdin can be not only for webpack serve

package.json
evilebottnawi: @ylemkimon I think we should update dev deps to latest too, but let's do it in separate PR

bin/cli-flags.js
ylemkimon: It just calls `process.exit()` on `Ctrl+D` (EOF): https://github.com/webpack/webpack-dev-server/blob/4ab1f21bc85cc1695255c739160ad00dc14375f1/lib/utils/createConfig.js#L83-L90

Now I think it should not just exit the process, but close the server, it should be on the dev-server's side, like `setupExitSignals`.

package.json
ylemkimon: I think we should enable renovate on the `v4` branch.

bin/cli-flags.js
evilebottnawi: hm, yes, also I think it should be supports not only for webpack serve, ref:
- https://github.com/webpack/webpack-cli/issues/789
- https://github.com/webpack/webpack-cli/issues/885

I think for v4 webpack-cli `--stdin` is broken, so I think we should fix it in both places :disappointed:

package.json
evilebottnawi: After merge it in master, it will be automatically, but to avoid spamming and weird problems with CI and tests better to do it in separate PR (I think)

shaodahong: yep, maybe need improve some tests

bin/cli-flags.js
evilebottnawi: What do you think? I am working on https://github.com/webpack/webpack-cli/issues/2121, then will review https://github.com/webpack/webpack-cli/pull/2126 and return to this, maybe we csn start to solve problem with `--stdin` parallels

ylemkimon: I've added `stdin` option along with `setupExitSignals`.

package.json
ylemkimon: @evilebottnawi Using webpack 5 as the dev dependency would mean we're compiling client sources with webpack 5. Wouldn't this cause problem with old browser compatibility?

package.json
```diff
@@ -108,11 +106,12 @@
...
-    "webpack-cli": "^3.3.12",
+    "webpack-cli": "^4.2.0",
     "webpack-merge": "^4.2.2"
   },
   "peerDependencies": {
-    "webpack": "^4.0.0 || ^5.0.0"
+    "webpack": "^4.0.0 || ^5.0.0",
+    "webpack-cli": "^4.0.0"
```
evilebottnawi, line 114: We should avoid this, and using only `peerDependenciesMeta`, because many boilerplates uses webpack-dev-server without CLI, so we will break them

lib/options.json
```diff
@@ -322,6 +322,9 @@
     "public": {
       "type": "string"
     },
+    "setupExitSignals": {
```
evilebottnawi, line 325: hm, why we don't do it by default? This option can be very misleading for developers

ylemkimon: This is the default for the CLI. However, for API this would lead to unexpected behavior, as this may close the process that is using the API. I think that should be up to the user of the API.

package.json
ylemkimon: Most package managers now support `peerDependenciesMeta`. WIthout `peerDependencies`, wouldn't it be impossible to enforce the version of `webpack-cli`?

package.json
evilebottnawi: Should be no, but we should set `target: ['web', 'es5']`

package.json
evilebottnawi: Yes, most of supported, so we need to use it, npm@7 automatically install `peerDependencies` :facepalm: But in some cases we don't need `webpack-cli`, for example CRA just use our server, but don't use `webpack-cli`

lib/options.json
evilebottnawi: yep, agree, we will improve docs here

package.json
ylemkimon, line 114:
```suggestion
    "webpack": "^4.0.0 || ^5.0.0"
```

evilebottnawi (approved):
Good job, thanks, I found a lot of strange deps like - `find-cache-dir`/`util`/`url`/many be other, can you look at code and maybe remove some of them, because I found them are unnecessary

ylemkimon:
@evilebottnawi `find-cache-dir` is used to store temporary self-signed certificates:
https://github.com/webpack/webpack-dev-server/blob/ccdce3585ade018211646097cb753c8eb49bf697/lib/utils/getCertificate.js#L11-L15 `util` and `url` are polyfills of Node modules for browsers.

ylemkimon:
Thank you for the review!

evilebottnawi:
@ylemkimon Found other bug(s) with CLI, we have https://github.com/webpack/webpack-dev-server/blob/master/bin/cli-flags.js and https://github.com/webpack/webpack-dev-server/blob/master/bin/options.js, `webpack serve` use `cli-flags.js`, also we don't supports all types, like `--open` without options, we need to exports all possible types, like `require('webpack').cli.getArguments` do it, I think it is the last problem with CLI

ylemkimon:
`v4`'s cli-flags seem to be set correctly, e.g., `--open` aceepts both Boolean and String: https://github.com/webpack/webpack-dev-server/blob/ccdce3585ade018211646097cb753c8eb49bf697/bin/cli-flags.js#L53-L58

evilebottnawi:
Oh, yes , I see, anyway we have any extra file, let's remove it

ylemkimon:
Where was `bin/options.js` originally used?

evilebottnawi:
in `bin` https://github.com/webpack/webpack-dev-server/blob/master/bin/webpack-dev-server.js#L19

Issue #2854, opened 2020-11-23
Upgrade yargs peer dependency to v16 to remediate prototype pollution vulnerability issue


jjloneman:
- Operating System: **macOS Catalina (10.15.7)**
- Node Version: **15.2.1**
- NPM Version: **7.0.12**
- webpack Version: **4.44.2**
- webpack-dev-server Version: **3.11.0**
- Browser: **Chrome 87**

- [x] This is a **bug**
- [x] This is a **modification** request

### Expected Behavior

No vulnerability issues.

### Actual Behavior

Vulnerability issues.

### For Bugs; How can we reproduce the behavior?

```
$ npx snyk test webpack-dev-server

Testing webpack-dev-server...

✗ High severity vulnerability found in y18n
  Description: Prototype Pollution
  Info: https://snyk.io/vuln/SNYK-JS-Y18N-1021887
  Introduced through: yargs@13.3.2
  From: yargs@13.3.2 > y18n@4.0.0



Organization:      jjloneman
Package manager:   npm
Open source:       yes
Project path:      webpack-dev-server

Tested webpack-dev-server for known vulnerabilities, found 1 vulnerability, 1 vulnerable path.
```

### For Features; What is the motivation and/or use-case for the feature?

**webpack-dev-server@3.11.0** has a peer dependency on **yargs@13.3.2** which pulls in **y18n@4.0.0** which is subject to prototype pollution - a high severity vulnerability which is reported in https://snyk.io/test/npm/webpack-dev-server/3.11.0.

The automatic enterprise dependency management tool that I use at work, when trying to request **react-scripts@4.0.1**, rejected my request due to **react-script**'s peer dependency on this version of **webpack-dev-server** which transitively pulls in the vulnerable **y18n** library (https://snyk.io/test/npm/react-scripts/4.0.1)

**webpack-dev-server** vulnerability report:

> ## Prototype Pollution
> Vulnerable module: y18n
> * **Introduced through:** yargs@13.3.2
> ### Detailed paths
> * **Introduced through:** webpack-dev-server@3.11.0 › yargs@13.3.2 › y18n@4.0.0
> Remediation: Upgrade to yargs@16.0.0.

alexander-akait:
Fixed https://github.com/webpack/webpack-dev-server/releases/tag/v4.0.0-beta.0, please try and put feedback

PR #2855, opened 2020-11-23
Upgrade `yargs` to fix vulnerability issue

jjloneman:
- https://snyk.io/test/npm/webpack-dev-server/3.11.0
- https://github.com/webpack/webpack-dev-server/issues/2854

- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes. Added test in `cli.test.js` to demonstrate that updated `yargs` module properly displays usage statement.

Additionally, the existing tests which pass command-line options to the `webpack-dev-server` cli in `cli.test.js` pass which prove that the updated `yargs` module functions as expected.

### Motivation / Use-Case

Fix prototype pollution vulnerability issue in `yargs`.

See https://github.com/webpack/webpack-dev-server/issues/2854.

### Breaking Changes

No breaking changes.

### Additional Info

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,9 @@
   "name": "webpack-dev-server",
   "version": "3.11.0",
   "description": "Serves a webpack app. Updates the browser on changes.",
-  "bin": "bin/webpack-dev-server.js",
+  "bin": {
+    "webpack-dev-server": "bin/webpack-dev-server.js"
+  },
   "main": "lib/Server.js",
   "files": [
     "bin",
@@ -68,7 +70,7 @@
     "webpack-dev-middleware": "^3.7.2",
     "webpack-log": "^2.0.0",
     "ws": "^6.2.1",
-    "yargs": "^13.3.2"
+    "yargs": "^16.1.1"
   },
   "devDependencies": {
     "@babel/cli": "^7.8.4",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,15625 @@
 {
   "name": "webpack-dev-server",
   "version": "3.11.0",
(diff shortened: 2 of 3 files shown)


evilebottnawi:
Breaking change and we can't update it

package.json
```diff
@@ -2,7 +2,9 @@
   "name": "webpack-dev-server",
   "version": "3.11.0",
   "description": "Serves a webpack app. Updates the browser on changes.",
-  "bin": "bin/webpack-dev-server.js",
+  "bin": {
+    "webpack-dev-server": "bin/webpack-dev-server.js"
```
jjloneman, line 6: Just realized that `npm install` automatically changed this line for me. Running `webpack-dev-server` as a binary still functions the same (no breaking changes - https://docs.npmjs.com/cli/v6/configuring-npm/package-json#bin).

I can revert this line if desired.

jjloneman:
> Breaking change and we can't update it

Is there a reason why it can't be updated?

evilebottnawi:
Because https://github.com/webpack/webpack-dev-server/blob/00ca2a4d1abcc52477c0fbea91db1eb3141e0d6c/package.json#L16

jjloneman:
@evilebottnawi Ah, makes sense. I'll open an issue in `yargs` to see if they can make a patch fix for `y18n` v4 for node 
6 then, then I'll update my PR with the fixed compatible version if they update it.

alexander-akait:
Thanks for the PR, fixed in master, today we will release `webpack-dev-server@4-beta.0` (I will do stable release on the next week), anyway it is always should be very stable (we just move some options in other places)

PR #2857, opened 2020-11-24
feat: remove CLI progress and move progress option into client

ylemkimon:
- [x] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case
* Remove unused `bin/options.js` in favor of `bin/cli-flags.js`.
* Removed CLI progress output and `profile` option
* Moved `progress` option to `client.progress` and added `--client-progress`.

### Breaking Changes
- `bin/options.js` is removed.
- `profile` (`--profile`) option is removed, to print profile data, set `progress: 'profile'` (`--progress profile`).
- `progress` (`--progress`) option is moved to `client`, set `client: {progress: true}` (`--client-progress`).

### Additional Info
Depends on https://github.com/webpack/webpack-cli/pull/2133.


evilebottnawi:
And other simple breaking change:
- union https://github.com/webpack/webpack-dev-server/blob/v4/lib/options.json#L295 and https://github.com/webpack/webpack-dev-server/blob/v4/lib/options.json#L298
`progress: boolean | 'profile'`

I apologize very much for that, I haven't looked into the code for many times and want to reduce complex as much as possible (especial unnecessary options, lets options === better DX)

ylemkimon:
> and progress should be moved in `client` options, because it is only for client

Aren't `progress` and `profile` for the webpack compilation?

lib/options.json
```diff
@@ -292,11 +292,15 @@
         }
       ]
     },
-    "profile": {
-      "type": "boolean"
-    },
     "progress": {
```
evilebottnawi, line 295: Let's move `progress` to `client` object, because this plugin only for `client`

ylemkimon: It seems it's also used for the CLI output: https://github.com/webpack/webpack-dev-server/blob/ccdce3585ade018211646097cb753c8eb49bf697/lib/Server.js#L86-L89

evilebottnawi: > Aren't progress and profile for the webpack compilation?

For use no, we just emit them to client using `sockets`

evilebottnawi: oh, we should remove this, this plugin is unnecessary, `webpack serve --profile` should enable this plugin on `webpack-cli` side, I think it is already works like this, but we need to test it, adding progress plugin is out of scope webpack-dev-server, also if somebody already use this plugin in this configuration we will adding plugin twice

ylemkimon: Should we add `clientProgress` or `consoleProgress`, or would it be too much?

evilebottnawi: Can we add tests for cli too to ensure `--progress` works fine via CLI for node (for browser I think we already have tests)

evilebottnawi: Like `webpack serve --progress` and look at output

ylemkimon: I cannot find a way to mock `stderr` with the TTY support. Wrapping `process.stderr` and passing `stderr: 'inherit'` to `execa` doesn't work.

bin/cli-flags.js
```diff
@@ -21,21 +21,25 @@ module.exports = {
...
-      name: 'progress',
-      type: Boolean,
-      describe: 'Print compilation progress in percentage',
+      describe: 'Print compilation progress in percentage in the browser',
       group: BASIC_GROUP,
+      processor(opts) {
+        opts.client = opts.client || {};
+        opts.client.progress = opts.clientProgress;
+        delete opts.clientProgress;
+      },
```
evilebottnawi, line 32: hm, I think webpack serve should do it automatically, where is the problem?

evilebottnawi: Oh, I see, good name, I have ideas for refactoring here, but let's do it late

ylemkimon: Currently, argument processing is done manually in https://github.com/webpack/webpack-cli/blob/master/packages/serve/src/createConfig.ts and commander doesn't support nested properties. For better manageability, I've opened https://github.com/webpack/webpack-cli/pull/2133 to enable `cli-flags` to process them.

lib/options.json
evilebottnawi: Why we need mock here? Why just do not run `execa` and get `stderr`?

ylemkimon: I thought it was TTY issue that it didn't print the progress. However, it seems webpack-cli doesn't add `WebpackCLIPlugin` when `webpack serve` is called.

bin/cli-flags.js
```diff
@@ -106,7 +115,6 @@ module.exports = {
       describe: 'Enable gzip compression',
       group: RESPONSE_GROUP,
     },
-    // findPort is currently not set up
```
hiroppy, line 109: why did you comment out?

ylemkimon: The TODO was resolved in #2845.

evilebottnawi: @ylemkimon Can we remove this?

evilebottnawi:
@ylemkimon @hiroppy Today I want to release `webpack-dev-server@4.0.0-beta.1` 
- Why not alpha? Because I think we should have done it earlier, but unfortunately I did not have enough time for this and I am very sorry
- Why not `rc` or `stable`?  I'm still afraid that we will need to make a pair breaking changes, but we really look stable

I want to hold `beta` this week and release `rc` or `stable` (after feedback) at the next week 

What do you think?

hiroppy:
> Why not alpha? Because I think we should have done it earlier, but unfortunately I did not have enough time for this and I am very sorry

yes, me too. Not your fault

we have many breaking changes so `beta` is better I think. I want to see this version for more than 2 weeks.(e.g. we don't check several platforms such as Electron so we will get many feedbacks..

bin/cli-flags.js
ylemkimon: @evilebottnawi Do you mean removing `port` from `options`?

shaodahong:
Release beta to collect actual usage status, in my local, it's looking good, but can't represent everyone, if release beta I will migrate as soon as possible

evilebottnawi:
:star: I need some improvements on CLI side (`webpack serve`), ETA is tomorrow

bin/cli-flags.js
evilebottnawi: Oh, my mistake, all fine

evilebottnawi (approved):
/cc @hiroppy

evilebottnawi:
@ylemkimon My idea right now - move `client` related options to the `client` option, for example `overlay` (https://github.com/webpack/webpack-dev-server/blob/v4/lib/options.json#L264), it is related only for `client`, why? So that in the future we can extract hmr/client logic into `webpack-hot-module-replacement-middleware` (maybe better name, feel free to provide good names) and avoid unnecessary breaking changes.

hiroppy (approved):

lib/Server.js
```diff
@@ -569,8 +563,8 @@ class Server {
         this.sockWrite([connection], 'liveReload');
       }
 
-      if (this.options.progress) {
-        this.sockWrite([connection], 'progress', this.options.progress);
+      if (this.options.client.progress) {
+        this.sockWrite([connection], 'progress', this.options.client.progress);
       }
 
       if (this.options.clientOverlay) {
```
ylemkimon, line 570: I just realized this should be `this.options.client.overlay`.

ylemkimon:
> @ylemkimon My idea right now - move `client` related options to the `client` option, for example `overlay` (https://github.com/webpack/webpack-dev-server/blob/v4/lib/options.json#L264), it is related only for `client`.

What do you think about:
- `injectClient`
- `injectHot`
- `useLocalIp`

lib/Server.js
evilebottnawi: yes, let's fix it in the separate PR

evilebottnawi:
@ylemkimon I need think about it, let's focus on `overlay`

ylemkimon:
Should I open a new PR for `overlay` or update this PR?

alexander-akait:
@ylemkimon New PR

alexander-akait:
Let's merge this

PR #2859, opened 2020-11-25
feat(client): allow sock host to use browser location's host

gluxon:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes!

### Motivation / Use-Case

This is templated off of commit 0835a19c/#2341, but for the browser location's hostname instead of its port.

This is useful when webpack-dev-server is being accessed behind a reverse proxy with a hostname that's not known at build time.

### Breaking Changes

No breaking changes.

### Additional Info

None

### DIFF
--- a/client-src/default/utils/createSocketUrl.js
+++ b/client-src/default/utils/createSocketUrl.js
@@ -74,10 +74,13 @@ function getSocketUrl(urlParts, loc) {
   // all of these sock url params are optionally passed in through
   // resourceQuery, so we need to fall back to the default if
   // they are not provided
-  const sockHost = query.sockHost || hostname;
+  let sockHost = query.sockHost || hostname;
   const sockPath = query.sockPath || '/sockjs-node';
   let sockPort = query.sockPort || port;
 
+  if (sockHost === 'location') {
+    sockHost = loc.hostname;
+  }
   if (sockPort === 'location') {
     sockPort = loc.port;
   }
--- a/test/client/utils/createSocketUrl.test.js
+++ b/test/client/utils/createSocketUrl.test.js
@@ -127,6 +127,16 @@ describe('createSocketUrl', () => {
       'https://asdf/sockjs-node',
     ],
     [
+      '?http://example.com:8096&sockHost=location',
+      'http://something.com',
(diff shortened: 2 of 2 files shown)


gluxon:
I'll sign the CLA soon. Just sent it to my company's legal team to review.

gluxon:
It just occurred to me that `location` is a valid hostname. 😕 I wonder if passing in a boolean param `useLocationHostname` would be acceptable.

evilebottnawi (commented):
What is use case?

jsg2021:
I believe you can set the host to `0.0.0.0` to get this right now.

gluxon:
Thanks to you both for taking a look so quickly!

jsg2021 I did notice the conditional that sets hostname to loc.host if it's `0.0.0.0`, but is there a way get this behavior without having webpack listening on all network interfaces?

@alexander-akait I'm trying to support local and remote development for my team a bit better. The app/framework I'm using allows developers to create publicly accessible URLs that expose services running within a container. (Ex: `something.dev.acme.local`)

The problem is that wds always tries to connect to `localhost:8080/sockjs-node` instead of the dynamically created URL, which fails since wds is being proxied.

Setting `public` is another option, but that requires people to tweaking configs and restarting. Ideally sockjs on the frontend just connects to the root-relative path `/sockjs-node` and no one has to tweak configs.

alexander-akait:
Can you switch on v4? Only security fixes for v3 (if it is possible)

gluxon:
Certainly. Although I'm not confident in this implementation anymore since we'd cause unexpected behavior for users that actually want to set their host to `location` (literally). Regardless of how rare that may be, it's a bit of a code smell.

I'll think about this for a bit and open a different PR against the `v4` branch. Thanks again!

jsg2021:
@gluxon I meant on the client include entry point:

```
webpack-dev-server/client?http://0.0.0.0:${devPort}
```

this won't effect the server binding

gluxon:
@jsg2021 Ah I misunderstood. That's a good workaround for `3.11.0`, thanks!

I noticed that using `window.location` is the new default from looking at #2869. Glad that change merged. Thanks @ylemkimon

Issue #1642, opened 2019-01-25
Hard to catch error on proxying in Node.js 10


kmannislands:
* Operating System: macOs 10.14.1
* Node Version: 10.15.0
* NPM Version: yarn 1.13.0
* webpack Version: 4.23.1
* webpack-dev-server Version: 3.1.14

- [X] This is a **bug**
- [ ] This is a **modification** request

### Code
The bug I'm reporting here showed up after switching to Node js v10.x. I believe there are differences in how the `net` core package raises exceptions/policy toward unhandled error events.

To repro, you need to proxy a connection that uses WebSockets, allow it to connect using chrome, then restart the target server of the proxy, for example with Nodemon. Want to avoid creating a minimal repo of this if possible since it's a bit comple and there seems to be a nice trail of ticket/changelogs.

This will cause the `WDS` process to die due to unhandled error event from `ECONNRESET` with the following stack trace:

```
HPM] Error occurred while trying to proxy request /foo from localhost:9090 to https://localhost:3333 (ECONNREFUSED) (https://nodejs.org/api/errors.html#errors_common_system_errors)
events.js:167
      throw er; // Unhandled 'error' event
      ^

Error: read ECONNRESET
    at TLSWrap.onStreamRead (internal/stream_base_commons.js:111:27)
Emitted 'error' event at:
    at emitErrorNT (internal/streams/destroy.js:82:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:50:3)
    at Object.apply (/Users/kieran/work/Atlas/ui/node_modules/harmony-reflect/reflect.js:2064:37)
    at process._tickCallback (internal/process/next_tick.js:63:19)
```
Before upgrading to node js 10, I believe something similar was happening but the exception wasn't fatal.

After some debugging/ticket searching it seems that what changed is that it is now an error for the https module to encounter an exception without a registered handler, which impacts the behavior of your dependency `node-http-proxy`, [like this ticket describes](https://github.com/nodejitsu/node-http-proxy/issues/1286).

Debugging the error, I discovered that WDS `proxy` config is really `http-proxy-middleware` config, so it is possible to suppress the error and keep the server up by modifying config by adding [an onError handler to proxy config](https://github.com/chimurai/http-proxy-middleware#http-proxy-events) for future readers, however this was fairly difficult to debug and ideally this package would register some sort of handler for this exception so that it isn't fatal.

[...]

PR #2883, opened 2020-11-30
fix: handle WebSocket errors by default (#1642)

maknapp:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No, not sure how to add tests for this.

### Motivation / Use-Case

Prevents webpack-dev-server from crashing if a proxied WebSocket has errors.

### Breaking Changes

No breaking changes.

### Additional Info
None.


lib/Server.js
```diff
@@ -242,6 +242,15 @@ class Server {
           ? proxyConfigOrCallback()
           : proxyConfigOrCallback;
 
+      // Catch WebSocket errors by default
+      if (proxyConfig.ws && typeof proxyConfig.onProxyReqWs !== 'function') {
+        proxyConfig.onProxyReqWs = (proxyReq, req, socket) => {
+          socket.on('error', (err) => {
+            this.logger.warn('WebSocket error', err.message);
```
alexander-akait, line 249: I think we should add better message here, like `[http-proxy-middleware] Web socket error:` to avoid misleading

alexander-akait (commented):
Yes, tests is hard there, maybe you can provide steps to manually tests and I will put in todo adding tests

lib/Server.js
maknapp: My output looks like this:
```
<i> [webpack-dev-server] [HPM] Upgrading to WebSocket
<w> [webpack-dev-server] WebSocket error read ECONNRESET
<i> [webpack-dev-server] [HPM] Upgrading to WebSocket
<i> [webpack-dev-server] [HPM] Upgrading to WebSocket
<w> [webpack-dev-server] WebSocket error write after end
<w> [webpack-dev-server] WebSocket error write after end
<i> [webpack-dev-server] [HPM] Upgrading to WebSocket
<w> [webpack-dev-server] WebSocket error write after end
<w> [webpack-dev-server] WebSocket error write after end
```

Are you saying to change it to `this.logger.warn('WebSocket error:', err.message);` to make it:
```
<w> [webpack-dev-server] WebSocket error: write after end
```
or add `[http-proxy-middleware]` also to make it:
```
<w> [webpack-dev-server] [http-proxy-middleware] WebSocket error: write after end
```

maknapp:
I was trying to write tests and then realized that this is for 4.0.0-beta versions and not 3.11.0 which is where I was seeing the issue. I did not realize that master was so far ahead. I tried to reproduce the issue on the 4.0.0-beta.0 release but websocket proxying does not work at all. The browser side always gets a connection error with "Invalid frame header".

alexander-akait:
@maknapp sorry for delay, maybe you can provide steps how you can reproduce it for v3?

maknapp:
This issue was fixed in http-proxy-middleware v2.0.6.

PR #2886, opened 2020-12-01
fix(status): allow an empty string of publichPath

hiroppy:
- [x] This is a **bugfix**
- [ ] This is a **feature**

### DIFF
--- a/lib/utils/status.js
+++ b/lib/utils/status.js
@@ -6,16 +6,16 @@ const runOpen = require('./runOpen');
 function status(uri, options, logger, useColor) {
   logger.info(`Project is running at ${colors.info(useColor, uri)}`);
 
-  if (options.dev && options.dev.publicPath) {
-    logger.info(
-      `webpack output is served from ${colors.info(
-        useColor,
-        options.dev.publicPath
-      )}`
-    );
-  }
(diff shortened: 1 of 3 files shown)


lib/utils/status.js
```diff
@@ -6,16 +6,16 @@ const runOpen = require('./runOpen');
...
-
   if (options.static && options.static.length > 0) {
+    if (typeof options.static[0].publicPath !== 'undefined') {
+      logger.info(
+        `webpack output is served from ${colors.info(
+          useColor,
+          options.static[0].publicPath
+        )}`
+      );
+    }
```
alexander-akait, line 17: It is different public paths :smile: I think we can output both, need very helpful message

hiroppy: I see, I'll check the `output.publicPath` together.

alexander-akait: I think we should keep `options.dev && options.dev.publicPath` and improve static output like:
```
Content not from webpack is served from 'directory' by `publicPath` URL
```

alexander-akait:
Fixed here https://github.com/webpack/webpack-dev-middleware/blob/master/src/options.json

Issue #2887, opened 2020-12-01
v4.0.0-beta.0: clientLogLevel is removed without replacement


akphi:
- Operating System: MacOS 11.0.1 (Big Sur)
- Node Version: 14.13.0
- NPM Version: 6.14.8
- Yarn version: 1.22.5
- webpack Version: 5.9.0
- webpack-dev-server Version: 4.0.0-beta.0
- Browser: Chrome Version 87.0.4280.67

- [ ] This is a **bug**
- [x] This is a **modification** request

### Expected Behavior
In version 3, we can specify the `clientLogLevel` option to remove info logging from WDS directly to Chrome console.

### Actual Behavior
This option has been removed, and when I read the changelog, I tried to follow by setting `infrastructureLogging.level` to `none` as well as setting `stats.logging` to `none` but Chrome console still show messages like

```
[HMR] Waiting for update signal from WDS...
[webpack-dev-server] Hot Module Replacement enabled.
[webpack-dev-server] Live Reloading enabled.
[webpack-dev-server] App updated. Recompiling...
[webpack-dev-server] App updated. Recompiling...
[webpack-dev-server] Content base changed. Reloading...
Navigated to http://127.0.0.1:3000/ide
[HMR] Waiting for update signal from WDS...
[webpack-dev-server] Hot Module Replacement enabled.
[webpack-dev-server] Live Reloading enabled.
```

### For Features; What is the motivation and/or use-case for the feature?
Could we bring back a way to hide these noisy messages or at least show them as debug-level messages in Chrome?

hiroppy:
Did you try to use `client.logging`?

akphi:
@hiroppy my bad here for not checking for all the possible ways. I was trying to read the code and arrived at this.

Thank you, I'll close this now

hiroppy:
No worries. I want to improve to respect `infrastructureLogging.level` when client.logging is empty.

akphi:
@hiroppy that being said, are you aware of any way for us to customize the format of `infrastructureLogging`?

From the doc, there seems to be no way since infra logging is no in the stats object. but I would like to customize what I see in the terminal without having to do what `create-react-app` does which is to implement its own dev server client

hiroppy:
at V3, you were able to customize a logger but from v4, we decided to use infrastructureLogging so you can't customize it.
see https://github.com/webpack/webpack/blob/master/lib/logging/createConsoleLogger.js

[...]

PR #2889, opened 2020-12-01
feat(client): respect infrastructureLogging.level on Browser side as well

hiroppy:
- [ ] This is a **bugfix**
- [x] This is a **feature**

### Motivation / Use-Case

I think the existence of both infrastructureLogging.level and client.logger is confusing. So if client.logger is empty but infrastructureLogging.level is set, we should respect infrastructureLogging.level on Browser side as well.

This is my proposal so what do you think? @alexander-akait 
If you agree with this, I'll add tests.

https://github.com/webpack/webpack-dev-server/issues/2887

### Breaking Changes

no

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -551,8 +551,13 @@ class Server {
         }
       });
 
-      if (this.options.client.logging) {
-        this.sockWrite([connection], 'logging', this.options.client.logging);
+      const clientLogLevel =
+        this.options.client.logging ||
+        (this.compiler.options.infrastructureLogging &&
+          this.compiler.options.infrastructureLogging.level);
+
+      if (clientLogLevel) {
+        this.sockWrite([connection], 'logging', clientLogLevel);
       }
 
       if (this.options.hot === true || this.options.hot === 'only') {


lib/Server.js
```diff
@@ -551,8 +551,13 @@ class Server {
         }
       });
 
-      if (this.options.client.logging) {
-        this.sockWrite([connection], 'logging', this.options.client.logging);
+      const clientLogLevel =
+        this.options.client.logging ||
+        (this.compiler.options.infrastructureLogging &&
```
alexander-akait, line 556: Compilers can be array here

alexander-akait: Ideally it should be on plugin level (and getting from the compilation), but it is require moer complex logic

ylemkimon: Note there is `getCompilerConfigArray()`  util.

alexander-akait: Yes, it should be easy to fix

hiroppy: Good catch, I'll fix this tomorrow.

snitin315:
Closing in favor of https://github.com/webpack/webpack-dev-server/pull/3613

PR #2890, opened 2020-12-01
feat: consider file URLs in WebSocket URL creation

kroko:
- [ ] This is a **bugfix**
- [x] This is a **feature**

No.

However here's a test repo <https://github.com/kroko/webpacktest-devserver-v4-case03> which shows

* when index.html is opened via file system currently devserver fails
* when index.html is opened via file system proposed changes would  allow for devserver to work (including HMR)

### Motivation / Use-Case

The use case is not easy to explain in short, but
* we face need to run webpack-dev-server on DOOH stands for a _remote validation and changes_
* the client's infrastructure for those devices is set up so that stuff is run from file system, basically everything is done by Chrome and `--allow-file-access-from-files`, ` --disable-web-security`, ` --autoplay-policy=no-user-gesture-required`, etc.
* This addition would allow to run webpack-dev-server when HTML is opened in browser from filesystem (`file://` protocol, that is browser opens i.e. `file:///some/path/to/public/index.html`)
* I do not see any downsides in introducing replacement also for `file://` protocol 😄 

### Breaking Changes

None

### DIFF
--- a/client-src/clients/WebsocketClient.js
+++ b/client-src/clients/WebsocketClient.js
@@ -8,7 +8,14 @@ const BaseClient = require('./BaseClient');
 module.exports = class WebsocketClient extends BaseClient {
   constructor(url) {
     super();
-    this.client = new WebSocket(url.replace(/^http/, 'ws'));
+    this.client = new WebSocket(
+      ((urlToChange) => {
+        let wsUrl = urlToChange;
+        wsUrl = wsUrl.replace(/^http/, 'ws');
+        wsUrl = wsUrl.replace(/^file/, 'ws');
+        return wsUrl;
+      })(url)
+    );
 
     this.client.onerror = (err) => {
       log.error(err);
--- a/client-src/clients/WebsocketClient.js
+++ b/client-src/clients/WebsocketClient.js
@@ -8,14 +8,7 @@ const BaseClient = require('./BaseClient');
 module.exports = class WebsocketClient extends BaseClient {
   constructor(url) {
     super();
-    this.client = new WebSocket(
-      ((urlToChange) => {
-        let wsUrl = urlToChange;
-        wsUrl = wsUrl.replace(/^http/, 'ws');
-        wsUrl = wsUrl.replace(/^file/, 'ws');
-        return wsUrl;
-      })(url)
-    );
+    this.client = new WebSocket(url.replace(/^(?:http|file)/, 'ws'));
 
     this.client.onerror = (err) => {
       log.error(err);


alexander-akait (commented):
Good fix, but we really need test for this

kroko:
> Good fix, but we really need test for this

Thanks.

I tried to check tests (as codecov is failing for this PR), but at the glance did not find what and where, so kind of skipped. 😕 

* <https://github.com/webpack/webpack-dev-server/blob/master/test/client/clients/WebsocketClient.test.js>
* <https://github.com/webpack/webpack-dev-server/blob/master/test/client/utils/createSocketUrl.test.js>

Please advise.

client-src/clients/WebsocketClient.js
```diff
@@ -8,7 +8,14 @@ const BaseClient = require('./BaseClient');
...
     super();
-    this.client = new WebSocket(url.replace(/^http/, 'ws'));
+    this.client = new WebSocket(
+      ((urlToChange) => {
+        let wsUrl = urlToChange;
+        wsUrl = wsUrl.replace(/^http/, 'ws');
+        wsUrl = wsUrl.replace(/^file/, 'ws');
+        return wsUrl;
+      })(url)
+    );
```
ylemkimon, line 18:
```suggestion
    this.client = new WebSocket(url.replace(/^(?:http|file)/, 'ws'));
```

alexander-akait:
@ylemkimon maybe you can help with tests? I think we need one good e2e tests to we can improve more them in future

client-src/clients/WebsocketClient.js
kroko: commited

ylemkimon:
Unit tests are currently done in: https://github.com/webpack/webpack-dev-server/blob/763cde106056f342dcc30b607308b378c479cdaa/test/client/clients/WebsocketClient.test.js#L34

I think file URL conversion can be tested with:
```diff
-  describe('client', () => {
+  describe.each([['http'], ['file']])('client (%s)', (protocol) => {
     it('should open, receive message, and close', (done) => {
       socketServer.on('connection', (connection) => {
         connection.send('hello world');

         setTimeout(() => {
           connection.close();
         }, 1000);
       });

-      const client = new WebsocketClient(`http://localhost:${port}/ws-server`);
+      const client = new WebsocketClient(`${protocol}://localhost:${port}/ws-server`);
```

Note you can run tests and update snapshots via `npm run test:only -- -u`.

E2E tests are in `test/e2e`, and to test `file` protocol, `runBrowser()` should be able to run with custom arguments: https://github.com/webpack/webpack-dev-server/blob/05bdb0ce0738c1d014c949575d15e77b6132ed25/test/helpers/run-browser.js#L6
```diff
-function runBrowser(config) {
+function runBrowser(config, args) {
   const options = {
     viewport: {
       width: 500,
       height: 500,
     },
     userAgent: '',
     ...config,
   };

   return new Promise((resolve, reject) => {
     let page;
     let browser;

     puppeteer
       .launch({
         headless: true,
         // args come from: https://github.com/alixaxel/chrome-aws-lambda/blob/master/source/index.js
-        args: puppeteerArgs,
+        args: args ? puppeteerArgs.concat(args) : puppeteerArgs,
```

and then you can run puppeteer with custom arguments, e.g., `runBrowser(null, ['--allow-file-access-from-files', '--disable-web-security'])`.

kroko:
@ylemkimon thanks for the direction.

I tried wrapping my head around your infrastructure. Failed to achieve passing tests.

Also, for the test to check all cases it should include 3 protocols, HTTP, HTTPS, FILE

``` javascript
describe.each([['http'], ['https'], ['file']])('client (%s)', (protocol) => { /* ... */});
```

I'm at a point, where I have the _feeling_ (as I cannot pinpoint at all lines in the code I scrolled over 😄  ) that to achieve more than HTTP test would require quite a lot of work. However, as in some of the aspects I have limited experience I could be wrong.

Currently as you know it only checks HTTP. 

I'm still stuck at

```
  ● WebsocketClient › client (https) › should open, receive message, and close
  ● WebsocketClient › client (file) › should open, receive message, and close
```

and tracing back through `log.error.mock.calls` stack ... I kind of give up, for me to solve this would probably take too much time.

alexander-akait:
Sorry for delay, can we continue?

ylemkimon:
Changes are implemented in #2954. E2E test needs to be added, though.

alexander-akait:
Fixed

PR #2892, opened 2020-12-01
feat: improve host output and fix open

### LINKED ISSUE #2874: v4.0.0-beta.0 and host option
- Operating System: macOS 11.0.1
- Node Version: v14.15.1
- NPM Version: 6.14.8
- webpack Version: 5.8.0
- Browser: Chrome last

- [x] This is a **bug**

### Question

How should I replace the `host` option in the new version?
Perhaps I did not understand this from the changelog.

webpack-dev-server v3:

```js
devServer: {
  host: 'some.site.loc',
  <...>
}
```

with webpack-dev-server v4 I try:

```js
devServer: {
  client: {
    host: 'some.site.loc',
  }
  <...>
}
```

Result from console:

```
<i> [webpack-dev-server] Project is running at https://127.0.0.1:9090/
```

When I try open `https://some.site.loc:9090` I get response `Invalid Host header`.

If I do not change the option (I leave it as it was in the previous version), then by default `https://127.0.0.1:9090` will still open, but `https:/some.site.loc:9090` also works. In this case, I would like the console to write the correct address and the correct address to open in the browser.

alexander-akait:
- [x] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

fixes #2874

### Breaking Changes

No

### Additional Info

inspired by CRA


alexander-akait:
When you use `0.0.0.0` or `::`, our output is:
```
<i> [webpack-dev-server] Local: http://localhost:8081/
<i> [webpack-dev-server] On Your Network: http://192.168.0.5:8081/
```
When it is not IP:
```
<i> [webpack-dev-server] Project is running at http://local-dev.io:8081/
```

lib/Server.js
```diff
@@ -580,17 +581,93 @@ class Server {
...
+    const prettyPrintUrl = (newHostname) =>
+      url.format({ protocol, hostname: newHostname, port, pathname: '/' });
+
+    let prettyHostname;
+    let lanUrlForTerminal;
+
+    if (hostname === '0.0.0.0' || hostname === '::') {
+      prettyHostname = 'localhost';
+
+      const localIP = internalIp.v4.sync();
```
ylemkimon, line 597:
```suggestion
      const localIP = hostname === '::' ? internalIp.v6.sync() : internalIp.v4.sync();
```

lib/Server.js
```diff
@@ -580,17 +581,93 @@ class Server {
...
+
+      if (
+        /^10[.]|^172[.](1[6-9]|2[0-9]|3[0-1])[.]|^192[.]168[.]/.test(localIP)
+      ) {
+        // Address is private, format it for later use
```
ylemkimon, line 602: I think this check is not needed, as `0.0.0.0` (`::`) will bind to public address, if available.

test/cli/__snapshots__/cli.test.js.snap
```diff
@@ -28,7 +28,7 @@ exports[`CLI --hot webpack 4 1`] = `
 `;
 
 exports[`CLI --hot webpack 5 1`] = `
```
ylemkimon, line 30: Note snapshots should be updated each for webpack 4 and 5.

lib/utils/createDomain.js
```diff
@@ -8,27 +8,27 @@ function createDomain(options, server) {
   // use location hostname and port by default in createSocketUrl
```
ylemkimon, line 8: `showStatus` now using its own logic, this is only used in client entries. I think this can be refactored in the future PR.

lib/Server.js
alexander-akait: It is for terminal output, developer should copy past it from terminal, IPv6 is not supported in many browsers

lib/Server.js
alexander-akait: You can have local public IP (it is very rare edge case, but it is possible), for example I can run it on my server with public IP

lib/utils/createDomain.js
alexander-akait: I think logic for terminal output and for entries is different, it is similar in some places, but it is for different purpose

lib/Server.js
ylemkimon: In that case, I think the public IP should be printed.

alexander-akait: In this case we print two IP - local and network

lib/Server.js
ylemkimon: `url` automatically formats IPv6 hostname with brackets, which most browsers support including IE 11, e.g., `http://[::]:8080`. I think IPv6 should take precendence over IPv4, if available.

lib/Server.js
ylemkimon: IMHO, I see no reason not to print the internal IP (the IP address of the NIC) if it's public. It's still a network IP, but just not a LAN.

lib/Server.js
alexander-akait: make sense, I hope this does not scare developers

lib/Server.js
alexander-akait: I don't have strong opinion here, it was inspired by CRA and them feedback, but maybe you are right

alexander-akait: Let's simplify it, we will change it in future after feedback

alexander-akait:
IPv6 is not supported by github actions, we can't test it :disappointed:

alexander-akait:
https://github.com/actions/virtual-environments/issues/668#issuecomment-624080758

lib/Server.js
```diff
@@ -580,17 +581,91 @@ class Server {
...
+      this.logger.info(
+        `Project is running at ${colors.info(useColor, localUrlForTerminal)}`
+      );
+    }
+
+    if (
+      this.options.dev &&
+      typeof this.options.dev.publicPath !== 'undefined'
+    ) {
+      this.options.info(
```
sprilukin, line 625: @alexander-akait looks like `this.options.info` should be replaced with `this.logger.info`

alexander-akait: Already fixed in master

PR #2897, opened 2020-12-02
fix: default value of the `static` option is `static`

### LINKED ISSUE #2893: v4.0.0-beta.0: `static.watch` is enabled by default causing liveReload to always reload page on change
- Operating System: MacOS 11.0.1 (Big Sur)
- Node Version: 14.13.0
- NPM Version: 6.14.8
- Yarn version: 1.22.5
- webpack Version: 5.9.0
- Browser: Chrome Version 87.0.4280.67

- [x] This is a **bug**

### Expected Behavior
A bit of background first. I use `react-refresh` in my project. In version 3, when I make a change in my JSX file, the page does not reload as HMR takes place. In version 4, the page proceeds with a full reload, causing me to lose all app states.

The doc for v3 mentions that `watchBaseContent` is [disabled by default](https://webpack.js.org/configuration/dev-server/#devserverwatchcontentbase) and thus, this effectively means `liveReload=true` [does not take any effect](https://webpack.js.org/configuration/dev-server/#devserverlivereload).

### Actual Behavior
When I explicitly set `static.watch=true` in v4, the issue is resolved. I took a look at the recent change and it seems that in v3, we did [this](https://github.com/webpack/webpack-dev-server/blob/v3.11.0/lib/Server.js#L573)

```js
if (this.options.watchContentBase) {
  runnableFeatures.push('watchContentBase');
}
```

whereas in v4, we did [this](https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L438)

```js
// Server.js
if (this.options.static) {
  runnableFeatures.push('static');
}

if (this.options.static) {
  runnableFeatures.push('staticServeIndex', 'staticWatch');
}

// normalizeOptions.js
const watchOptionsConfig = configArr.find(
    (config) => config.watch !== false && config.watchOptions
  );
  const watchOptions = watchOptionsConfig
    ? watchOptionsConfig.watchOptions
    : {};

  const defaultOptionsForStatic = {
    directory: process.cwd(),
    staticOptions: {},
    publicPath: ['/'],
    serveIndex: { icons: true },
    // Respect options from compiler watchOptions
    watch: watchOptions,
  };
```

@hiroppy Maybe I misunderstood here, but this will try to get the `watchOptions` config from `webpack` and thus almost guarantee to be truthy if we're in `development` mode. Should we rethink this behavior or clarify it better in the doc/changelog since not everyone uses `webpack-dev-server` use it for serving static files?

### How to Reproduce The Issue
I have a reproducible repo, please take a look at this [PR](https://github.com/akphi/config-tester/pull/12). After upgrading to `webpack-dev-server@4.0.0-beta.0`. If you run the app using `yarn start` and try to hit the `+` button a few time to increase the age, then you try to make some trivial change in `App.tsx` the page is reloaded, and the age is reset to `10`.

On the other hand, if we add the following to `devServer` section in `webpack.config.js`, things will behave as expected:
```javascript
...
  devServer: {
      static: {
        watch: false,
      },
  ...
  }
```

alexander-akait:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Existing

### Motivation / Use-Case

fixes #2893

### Breaking Changes

Yes

BREAKING CHANGE: default value of the `static` option is `path.resolve(process.cwd(), 'dist')`, previously `process.cwd()`

### Additional Info

no


lib/utils/normalizeOptions.js
```diff
@@ -14,7 +15,7 @@ function normalizeOptions(compiler, options) {
     : {};
 
   const defaultOptionsForStatic = {
-    directory: process.cwd(),
+    directory: path.resolve(process.cwd(), 'static'),
```
alexander-akait, line 18: I don't use check directory is exist to allow ability adding `static` dynamically and working with it without problems

hiroppy (approved):

alexander-akait (commented):
After discussion with socka we decide to use `public` or `static`, I prefer second, there is some bad side effects to using `dist`

alexander-akait:
Let's merge, I think we need release soon

Issue #2932, opened 2020-12-16
Cannot mute `[HMR] Waiting for update signal from WDS...` in browser console


noelebrun:
- Operating System: macOS 11.1
- Node Version: v14.12.0
- NPM Version: v6.14.8
- webpack Version: v5.10.1
- webpack-dev-server Version: 4.0.0-beta.0
- Browser: Chrome & Safari

- [x] This is a **bug**
- [ ] This is a **modification** request

### Code

```js
{
  ...
  devServer: {
    client: {
      logging: 'none',
     ...
    },
    ...
  },
}
```

### Expected Behavior

Setting client logging to `none` should suppress completely all client log, including from the hot chunk.

### Actual Behavior

The log `[HMR] Waiting for update signal from WDS...` is still outputted to browser console when loading the page.

### For Bugs; How can we reproduce the behavior?

Simply start webpack-dev-server with any kind of configuration (just don't pass `hot: false`), and check the browser console when the page has loaded.

From what I understood, that's because [the default logging level of the hot chunk is `info`](https://github.com/webpack/webpack/blob/master/hot/log.js#L1), and it is overwritten by webpack-dev-server client by calling [the exported `setLogLevel` function](https://github.com/webpack/webpack/blob/master/hot/log.js#L45).

But that function is only called once the webpack-dev-server client receives [a `logging` command](https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/index.js#L73) that is send [when creating the socket server](https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L560). That only happens [after the web server has been started and is listening](https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L687).

That specific logging happens [as soon as the hot chunk is loaded](https://github.com/webpack/webpack/blob/master/hot/dev-server.js#L58), which is before the webpack-dev-server client receives that `logging` command from the socket server.

I don't have enough context to know which would be the best way to solve this, but let's start a discussion about it in here.

noelebrun:
I guess it wouldn't make sense for the hot chunk to wait for anything from webpack-dev-server client before logging since the whole purpose of this log is that inform that it's waiting for that.

Maybe simply adding a timeout and changing the log to something like `Nothing received from WDS after *X*s` if nothing was received by then?

alexander-akait:
[...]

PR #2946, opened 2020-12-27
fix: apply client logging options directly on client loading

noelebrun:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Updating logs snapshots was enough to test the new expected behavior. Some client tests were refactored as a result of some refactor in the client utils.

### Motivation / Use-Case

Following the issue reported in #2932, client logging options needed to be moved to initial client options (the ones passed in the entry URL query).

### Breaking Changes

None

### Additional Info

Since the query parsing was done directly in the socket URL generation function, it had to be moved to a separate util. This split also simplified the unit tests of those utils.


client-src/default/index.js
```diff
@@ -27,12 +28,24 @@ const options = {
...
+const socketUrl = createSocketUrl(urlOptions);
 
 self.addEventListener('beforeunload', () => {
   status.isUnloading = true;
 });
 
+if (urlOptions.query.logging) {
+  // this is needed because the HMR logger operate separately from
+  // dev server logger
+  const hotCtx = require.context('webpack/hot', false, /^\.\/log$/);
```
alexander-akait, line 41: Why do not use hot directly?

noelebrun: Indeed, wasn't sure if it would break something but seems like it was some legacy check to previously support old Webpack versions.

Updated it with a classic import now

client-src/default/utils/createSocketUrl.js
```diff
@@ -3,43 +3,19 @@
...
-}
 
 /*
  * Gets socket URL based on Script Source/Location
  * (scriptSrc: URL, location: URL) -> URL
  */
-function getSocketUrl(urlParts, loc) {
+function createSocketUrl(urlParts, loc) {
+  // Use parameter to allow passing location in unit tests
+  if (typeof loc === 'string' && loc !== '') {
```
alexander-akait, line 13: Is it only for tests?

noelebrun: Yep, it is. I kept it how it was before. (I've tried removing it and overriding `self.location` in the test but that didn't work. Probably because `self` doesn't work as expected in the test environment?)

alexander-akait: hm, weird, self should work, jest allows to mock this

noelebrun: I guess that's why it has already been mocked before

alexander-akait: I think we should avoid code for testing from runtime...

PR #2954, opened 2021-01-02
fix: support file: and chrome-extension: protocols in client

vespakoen:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes, I added tests to check if the protocol was changed, it is poking at variables inside of WebSocket and SockJS though, this is probably not the most stable way to do this, but I am not sure how to mock those or if that is even necessary.

### Motivation / Use-Case

When using node-webkit, or electron, the protocol can be `file:` or `chrome-extension:`, causing the websocket or sockjs to try to connect to an incorrect url, for example: `file://localhost:8080/ws` or `chrome-extension://localhost:8080/ws`.

This PR changes it to the correct behaviour.

### Breaking Changes

This PR should not be a breaking change.

### Additional Info

I just noticed I also replace `https` with `ws`, this should probably be an extra case where it is replaced with `wss`.
**EDIT: I removed the https case**

### DIFF
--- a/client-src/clients/WebsocketClient.js
+++ b/client-src/clients/WebsocketClient.js
@@ -8,7 +8,8 @@ const BaseClient = require('./BaseClient');
 module.exports = class WebsocketClient extends BaseClient {
   constructor(url) {
     super();
-    this.client = new WebSocket(url.replace(/^http/, 'ws'));
+    const wsUrl = url.replace(/^(?:http|chrome-extension|file)/i, 'ws');
+    this.client = new WebSocket(wsUrl);
 
     this.client.onerror = (err) => {
       log.error(err);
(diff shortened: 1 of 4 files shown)


client-src/clients/WebsocketClient.js
```diff
@@ -8,7 +8,8 @@ const BaseClient = require('./BaseClient');
 module.exports = class WebsocketClient extends BaseClient {
   constructor(url) {
     super();
-    this.client = new WebSocket(url.replace(/^http/, 'ws'));
+    const wsUrl = url.replace(/http|https|chrome-extension|file/gi, 'ws');
```
ylemkimon, line 11:
```suggestion
    const wsUrl = url.replace(/^(?:http|chrome-extension|file)/i, 'ws');
```

alexander-akait (approved):

PR #2992, opened 2021-02-05
chore: update webpack and webpack-cli

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Yes, updated tests.

### Motivation / Use-Case
update `webpack` and `webpack-cli` to lates version.

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/test/cli/cli.test.js
+++ b/test/cli/cli.test.js
@@ -66,7 +66,8 @@ runCLITest('CLI', () => {
   });
 
   webpack5Test('--hot webpack 5', (done) => {
-    testBin('--hot')
+    // host doesn't default to `localhost`
+    testBin('--hot --host localhost')
       .then((output) => {
         expect(output.exitCode).toEqual(0);
         expect(normalizeOutput(output.stderr)).toMatchSnapshot();
@@ -76,7 +77,8 @@ runCLITest('CLI', () => {
   });
 
   webpack5Test('--no-hot webpack 5', (done) => {
-    testBin('--no-hot')
+    // host doesn't default to `localhost`
+    testBin('--no-hot --host localhost')
       .then((output) => {
         expect(output.exitCode).toEqual(0);
         expect(normalizeOutput(output.stderr)).toMatchSnapshot();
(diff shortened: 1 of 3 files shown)


test/cli/cli.test.js
```diff
@@ -117,15 +119,6 @@ runCLITest('CLI', () => {
...
   });
 
-  it('unspecified host and port', (done) => {
-    testBin('')
-      .then((output) => {
-        expect(/http:\/\/localhost:[0-9]+/.test(output.stderr)).toEqual(true);
-        done();
-      })
-      .catch(done);
-  });
```
alexander-akait, line 127: We should not delete this tests, let's improve testing here

snitin315: Now we don't default host so we can change the assertion to false.

alexander-akait: host should be in IPv6 here

snitin315: I see `running at http:/` without specifying host and port -

Screenshot at 2021-02-06 10-31-20

alexander-akait: Bug :smile:

alexander-akait: We need to fix it, because it is very bad message

snitin315: Looking into it.

snitin315: I think we need to use `::` by default here.

alexander-akait: Ideally we should print message like `Project is running at https://server.com` or `Project is running at IPv4 (IPv6)`, it is in my roadmap, give me today to improve this

snitin315: Do we need an update here? 
https://github.com/webpack/webpack-dev-server/blob/b0868c399ee965727a97de303452b9841fd315e3/lib/Server.js#L592-L609

alexander-akait: Yes, we need improve it here

snitin315:
```js
if (typeof hostname === 'undefined') { 
  prettyHostName = ?? // what should be printed here?
}
```

alexander-akait: hm, I need to look at code, why hostname here undefined :confused:

alexander-akait (approved):
Let's merge as is, to unblock other deps updating and solve messages in the next PR

Issue #2969, opened 2021-01-19
Application with multiple iframes performances problems on hot reload


paztis:
- Operating System: MacOS BigSur
- Node Version: 14.15
- NPM Version: 6.14.8
- webpack Version: 4.44.2
- webpack-dev-server Version: 3.11.1
- Browser: chrome 88.0.4324.87

- [ ] This is a **bug**
- [x] This is a **modification** request

### Code

```js
// webpack.config.js
```

```js
// additional code, remove if not needed.
```

### Expected Behavior
I've an application that contains 2 sub iframes
Webpack generates 3 html files for the application
In dev mode, in case of code change, only the top level window might listne the events / try to refresh the page
or the main window need a way to advert the subframes to avoid listening
 

### Actual Behavior
in dev mode, all the frames + root window are listening the hot reload changes.
In case of code change, the frames + root window tries to reload the page in same time
Chrome devTools frequently crashes because of to many logs (3x the progress logs) + interrupted sourcemaps decoding

### For Bugs; How can we reproduce the behavior?
create an app with an iframe, both pointing to html pages managed by webpack

alexander-akait:
Do not ignore section how we can reproduce it, i.e. core and configuration, otherwise we can't help

paztis:
You want me to produce a full application to explain you the problem ?

> Expected, why do you avoid changes here?
In case of liveReload (page reload), the subframe is destroyed during the top window unload
Then a new subframe is created. 
Devtools encounter problems because it starts to decode sourcemap then destroy and load a new frame

I search a way to inform the subframes not to listen to socket events or at least to not refresh code.
But if I disable hot or liveReload, I do it also for top level window.

There's no way to provide different configs for each page produced by the same webpack build

alexander-akait:
> You want me to produce a full application to explain you the problem ?

Minimum or configuration

paztis:
original comment updated with webpack config and index files

alexander-akait:
You need to use `injectHot` as function and return false for `index-frame1` and `index-frame2`

paztis:
With injectHot and injectClient I only have the compile config as argument.

How do I do a distinction between chunk entry of root index and chuck
entries of frame index ? More it seems only called once.

Le mer. 20 janv. 2021 à 12:35, Alexander Akait <notifications@github.com> a
écrit :

paztis:
any other idea @alexander-akait ?

alexander-akait:
[...]

PR #2995, opened 2021-02-06
Add support of injectHot and injectClient on specific chunks

paztis:
- [ ] This is a **bugfix**
- [x] This is a **feature**

yes

### Motivation / Use-Case
In case of application with multiple iframes, only the top level frame has to listen socket for refresh
This PR add the support of chunk target in the injectClient and injectHot options. You are able to limit the client injection only in specific chunks   

Resolving of #2969

### Breaking Changes
none

### DIFF
--- a/lib/options.json
+++ b/lib/options.json
@@ -161,6 +161,13 @@
           "type": "boolean"
         },
         {
+          "type": "array",
+          "items": {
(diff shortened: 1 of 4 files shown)


paztis:
any news ?

PR #3011, opened 2021-02-13
fix: show detailed error in overlay

### LINKED ISSUE #2973: WP5: Warnings logged as objects in browser
I upgraded webpack to version 5 and now warnings and errors are logged broken as objects in browser console:
image

Before wordpress 5 upgrade it looked like this:
image

I could track this error down to [this location](https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L916) (which is logged in the client [here](https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/index.js#L116).
It seems that warnings from the stats object are now returned as object `{ file: string; message: string; }` instead of just strings like before.

I did not check if any plugin might be responsible for this and did not try a minimal sample yet.

Does any1 else experience the same issue?

snitin315:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

Fixes #2973 

### Breaking Changes
None

### DIFF
--- a/client-src/default/index.js
+++ b/client-src/default/index.js
@@ -125,7 +125,7 @@ const onSocketMessage = {
   },
   errors(errors) {
     log.error('Errors while compiling. Reload prevented.');
-    const strippedErrors = errors.map((error) => stripAnsi(error));
+    const strippedErrors = errors.map((error) => stripAnsi(error.message ? error.message : error));
     sendMessage('Errors', strippedErrors);
     for (let i = 0; i < strippedErrors.length; i++) {
       log.error(strippedErrors[i]);


alexander-akait (approved):

alexander-akait:
We should improve testing here, I will do it in future

olee:
@snitin315 would you mind releasing a new beta with this fix and potentially others which have been committed since beta-0 (which was released over 3 months ago already)?

snitin315:
I think we are mostly done and can publish a new release this week. 
/cc @alexander-akait

PR #3013, opened 2021-02-14
fix: show correct url in output status

snitin315:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

https://github.com/webpack/webpack-dev-server/pull/2992#discussion_r571360196

Screenshot at 2021-02-14 12-43-58

### Breaking Changes

None.

### Additional Info
No


lib/utils/normalizeOptions.js
```diff
@@ -73,6 +73,9 @@ function normalizeOptions(compiler, options) {
   options.liveReload =
     typeof options.liveReload !== 'undefined' ? options.liveReload : true;
 
+  options.host =
+    typeof options.host !== 'undefined' ? options.host : 'localhost';
```
rishabh3112, line 77:
```suggestion
    typeof options.host !== 'undefined' ? options.host : null;
```

rishabh3112 (changes requested):
Needs update in snapshots as well.

lib/utils/normalizeOptions.js
snitin315: `null` won't fix the issue here - https://github.com/webpack/webpack-dev-server/pull/2992#discussion_r571360196 . I'm still unsure about this approach

rishabh3112: Thanks for updating issue.
My suggestion was based on test here https://github.com/webpack/webpack-dev-server/blob/master/test/server/host-option.test.js

Ideally, https://github.com/webpack/webpack-dev-server/blob/master/test/server/host-option.test.js#L18 and https://github.com/webpack/webpack-dev-server/blob/master/test/server/host-option.test.js#L48 should have same behaviour. But that maybe a breaking change.

alexander-akait (dismissed):

alexander-akait:
Please fix lint problem

alexander-akait (approved):

snitin315:
I shouldn't be able to commit if there is a lint issue until I bypass with `--no-verify` 😕 .

snitin315:
I will fix ci tomorrow.

alexander-akait (commented):
Can you test locally? Using a lot of CI reduce available machines for other projects

snitin315:
Sorry about that, sometimes it passes locally but fails on CI. I will keep this in my mind and will use my fork's CI in the future in such a case. 

This one should be good now.

lib/Server.js
```diff
@@ -588,14 +588,18 @@ class Server {
...
     const prettyPrintUrl = (newHostname) =>
       url.format({ protocol, hostname: newHostname, port, pathname: '/' });
 
     let prettyHostname;
     let lanUrlForTerminal;
 
-    if (hostname === '0.0.0.0' || hostname === '::') {
+    if (
+      hostname === '0.0.0.0' ||
+      hostname === '127.0.0.1' ||
```
alexander-akait, line 600: In theory `127.0.0.1` can be not localhost (very rare edge case), also localhost can be not `127.0.0.1` (also very rare edge case). Why we added this?

snitin315: because without it we use `webpack serve --host localhost` it was printing `server running at http://127.0.0.1:8080` instead of `http://localhost:8080`. also failing some snapshots. I thought they should be the same.

Screenshot at 2021-02-15 19-03-26

alexander-akait: Here right solution https://github.com/indutny/node-ip/blob/master/lib/ip.js#L320 just use `ip.isLoopback()`

alexander-akait: Should be easy to fix :star:

snitin315: Like this ?

```suggestion
      ip.isLoopback(hostname) ||
```

alexander-akait: Yep, we also can remove `hostname === '::'` and maybe `0.0.0.0` need test it (manually as minimum)

alexander-akait (approved):

alexander-akait:
Thanks

test/cli/cli.test.js
```diff
@@ -6,42 +6,7 @@ const internalIp = require('internal-ip');
...
 
-// skip if webpack-dev-server is not linked
-let runCLITest = describe;
-let basePath;
-
-try {
-  basePath = path.join(require.resolve('webpack-dev-server'), '..', '..');
-} catch {
-  runCLITest = describe.skip;
-}
```
ylemkimon, line 17: This'd make tests fail where `webpack-dev-server` is not `npm link`ed.

snitin315: We link it on CI before running tests.

ylemkimon: But for developers testing locally, it may be a hurdle.

PR #3014, opened 2021-02-14
docs: add CLI options

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

update docs

### Breaking Changes
None

### Additional Info
None


rishabh3112 (changes requested):
Lets keep a single source of truth for flags. Either in CLI or in dev-server. I would prefer to have them in CLI as that is the case with webpack core as well.

snitin315:
ideally, we need to improve it on https://webpack.js.org, but these flags correspond to the `master` branch of `webpack-dev-server`. We can mention in `webpack-cli` or official docs after release only (many options removed since version `3.11.2`).

rishabh3112:
> these flags correspond to the `master` branch of `webpack-dev-server`. We can mention in `webpack-cli` or official docs after release only (many options removed since version `3.11.2`).

That is not the problem, we don't document the flags for `webpack` (i.e. bundle command) in main webpack repository, same should be with `webpack serve` as well. We should mention removed flags as `deprecated` in serve's documentation.

> ideally, we need to improve it on https://webpack.js.org,
Yup, but that should be done on release basis.

alexander-akait (approved):

PR #3041, opened 2021-02-22
feat: allow `Boolean` type for --firewall

snitin315:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yet to add some.

### Motivation / Use-Case

Allow `webpack serve --firewall` and `webpack serve --no-firewall`.

### Breaking Changes
None

### Additional Info
I will add some tests if this feature is accepted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,8 @@ Options:
   --history-api-fallback     Fallback to /index.html for Single Page Applications.
   --compress                 Enable gzip compression.
   --public <value>           The public hostname/ip address of the server.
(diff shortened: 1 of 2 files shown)


alexander-akait (approved):

alexander-akait:
Thanks

snitin315:
Should we allow a negative flag for each boolean? like we do for core flags.

alexander-akait:
I think yes, but only where it makes sense

PR #3050, opened 2021-02-25
feat: add more negative flags

snitin315:
- [ ] This is a **bugfix**
- [x] This is a **feature**

No

### Motivation / Use-Case

Add some more negative flags.

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/bin/cli-flags.js
+++ b/bin/cli-flags.js
@@ -30,11 +30,15 @@ module.exports = {
       name: 'https',
       type: Boolean,
       description: 'Use HTTPS protocol.',
+      negatedDescription: 'Do not use HTTPS protocol.',
+      negative: true,
     },
     {
       name: 'http2',
       type: Boolean,
       description: 'Use HTTP/2, must be used with HTTPS.',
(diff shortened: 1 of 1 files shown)


alexander-akait (approved):

alexander-akait:
Let's add tests

ylemkimon:
I think we can programmatically build the `negatedDescription`, for instance:
1. if the `description` starts with `Enable`, replace it with `Disable`
2. otherwise, prepend it with `Do not ` and change the first letter to the lowercase
3. if the ` or ` or comma exists, use the segment before it

snitin315:
It can introduce potential bugs as we have more than 100 negative flags (including flags from the core webpack) and the descriptions can change often. 

So we simply set a specific description for some flags if we need - 

https://github.com/webpack/webpack-cli/blob/3772148fda2b52d6498ad9d4f43e2dae0eb8f3a2/packages/webpack-cli/lib/webpack-cli.js#L219

snitin315:
recreating on master.

PR #3090, opened 2021-03-16
fix: port fallback logic in domain creator

anshumanv:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case
Fix logic

### Breaking Changes
No

### DIFF
--- a/lib/utils/createDomain.js
+++ b/lib/utils/createDomain.js
@@ -2,6 +2,7 @@
 
 const url = require('url');
 const ip = require('internal-ip');
+const defaultPort = require('../../lib/utils/defaultPort');
 
 function createDomain(options, server) {
   const protocol = options.https ? 'https' : 'http';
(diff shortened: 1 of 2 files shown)


lib/utils/createDomain.js
```diff
@@ -17,7 +18,7 @@ function createDomain(options, server) {
     hostname = '0.0.0.0';
   }
 
-  const port = server ? server.address().port : 0;
+  const port = server ? server.address().port : defaultPort;
```
ylemkimon, line 21: The intention was for client to use the `location` port: https://github.com/webpack/webpack-dev-server/blob/723e1ec394d2f1bdeda2e80dc8748b6da1c0a3a3/client-src/default/utils/createSocketUrl.js#L48-L50

anshumanv: hmm yes missed this, @alexander-akait  better alternate?

alexander-akait: oh, good catch, we need refactor logic here, it is more complex

alexander-akait:
Fixed in master

PR #3103, opened 2021-03-24
chore: make client overlay default to be true

rishabh3112:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes, will add

### Motivation / Use-Case
suggested by @alexander-akait 

### Breaking Changes
No

### Additional Info
No

### DIFF
--- a/lib/utils/normalizeOptions.js
+++ b/lib/utils/normalizeOptions.js
@@ -98,6 +98,11 @@ function normalizeOptions(compiler, options) {
     options.client = {};
   }
 
+  // Enable client overlay by default
+  if (typeof options.client.overlay === 'undefined') {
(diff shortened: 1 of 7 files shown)


snitin315:
Let's fix CI.

rishabh3112:
> Let's fix CI.

Any suggestions, I updated snapshots but it is now regressing on hash values, which will be different every time.

snitin315:
You can replace hash in a snapshot with `<hash>`, like we do for ports https://github.com/webpack/webpack-dev-server/blob/5f649010f96bb8cee551fa5c4cb53ce7ea740406/test/helpers/test-bin.js#L61

rishabh3112:
Okay, I will try

PR #3107, opened 2021-03-25
refactor: cli flags

snitin315:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case
Avoid additional negative flags.

### Breaking Changes
No

### DIFF
--- a/bin/cli-flags.js
+++ b/bin/cli-flags.js
@@ -78,7 +78,8 @@ module.exports = {
       type: Boolean,
       configs: [
         {
-          type: 'boolean',
+          type: 'enum',
+          values: [true],
         },
       ],
       description: 'Broadcasts the server via ZeroConf networking on start.',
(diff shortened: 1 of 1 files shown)


anshumanv (approved):

bin/cli-flags.js
```diff
@@ -88,7 +89,8 @@ module.exports = {
       type: Boolean,
```
anshumanv, line 89: I think we no longer need these types, lets remove

snitin315: I'm afraid we will lose compatibility with the webpack-cli version <= 4.5.0

anshumanv: yep lets keep it

alexander-akait:
Please don't merge PRs without my approve on webpack-dev-server, this packages is not stable, fir example, developer can set `bonjour: true` in config and use `--no-bonjour` to disable it, same for `open` and `client-progress`

snitin315:
Oh yes, my bad. sorry about that, I will be more careful in the future. Should I revert one this then?

alexander-akait:
Yes, feel free to send a PR

PR #3108, opened 2021-03-25
feat: enable overlay by default

snitin315:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case
Fixes #3103

### DIFF
--- a/lib/utils/normalizeOptions.js
+++ b/lib/utils/normalizeOptions.js
@@ -98,6 +98,11 @@ function normalizeOptions(compiler, options) {
     options.client = {};
   }
 
+  // Enable client overlay by default
+  if (typeof options.client.overlay === 'undefined') {
+    options.client.overlay = true;
+  }
+
   options.client.path = `/${
     options.client.path ? options.client.path.replace(/^\/|\/$/g, '') : 'ws'
   }`;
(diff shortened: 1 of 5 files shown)


alexander-akait (approved):

jordanbtucker:
Can this be promoted to BREAKING CHANGE?

This will require users to change their existing configuration to continue using the package in the same way they were using it before (without an overlay if not specified), therefore it's a BREAKING CHANGE.

Although in isolation, this shouldn't really cause any issues. If someone upgrades from v3 to v4, at best it could confuse them if they didn't have this turned on in v3. At worst, this could cause conflicts with third-party overlays in existing projects when they upgrade to v4, essentially *breaking* the development process. So just because a v3 config works with v4, it doesn't mean that this feature doesn't have repercussions when you change a default.

If this isn't specifically called out as a BREAKING CHANGE in the release notes, then it can be missed by developers upgrading from v3 to v4 and take a while to track down the cause (increasing the number of issues opened on this repo).

alexander-akait:
@jordanbtucker Yep, we just missing this in changelog, anyway we will provide full list of breaking changes and other notes for all options for stable v4 release, so you will do upgrade easy

PR #3112, opened 2021-03-26
feat: improve `host` option

snitin315:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

Improve `host` option by merging with `use-local-ip`. 

Now `host` option can accept - `local-ip`, `local-ipv4`, `local-ipv6`.

### Breaking Changes

`useLocalIp` option was removed.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -519,12 +519,6 @@ class Server {
     }
 
     this.server.on('error', (error) => {
-      if (error.syscall === 'getaddrinfo') {
-        this.logger.error(
-          `Address "${this.hostname}" is not available. Try with a different value for "host" option.\n ${error}`
-        );
-        process.exit(2);
-      }
       throw error;
     });
   }
@@ -731,11 +725,12 @@ class Server {
 
   listen(port, hostname, fn) {
     if (hostname === 'local-ip') {
-      this.hostname = internalIp.v4.sync() || internalIp.v6.sync();
+      this.hostname =
+        internalIp.v4.sync() || '0.0.0.0' || internalIp.v6.sync() || '::';
     } else if (hostname === 'local-ipv4') {
-      this.hostname = internalIp.v4.sync();
+      this.hostname = internalIp.v4.sync() || '0.0.0.0';
     } else if (hostname === 'local-ipv6') {
-      this.hostname = internalIp.v6.sync();
+      this.hostname = internalIp.v6.sync() || '::';
     } else {
       this.hostname = hostname;
     }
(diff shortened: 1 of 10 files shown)


lib/utils/createDomain.js
```diff
@@ -9,8 +9,12 @@ function createDomain(options, server) {
...
   let hostname;
 
-  if (options.useLocalIp) {
-    hostname = ip.v4.sync() || '0.0.0.0';
+  if (options.host === 'local-ip') {
+    hostname = ip.v4.sync() || ip.v6.sync();
+  } else if (options.host === 'local-ipv4') {
+    hostname = ip.v4.sync();
+  } else if (options.host === 'local-ipv6') {
+    hostname = ip.v6.sync();
```
alexander-akait, line 17: What about of `ipv6` is not available, what is the error? Maybe we need improve error message, need test, at least manually

alexander-akait: Same for ipv4

snitin315: It's like this, for now, WIP on improvements.

Screenshot 2021-03-26 at 7 44 31 PM

lib/Server.js
```diff
@@ -519,6 +519,11 @@ class Server {
     }
 
     this.server.on('error', (err) => {
+      if (err.code === 'EADDRNOTAVAIL') {
+        this.logger.error(
+          `Address ${this.hostname} is not available. Try with a different value for "host" option`
+        );
+      }
       throw err;
```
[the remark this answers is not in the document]
alexander-akait: Let's improve:
-  this.logger.error(`You fancy message.\n${error}`), never eat the original message
- let's add test, as minimum for CLI

alexander-akait (commented):
Need some improve

alexander-akait:
Also need rebase

alexander-akait:
Small improvements here, I was wrong about handling error, let's keep as is, because Node.js output the same message

lib/Server.js
```diff
@@ -724,7 +724,16 @@ class Server {
   }
 
   listen(port, hostname, fn) {
-    this.hostname = hostname;
+    if (hostname === 'local-ip') {
+      this.hostname =
+        internalIp.v4.sync() || '0.0.0.0' || internalIp.v6.sync() || '::';
```
ylemkimon, line 729: Isn't this a dead code, since '0.0.0.0' can be never `false`?

alexander-akait: Oh, yep, we need change it to `internalIp.v4.sync() || internalIp.v6.sync() || '0.0.0.0'`

anshumanv: fixed here https://github.com/webpack/webpack-dev-server/pull/3128

PR #3137, opened 2021-03-31
refactor: code

alexander-akait:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

union `open` and `openPage`.

### Breaking Changes

Yes, `openPage` was removed

### Additional Info

```ts
type OpenTarget = string | string[];
// i.e. `'google-chrome'` or `['google-chrome', '--incognito']`
type OpenApp = string | string[];
type OpenObject = { target?: OpenTarget, app?: OpenApp };
type Open = boolean | OpenTarget | OpenObject | (OpenTarget | OpenObject)[] ;
```


alexander-akait:
Also fixed cases when developer uses `['google-chrome', '--incognito']`

alexander-akait:
Hm, maybe we don't need `--open` anymore for CLI, maybe we can use `--open-target`, but let's keep for compatibility right now, we will think about this for the next major release

lib/Server.js
```diff
@@ -719,7 +719,7 @@ class Server {
     if (this.options.open || this.options.openPage) {
```
snitin315, line 719:
```suggestion
    if (this.options.open) {
```

alexander-akait:
I thought and decided that we will keep `--open` cli option for compatibility purpose, we will remove it in the next major release (we already do a lot of breaking changes and I think we will do it more), so this release will be painful for some developers, let's reduce it where possible

snitin315:
Makes sense 👍

PR #3154, opened 2021-04-05
chore: migrate error messages on descriptions

anshumanv:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

NA

### Motivation / Use-Case
Error messages are unused.

### Breaking Changes
No

### DIFF
--- a/lib/options.json
+++ b/lib/options.json
@@ -136,7 +136,8 @@
   },
   "properties": {
     "bonjour": {
-      "type": "boolean"
+      "type": "boolean",
+      "description": "Broadcasts the server via ZeroConf networking on start."
     },
     "client": {
       "type": "object",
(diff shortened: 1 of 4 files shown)


lib/options.json
```diff
@@ -205,13 +206,16 @@
...
     },
     "compress": {
-      "type": "boolean"
+      "type": "boolean",
+      "description": "https://webpack.js.org/configuration/dev-server/#devservercompress"
     },
     "dev": {
-      "type": "object"
+      "type": "object",
+      "description": "https://webpack.js.org/configuration/dev-server/#devserverdev-"
```
snitin315, line 218: I think we should use proper descriptions instead of links. Links are subject to change as we improve docs.

anshumanv: yeah we could but these show up in validation errors which can give more visibility of the error on the detailed docs page, I do have proper desc in older commit we can revert to that if we want /cc @alexander-akait

alexander-akait: We need keep description and link(s), only links is bad, especial for local usage, I don't want to open link for simple description

anshumanv: yep lets do it

anshumanv: done

alexander-akait:
I will finish, also we need to add `description` for all properties, includes nested, for example `client.overlay`

alexander-akait:
Thanks

PR #3159, opened 2021-04-07
fix: respect the `client.logging` option for HMR logging

### LINKED ISSUE #2932: Cannot mute `[HMR] Waiting for update signal from WDS...` in browser console

alexander-akait:
- [x] This is a **bugfix**
- [ ] This is a **feature**

fixes #2946 fixes #2932

### Motivation / Use-Case

Allow to disable extra logging from webpack

### Breaking Changes

No

### Additional Info

- I keep logging api, so other modules can change it in runtime
- fix broken `currentScript` parsing
- support `new URL()` and Node.js legacy format
- refactor
- more tests


alexander-akait:
Will finish tomorrow, need improve client for `file` protocol and test https with localhost

snitin315:
@alexander-akait  what is still in TODO here?

alexander-akait:
WIP on tests, broken only on Node.js 10 due bug in `url` module implementation

snitin315:
Node 10 is about to reach its end of life 😞

alexander-akait:
Yep, but I think webpack-dev-server v4 will have short time, we should refactor it to plugin, I'll show you how to do it

alexander-akait:
We will improve coverage after refactor `public`/`client.sock*` properties

PR #3163, opened 2021-04-08
fix script name

kakty3:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

Change script name to match it with `npm run` command

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ script as such:
 ```json
 {
   "scripts": {
-    "serve": "webpack serve"
+    "start:dev": "webpack serve"
   }
 }
 ```


snitin315:
Let's change the command to `npm run serve` instead.

alexander-akait:
No need, you can modify self this in own `package.json`

snitin315:
Actually, we have shown `serve` in package.json snippet and then running `npm run start:dev` in readme https://github.com/webpack/webpack-dev-server#with-npm-scripts

alexander-akait:
@snitin315 oh, yes, let's fix it, ideally we should write here two solutions:
1. Using `webpack serve` without commands in `package.json`
2. Custom name and run it

snitin315:
I will update soon 👍🏻

PR #3169, opened 2021-04-09
refactor: rename client properties

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**
- [x] This is a **code refactor**
- [x] This is a **test update**
- [ ] This is a **docs update**
- [ ] This is a **metadata update**

### For Bugs and Features; did you add new tests?
Yes updated existing ones

### Motivation / Use-Case
following changes for `client` option -
 
`host` to `webSocketHost`
`port` to `webSocketPort`
`path` to `webSocketPath`

### Breaking Changes
`client.host`, `client.port`, `client.path` were removed.

### Additional Info
None

PR #3181, opened 2021-04-13
refactor: union `public`/`client.host`/`client.port`/`client.path`

alexander-akait:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Shortly - union `public`/`client.host`/`client.port`/`client.path` into `client.webSocketUrl`

fixes #3169

### Motivation / Use-Case

Better DX, simplify logic

### Breaking Changes

Yes, options was removed, also you need put `firewall: ['domain.com']` when your proxy dev server (**always**)

### Additional Info

Possible values:
- `webSocketUrl: 'ws://0.0.0.0:0/ws'` - full specified URL, `0.0.0.0` for using `self.location.hostname`, `:0` for using `self.location.port`
- `webSocketUrl: 'wss://192.168.0.1:8082/custom'` - wss, custom hostname, custom port and custom pathname

Summary: 
- When you use `0.0.0.0`, `self.location.hostname` will be used
- When you use `:0` or don't specify it, `self.location.port` will be used
- Default `pathname` is `/ws`, we will improve supporting in future, if you don't `pathname` `/ws` will be used by default


lib/utils/DevServerPlugin.js
```diff
@@ -62,7 +39,9 @@ class DevServerPlugin {
     /** @type {string} */
     const clientEntry = `${require.resolve(
       '../../client/'
-    )}?${domain}${host}${path}${port}${logging}`;
+    )}?${domain}${logging}`;
+
+    console.log(domain);
```
snitin315, line 44:
```suggestion
```

lib/utils/createDomain.js
```diff
@@ -2,30 +2,24 @@
...
-
-  const port = server ? server.address().port : 0;
 
   // use explicitly defined public url
   // (prefix with protocol if not explicitly given)
-  if (options.public) {
-    return /^[a-zA-Z]+:\/\//.test(options.public)
-      ? `${options.public}`
-      : `${protocol}://${options.public}`;
+  if (options.client.socketUrl) {
```
snitin315, line 10:
```suggestion
  if (options.client && options.client.socketUrl) {
```

alexander-akait: We set `options.client` to `{}` in `normalizeOptions`

snitin315: All good then 👍🏻

snitin315:
Also, we need to remove `--public` in favor of `--client-socket-url`.

https://github.com/webpack/webpack-dev-server/blob/c4279c20f2cd6a02f60f7d1114e90808a46f8fda/bin/cli-flags.js#L234

alexander-akait:
Yep, still WIP, we need more things here

alexander-akait:
@snitin315 I think maybe we need to use `webSocketUrl` instead `socketUrl`, what do you think?

snitin315:
Yes, let's go with `webSocketUrl` 👍.

alexander-akait:
We need fix line https://github.com/webpack/webpack-cli/blob/master/packages/serve/src/startDevServer.ts#L77 in `webpack-cli`, we should not set port, because by default we use `0`, i.e. `self.location.port`, but we need to use this only for current version (this PR), we can check this on webpack-cli side like `const isModernClient = devServerCliOptions['client-web-socker-url']` (pseudo code) and do not set `client.port` if `isModernClient` is `true`

snitin315:
> We need fix line https://github.com/webpack/webpack-cli/blob/master/packages/serve/src/startDevServer.ts#L77 in `webpack-cli`, we should not set port now, because by default we use `0`, i.e. `self.location.port`, but we need to use this only for current version (this PR), we can check this on webpack-cli side like `const isModernClient = devServerCliOptions['client-web-socker-url']` (pseudo code) and do not set `client.port` if `isModernClient` is `true`

I will take care of it

lib/Server.js
```diff
@@ -937,19 +937,6 @@ class Server {
...
-      const idxPublic = this.options.public.indexOf(':');
-      const publicHostname =
-        idxPublic >= 0
-          ? this.options.public.substr(0, idxPublic)
-          : this.options.public;
-
-      if (hostname === publicHostname) {
-        return true;
-      }
-    }
```
alexander-akait, line 951: We have mixed logic for proxied server and socket url, it is weird, when you proxy dev server you need specify `host` in `firewall`

lib/servers/SockJSServer.js
```diff
@@ -50,7 +50,7 @@ module.exports = class SockJSServer extends BaseServer {
     });
 
     this.socket.installHandlers(this.server.server, {
-      prefix: this.server.options.client.path,
+      prefix: '/ws',
```
alexander-akait, line 53: Will be improved in future

lib/servers/WebsocketServer.js
```diff
@@ -12,7 +12,7 @@ module.exports = class WebsocketServer extends BaseServer {
 
     this.wsServer = new ws.Server({
       noServer: true,
-      path: this.server.options.client.path,
+      path: '/ws',
```
alexander-akait, line 15: Will be improved in future

test/server/clientOptions-option.test.js
```diff
@@ -43,33 +37,6 @@ describe('client option', () => {
...
-    });
-
-    it('sets the sock path correctly and strips leading and trailing /s', () => {
-      expect(server.options.client.path).toEqual(path);
-    });
-
-    it('responds with a 200 second', (done) => {
-      req.get(path).expect(200, done);
-    });
-  });
```
alexander-akait, line 71: unnecessary test, we already have tests in `e2e`

alexander-akait:
Maybe:
```ts
{
  client: {
    websocketTransport: 
      'ws' | 
      'sockjs' | 
      { 
        client: { type: 'ws' | string, options: Record<any, any> }, 
        server: { type: 'ws' | string, options: Record<any, any> } 
      }
  }
}
```

Allow to set options for websocket server:
```js
{
  client: { 
    websocketTransport: { 
      client: {  
        // type: 'ws', can be omitted, because default value is `ws`
        options: { socketUrl: 'ws://dev.site/custom' }
      },
      server: {  
        // type: 'ws', can be omitted, because default value is `ws`
        options: { path: '/custom' }
      }  
    }
  }
}
```

But multiple `client` is some misleading here... need think more, maybe anybody have idea.

alexander-akait:
Done here https://github.com/webpack/webpack-dev-server/pull/3309

PR #3202, opened 2021-04-19
feat: bonjour options

anshumanv:
- [ ] This is a **bugfix**
- [x] This is a **feature**

wip

### Motivation / Use-Case
Allows options from bonjour for more configuration - https://github.com/watson/bonjour#initializing

### Breaking Changes
no

### DIFF
--- a/lib/options.json
+++ b/lib/options.json
@@ -140,7 +140,14 @@
   },
   "properties": {
     "bonjour": {
-      "type": "boolean",
+      "anyOf": [
+        {
+          "type": "boolean"
+        },
+        {
+          "type": "object"
+        }
+      ],
       "description": "Broadcasts the server via ZeroConf networking on start. https://webpack.js.org/configuration/dev-server/#devserverbonjour"
     },
     "client": {
--- a/lib/utils/runBonjour.js
+++ b/lib/utils/runBonjour.js
@@ -9,7 +9,7 @@ function runBonjour(options) {
     port,
     type: https ? 'https' : 'http',
     subtypes: ['webpack'],
-    ...(typeof options.bonjour === 'object' && options.bonjour),
+    ...options.bonjour,
   });
 
   process.on('exit', () => {
--- a/lib/utils/runBonjour.js
+++ b/lib/utils/runBonjour.js
@@ -1,14 +1,15 @@
 'use strict';
 
-function runBonjour({ port }) {
+function runBonjour(options) {
   const bonjour = require('bonjour')();
   const os = require('os');
-
+  const { https, port } = options;
   bonjour.publish({
     name: `Webpack Dev Server ${os.hostname()}:${port}`,
     port,
-    type: 'http',
+    type: https ? 'https' : 'http',
     subtypes: ['webpack'],
+    ...(typeof options.bonjour === 'object' && options.bonjour),
   });
 
   process.on('exit', () => {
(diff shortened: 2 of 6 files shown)


anshumanv:
@alexander-akait tests for object bonjour config and use `https` when passed in `devServer` options, anything else?

lib/options.json
```diff
@@ -140,7 +140,14 @@
...
   "properties": {
     "bonjour": {
-      "type": "boolean",
+      "anyOf": [
+        {
+          "type": "boolean"
+        },
+        {
+          "type": "object"
+        }
```
snitin315, line 149: Need to add `"properties"` as well. https://www.npmjs.com/package/bonjour#var-service--bonjourpublishoptions

anshumanv: Could be problematic if they change in the package, we'll have to update manually and do release everytime

snitin315: I would prefer to keep it as `boolean` only and respect `options.https`

snitin315: It also seems this package doesn't go under changes anymore. Last publish 5 years ago.

anshumanv: we need to support all options

anshumanv: hm then we can have properties, lets do it

lib/options.json
```diff
@@ -145,7 +145,34 @@
...
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "txt": {
+              "type": "object"
+            }
+          },
+          "additionalProperties": false
```
alexander-akait, line 175: I still think we don't need this, new options can be added and we don't need do extra work here it is 3rd party code

anshumanv: https://github.com/webpack/webpack-dev-server/pull/3202#discussion_r616536873

anshumanv: hasn't done any changes since 5y

alexander-akait: Tomorrow all can be changed

anshumanv: lets avoid it then 😮

anshumanv: done

snitin315: yes, lets keep only `boolean` and set `https` when `options.https` else `http`

anshumanv: https://github.com/webpack/webpack-dev-server/pull/3202#discussion_r616536946 we need support all options

alexander-akait: `protocol (string, optional) - udp or tcp (default)` and `txt (object, optional) - a key/value object to broadcast as the TXT record` can be useful too in some cases

lib/utils/runBonjour.js
```diff
@@ -1,14 +1,15 @@
...
   const os = require('os');
-
+  const { https, port } = options;
   bonjour.publish({
     name: `Webpack Dev Server ${os.hostname()}:${port}`,
     port,
-    type: 'http',
+    type: https ? 'https' : 'http',
     subtypes: ['webpack'],
+    ...(typeof options.bonjour === 'object' && options.bonjour),
```
alexander-akait, line 12: No need extra check (`...undefined` works fine), just use `...options.bonjour`

anshumanv: done

alexander-akait (approved):

snitin315 (approved):

PR #3215, opened 2021-04-22
fix: overlay with warnings

alexander-akait:
- [x] This is a **bugfix**
- [ ] This is a **feature**

bug, existing

### Motivation / Use-Case

When we set `overlay: true` we mean `overlay: { warnings: true, errors: true }`, but current logic is not show warnings in overlay by defualt, also we have bug when we show warning and invalidate code warnings just hide

### Breaking Changes

In theory no, because `overlay: true` should show warnings and errors

### Additional Info

No

### DIFF
--- a/client-src/index.js
+++ b/client-src/index.js
@@ -21,9 +21,8 @@ const defaultOptions = {
   hotReload: true,
   liveReload: false,
   initial: true,
-  useWarningOverlay: false,
-  useErrorOverlay: false,
-  useProgress: false,
+  progress: false,
+  overlay: false,
 };
 const parsedResourceQuery = parseURL(__resourceQuery);
 const options = defaultOptions;
@@ -84,8 +83,8 @@ const onSocketMessage = {
   invalid() {
     log.info('App updated. Recompiling...');
 
-    // fixes #1042. overlay doesn't clear if errors are fixed but warnings remain.
-    if (options.useWarningOverlay || options.useErrorOverlay) {
+    // Fixes #1042. overlay doesn't clear if errors are fixed but warnings remain.
+    if (options.overlay) {
       overlay.clear();
     }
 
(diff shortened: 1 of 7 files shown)


lib/Server.js
```diff
@@ -986,17 +986,20 @@ class Server {
...
   }
 
-  // send stats to a socket or multiple sockets
+  // Send stats to a socket or multiple sockets
   sendStats(sockets, stats, force) {
     const shouldEmit =
       !force &&
       stats &&
       (!stats.errors || stats.errors.length === 0) &&
+      (!stats.warnings || stats.warnings.length === 0) &&
```
alexander-akait, line 995: Here we fix bug, when if we have previously warning and run invalidate without change code (for example using `API` to invalidate), overlay with warnings hid and did not show again

client-src/index.js
```diff
@@ -21,9 +21,8 @@ const defaultOptions = {
   hotReload: true,
   liveReload: false,
   initial: true,
-  useWarningOverlay: false,
-  useErrorOverlay: false,
-  useProgress: false,
+  progress: false,
+  overlay: false,
```
alexander-akait, line 25: Better keep options here as is,in future we can improve this and allow to using this `http://192.168.0.5?overlay&no-progress`

snitin315 (commented):
Let's update snapshot

jordanbtucker:
This is a breaking change, by the way. It doesn't really matter what `{overlay: true}` *should* have done. It has always meant `{overlay: {errors: true}}` according to the [documentation](https://webpack.js.org/configuration/dev-server/#devserveroverlay). If you're making a change to the public API, then it's a breaking change. So, if this is merged, it should be called out as a BREAKING CHANGE in the release notes.

alexander-akait:
@jordanbtucker no problems, docs will be updated together with stable release

snitin315 (approved):
LGTM

Birch-san:
to get the old behaviour (show only errors in overlay): am I understanding correctly that the new convention is supposed to be:

```js
overlay: {
  warnings: false,
  errors: true
}
```

from this PR, it looks like this object-style syntax is supposed to be supported. but it's rejected in schema validation:

```
configuration has an unknown property 'overlay'. These properties are valid:
object { bonjour?, client?, compress?, …
```

I've also tried `overlay: false`, but this isn't supported either.

As a result, my application is covered with a full-screen overlay warning me of my bundle size. this is not something I can trivially fix — I would rather hide the overlay, but don't know how.

I'm using `4.0.0-beta.3`.

Birch-san:
ah, I see; it's:

```js
client: {
  overlay:  {
    warnings: false,
    errors: true
  }
}
```

alexander-akait:
@Birch-san Yes, we will update docs in future, anyway why you disable warnings? What is the problem?

Birch-san:
disabled warnings because it's a warning about "chunk size is bigger than recommended". we know that 50MB is a big chunk, but we don't really have the option of making it smaller — we're limiting our app to 6 chunks to fit into browser's HTTP 1.1 concurrent request limit.

correct solution is to serve via HTTP/2 (so we can have as many chunks as we like), but requires us to get SSL certificates, so there's a few things we'd need to get right. we've accepted for now that it's easier to keep the big chunks.

out of interest: how bad is it to have big chunks? what kind of problems happen?

Birch-san:
> disable/change/filter this warning

thanks; I'm still happy to see the warning _in the webpack CLI_. only wanted it excluded from the full-screen overlay.

> splitting them is not simple story for some projects, need look at code

can't show code. will describe broad approach…

we currently chunk based on how long they're likely to live in the browser's cache (or whether they're conditionally-used features):

- libraries that change rarely
- libraries that change semi-frequently
- dynamically-imported libraries
- everything else

I think those categories are good (are they? 🤔 ), but if we used _more_ chunks… probably I'd use the same _categories_, but split each category into 250kB chunks?

it looks like maybe it wouldn't be impossible for us to enable HTTP/2 and use small chunks… are there any problems that can happen when you have **loads** of small chunks? or is this the recommended way to do chunking?

PR #3219, opened 2021-04-23
fix: double require problem

rishabh3112:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case
Fixing double require problem in bin where we check if `webpack-cli` is installed or not. But were doing a `require.resolve`, then installing and then again requiring `webpack-cli`. Later import will create error


alexander-akait:
Fixed by https://github.com/webpack/webpack-dev-server/pull/3400, anyway thanks for PR

PR #3228, opened 2021-04-26
feat: add `--client-need-client-entry` and `--client-need-hot-entry`

snitin315:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes.

### Motivation / Use-Case

Addded -

- `--client-need-client-entry`
-  `--no-client-need-client-entry`
- `--client-need-hot-entry`
- `--no-client-need-hot-entry`

### Breaking Changes
None

### Additional Info
No


bin/cli-flags.js
```diff
@@ -123,6 +123,41 @@ module.exports = {
         delete opts.clientOverlay;
       },
     },
+    {
+      name: 'client-need-client-entry',
```
snitin315, line 127: Maybe we need a better name, double `client` can be confusing.

alexander-akait: Yep, good catch I will think about it

snitin315: Maybe: 

```js
client: {
   appendDefaultEntries: true,
   appendHotEntries: true
}
```

alexander-akait: hm, give me time to think about these options, maybe we need `{ client: (entry) => { return clientOption; } }`, so you will setup hot/client entry for multiple entrypoints

evenstensberg (approved):

alexander-akait:
@snitin315 I found solution:
- let's rename `needHotEntry` -> `hotEntry: boolean` (`--client-hot-entry`), default value `const hotEntry = options.client.hotEntry || options.hot`
- let's remove `needClientEntry` in favor `client: boolean | ClientOptions | function(entryName): ClientOptions`, so you can disable `hotEntry` or whole client by entries using `function`

Maybe in future we will allow to write to only specific client (we already have this information internally, nobody ask this, so let's postpone)

snitin315:
Nice solution. I will update 👍

PR #3251, opened 2021-05-06
feat: implement dev server as plugin

rishabh3112:
- [x] This is a **bugfix**
- [x] This is a **feature**

### Motivation / Use-Case
The Dev Server should be implemented as a plugin for long run. This would simplify implementation of `webpack serve` as will give user more flexibility in multicompiler case.

The current implementation code present in `DevServerPlugin.js` implements few parts but still relies on `Server.js` to be applied.

I am currently implementing it separately at `WebpackDevServerPlugin.js`  to simplify testing. Later on we will export only this plugin instead on `Serve.js`.

### Breaking Changes
Yes, it would be a breaking change.

### Additional Info
@alexander-akait suggested this feature. Rest, we can chat on this PR.


alexander-akait (commented):
Logic should be in `apply` function here https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js, no need to duplicate this, all method already exists in server

lib/Server.js
```diff
@@ -39,18 +39,21 @@ if (!process.env.WEBPACK_SERVE) {
 }
 
 class Server {
-  constructor(compiler, options = {}) {
```
alexander-akait, line 42: We need keep compatibility here, just ignore compiler from constructor for apply

alexander-akait: Ideally after stable release v4, we will release 4.1/4.2 with plugin support, and for v5 we drop compiler from `constructor`

alexander-akait: Sounds good, we need do it before stable release, let's do it separately and fix in webpack-cli too, maybe adding hidden static option `_newConstructor`

rishabh3112: On it 👍🏽

rishabh3112:
## Status
Waiting merge of https://github.com/webpack/webpack-dev-server/pull/3273

lib/Server.js
```diff
@@ -39,24 +39,34 @@ if (!process.env.WEBPACK_SERVE) {
...
+      this.apply(options);
+    }
+  }
+
+  apply(compiler) {
+    this.compiler = compiler;
+    this.logger = this.compiler.getInfrastructureLogger(this.name);
+
     normalizeOptions(this.compiler, this.options);
     this.applyDevServerPlugin();
```
alexander-akait, line 71: Ideally I think we need inline this helper here, let's do it after https://github.com/webpack/webpack-dev-server/pull/3260/ and some refactoring

Anyway good job and thanks

rishabh3112: Yeah removing `DevServerPlugin` and perform ops directly on child compilers is next step to do in my list.

lib/Server.js
```diff
@@ -74,14 +84,218 @@ class Server {
...
+        portString = `&port=${options.webSocketServer.options.port}`;
+      } else {
+        portString = `&port=${options.port}`;
+      }
+
+      /** @type {string} */
+      let pathString = '';
+
+      // Only add the path if it is not default
+      if (options.client && options.client.path && options.client.path) {
```
snitin315, line 131:
```suggestion
      if (options.client && options.client.path) {
```

rishabh3112:
@alexander-akait what should I do next in this PR. I am done on my end 😅

alexander-akait (commented):
Can we avoid a lot of change, maybe better create a new PR

rishabh3112:
> Can we avoid a lot of change, maybe better create a new PR
Yup, will create a new PR for this now.

PR #3261, opened 2021-05-07
refactor: avoid new public options on class

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

rely on `this.options.host`

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -14,6 +14,8 @@ const routes = require('./utils/routes');
 const getSocketServerImplementation = require('./utils/getSocketServerImplementation');
 const getCompilerConfigArray = require('./utils/getCompilerConfigArray');
 const setupExitSignals = require('./utils/setupExitSignals');
+const getStatsOption = require('./utils/getStatsOption');
+const getColorsOption = require('./utils/getColorsOption');
 const schema = require('./options.json');
 
 if (!process.env.WEBPACK_SERVE) {
@@ -40,6 +42,7 @@ class Server {
     this.wsHeartbeatInterval = 30000;
 
     normalizeOptions(this.compiler, this.options);
+
     this.applyDevServerPlugin();
 
     this.SocketServerImplementation = getSocketServerImplementation(
@@ -629,8 +632,6 @@ class Server {
   }
 
   showStatus() {
-    const getColorsOption = require('./utils/getColorsOption');
-
     const useColor = getColorsOption(getCompilerConfigArray(this.compiler));
     const protocol = this.options.https ? 'https' : 'http';
     const { address, port } = this.server.address();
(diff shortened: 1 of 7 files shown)


snitin315:
I will update 👍

snitin315:
> const devServer = new DevServer({ host: '192.168.0.2' }); devServer.listen(80, '127.0.0.1'); and throw an error like we do with the port option

In the case of `port` we are just logging a warning, should we throw an error instead? 

https://github.com/webpack/webpack-dev-server/blob/297a64925cf375092df2b3d813c5f43cb6b6d4e0/lib/Server.js#L764-L768

alexander-akait:
Better warning here

lib/Server.js
```diff
@@ -744,36 +744,46 @@ class Server {
...
-  listen(port, hostname, fn) {
-    if (hostname === 'local-ip') {
-      this.hostname = internalIp.v4.sync() || internalIp.v6.sync() || '0.0.0.0';
-    } else if (hostname === 'local-ipv4') {
-      this.hostname = internalIp.v4.sync() || '0.0.0.0';
-    } else if (hostname === 'local-ipv6') {
-      this.hostname = internalIp.v6.sync() || '::';
-    } else {
-      this.hostname = hostname;
+  listen(port, hostname = this.options.host, fn) {
```
alexander-akait, line 753: Why we need default assign? It can be some misleading and we may use it by mistake like we do with port before, I prefer to use always  `this.options.host` and `this.options.port`

snitin315: > const devServer = new DevServer({ host: '192.168.0.2' }); devServer.listen(80); should work and listen 192.168.0.2 

Else this case will throw error too.

alexander-akait: I think we can refactor it to avoid using `hostname`, I mean we need simplify logic here, code is good

snitin315: Hmm, we can do `listen(port, fn)`

alexander-akait: We will do it late, just assign `hostname` to `this.options.host` after check

snitin315:
I will fix CI in near future

test/helpers/test-server.js
```diff
@@ -27,11 +27,12 @@ function startFullSetup(config, options, done) {
...
     port = 8080;
   }
-  const host = Object.prototype.hasOwnProperty.call(options, 'host')
+
+  options.host = Object.prototype.hasOwnProperty.call(options, 'host')
     ? options.host
     : 'localhost';
 
-  server.listen(port, host, (err) => {
+  server.listen(port, options.host, (err) => {
```
alexander-akait, line 35: hm, why we change it?

snitin315: I don't exactly recall why but some test cases were failing throwing `different hosts` error. 

We might not need it anymore, I will check if we still need this change.

snitin315: I see for the cases when `host` is `undefined` or not `specified` in config. We can directly pass `options.host` without defaulting to `localhost` above. I will push an update.

snitin315: Done

alexander-akait (commented):
I will test and review it in near future

alexander-akait:
@snitin315 need rebase too

snitin315:
Unfortunately I'm off my system, it's already night here. I will rebase in near future.

snitin315:
Done ✅

Issue #2758, opened 2020-10-06
HMR/Live Reloading broken after Webpack 5 rc.0 -> rc.1 update


wojtekmaj:
- Operating System: macOS Big Sur 
- Node Version: 14.5.0
- NPM Version: 6.14.5
- webpack Version: 5.0.0-rc.3
- webpack-dev-server Version: ^3.10.3
- Browser: Chrome

- [x] This is a **bug**
- [ ] This is a **modification** request

After upgrading from Webpack `5.0.0-beta.29` to `5.0.0-rc.3`, I've noticed that HMR/Live Reloading stopped working, even though no other changes to the code were made.

I investigated by updating it step by step and here's my outcome:

| Version | Works? |
| -- | -- |
| 5.0.0-beta.29 | Yes ✅  |
| 5.0.0-beta.30 | Yes ✅  |
| 5.0.0-beta.31 | Yes ✅  |
| 5.0.0-beta.32 | Yes ✅  |
| 5.0.0-beta.33 | Yes ✅  |
| 5.0.0-rc.0 | Yes ✅  |
| 5.0.0-rc.1 | No ❌  |
| 5.0.0-rc.2 | No ❌  |
| 5.0.0-rc.3 | No ❌  |

### Code

```js
// webpack.config.js
const webpack = require('webpack');
const path = require('path');

const HtmlWebpackPlugin = require('html-webpack-plugin');

const isProduction = process.env.NODE_ENV === 'production';
const isDevelopment = !isProduction;

const publicPath = process.env.PUBLIC_URL || '/';

module.exports = {
  mode: isProduction ? 'production' : 'development',
  bail: isProduction,
  entry: {
    src: [
      './src/index.jsx',
    ],
  },
  output: {
    path: path.resolve(__dirname, 'dist'),
    publicPath,
  },
  resolve: {
    extensions: ['.js', '.jsx'],
  },
  module: {
    rules: [
      {
        test: /\.jsx?$/,
        include: path.resolve(__dirname, 'src'),
        use: 'babel-loader',
      },
    ].filter(Boolean),
  },
  plugins: [
    new HtmlWebpackPlugin({
      template: './src/index.html'
    }),
  ].filter(Boolean),
  devServer: {
    historyApiFallback: true,
    hot: true,
    port: 3000,
  },
};
```

### Expected Behavior

HMR to work on Webpack version `5.0.0-rc.1` to `5.0.0-rc.3`.

[...]

PR #3271, opened 2021-05-11
Fix HMR/Live Reloading when running against webpack v5 and have an array of targets when using webpack-dev-server v3

greyepoxy:
- [X] This is a **bugfix**
- [ ] This is a **feature**

Updated the existing `addEntries` web targets test

### Motivation / Use-Case

Since v4 is not out yet (where this issue is fixed) this fixes the live reload issue https://github.com/webpack/webpack-dev-server/issues/2758 that occurs when a user is using webpack v5.

### Breaking Changes

None

### Additional Info

Tested against the examples and my own private webpack v5 config to confirm that it detected that the config was a web config.

Looking at the fix in the v4 branch I suspect this will not work as well as that change. Specifically if someone uses the new `browserlists` target option this change won't detect that as a web config. Since this change only needs to hold people over until v4 is released I believe its okay since in the example above the user could just add `web` to their target array.

### DIFF
--- a/lib/utils/addEntries.js
+++ b/lib/utils/addEntries.js
@@ -117,15 +117,24 @@ function addEntries(config, options, server) {
 
     // eslint-disable-next-line no-shadow
     [].concat(config).forEach((config) => {
+      // In webpack v5 target can be a string, an array of strings, null, or undefined
+      /** @type {Array<string>} */
+      const targets = Array.isArray(config.target)
+        ? config.target
+        : [config.target];
+
       /** @type {Boolean} */
-      const webTarget = [
-        'web',
-        'webworker',
-        'electron-renderer',
-        'node-webkit',
-        undefined, // eslint-disable-line
-        null,
-      ].includes(config.target);
+      const webTarget = targets.some((target) => {
+        return [
+          'web',
+          'webworker',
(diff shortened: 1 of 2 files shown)


snitin315:
You can use `v4.0.0-beta.3`, Just a couple of more PRs and we will do a stable release in the near future.

greyepoxy:
Yes I am aware this is fixed in v4 but due to that version still being in beta my company will not update (Tableau) due to the risk of issue. I was hoping to send this request a few weeks ago but only got permission to make the commit recently.

alexander-akait:
Make sense as patch release, I will try to do release in near future

greyepoxy:
@alexander-akait just checking that you are not waiting on me, thank you

alexander-akait:
Yep, I am here, I am working on `rc` release for v4, after this I will backport some fixes to v3 too

bartvandenende-wm:
it would be great indeed to see this change backported to 3.x for compatibility

getsetgopi:
Since my company still supports IE11. It fails when I use ```target: ["web", "es5"]```. I waiting for the fix where live reload and this works together.

alexander-akait:
A long time ago was fixed

PR #3273, opened 2021-05-12
refactor: change order of params for Server

rishabh3112:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Will update tests accordingly

### Motivation / Use-Case
We are moving to plugin interface for the server. The ideal signature for that would be `constructor(options={})` but we need compiler in params as well for backward compatibility. For that reason, this PR aims to change the order of params so that plugin implementation can ignore compiler param.

### Breaking Changes
Yes, we will need to fix this in serve package after patch release here

### Additional Info
https://github.com/webpack/webpack-dev-server/pull/3251/files#r629851714


anshumanv (dismissed):
Thanks!! 🌟

rishabh3112:
> Thanks!! :star2:

Hey! No need to thank as such but appreciated 😄.

lib/Server.js
```diff
@@ -39,7 +39,12 @@ if (!process.env.WEBPACK_SERVE) {
 }
 
 class Server {
-  constructor(compiler, options = {}) {
+  constructor(options = {}, compiler) {
+    // TODO: remove this after plugin support is published
+    if (options.hooks) {
```
alexander-akait, line 44: Let's rename `options.plugin`, but will not docs it, it is internal

rishabh3112: I didn't get what you want to say here 😅

alexander-akait: Oh I see, you just check `options.hooks` to if it is exists, it is compiler, otherwise options, right?

rishabh3112: Yup

alexander-akait: Good idea

alexander-akait (dismissed):

alexander-akait (commented):
Oh, can we test multi compiler mode too or we already have tests?

rishabh3112:
> Oh, can we test multi compiler mode too or we already have tests?

Nothing is changed in this PR for multi-compilers though, will add similar test for multi-compiler too anyway.

snitin315 (approved):
LGTM

alexander-akait (approved):

alexander-akait:
Thanks

alexander-akait:
Now we can continue development dev server as plugin and don't worry about breaking change for v4

PR #3299, opened 2021-05-15
refactor: imports

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No, Already present.

### Motivation / Use-Case

Reduce import at startup. 

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -171,6 +171,8 @@ class Server {
   }
 
   setupProxyFeature() {
+    const { createProxyMiddleware } = require('http-proxy-middleware');
+
     /**
      * Assume a proxy configuration specified as:
      * proxy: {
@@ -243,8 +245,6 @@ class Server {
       // It is possible to use the `bypass` method without a `target`.
       // However, the proxy middleware has no use in this case, and will fail to instantiate.
       if (proxyConfig.target) {
-        const { createProxyMiddleware } = require('http-proxy-middleware');
-
         return createProxyMiddleware(context, proxyConfig);
       }
     };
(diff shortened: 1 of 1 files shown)


lib/Server.js
```diff
@@ -246,6 +243,8 @@ class Server {
       // It is possible to use the `bypass` method without a `target`.
       // However, the proxy middleware has no use in this case, and will fail to instantiate.
       if (proxyConfig.target) {
+        const { createProxyMiddleware } = require('http-proxy-middleware');
```
alexander-akait, line 246: Move it here https://github.com/webpack/webpack-dev-server/blob/d8ffe3927b235e542eae23bab6e9dfdbea8e5bf0/lib/Server.js#L173 better we still will use it due https://github.com/webpack/webpack-dev-server/blob/d8ffe3927b235e542eae23bab6e9dfdbea8e5bf0/lib/Server.js#L421

snitin315: On it.

lib/Server.js
```diff
@@ -980,6 +985,9 @@ class Server {
   }
 
   serveMagicHtml(req, res, next) {
+    const getFilenameFromUrl =
+      require('webpack-dev-middleware/dist/utils/getFilenameFromUrl').default;
```
alexander-akait, line 989: I think we should add this method to public `webpack-dev-middleware` API here https://github.com/webpack/webpack-dev-middleware/blob/master/src/index.js#L78, in future we should avoid import non public API

snitin315: Makes sense, will do 👍🏻

anshumanv (approved):
👍

lib/Server.js
```diff
@@ -2,36 +2,18 @@
...
 const path = require('path');
 const url = require('url');
-const http = require('http');
-const https = require('https');
 const fs = require('graceful-fs');
 const ipaddr = require('ipaddr.js');
 const internalIp = require('internal-ip');
 const killable = require('killable');
-const chokidar = require('chokidar');
 const express = require('express');
```
alexander-akait, line 9: I think in future we reduce them more, good right now

alexander-akait (approved):

lib/Server.js
```diff
@@ -546,6 +544,9 @@ class Server {
   }
 
   createServer() {
+    const https = require('https');
+    const http = require('http');
```
ylemkimon, line 548: I'm not sure it'd make a difference, but I think we can move these below just like `require('spdy')`.

alexander-akait: Feel free to send PR

PR #3300, opened 2021-05-16
docs: add `--web-socket-server`

snitin315:
- [ ] This is a **bugfix**
- [x] This is a **feature**

yes

### Motivation / Use-Case

Add `--web-socket-server`.

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/examples/cli/web-server-socket/README.md
+++ b/examples/cli/web-server-socket/README.md
@@ -1,0 +1,24 @@
+# CLI: web-socket-server
+
+To create a custom server implementation.
+
+## sockjs
+
+This mode uses [SockJS-node](https://github.com/sockjs/sockjs-node) as a server.
+
+```console
+npx webpack serve --web-socket-server sockjs --open-target
+```
+
+## ws
+
+This mode uses [ws](https://github.com/websockets/ws) as a server.
+
+```console
+npx webpack serve --web-socket-server ws --open-target
+```
+
+### What Should Happen
+
+1. The script should open `http://localhost:8080/` in your default browser.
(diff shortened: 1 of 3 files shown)


alexander-akait (dismissed):

anshumanv (commented):
Can add some examples

alexander-akait:
Yep, or fix exiting example(s)

snitin315:
I will update 👍🏻

snitin315:
Done ✅

alexander-akait:
Let's improve this, ideally we should use to get all arguments, but because we supported webpack v4, we need specify them, let's add test for this, fix schema and add necessary args:

```
const schema = require('./lib/options');

console.log(webpack.cli.getArguments(schema))
```

alexander-akait:
Here fix for schema https://github.com/webpack/webpack-dev-server/pull/3315, feel free to merge when it will be green and improve CLI args (should be easy)

snitin315:
I think we can introduce the following in public API -

- `Server.getCliArguments()` - return all cli-flgs, 
- `Server.processArgs(args, config, values)` - process args via `webpack.cli.processArguments(args, config, values)` .

Then use these in `webpack-cli`, and we can get rid of `cli-flags.js` in v5.

alexander-akait:
@snitin315 Yes, this is the right direction, but we still support v4, so we don't have this API, we will implement it in the next major release, now I think we can copy paste them in our flags and add tests to ensure we nothing missing

alexander-akait (approved):

PR #3302, opened 2021-05-17
feat: add --client & --no-client

snitin315:
- [ ] This is a **bugfix**
- [x] This is a **feature**

### Motivation / Use-Case

`client: boolean | ClientOptions | function(entryName): ClientOptions`

Add `--client` and `--no-client`

### Breaking Changes

`client.needClientEntry` was removed in favor of `client: boolean | ClientOptions | function(entryName): ClientOptions`

### Additional Info
None


alexander-akait:
Let's close, we remove these options in favor manual setup also tests were added, `function(entryName): ClientOptions` is workaround in most cases, better to fix problem instead add workarounds, anyway for complex setup you can use manual setup as workaround

PR #3303, opened 2021-05-17
feat: enable `compress` by default

snitin315:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case
Enable `compress` by default

### Breaking Changes

Maybe

### Additional Info
No


alexander-akait:
hm, do you think it should be `true` by default?

snitin315:
Yes, I also see it in TODO https://github.com/webpack/webpack-dev-server/projects/1#card-19879364

alexander-akait (dismissed):

anshumanv (commented):
need update docs

alexander-akait (approved):

examples/cli/compression/README.md
```diff
@@ -3,19 +3,38 @@
...
 2. Files being sent to the browser from the `webpack` bundle should be gzipped.
 3. Open the console in your browser's devtools and select the _Network_ tab.
 4. Find `bundle.js`. The response headers should contain `Content-Encoding: gzip`.
 
+## `--no-compress`
+
+To run this example, run this command in your console or terminal:
+
+```console
+npm run webpack-dev-server -- --open --no-compress
```
anshumanv, line 28: we run webpack serve right?

snitin315: Actually it is script https://github.com/webpack/webpack-dev-server/blob/88468b67ed3f66470acc7816a25e49e07a0fadbf/package.json#L32

anshumanv (approved):
nicenicenice

examples/cli/compression/README.md
anshumanv: looks good then

alexander-akait: Let's change it to `npx webpack serve` in future PRs

anshumanv: will take care

Den-dp:
@snitin315 I thought that reading (and writing) the content served in the memory - is the fastest thing that can be possible.

And adding compression for the in-memory server will only lead to more CPU usage, and may slow down a feedback loop for the developer (longer refresh time).

In theory, after this change, we will pay more CPU load for in-memory content compression, and un-compression (by Express and the browser). 🤔

alexander-akait:
@Den-dp yes, but in other case HMR and live reloading will be faster because you will get compressed sources, we collect feedback and maybe revisit it, but I think it is better to have fast development env

Den-dp:
@alexander-akait looks like @sokra also hit a similar problem with enabling compression by default https://github.com/webpack/webpack/pull/13962

alexander-akait:
it is another problem

PR #3309, opened 2021-05-17
feat: add webSocketURL option

anshumanv:
- [x] This is a **bugfix**
- [ ] This is a **feature**

migrated existing

### Motivation / Use-Case
Move out client.host/port/path to a new option `webSocketURL`

### Breaking Changes
Yes


lib/options.json
```diff
@@ -285,6 +263,40 @@
...
+              "anyOf": [
+                {
+                  "type": "number"
+                },
+                {
+                  "type": "string"
+                },
+                {
+                  "enum": ["auto"]
+                }
```
snitin315, line 292:
```suggestion
```

lib/utils/DevServerPlugin.js
```diff
@@ -35,34 +35,46 @@ class DevServerPlugin {
...
-    }
+    let webSocketURL = '';
+
+    if (typeof options.webSocketURL === 'string') {
+      webSocketURL = options.webSocketURL;
+    } else if (typeof options.webSocketURL === 'object') {
+      /** @type {string} */
+      let host = '';
+
+      if (options.webSocketURL && options.webSocketURL.host) {
```
snitin315, line 46: Already check above for object

```suggestion
      if (options.webSocketURL.host) {
```

lib/utils/DevServerPlugin.js
```diff
@@ -35,34 +35,46 @@ class DevServerPlugin {
...
+      let port = '';
 
-    if (options.client && options.client.port) {
-      port = `&port=${options.client.port}`;
-    } else if (options.webSocketServer.options.port && !isSockJSType) {
-      port = `&port=${options.webSocketServer.options.port}`;
-    } else {
-      port = `&port=${options.port}`;
-    }
+      if (options.webSocketURL && options.webSocketURL.port) {
```
snitin315, line 55:
```suggestion
      if (options.webSocketURL.port) {
```

lib/utils/DevServerPlugin.js
```diff
@@ -35,34 +35,46 @@ class DevServerPlugin {
...
-    /** @type {string} */
-    let path = '';
+      /** @type {string} */
+      let path = '';
+
+      // Only add the path if it is not default
+      if (
+        options.webSocketURL &&
+        options.webSocketURL.path &&
+        options.webSocketURL.path
```
snitin315, line 70:
```suggestion
        options.webSocketURL.path
```

lib/utils/DevServerPlugin.js
```diff
@@ -35,34 +35,46 @@ class DevServerPlugin {
...
+      ) {
+        path = `&path=${options.client.path}`;
```
snitin315, line 72:
```suggestion
        path = `&path=${options.webSocketURL.path}`;
```

anshumanv:
no need review drafts, let me finish

lib/options.json
anshumanv: why?

snitin315: As per the thread in slack. 

> webSocketURL: STRING | { host: string, port: number, path: string }

lib/options.json
```diff
@@ -306,6 +284,34 @@
       "description": "Specifies client properties. https://webpack.js.org/configuration/dev-server/#devserverclient",
       "additionalProperties": false
     },
+    "webSocketURL": {
```
alexander-akait, line 287: It should be in the `client` options, because this option only for client

lib/utils/DevServerPlugin.js
```diff
@@ -35,36 +35,44 @@ class DevServerPlugin {
     // TODO show warning about this
     const isSockJSType = options.webSocketServer.type === 'sockjs';
 
-    /** @type {string} */
-    let host = '';
+    let webSocketURL = '';
+
+    if (typeof options.webSocketURL === 'string') {
+      webSocketURL = options.webSocketURL;
```
alexander-akait, line 41: We need normalize `webSocketURL`, i.e. parse and get `host`/`port`/`path`

alexander-akait: Same for below code

test/server/__snapshots__/Server.test.js.snap.webpack5
```diff
@@ -45,8 +43,7 @@ Array [
   Array [
     "client",
     "index.js?http:",
-    "0.0.0.0&path=",
-    "ws&port=8100",
+    "0.0.0.0",
   ],
   Array [
```
alexander-akait, line 48: Existing snapshots should not be changed, it is big breaking change

anshumanv: oh yep need normalize

lib/utils/DevServerPlugin.js
anshumanv: fixed

lib/options.json
anshumanv: yep maybe I was confused before, fixed now 👍

anshumanv:
@alexander-akait @snitin315 CI should be green now 😎

lib/options.json
```diff
@@ -301,6 +279,34 @@
...
           "description": "Tells devServer to inject a Hot Module Replacement entry."
+        },
+        "webSocketURL": {
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "object",
+              "properties": {
```
snitin315, line 290:
```suggestion
              "type": "object",
              "additionalProperties": false,
              "properties": {
```

lib/options.json
```diff
@@ -301,6 +279,34 @@
...
+                  "type": "string",
+                  "description": "Tells clients connected to devServer to use the provided path to connect."
+                },
+                "port": {
+                  "anyOf": [
+                    {
+                      "type": "number"
+                    }
+                  ],
+                  "description": "Tells clients connected to devServer to use the provided path to connect."
```
snitin315, line 305:
```suggestion
                  "description": "Tells clients connected to devServer to use the provided port."
```

lib/utils/DevServerPlugin.js
```diff
@@ -35,36 +35,44 @@ class DevServerPlugin {
     // TODO show warning about this
     const isSockJSType = options.webSocketServer.type === 'sockjs';
 
-    /** @type {string} */
-    let host = '';
+    let webSocketURL = '';
+
+    if (typeof options.client.webSocketURL === 'string') {
+      webSocketURL = options.client.webSocketURL;
```
snitin315, line 41: So user input should be like this in case of string `&host=0.0.0.0&port=8080&path=foo`?

anshumanv: Yeah I think that's the only possible way since it's resolving a client path before it, not very convenient for user I agree

/cc @alexander-akait should confirm

lib/options.json
anshumanv: yep lets improve it

lib/utils/DevServerPlugin.js
anshumanv: It doesn't make sense actually, most likely user will pass an absolute URL and we can use it directly

snitin315: It should work like `public`. And then we remove options.public in favor of `webSocketURL`

alexander-akait: > It should work like public. And then we remove options.public in favor of webSocketURL

Yes, right direction

anshumanv: yep lets update

alexander-akait:
@anshumanv Do you need help? I can finish, it is blocker for the next `beta` release

anshumanv:
@alexander-akait should be green now, feel free to make any changes or lmk I can do it, thanks for help 👍

snitin315 (commented):
Need to remove `public` option.

test/e2e/ClientOptions.test.js
```diff
@@ -38,7 +38,7 @@ for (const webSocketServerType of webSocketServerTypes) {
     });
   };
 
-  describe(`should work behind proxy, when hostnames are same and ports are different (${webSocketServerType})`, () => {
+  describe.only(`should work behind proxy, when hostnames are same and ports are different (${webSocketServerType})`, () => {
```
bompus, line 41: don't forget to remove the .only :)

anshumanv: oh yep, good catch!

alexander-akait:
Let's finish :+1:

alexander-akait:
Finished... We still have some edge cases and places where we can simplify code, but I will do it in the next PRs

snitin315:
Also need to remove `--public` from cli-flags.

alexander-akait:
Oh, yes, forgot about it

alexander-akait:
I will remove this option in this PR and let's add all options here https://github.com/webpack/webpack-dev-server/pull/3325

alexander-akait:
All our tests are complete horror, we need to rewrite them, I spent half a day to finish them

alexander-akait:
@snitin315 @anshumanv Can you review?

test/server/utils/normalizeOptions.test.js
```diff
@@ -85,7 +87,9 @@ describe('normalizeOptions', () => {
       multiCompiler: false,
       options: {
         client: {
-          path: '/custom/path/',
+          webSocketURL: {
+            path: '/custom/path/',
+          },
```
snitin315, line 92: We should add one case with `webSocketURL: 'ws://myhost.com:8080/path'`, and check it normalizes host, port and path correctly from the string.

lib/utils/normalizeOptions.js
```diff
@@ -104,6 +104,18 @@ function normalizeOptions(compiler, options, logger) {
...
   }
 
+  if (typeof options.client.webSocketURL === 'undefined') {
+    options.client.webSocketURL = {};
+  } else if (typeof options.client.webSocketURL === 'string') {
+    const parsedURL = new URL(options.client.webSocketURL);
+
+    options.client.webSocketURL = {
+      host: parsedURL.hostname,
+      port: parsedURL.port,
```
snitin315, line 114: Hmm, I think port here will be type `string`,  schema have only type `number`

test/server/utils/normalizeOptions.test.js
bompus: Will the webSocketURL string allow me to provide wss:// and actually have the browser request using wss://, or will the normalization turn it back into ws:// if https is false? I'm running a reverse https -> http proxy in front, so the dev server should be https:false, but currently doing so, the websocket logic attempts to use ws:// even though the address bar is an https:// url. Just checking to see if this string webSocketURL option will indeed allow me to now force wss:// to be used for the websocket requests.

alexander-akait: @bompus you can use `ws` and `wss`, and even `http`/`https` (we normalize them to `ws`/`wss` for backward compatibility). But you should use `webSocketURL` only when you proxy dev server, if you set `https: true` we will use `wss` by default

alexander-akait: @snitin315 we have this test (in `test/e2e/web-socket-server-and-url.test.js`):
```
client: {
  webSocketURL: `ws://myhost.test:${port2}/foo/test/bar/`,
},
`

File: lib/utils/normalizeOptions.js (hunk 14 above)
Comment (alexander-akait) [reply lib/utils/normalizeOptions.js:114]: Yes, I think about it, here two solution:
- allow to set port as `string` everywhere
- do not allow use `port` as `string` everywhere

Let's fix it in other PR, I found we allow to set `host` to `null`, it is wrong, also we need allow to set port as `string` for websocket server too

File: test/server/utils/normalizeOptions.test.js (hunk 13 above)
Comment (snitin315) [reply test/server/utils/normalizeOptions.test.js:92]: All good then 👍🏻

File: lib/utils/normalizeOptions.js (hunk 14 above)
Comment (snitin315) [reply lib/utils/normalizeOptions.js:114]: > allow to set port as string everywhere

+1, I can send a PR for the same 👍🏻

**Review by snitin315 — APPROVED** [review]

**Review by anshumanv — COMMENTED** [review]
looks good 👍

File: test/server/utils/normalizeOptions.test.js (hunk 13 above)
Comment (bompus) [reply test/server/utils/normalizeOptions.test.js:92]: @alexander-akait I tried this on master after it was merged and browser is still told to request ws:// instead of wss:// , even when wss:// is specified on webSocketURL. Am I missing something in the config, or is this as expected?

My config is:
```
devServer: {
      https: false,
      host: '127.0.0.1',
      port: 8888,
      firewall: false,
      client: {
        webSocketURL: 'wss://myhost.test/ws/'
      }
}
```

alexander-akait: hm, why you need to change `protocol`?

alexander-akait: webpack dev server without https and proxy server under https?

bompus: The dev server is proxied via a protected cloudflared argo tunnel (cloudflare https -> nginx http -> webpack-dev-server http), so the local dev server needs to listen for http instead of https. This is done so dev matches production closer to prevent surprises.

This works in a previous version of v4 (pre-beta3?) and also in v3.

It would be useful if you just detected the window.location.protocol and used that as the default for ws:// or wss://, which is probably what was done previously. With the recent changes, it seems that everything is being toggled off that https:[true/false] option.

I'm running https:true for now to resolve the issue, but it'd be nice to go back to the old config so the reverse proxy matches production and hits all the same cloudflare https -> nginx http logic.

alexander-akait: @bompus fair enough, my idea - adding `protocol` option to `webSocketURL` (you can use it in `string` notation too), also provide `protocol: 'auto'` to use from `location`, so passing `auto://0.0.0.0:0/` will means all from `location`, what do you think?

bompus: @alexander-akait that sounds good to me, as I had already expected it to use `ws` or `wss` based on what had been specified in webSocketURL. `auto` works as well, and I assume one would need to use `auto://0.0.0.0:0/wspath/` if they wanted a specific path?

alexander-akait: >  I assume one would need to use auto://0.0.0.0:0/wspath/ if they wanted a specific path?

Yes, maybe make sense to add `auto` to path too

bompus: Great. This should solve my use case 👍

alexander-akait: Put in todo, will solve tomorrow

bompus: @alexander-akait did you get a chance to start on this yet, or still pending? Just want to make sure it doesn't get buried.

alexander-akait: in my todo on tomorrow, don't worry :+1:

PR #3313, opened 2021-05-18
fix: imporove processing of CLI flags

### LINKED ISSUE #3306: --https- namespaced CLI commands don't seem to work. --https-key --https-cert .  ] TypeError: Cannot create property 'cert' on boolean 'true'
- Operating System: OS X
- Node Version: 14.15.3
- NPM Version: 6.14.9
- webpack Version: 5
- webpack-dev-server Version: 4.0.0.beta.3
- Browser: Brave

- [x] This is a **bug**

### Code

Running webpack serve via the CLI with https cert options seems to error

```
webpack serve --https --https-cert ./.ssl/cert.pem --https-key ./.ssl/key.pem",
```
The above will error with:
```
[webpack-cli] TypeError: Cannot create property 'cert' on boolean 'true'
```

I tried omitting the `--https` as I wasn't sure if that was not necessary due to the namespced key and cert options being passed.  That gets past that previous point of error in the compilation but then errors 

```
[webpack-cli] Invalid configuration object. Object has been initialized using a configuration object that does not match the API schema.
 - configuration has an unknown property 'stats'. These properties are valid:
   object { bonjour?, client?, compress?, devMiddleware?, firewall?, headers?, historyApiFallback?, host?, hot?, http2?, https?, liveReload?, onAfterSetupMiddleware?, onBeforeSetupMiddleware?, onListening?, open?, port?, proxy?, public?, setupExitSignals?, static?, transportMode?, watchFiles? }
```

### Expected Behavior

Be able to pass `--https --https-key ../path --https-cert ../path` options

### Actual Behavior

Errors

snitin315:
- [X] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

Fix #3306 

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/test/cli/cli.test.js
+++ b/test/cli/cli.test.js
@@ -504,6 +504,15 @@ describe('CLI', () => {
       .catch(done);
   });
 
+  it(' --open --open-target index.html', (done) => {
+    testBin('--open --open-target index.html')
+      .then((output) => {
+        expect(output.exitCode).toEqual(0);
+        done();
+      })
+      .catch(done);
+  });
+
   it('--open-target /first.html second.html', (done) => {
     testBin('--open-target /first.html second.html')
       .then((output) => {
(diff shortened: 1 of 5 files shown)


alexander-akait (commented):
I think we need to write common function which allow to use, not only https, also `open`, `--open --open-target` and other cli options

test/cli/cli.test.js
```diff
@@ -209,6 +209,26 @@ describe('CLI', () => {
...
+      .then((output) => {
+        expect(output.exitCode).toEqual(0);
+        expect(
+          normalizeStderr(output.stderr, { ipv6: true, https: true })
+        ).toMatchSnapshot();
+        done();
+      })
+      .catch(done);
+  });
+
```
alexander-akait, line 231: Can we add `--open --open-target` test here too?

alexander-akait (approved):
Feel free to merge when CI will be green

PR #3330, opened 2021-05-23
refactor: client

anshumanv:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

existing

### Motivation / Use-Case

Closing stale PR https://github.com/webpack/webpack-dev-server/pull/2431

### Breaking Changes

no

### Additional Info

Close https://github.com/webpack/webpack-dev-server/pull/2431

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -600,10 +600,8 @@ class Server {
       this.sockets.push(connection);
 
       this.socketServer.onConnectionClose(connection, () => {
-        const idx = this.sockets.indexOf(connection);
-
-        if (idx >= 0) {
-          this.sockets.splice(idx, 1);
+        if (this.sockets.includes(connection)) {
+          this.sockets = this.sockets.filter((socket) => socket !== connection);
         }
       });
 


snitin315 (approved):

anshumanv:
There's one more instance https://github.com/webpack/webpack-dev-server/blob/84e1d3426684849caca2104f834e4c83d86b1e17/lib/Server.js#L1018

But I feel it's better as is

client-src/index.js
```diff
@@ -66,7 +66,7 @@ self.addEventListener('beforeunload', () => {
 if (typeof window !== 'undefined') {
   const qs = window.location.search.toLowerCase();
 
-  options.hotReload = qs.indexOf('hotreload=false') === -1;
+  options.hotReload = !qs.includes('hotreload=false');
```
alexander-akait, line 69: Don't no use modern function in client due some developers still need IE 11 support and other old browsers https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

anshumanv: but we transpile them using babel right?

alexander-akait: Yes, but it is increase our bundle, ideally we should keep this small as possible

anshumanv: gotcha lets change this then

anshumanv: startsWith will also come after transpile, remove it too? I think it won't affect the bundle that much as polyfill will be really small..

alexander-akait: Yes, they are small, but why do this if we can use alternatives and not increase the size at all

anshumanv: makes sense, should I leave change in `Server.js` or close?

alexander-akait:
Done in master

PR #3344, opened 2021-05-25
chore: fix sec vuln

anshumanv:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

NA

### Motivation / Use-Case
Fixes security vuln in next

### Breaking Changes
Shouldn't, bonjour was downgraded to secure version

### Additional Info
We may need to patch for v3 too, will do after confirmation

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "license": "MIT",
       "dependencies": {
         "ansi-html": "^0.0.7",
-        "bonjour": "^3.5.0",
+        "bonjour": "^3.3.0",
         "chokidar": "^3.5.1",
         "compression": "^1.7.4",
         "connect-history-api-fallback": "^1.6.0",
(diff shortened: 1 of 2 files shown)


ylemkimon:
Note `^3.3.0` would be vulnerable to the same bug, as the `dns-packet` was lifted out of `multicast-dns` and uses the same implementation.

anshumanv:
yep will send for v3 as well

ylemkimon:
@anshumanv I mean `bonjour@^3.3.0` by `^3.3.0`. Downgrading the dependency wouldn't help.

ylemkimon:
@anshumanv Advisories are not published, but as it uses the same implementation, it'd be vulnerable to the same bug. I've contacted the maintainer.

anshumanv:
Gotcha, then this won't solve it, thanks for heads up! Closing :)

PR #3354, opened 2021-05-27
feat: allow string value for `port`

snitin315:
- [ ] This is a **bugfix**
- [X] This is a **feature**

### Motivation / Use-Case

Allow string for `client.webSocketURL.port`

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/lib/utils/normalizeOptions.js
+++ b/lib/utils/normalizeOptions.js
@@ -103,8 +103,10 @@ function normalizeOptions(compiler, options, logger) {
       },
     };
 
-    if (typeof options.webSocketServer.port === 'string') {
-      options.webSocketServer.port = Number(options.webSocketServer.port);
+    if (typeof options.webSocketServer.options.port === 'string') {
+      options.webSocketServer.options.port = Number(
+        options.webSocketServer.options.port
+      );
     }
   }
 
@@ -123,10 +125,8 @@ function normalizeOptions(compiler, options, logger) {
       path: parsedURL.pathname,
     };
   } else if (typeof options.client.webSocketURL.port === 'string') {
-      options.client.webSocketURL.port = Number(
-        options.client.webSocketURL.port
-      );
-    }
+    options.client.webSocketURL.port = Number(options.client.webSocketURL.port);
+  }
 
   // Enable client overlay by default
   if (typeof options.client.overlay === 'undefined') {
--- a/lib/utils/normalizeOptions.js
+++ b/lib/utils/normalizeOptions.js
@@ -95,15 +95,6 @@ function normalizeOptions(compiler, options, logger) {
       options: defaultWebSocketServerOptions,
     };
   } else {
-    if (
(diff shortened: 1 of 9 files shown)


test/validate-options.test.js
```diff
@@ -149,9 +152,6 @@ const tests = {
       {
         webSocketURL: { host: '' },
       },
-      {
-        webSocketURL: { port: '' },
-      },
```
alexander-akait, line 154: Keep it, should be fail when `minLength: 0` (in schema)

anshumanv (commented):
Can we add a simple e2e so it doesn't fail during compilation?

snitin315:
Added 👍🏻

anshumanv (dismissed):

alexander-akait (commented):
We also can have `port: string` and `webSocketServer: { type: 'ws', options: { port: string } }`, let's add these tests too, maybe will be great to move it in normalize and convert to `Number` to avoid working with different types:
```
if (typeof this.options.port === 'string' && this.options.port !== 'auto') {
  this.options.port = Number(this.options.port)
}

// Same for `websocketUrl.port` and `webSocketServer.options.port`
```

lib/utils/normalizeOptions.js
```diff
@@ -91,6 +95,15 @@ function normalizeOptions(compiler, options, logger) {
...
     };
   } else {
+    if (
+      options.webSocketServer.options &&
+      typeof options.webSocketServer.options.port === 'string'
+    ) {
+      options.webSocketServer.options.port = Number(
+        options.webSocketServer.options.port
+      );
+    }
```
alexander-akait, line 105: Make sense to simplify it moving here https://github.com/webpack/webpack-dev-server/blob/5adf5d3a4de11451ff65c00d63c0d89d6f4e2ddf/lib/utils/normalizeOptions.js#L113 and use

```js
// i.e. normalize after normalize :)
if (typeof options.webSocketServer.port === 'string') {
  options.webSocketServer.port = Number(options.webSocketServer.port);
}
```

lib/utils/normalizeOptions.js
```diff
@@ -111,9 +124,14 @@ function normalizeOptions(compiler, options, logger) {
 
     options.client.webSocketURL = {
       host: parsedURL.hostname,
-      port: parsedURL.port,
+      port: Number(parsedURL.port),
       path: parsedURL.pathname,
     };
+  } else if (
+    typeof options.client.webSocketURL === 'object' &&
```
alexander-akait, line 131: Wrong check and unnecessary, we always has `Object` here

alexander-akait: Just use `else`

alexander-akait (approved):
Feel free to merge when CI will be green

alexander-akait (approved):

anshumanv (approved):

Issue #3365, opened 2021-05-29
The "poll" property name on watchFiles option 


muzuiget:
- [x] This is a **bug**
- [ ] This is a **modification** request

The `watchFiles` option is added #3136, and the testcase code show that we should use the property name `usePolling`.

https://github.com/webpack/webpack-dev-server/blob/12f045f709850ee6879da8e4379510594d3af49c/test/options.test.js#L409

But it isn't work.

After reading the code:

https://github.com/webpack/webpack-dev-server/blob/73896127cef090124f75a1096b2968bda7aa2cb1/lib/Server.js#L976

We should use `poll` instead.

anshumanv:
Good catch, will fix

anshumanv:
I think staticOptions use `.poll` and chokidar options use `usePolling`, we need to handle both properties

anshumanv:
@alexander-akait I think we should use same properties as chokidar options, presently we use `poll` to calculate `internal` option value, we can just read `usePolling` and `interval` and pass it to config, i.e no resolving on our side

alexander-akait:
Ideally we should respect `watchOptions` from webpack, but rename them if needed

anshumanv:
we can add fallback logic to support both maybe -

`usePolling: (!!watchOptions.poll || watchOptions.usePolling)`

PR #3366, opened 2021-05-29
fix: polling usage in watchFiles option

anshumanv:
- [x] This is a **bugfix**
- [ ] This is a **feature**

yet to

### Motivation / Use-Case
Fix https://github.com/webpack/webpack-dev-server/issues/3365

### Breaking Changes
no

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -973,9 +973,12 @@ class Server {
     // https://github.com/webpack/watchpack/blob/master/lib/DirectoryWatcher.js#L49
     // this isn't an elegant solution, but we'll improve it in the future
     // eslint-disable-next-line no-undefined
-    const usePolling = !!watchOptions.poll || watchOptions.usePolling;
+    const usePolling = watchOptions.usePolling || Boolean(watchOptions.poll);
     const interval =
-      typeof watchOptions.poll === 'number'
+      // eslint-disable-next-line no-nested-ternary
+      typeof watchOptions.interval !== 'undefined'
+        ? watchOptions.interval
+        : typeof watchOptions.poll === 'number'
         ? watchOptions.poll
         : // eslint-disable-next-line no-undefined
           undefined;
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -973,7 +973,7 @@ class Server {
     // https://github.com/webpack/watchpack/blob/master/lib/DirectoryWatcher.js#L49
     // this isn't an elegant solution, but we'll improve it in the future
     // eslint-disable-next-line no-undefined
-    const usePolling = watchOptions.poll ? true : undefined;
+    const usePolling = !!watchOptions.poll || watchOptions.usePolling;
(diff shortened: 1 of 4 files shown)


anshumanv:
Not sure how to test this, maybe mock chokidar and check it's getting called by the correct config?

alexander-akait:
> maybe mock chokidar

Yep, right direction

lib/Server.js
```diff
@@ -973,7 +973,7 @@ class Server {
     // https://github.com/webpack/watchpack/blob/master/lib/DirectoryWatcher.js#L49
     // this isn't an elegant solution, but we'll improve it in the future
     // eslint-disable-next-line no-undefined
-    const usePolling = watchOptions.poll ? true : undefined;
+    const usePolling = !!watchOptions.poll || watchOptions.usePolling;
```
alexander-akait, line 976: Prefer to use `Boolean` instead `!!`

anshumanv: yep correct 

```
const usePolling = watchOptions.usePolling || Boolean(watchOptions.poll);
const interval =
  typeof watchOptions.poll === 'number'
    ? watchOptions.poll
    : // eslint-disable-next-line no-undefined
      undefined;
```

this will handle the above cases 👍

alexander-akait:
```
const usePolling = watchOptions.usePolling || Boolean(watchOptions.poll);
const interval =
  typeof watchOptions.internal !== 'undefined'
    ? watchOptions.internal
    : 
      typeof watchOptions.poll === number 
        ? watchOptions.poll
        : undefined;
```

anshumanv: oh yep we should respect user interval too

anshumanv: fixed @alexander-akait

test/server/watchFiles-option.test.js
```diff
@@ -298,6 +302,21 @@ describe("'watchFiles' option", () => {
...
+        followSymlinks: false,
+        atomic: false,
+        alwaysStat: true,
+        ignorePermissionErrors: true,
+        // eslint-disable-next-line no-undefined
+        ignored: undefined,
+        usePolling: true,
+        interval: 400,
+      });
+    });
```
alexander-akait, line 318: Let's add do this check in each test, to avoid future regressions, also let's add test:
- `poll: true`
- `usePolling: true` without interval
- `usePolling: false` and `poll: true` in this case we should prefer use `false` value (maybe need typeof)
- `usePolling: true` and `interval: 200` and `poll: 400` (should be preferred `200`)

anshumanv: yep wip

anshumanv: done all

alexander-akait (approved):
/cc @snitin315

snitin315 (approved):
Looks Good!

PR #3392, opened 2021-06-06
refactor: move `runBonjour`and `runOpen` utils  to `Server`

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Already present.

### Motivation / Use-Case

Inline utils with the server.

### Breaking Changes
None

### Additional Info
After this


anshumanv (dismissed):

rishabh3112 (approved):

lib/Server.js
```diff
@@ -538,6 +538,110 @@ class Server {
...
+      subtypes: ['webpack'],
+      ...options.bonjour,
+    });
+
+    process.on('exit', () => {
+      bonjour.unpublishAll(() => {
+        bonjour.destroy();
+      });
+    });
+  }
```
alexander-akait, line 559: It should not be static function, it is not public API

lib/Server.js
```diff
@@ -538,6 +538,110 @@ class Server {
...
+                        )}" arguments`
+                      : ''
+                  }`
+                : ''
+            }. If you are running in a headless environment, please do not use the "open" option or related flags like "--open", "--open-target", and "--open-app".`
+          );
+        });
+      })
+    );
+  }
```
alexander-akait, line 643: Ideally will be great to inline inside code

alexander-akait: Also I think we need better name...

snitin315: Maybe `openBrowser`?

alexander-akait: Yep sounds good

lib/Server.js
```diff
@@ -538,6 +538,110 @@ class Server {
...
+
+    if (Array.isArray(options)) {
+      options.forEach((item) => {
+        openTasks.push(...getOpenTask(item));
+      });
+    } else {
+      openTasks.push(...getOpenTask(options));
+    }
+
+    return Promise.all(
```
alexander-akait, line 609: I think we don't need `return` here, because we don't use returned result :smile:

snitin315: Yes, good catch.

rishabh3112 (approved):

rishabh3112 (approved):

alexander-akait:
@snitin315 I will solve conflicts, anyway we should remove more utils, some of them we can move in normalize, some as static methods

alexander-akait:
Found more cases, WIP

alexander-akait:
@snitin315 Feel free to remove more utils, ideally we should keep only two utils after refactor:
- `normalizeOptions.js` (we will move this as method in future)
- `DevServerPlugin.js` (will be removed when we will have plugin support here)

Both will be remove in the one future PR

Also because we already do many breaking change and will write migration guide, make sense to remove `findPort.js` and say in docs what you need the latest version of `webpack-cli` to using `webpack-dev-server`

snitin315:
Will take care of it 👍

alexander-akait:
@snitin315 But let's do it in separate PR :+1: I want to merge this and continue

PR #3468, opened 2021-06-21
fix: move wsHeartbeatinterval to socket server

anshumanv:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

existing

### Motivation / Use-Case
doesn't make sense in main server class

### Breaking Changes
no


alexander-akait (dismissed):

lib/servers/WebsocketServer.js
```diff
@@ -40,7 +42,7 @@ module.exports = class WebsocketServer extends BaseServer {
         socket.isAlive = false;
         socket.ping(noop);
       });
-    }, this.server.webSocketHeartbeatInterval);
+    }, server.webSocketHeartbeatInterval || this.webSocketHeartbeatInterval);
```
anshumanv, line 45: @alexander-akait had to make this change because we want to tune this value from test, can you think of any better approach?

alexander-akait: We need rewrite our test on e2e, because it is stupid test

PR #3474, opened 2021-06-22
refactor: manage web socket connections from base server class

anshumanv:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

existing 

### Motivation / Use-Case
we'll manage all socket connection and events from the base server class

### Breaking Changes
no

### Additional Info

move `this.(web)sockets` from `dev` server + `servers` and move into `WebsocketServer`

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -202,7 +202,7 @@ class Server {
       }
 
       this.sendMessage(
-        this.webSocketServer.webSocketConnections,
+        this.webSocketServer && this.webSocketServer.webSocketConnections,
         'progress-update',
         {
           percent,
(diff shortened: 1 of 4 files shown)


lib/Server.js
```diff
@@ -202,7 +202,7 @@ class Server {
       }
 
       this.sendMessage(
-        this.webSocketServer.webSocketConnections,
+        this.webSocketServer && this.webSocketServer.webSocketConnections,
```
alexander-akait, line 205: No need to check `this.webSocketServer`, it will be always non null-able here, better rewrite tests

PR #3522, opened 2021-07-12
refactor: exlcude `hot` for cli

snitin315:
- [X] This is a **bugfix**
- [ ] This is a **feature**

Already present, CI should be green.

### Motivation / Use-Case

`webpack-cli` already defines `--hot`, so it is kind of just duplicate code.

Also right now it shows `--hot` two times because of this in suggestions on passing a wrong cli option.

## Before 
Screenshot 2021-07-12 at 5 35 42 PM

## After
Screenshot 2021-07-12 at 5 35 51 PM

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/bin/cli-flags.js
+++ b/bin/cli-flags.js
@@ -374,30 +374,6 @@ module.exports = {
     simpleType: 'string',
     multiple: false,
   },
-  hot: {
-    configs: [
-      {
-        type: 'boolean',
-        multiple: false,
-        description:
-          'Enables Hot Module Replacement. https://webpack.js.org/configuration/dev-server/#devserverhot',
-        path: 'hot',
-      },
-      {
-        type: 'enum',
-        values: ['only'],
-        multiple: false,
-        description:
-          'Enables Hot Module Replacement. https://webpack.js.org/configuration/dev-server/#devserverhot',
-        path: 'hot',
-      },
-    ],
-    description:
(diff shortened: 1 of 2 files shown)


alexander-akait:
here interesting case, I think we should solve it on webpack-cli level, here https://github.com/webpack/webpack-dev-server/pull/3522, for webpack-dev-server we support the `hot` option, but webpack-cli should merge them

snitin315:
It looks like we may need it here, and should update code in webpack-cli to not show `--hot` multiple times.

alexander-akait:
Yep, I think we should fix it in webpack-cli

snitin315:
hmm, will fix it 👍🏻

alexander-akait:
To be honestly I think we remove `--hot` from `webpack-cli`, for developers who want to use hot module replacement for non browser env should add plugin manually

snitin315:
I think the same, let's remove from cli

alexander-akait:
Yep, I think we can start to create the next branch in webpack-cli and start to working, drop webpack v4, drop node v10 and some other stuff, feel free to start and I will add small readmap (there are no a lot of work in webpack-cli for the next major release)

snitin315:
Will do 👍

alexander-akait:
@snitin315 Let's close here, it will be for webpack-cli@next

snitin315:
Yep  we can close.

PR #3533, opened 2021-07-19
Chore update strip ansi

alexander-akait:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

update `strip-ansi` + test stabilization

### Breaking Changes

No

### Additional Info

Found one problem, we neeed to allow set `webSocketServer: false`, I will send a PR

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "serve-index": "^1.9.1",
     "sockjs": "^0.3.21",
     "spdy": "^4.0.2",
-    "strip-ansi": "^6.0.0",
+    "strip-ansi": "^7.0.0",
     "url": "^0.11.0",
     "webpack-dev-middleware": "^5.0.0",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -97,6 +97,7 @@
     "sockjs-client": "^1.5.1",
     "standard-version": "^9.3.0",
     "style-loader": "^2.0.0",
+    "strip-ansi-v6": "npm:strip-ansi@^6.0.0",
```
alexander-akait, line 100: We need to migrate our test on ECMA modules, but let's do it late

snitin315: Will take care 👍

alexander-akait: Ideally we should do it for all our repos in webpack/webpack-contrib except webpack itself

PR #3542, opened 2021-07-20
refactor: remove unused method

alexander-akait:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

existing

### Motivation / Use-Case

we don't use it

### Breaking Changes

potentially yes, but it should be not used directly

### Additional Info

No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -1123,11 +1123,6 @@ class Server {
     return statsObj.toJson(stats);
   }
 
-  use() {
-    // eslint-disable-next-line prefer-spread
-    this.app.use.apply(this.app, arguments);
-  }
-
   setContentHeaders(req, res, next) {
     let { headers } = this.options;
     if (headers) {


alexander-akait:
It is breaking change for v4, use `devServer.server`

zbrcz:
Not sure I understand.

We are doing something like this:

```
const server = new WebpackDevServer(/* ... */);
server.use((req, res) => {
  res.end(/* ... */);
});
```

Now we get "TypeError: server.use is not a function". What to use instead of `server.use` to be able to tap into the request handling process?

Thanks.

alexander-akait:
```
const devServer = new WebpackDevServer();

await devServer.start();

devServer.server.use(() => {});
```

zbrcz:
Unfortunately, that just gives "TypeError: Cannot read property 'use' of undefined"... `devServer.server` is not defined.

alexander-akait:
Check your code https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L1601

zbrcz:
It seems the error I reported in my previous message was due to some mistake in my code. Even after fixing it, I kept getting "TypeError: devServer.server.use is not a function", though. Ended up using `devServer.app.use`. Thanks for all your help!

Issue #3552, opened 2021-07-22
Overlay HTML


joacub:
Overlay is showing the html entities instead of html processed by browser.

<img width="1307" alt="image" src="https://user-images.githubusercontent.com/2091228/126585058-05d0de86-599f-4300-a190-eadafa300fa1.png">

snitin315:
Looking into it.

alexander-akait:
Please respect time of other developers, what `normal view`?

snitin315:
The issue here is `'` is being shown as `&apos;` -

for example `Can't -> Can&apos;t` 

<img width="1404" alt="Screenshot 2021-07-22 at 4 44 46 PM" src="https://user-images.githubusercontent.com/46647141/126630716-b2bd61ce-6fc1-43ba-b7a0-8cf8953f55b9.png">

alexander-akait:
oh, I see, let's use `pre` tag for this, maybe it will be even better

snitin315:
> oh, I see, let's use `pre` tag for this, maybe it will be even better

No luck with `pre` tag. 

<img width="1008" alt="Screenshot 2021-07-22 at 5 44 31 PM" src="https://user-images.githubusercontent.com/46647141/126637450-6019a3ed-0694-4e19-a095-09b0a0e07f55.png">

joacub:
> Please respect time of other developers, what is `normal view`?

What do you mean ? Is not enough clear a picture with the html tags in the text ?

why you are so like that ? Don’t even see the picture ?

alexander-akait:
@joacub because using the issue template for everyone it is necessary, I am not like that, you will report a bug so we ask you to use the issue template, this is a banal respect

joacub:
this is a really easy thing html tags are encoded yo htmlentities that's it, should I have to filled a big extensive template for such a simple things,

By the way here is other picture showing the HTML tags, is not just the ', ", is all html, the content is encoded to htmlentities:


<img width="1298" alt="image" src="https://user-images.githubusercontent.com/2091228/126671212-63a701f0-82a9-438e-9097-635bc6719532.png">

alexander-akait:
> this is a really easy thing html tags are encoded to htmlentities that's it, should I have to filled a big extensive template for such a simple things,

[...]

PR #3553, opened 2021-07-22
fix: overlay content

snitin315:
- [X] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case
Fix https://github.com/webpack/webpack-dev-server/issues/3552

#### Before 

Screenshot 2021-07-22 at 4 44 46 PM

#### After

Screenshot 2021-07-22 at 4 51 23 PM

### Breaking Changes
No

### Additional Info
No

### DIFF
--- a/client-src/overlay.js
+++ b/client-src/overlay.js
@@ -4,7 +4,7 @@
 // They, in turn, got inspired by webpack-hot-middleware (https://github.com/glenjamin/webpack-hot-middleware).
 
 const ansiHTML = require("ansi-html");
-const { encode } = require("html-entities");
+const { encode, decode } = require("html-entities");
 
 const colors = {
   reset: ["transparent", "transparent"],
@@ -138,7 +138,7 @@ function show(messages, type) {
 
       // Make it look similar to our terminal.
       const errorMessage = message.message || messages[0];
-      const text = ansiHTML(encode(errorMessage));
+      const text = decode(ansiHTML(encode(errorMessage)), { level: "html5" });
       const messageTextNode = document.createTextNode(text);
 
       entryElement.appendChild(typeElement);
--- a/client-src/overlay.js
+++ b/client-src/overlay.js
@@ -4,7 +4,6 @@
 // They, in turn, got inspired by webpack-hot-middleware (https://github.com/glenjamin/webpack-hot-middleware).
 
 const ansiHTML = require("ansi-html");
-const { encode, decode } = require("html-entities");
 
 const colors = {
   reset: ["transparent", "transparent"],
@@ -138,7 +137,7 @@ function show(messages, type) {
 
       // Make it look similar to our terminal.
       const errorMessage = message.message || messages[0];
-      const text = decode(ansiHTML(encode(errorMessage)), { level: "html5" });
+      const text = ansiHTML(errorMessage);
       const messageTextNode = document.createTextNode(text);
 
       entryElement.appendChild(typeElement);
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,6 @@
         "del": "^6.0.0",
         "express": "^4.17.1",
         "graceful-fs": "^4.2.6",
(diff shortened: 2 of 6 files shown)


client-src/overlay.js
```diff
@@ -138,7 +137,7 @@ function show(messages, type) {
 
       // Make it look similar to our terminal.
       const errorMessage = message.message || messages[0];
-      const text = ansiHTML(encode(errorMessage));
```
alexander-akait, line 141: It is not safe, here can be XSS

snitin315: Maybe we need [decode(text, {level: 'html5'})](https://github.com/mdevils/html-entities#decodetext-options) afterwards

alexander-akait: hm, maybe, just try it

alexander-akait (commented):
Also we need test it

snitin315:
> Also we need test it

WIP on tests

snitin315:
Ready for review. /cc @alexander-akait

alexander-akait:
Thanks, will look at near future

alexander-akait:
I was wrong about XSS due `Creates a new Text node. This method can be used to escape HTML characters.` https://developer.mozilla.org/en-US/docs/Web/API/Document/createTextNode

alexander-akait:
Fixed, anyway thanks

alexander-akait (approved):
Less deps :star:

joacub:
the correct fix is this ====>
`var errorMessage = message.message || messages[0];
      var text = ansiHTML(errorMessage);
      var messageTextNode = document.createElement('div');`

html still showing bad, with this looks perfect, see the imagen bellow:

image

alexander-akait:
It is xss, no it is not good

alexander-akait:
Current solution has the same output

joacub:
> Current solution has the same output

Impossible, html in a text node will be show up like entities….

joacub (commented):
Html in text node will appear like `<span>raw</span>`

alexander-akait:
Please provide screenshot, again, always use the issue template, our tests are find

joacub:
> Please provide screenshot, again, always use the issue template, our tests are fine

Did you test with html content ?

alexander-akait:
We should not accept HTML content from error/warning message, otherwise you can inject `<script>` and do XSS

joacub:
Look is very simple, text node can no allocate html only text….

alexander-akait:
Because it is XSS

joacub:
> Because it is XSS

Inject xss to my self 🤣

alexander-akait:
other plugin which you can install and use can emit HTML content and steel your environments, it is not joke

joacub:
Perfect, is a 1% of the time that can happen maybe less, there is other solution to show correctly this? cause is impossible to read anything

alexander-akait:
No one plugin should not emit HTML content in warning/error messages

alexander-akait:
In your screenshot only text content and it should be good in overlay

joacub:
> No one plugin should not emit HTML content in warning/error messages

Eslint do and do nice and read is bestifully, so best way is disable this not useful overlay

alexander-akait:
> Eslint do and do nice and read is bestifully

HTML content? don't talk nonsense, eslint uses only ansi colors, I don't see HTML content

joacub:
> > Eslint do and do nice and read is bestifully
> 
> HTML content? don't talk nonsense, eslint uses only ansi colors (yes, you can write custom output, but I don't see it is good idea), I don't see HTML content

Eslint return html you check in the other issue the screenshot with html …..

alexander-akait:
No HTML in any place

joacub:
Loook this

alexander-akait:
Okay, please open a new issue with **reproducible test repo** and format this as issue template, original problem with html entries was fixed, what is why I ask ALWAYS use the issue template

PR #3555, opened 2021-07-22
ci: debug

alexander-akait:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No need

### Motivation / Use-Case

we should use client bundled by webpack v5 to test on webpack v4, need check it, if we don't do it we need to fix it here

### Breaking Changes

No

### Additional Info

No


alexander-akait:
Due:
```
(function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory();
	else if(typeof define === 'function' && define.amd)
		define([], factory);
	else {
		var a = factory();
		for(var i in a) (typeof exports === 'object' ? exports : root)[i] = a[i];
	}
})(window, function() {
```

we use bundled client for webpack v4, but it is not right

alexander-akait:
npm is the worst package manager ever I see, just ignore `--ignore-scripts` and run it without any output...

alexander-akait:
`npm --ignore-scripts` is magic, on linux npm always run scripts, on windows never, setting false options, disable all scripts on linux, but run scripts on windows, magic :mage: maybe time to migration on yarn from stupid npm

alexander-akait:
I hate npm, really hate, even `npm config set ignore-scripts true` doesn't work

alexander-akait:
We have three unstable tests:
- one in CLI for IPv6
- two in e2e - in `web-socket-communication.test.js` (need investigate) and in `hot-and-live-reload.test.js` sometimes it output 2/3/4 errors

First and third is very rare, need investigate why it is happens (maybe problem in regexp), second should be fixed there

PR #3563, opened 2021-07-23
fix: show deprecation warning for incorrect usage of Node.js API

snitin315:
- [x] This is a **bugfix**
- [ ] This is a **feature**

WIP ON TESTS

### Motivation / Use-Case

SHOW DEPRECATION WARNING

### Breaking Changes
NO

### Additional Info
NO

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -19,6 +19,11 @@ class Server {
   constructor(options = {}, compiler) {
     // TODO: remove this after plugin support is published
     if (options.hooks) {
+      process.emitWarning(
+        "Using 'compiler' as the first argument is deprecated. Please use 'options' as the first argument and 'compiler' as the second argument.",
+        "DeprecationWarning",
+        "DEP_WEBPACK_DEV_SERVER_API"
+      );
       [options, compiler] = [compiler, options];
     }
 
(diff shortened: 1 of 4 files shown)


lib/Server.js
```diff
@@ -20,6 +20,13 @@ class Server {
     // TODO: remove this after plugin support is published
     if (options.hooks) {
       [options, compiler] = [compiler, options];
+
+      this.compiler = compiler;
+      this.logger = this.compiler.getInfrastructureLogger("webpack-dev-server");
```
alexander-akait, line 25: don't do it, you should use https://nodejs.org/api/util.html#util_util_deprecate_fn_msg_code

snitin315: Not sure how to use `util.deprecate` on the constructor, I used `process.emitWarning` instead - 

Screenshot 2021-07-24 at 9 42 18 AM

ylemkimon: @snitin315 You can pass an empty function to `util.deprecate()` and call it.

snitin315: @ylemkimon thanks 👍🏻

alexander-akait (approved):
I will fix tests today

PR #3566, opened 2021-07-23
chore: use default node cache for deps

anshumanv:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

yet to

### Motivation / Use-Case
use inbuilt node cache

### Breaking Changes
no

### Additional Info
https://github.com/actions/setup-node#caching-packages-dependencies

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -35,6 +35,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
+          cache: "npm"
 
       - name: Use latest NPM
         run: sudo npm i -g npm
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -69,7 +69,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
-          cache: 'npm'
+          cache: "npm"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/nodejs.yml
```diff
@@ -78,20 +79,6 @@ jobs:
...
-
-      - name: Cache Node.js modules
-        id: cache
-        uses: actions/cache@v2
-        with:
-          path: ${{ steps.npm-cache-dir.outputs.dir }}
-          key: ${{ runner.os }}-${{ matrix.webpack-version }}-node-${{ hashFiles('**/package-lock.json') }}
-          restore-keys: |
-            ${{ runner.os }}-${{ matrix.webpack-version }}-node-
-
```
alexander-akait, line 94: We need `webpack-version` here, otherwise cache was wrong

anshumanv: if I understand correctly, we probably won't need it because we do an install after deps are restored so it'll do a fresh install for webpack version

snitin315 (commented):
one more instance - 
https://github.com/webpack/webpack-dev-server/blob/cce64d80e6326740af36b33d2bc0a0b9c77bd972/.github/workflows/nodejs.yml#L35

anshumanv:
it is single job, no need cache there

snitin315 (dismissed):

PR #3579, opened 2021-07-31
fix(overlay): render ansi formattet error messages correctly

### LINKED ISSUE #3575: Error overlay html encodes the ansi html formatting the error output
- [x] This is a **bug**

### Code

Reproduced in the [CRA webpack 5 update branch](https://github.com/facebook/create-react-app/pull/11201) using webpack error overlay in the WP5 CRA pr (replacing the CRA error overlay)
 _(The "wp5" branch is on the way to "main")_

### Expected Behavior
image

### Actual Behavior
image

[ansiHTML](https://github.com/webpack/webpack-dev-server/blob/master/client-src/overlay.js#L138) returns error formattet using html as a text string but because it's handed of to [ document.createTextNode](https://github.com/webpack/webpack-dev-server/blob/master/client-src/overlay.js#L139) the [browser will escape the string](https://developer.mozilla.org/en-US/docs/Web/API/Document/createTextNode) problem is that `errorMessage` is already escaped _(as it should be to prevent XSS/script injection)_ so no need for the double escaping.

Suggestion:
Change
```js
...
      const text = ansiHTML(errorMessage);
      const messageTextNode = document.createTextNode(text);
...
```
To:
```js
...
      var text = ansiHTML(encode(errorMessage));
      var messageTextNode = document.createElement('div');
      messageTextNode.innerHTML = text;
...
```

raix:
- [x] This is a **bugfix**
- [ ] This is a **feature**

yes, added a test to replicate the issue / cover ansi formatting

### Motivation / Use-Case

Fixes the ansi formatting in the overlay - want to use the overlay in the CRA project see #3575 for more info an ways to replicate

### Breaking Changes

none

### Additional Info

fixed #3575

### DIFF
--- a/client-src/overlay.js
+++ b/client-src/overlay.js
@@ -2,6 +2,7 @@
 // They, in turn, got inspired by webpack-hot-middleware (https://github.com/glenjamin/webpack-hot-middleware).
 
 const ansiHTML = require("ansi-html");
+const { encode } = require("html-entities");
 
 const colors = {
   reset: ["transparent", "transparent"],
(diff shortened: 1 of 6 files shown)


raix:
Seems to be an issue with the test suite...

alexander-akait (commented):
Thanks, I will look at this in near future

alexander-akait:
Will fix the flaky test in other PR

PR #3647, opened 2021-08-14
chore: warning when proxy property is undefined

evenstensberg:
- [ ] This is a **bugfix**
- [x?] This is a **feature**

No

### Motivation / Use-Case

When the `proxy` property target is undefined, the user doesn't know, and it's nice to know its being skipped.

### Breaking Changes
No

### Additional Info

:)

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -832,6 +832,9 @@ class Server {
       if (proxyConfig.target) {
         return createProxyMiddleware(context, proxyConfig);
       }
+      this.logger.warn(
+        `The "target" value specified in the "proxy" property ${context} is not defined and will be skipped.`
+      );
     };
     /**
      * Assume a proxy configuration specified as:


lib/Server.js
```diff
@@ -832,6 +832,9 @@ class Server {
       if (proxyConfig.target) {
         return createProxyMiddleware(context, proxyConfig);
       }
+      this.logger.warn(
+        `The "target" value specified in the "proxy" property ${context} is not defined and will be skipped.`
```
snitin315, line 836: for better visibility 

```suggestion
        `The "target" value specified in the "proxy" property "${context}" is not defined and will be skipped.`
```

Can you also add a simple test case in https://github.com/webpack/webpack-dev-server/blob/master/test/server/proxy-option.test.js ?

alexander-akait (changes requested):
Bad idea, we should allow to use proxy without target, it is valid case

alexander-akait:
WIP on fix

PR #3650, opened 2021-08-16
fix: update webpack peer dependency to ^4.37.0

dtinth:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No, this is a metadata update.

### Motivation / Use-Case

When webpack-dev-server is run on any webpack version < 4.37, this error appears.

    [webpack-cli] TypeError: compiler.getInfrastructureLogger is not a function

This is because `compiler.getInfrastructureLogger` is introduced in webpack v4.37, therefore webpack-dev-server is not compatible with any version of webpack below it.

### Breaking Changes

None.

### Additional Info

We hit this error when using `webpack-dev-server@4.0.0-rc.0` with `webpack@4.33.0`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "webpack-merge": "^5.8.0"
   },
   "peerDependencies": {
-    "webpack": "^4.0.0 || ^5.0.0"
+    "webpack": "^4.37.0 || ^5.0.0"
   },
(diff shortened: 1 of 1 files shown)


jiengsad:
这个问题现在依然存在

PR #3651, opened 2021-08-16
fix: proxy logging and allow to pass options without the `target` option

### LINKED ISSUE #3486: Proxy without target
- [ ] This is a **bug**

### Code

```js
devServer: {
    proxy: {
        '/api': {
            target: 'dummy', // shouldn't be necessary
            router: request => ...
        },
    },
},
```

### Expected Behavior

I expect to be able to use the `router` feature of `http-proxy-middleware` without having to specify a `target`.

### Actual Behavior

When `target` is missing, the proxy middleware is not used at all: https://github.com/webpack/webpack-dev-server/blob/516d7b1b5c44ca9b934b91c777120396d895a7d2/lib/Server.js#L281-L283

Which is also noted in the [documentation](https://webpack.js.org/configuration/dev-server/#devserverproxy):
>  Note that some of `http-proxy-middleware`'s features do not require a `target` key, e.g. its `router` feature, but you will still need to include a `target` key in your configuration here, otherwise `webpack-dev-server` won't pass it along to `http-proxy-middleware`).

Recently, the validation in `http-proxy-middleware` has been [updated to allow for configurations without a `target` key](https://github.com/chimurai/http-proxy-middleware/pull/512). I suggest to do the same here and extend the condition to `proxyConfig.target || proxyConfig.router`.

alexander-akait:
- [x] This is a **bugfix**
- [ ] This is a **feature**

fixes #3486 fixes #3647

### Motivation / Use-Case

fixes

### Breaking Changes

No

### Additional Info

TODO for the next major release

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -492,8 +492,12 @@ class Server {
      * }
      */
     if (typeof options.proxy !== "undefined") {
+      // TODO remove in the next major release, only accept `Array`
       if (!Array.isArray(options.proxy)) {
-        if (Object.prototype.hasOwnProperty.call(options.proxy, "target")) {
+        if (
+          Object.prototype.hasOwnProperty.call(options.proxy, "target") ||
+          Object.prototype.hasOwnProperty.call(options.proxy, "router")
+        ) {
           options.proxy = [options.proxy];
         } else {
           options.proxy = Object.keys(options.proxy).map((context) => {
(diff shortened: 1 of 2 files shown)


snitin315 (approved):

alexander-akait:
I will fix random tests fail late

PR #3657, opened 2021-08-17
refactor: remove `killable`

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

less deps

### Breaking Changes
No

### Additional Info
No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -1142,9 +1142,6 @@ class Server {
   }
 
   createServer() {
-    const https = require("https");
-    const http = require("http");
-
     if (this.options.https) {
       if (this.options.http2) {
         // TODO: we need to replace spdy with http2 which is an internal module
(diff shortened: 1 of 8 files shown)


lib/Server.js
```diff
@@ -61,6 +60,31 @@ class Server {
...
+      server.close(cb);
+      sockets.forEach((socket) => {
+        socket.destroy();
+      });
+      // reset so the server can be restarted
+      sockets = [];
+    };
+
+    return server;
+  }
```
alexander-akait, line 86: Let's avoid extra method and put this code inside close, no need for this

lib/Server.js
```diff
@@ -1738,6 +1735,27 @@ class Server {
...
+    });
+
+    this.server.kill = (cb) => {
+      this.server.close(cb);
+      sockets.forEach((socket) => {
+        socket.destroy();
+      });
+      // reset so the server can be restarted
+      sockets = [];
+    };
```
alexander-akait, line 1757: Move this to close, we don't need this here

lib/Server.js
```diff
@@ -1738,6 +1735,27 @@ class Server {
...
+
+    this.server.on("connection", (socket) => {
+      // add socket to list
+      sockets.push(socket);
+
+      socket.once("close", () => {
+        // remove socket from list
+        sockets.splice(sockets.indexOf(socket), 1);
+      });
+    });
```
alexander-akait, line 1748: Move this where we create `server`

lib/Server.js
snitin315: Here? https://github.com/webpack/webpack-dev-server/blob/ae8c523854700d63baca335312ccc55df4bfbc42/lib/Server.js#L1790

PR #3719, opened 2021-08-22
refactor: use async API

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

use async API

### Breaking Changes
No

### Additional Info
No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -545,11 +545,18 @@ class Server {
       if (!options.https.key || !options.https.cert) {
         const certificateDir = Server.findCacheDir();
         const certificatePath = path.join(certificateDir, "server.pem");
-        let certificateExists = fs.existsSync(certificatePath);
+        let certificateExists;
+
+        try {
+          const certificate = await fs.promises.stat(certificatePath);
+          certificateExists = certificate.isFile();
+        } catch {
+          certificateExists = false;
+        }
 
         if (certificateExists) {
           const certificateTtl = 1000 * 60 * 60 * 24;
-          const certificateStat = fs.statSync(certificatePath);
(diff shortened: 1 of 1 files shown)


lib/Server.js
```diff
@@ -405,13 +405,29 @@ class Server {
...
+              if (error) {
+                this.logger.error(error);
+                process.exit(2);
+              }
+            }
+          );
         }
 
-        fakeCert = fs.readFileSync(certificatePath);
+        fakeCert = fs.readFile(certificatePath);
```
anshumanv, line 430:
```suggestion
        fakeCert = await fs.readFile(certificatePath);
```

lib/Server.js
```diff
@@ -405,13 +405,34 @@ class Server {
...
+              if (error) {
+                this.logger.error(error);
+              }
+            }
+          );
         }
 
-        fakeCert = fs.readFileSync(certificatePath);
+        fakeCert = await new Promise((resolve, reject) => {
+          fs.readFile(certificatePath, (error, data) => {
```
alexander-akait, line 429: maybe we can use `fs/promise`?

snitin315: Yes, that would be better.

lib/Server.js
```diff
@@ -2033,10 +2045,10 @@ class Server {
         const net = require("net");
         const socket = new net.Socket();
 
-        socket.on("error", (error) => {
+        socket.on("error", async (error) => {
```
alexander-akait, line 2048: It can be async

alexander-akait (approved):

alexander-akait:
In future we should avoid sync API in favor async where it possible

PR #3720, opened 2021-08-22
fix: normalize for allowedHosts

anshumanv:
- [x] This is a **bugfix**
- [ ] This is a **feature**

not yet, will add once this fix is validated

### Motivation / Use-Case
Fix https://github.com/webpack/webpack-cli/issues/2909

### Breaking Changes
no

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -190,27 +190,23 @@ class Server {
     };
 
     if (typeof options.allowedHosts === "undefined") {
-      // allowedHosts allows some default hosts picked from
-      // `options.host` or `webSocketURL.hostname` and `localhost`
+      // AllowedHosts allows some default hosts picked from `options.host` or `webSocketURL.hostname` and `localhost`
       options.allowedHosts = "auto";
     }
-
-    if (
+    // We store allowedHosts as array when supplied as string
+    else if (
       typeof options.allowedHosts === "string" &&
       options.allowedHosts !== "auto" &&
       options.allowedHosts !== "all"
     ) {
-      // we store allowedHosts as array when supplied as string
       options.allowedHosts = [options.allowedHosts];
     }
-
-    // for cases when it's a singleton array with `all` or `auto`
-    if (
-      Array.isArray(options.allowedHosts) &&
-      options.allowedHosts.length === 1 &&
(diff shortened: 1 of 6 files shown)


lib/Server.js
```diff
@@ -204,6 +204,15 @@ class Server {
...
 
+    // for cases when it's a singleton array with `all` or `auto`
+    if (
+      Array.isArray(options.allowedHosts) &&
+      options.allowedHosts.length === 1 &&
+      (options.allowedHosts[0] === "auto" || options.allowedHosts[0] === "all")
+    ) {
+      options.allowedHosts = options.allowedHosts[0];
+    }
+
```
snitin315, line 215: I think we should also account for `allowedHosts: ['host1.com', 'host2.com', 'all']`:

https://github.com/webpack/webpack-dev-server/blob/8a3f0d8b9d41a4f431ceb4700341ac9e6e585563/lib/Server.js#L1516

```diff
- if (this.options.allowedHosts === "all") { 
+ if (
+ this.options.allowedHosts === "all" || 
+ (Array.isArray(this.options.allowedHosts) && this.options.allowedHosts.includes("all"))
+ ) { 
```

What do you think?

anshumanv: `all` and other hosts are mutually exclusive, I don't think there would be a use case for specifying other hosts with all :thinking: Maybe this would be the case for `auto` when we need to allow localhost along with other hosts

snitin315: 👍🏻

alexander-akait (commented):
Let's add tests too

lib/Server.js
anshumanv: /cc @alexander-akait need slight attention 

is this okay or should we handle 'auto' too here

alexander-akait: I will look at this in near future

anshumanv: will add tests once you confirm

anshumanv: will add tests once you confirm

alexander-akait: I think it was mistake adding `auto` (because we already allow to handle from localhost/ip/etc), we should revisit this in the next major release, I will finish PR

snitin315 (approved):

snitin315 (approved):

PR #3772, opened 2021-09-01
fix: improve description of `magicHtml` option

snitin315:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

 improve the description of `magicHtml` option

### Breaking Changes
None

### Additional Info
No


alexander-akait:
@snitin315 webpack can generate `.mjs` when you use `outputModules: true`, I think we should fix it too

snitin315:
> @snitin315 webpack can generate `.mjs` when you use `outputModules: true`, I think we should fix it too

On it.

PR #3798, opened 2021-09-05
Replace ansi-html and fix the ReDoS vulnerability

### LINKED ISSUE #3576: ReDoS Vulnerability
- [x] This is a **bug**

This is a vulnerability in ansi-html, which this package depends on.  To repro, you can use the command below as described https://github.com/Tjatse/ansi-html/issues/19.  

require('ansi-html')('\x1b[0m\x1b[' + '0'.repeat(35))

Are there any plans to replace/remove this package, with a more active project?  The library maintainers have been made aware of this change, but it does not look like it is being maintained.  This is preventing us from moving our source code to Production, as ansi-html is being flagged.

carlobeltrame:
- [x] This is a **bugfix**
- [ ] This is a **feature**

I moved and simplified the code from an unmaintained and vulnerable dependency to this project and fixed a ReDoS vulnerability, as suggested by @alexander-akait. I did not move the tests from that project - let me know if I should do that.

### Motivation / Use-Case

Fixes #3576
As suggested by @alexander-akait in https://github.com/webpack/webpack-dev-server/issues/3576#issuecomment-903732624

### Breaking Changes

None

### Additional Info
If the Apache license and altogether slightly awkward position of this code in webpack-dev-server doesn't seem right, an alternative would be to use the new [ansi-html-community](https://www.npmjs.com/package/ansi-html-community) package, which includes the CVE fix and promises to be better maintained. I'd be happy to create a PR for that as well. Let me know what you think.

### DIFF
--- a/client-src/modules/ansi-html/LICENSE
+++ b/client-src/modules/ansi-html/LICENSE
@@ -1,0 +1,202 @@
+Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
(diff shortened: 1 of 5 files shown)


YipingRuan:
Well done..thank you.!
Maybe fix the commit message and we are all saved?

refactor : Remove ansi-html dependency ?

carlobeltrame:
Oh, right, semantic commit messages. Done.

package-lock.json
```diff
@@ -1,17939 +1,8 @@
 {
   "name": "webpack-dev-server",
   "version": "4.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
snitin315, line 4: We use `"lockfileVersion": 2`, it would be nice if you can commit package-lock.json with npm@7, to avoid large diffs.

client-src/modules/ansi-html/LICENSE
```diff
@@ -0,0 +1,202 @@
+Apache License
```
snitin315, line 1: We use MIT 😕 , would it be a problem?

carlobeltrame: Yes, ansi-html was published under Apache 2.0. I cannot simply re-license code that isn't mine. Apache 2.0 is more restrictive than MIT, but it doesn't stop us from copying and adapting code, as long as the license is preserved.

package-lock.json
carlobeltrame: Of course, my bad. Done.

client-src/modules/ansi-html/index.js
```diff
@@ -0,0 +1,122 @@
+/**
```
carlobeltrame, line 1: Not sure if `modules` is the correct directory for this, I see that webpack builds some separate bundles based on subdirectories of `modules`. Let me know if there is a better suited location.

client-src/modules/ansi-html/LICENSE
alexander-akait: We can't accept this, sorry

alexander-akait: WE need rewrite code in this case on our solution

carlobeltrame: In that case it might be far easier to just change the dependency from `ansi-html` to `ansi-html-community`. Since the original author of `ansi-html` is not reacting, someone else created a new npm package with the same code plus the fix.

alexander-akait: Yes, we can do it too

carlobeltrame:
I have created a new PR #3801 which does not copy and adapt the code of ansi-html, but rather uses a fork of ansi-html that should be better maintained in the future.
Closing this PR in favor of #3801.

PR #3830, opened 2021-09-10
fix: don't reserve unused port too early

alan-agius4:
- [x] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case
At the moment, when starting 2 processes at the same time and using `port: 0` or `port auto` in the majority of cases one of the servers will fail because both server will end up being assigned the same port.

The reason for this is that a number of precious milliseconds elapse between the `getFreePort` and `listen` call. The main reason for this is the `initialize` method.

```
../Users/cli-reproductions/node_modules/webpack-dev-server/lib/Server.js:1544
      throw error;
Error: listen EADDRINUSE: address already in use 127.0.0.1:8080
    at Server.setupListenHandle [as _listen2] (net.js:1320:16)
    at listenInCluster (net.js:1368:12)
    at GetAddrInfoReqWrap.doListen [as callback] (net.js:1505:7)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:71:8) {
  code: 'EADDRINUSE',
  errno: -48,
  syscall: 'listen',
  address: '127.0.0.1',
  port: 8080
}
```

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -2071,6 +2071,7 @@ class Server {
 
   async start() {
     await this.normalizeOptions();
+    await this.initialize();
 
     if (this.options.ipc) {
       await new Promise((resolve, reject) => {
@@ -2102,8 +2103,6 @@ class Server {
       this.options.host = await Server.getHostname(this.options.host);
       this.options.port = await Server.getFreePort(this.options.port);
     }
-
-    await this.initialize();
 
     const listenOptions = this.options.ipc
       ? { path: this.options.ipc }


alexander-akait:
hm, moving `this.initialize` will not solve this, we need something another...

alan-agius4:
I also though that, but I couldn’t replicate the failure after I move it. 

So I figured it was enough.

alexander-akait:
it will not work because we need host and port before `initialize`, otherwise client entry will be broken

alexander-akait:
I will run tests and you will see

alan-agius4:
@alexander-akait oh I now understand.

alexander-akait:
But, yes, we need to solve this, right now I am working on better supporting multiple dev server, it can be problem, need to solve

alexander-akait:
Something like:
```
const server = net.createServer();
	server.unref();
	server.on('error', reject);
	server.listen(options, () => {
		const {port} = server.address();
		server.close(() => {
			resolve(port);
		});
	});
```

and try again if it fails

alan-agius4:
Yeah, that should work. I can probably that a look at that next week. Unless, you want to include this in your work to support multiple dev-server instances.

For the time being, I’ll close this PR to reduce noise.

alexander-akait:
Let's keep open, I will look at solution in near future

PR #3847, opened 2021-09-15
feat: allow array for `headers` option

snitin315:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

allow array for `headers` option :- 

```js
          headers: [
            {
              key: "X-Foo",
              value: "value1",
            },
            {
              key: "X-Bar",
              value: "value2",
            },
          ],
```

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -1847,7 +1847,7 @@ class Server {
 
       const allHeaders = [];
 
-      if (typeof headers === "object" && !Array.isArray(headers)) {
+      if (!Array.isArray(headers)) {
         // eslint-disable-next-line guard-for-in
         for (const name in headers) {
           allHeaders.push({ key: name, value: headers[name] });
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -490,7 +490,20 @@ class Server {
       options.devMiddleware = {};
     }
 
-    // No need to normalize `headers`
+    // normalize `headers`
+
+    if (
+      typeof options.headers !== "undefined" &&
+      Array.isArray(options.headers)
+    ) {
+      const allHeaders = {};
+
+      options.headers.forEach((header) => {
+        allHeaders[header.key] = header.value;
+      });
+
+      options.headers = allHeaders;
+    }
 
     if (typeof options.historyApiFallback === "undefined") {
       options.historyApiFallback = false;
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -491,17 +491,27 @@ class Server {
     }
 
     // normalize `headers`
-    if (
-      typeof options.headers !== "undefined" &&
-      Array.isArray(options.headers)
-    ) {
+    if (typeof options.headers !== "undefined") {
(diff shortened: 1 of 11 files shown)


test/e2e/headers.test.js
```diff
@@ -66,6 +66,78 @@ describe("headers option", () => {
...
+
+      expect(response.status()).toMatchSnapshot("response status");
+
+      expect(consoleMessages.map((message) => message.text())).toMatchSnapshot(
+        "console messages"
+      );
+
+      expect(pageErrors).toMatchSnapshot("page errors");
+    });
+  });
```
alexander-akait, line 139: Let's add test there `function () => { return [{key: 'key', value: 'value'}] }`

lib/Server.js
```diff
@@ -490,7 +490,29 @@ class Server {
...
+
+        if (Array.isArray(returnedHeaders)) {
+          returnedHeaders.forEach((header) => {
+            allHeaders[header.key] = header.value;
+          });
+
+          options.headers = allHeaders;
+        }
+      }
+    }
```
alexander-akait, line 515: We can't normalize function here, because https://github.com/webpack/webpack-dev-server/blob/6b6ace1e576e69976b00eaf42c18b61b2faf1345/lib/Server.js#L1867, ideally we should convert object to array with key/value, but if it is function run it https://github.com/webpack/webpack-dev-server/blob/6b6ace1e576e69976b00eaf42c18b61b2faf1345/lib/Server.js#L1867 and if developer return object convert it to array too, then write headers

snitin315: >  ideally we should convert object to array with key/value, but if it is function run it

you mean provided `headers: { "x-foo": "yes"}` in config, normalize it to `headers: [{ key: 'x-foo', value: "yes"}]` ?

alexander-akait: Yes

alexander-akait: Any object to array, plus if developer return object from function convert too

snitin315: I will refactor 👍🏻

rishabh3112 (dismissed):

lib/Server.js
```diff
@@ -1845,10 +1845,19 @@ class Server {
         headers = headers(req, res, this.middleware.context);
       }
 
-      // eslint-disable-next-line guard-for-in
-      for (const name in headers) {
-        res.setHeader(name, headers[name]);
+      const allHeaders = [];
+
+      if (typeof headers === "object" && !Array.isArray(headers)) {
```
alexander-akait, line 1850: Simplify it, `!Array.isArray(headers)` enough

snitin315:
```suggestion
      if (!Array.isArray(headers)) {
```

alexander-akait (approved):

rishabh3112 (approved):

PR #3850, opened 2021-09-16
chore: convert dev server to webpack plugin

rishabh3112:
- [ ] This is a **bugfix**
- [x] This is a **feature**

 WIP

### Motivation / Use-Case
Dev Server should be a plugin

### Breaking Changes
Nope

### Additional Info
N/A

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -36,7 +36,6 @@ class Server {
     this.webSocketProxies = [];
     this.sockets = [];
 
-    // TODO: remove this after plugin support is published
     if (compiler) {
       this.compiler = compiler;
     }
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -36,8 +36,9 @@ class Server {
     this.webSocketProxies = [];
     this.sockets = [];
 
+    // TODO: remove this after plugin support is published
     if (compiler) {
-      this.apply(compiler);
+      this.compiler = compiler
     }
   }
 
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -15,6 +15,8 @@ if (!process.env.WEBPACK_SERVE) {
   process.env.WEBPACK_SERVE = true;
 }
 
+const pluginName = "webpack-dev-server";
+
 class Server {
   constructor(options = {}, compiler) {
     // TODO: remove this after plugin support is published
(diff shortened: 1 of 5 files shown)


rishabh3112:
@alexander-akait let me know if there any custom hooks required to be implemented here.

test/server/Server.test.js
```diff
@@ -50,6 +50,17 @@ describe("Server", () => {
...
+    it("should accept plugin API", (done) => {
+      const compiler = webpack(config);
+      const server = new Server(baseDevConfig);
+      server.apply(compiler);
+
+      compiler.hooks.done.tap("webpack-dev-server", () => {
+        expect(true).toBe(true);
+        server.stopCallback(done);
+      });
+    });
```
alexander-akait, line 62: We need e2e tests, no need to use unit tests

lib/Server.js
```diff
@@ -35,7 +35,16 @@ class Server {
     // Keep track of websocket proxies for external websocket upgrade.
     this.webSocketProxies = [];
     this.sockets = [];
+
+    // TODO: remove this after plugin support is published
+    if (compiler) {
```
alexander-akait, line 40: I think we should keep it here to avoid breaking change

rishabh3112: Sure, will remove comment.

test/server/Server.test.js
rishabh3112: Done!

lib/Server.js
rishabh3112: Done.

lib/Server.js
```diff
@@ -126,6 +131,20 @@ class Server {
     return path.resolve(dir, "node_modules/.cache/webpack-dev-server");
   }
 
+  apply(compiler) {
+    this.compiler = compiler;
+
+    let started = false;
+
+    compiler.hooks.watchRun.tapPromise(pluginName, async () => {
```
rishabh3112, line 139: We should enable `watch` before taping into `watchRun` hook.

alexander-akait: Yes, plugin should work only for `webpack watch`/`webpack serve`, but I am afraid it is so late to override fs, need think more about it

rishabh3112: ping @alexander-akait

alexander-akait: Yep, in my roadmap on the next week

bjohansebas:
fixed by #5650

PR #3867, opened 2021-09-22
fix: display file name in overlay

githoniel:
- [X] This is a **bugfix**
- [ ] This is a **feature**

since there is no test for overlayer yet, so I did not add test for it
### Motivation / Use-Case

we are not display filename in webpack5 for error now

bug

compare to error ovelay in webpack-dev-server V3

fixed

### Breaking Changes

no

### Additional Info

add filename for error

### DIFF
--- a/client-src/overlay.js
+++ b/client-src/overlay.js
@@ -130,9 +130,9 @@ function show(messages, type) {
     messages.forEach((message) => {
       const entryElement = document.createElement("div");
       const typeElement = document.createElement("span");
-
+      // ts-loader will output `error.file` for error file path, not in message
       typeElement.innerText =
-        (type === "warnings" ? "Warning:" : "Error:") + message.file || "";
+        (type === "warnings" ? "Warning:" : "Error:") + (message.file || "");
       typeElement.style.color = `#${colors.red}`;
 
       // Make it look similar to our terminal.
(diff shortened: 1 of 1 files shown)


alexander-akait:
Please update tests

githoniel:
> Please update tests

should I add test for it or just fix the lint?

snitin315:
> > Please update tests
> 
> should I add test for it or just fix the lint?

Both, CI should be all green.

client-src/overlay.js
```diff
@@ -131,7 +131,8 @@ function show(messages, type) {
       const entryElement = document.createElement("div");
       const typeElement = document.createElement("span");
 
-      typeElement.innerText = type === "warnings" ? "Warning:" : "Error:";
+      typeElement.innerText =
+        (type === "warnings" ? "Warning:" : "Error:") + message.file || "";
```
snitin315, line 135: I think `message.file` is `undefined` for webpack 4. So we should only add it when it's defined.

githoniel: I have add comment for it.  file only work for typescript

snitin315 (commented):
I think we should add a test case for this in https://github.com/webpack/webpack-dev-server/blob/master/test/e2e/overlay.test.js

alexander-akait:
@snitin315 Yep, I think we can merge it, I will send test cases today, maybe we need more fields here

snitin315 (approved):

PR #3905, opened 2021-09-30
perf: remove assets from stats

markjm:
- [ x ] This is a **bugfix**

no, I am willing to, but am unsure exactly where this functionality is currently being tests

### Motivation / Use-Case

stats.toJson call is more expensive with `assets` enabled, but it is not needed. Instead, we still keep track of hash so as to keep to only refresh the page when there is a real change (saving file with no changes will not refresh the page)

Before
image

After
image

### Breaking Changes

none

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -36,13 +36,14 @@ class Server {
     this.webSocketProxies = [];
     this.sockets = [];
     this.compiler = compiler;
+    this.currentHash = null;
   }
 
   static get DEFAULT_STATS() {
     return {
       all: false,
       hash: true,
-      assets: true,
+      assets: false,
       warnings: true,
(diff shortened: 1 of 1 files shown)


markjm:
cc @alexander-akait here are the changes with snippets of the traces. Saves a little under 100ms in my project

lib/Server.js
```diff
@@ -36,13 +36,14 @@ class Server {
...
     this.compiler = compiler;
+    this.currentHash = null;
   }
 
   static get DEFAULT_STATS() {
     return {
       all: false,
       hash: true,
-      assets: true,
+      assets: false,
```
alexander-akait, line 46: we can omit this because `all: false`

PR #3906, opened 2021-10-02
fix: perf

alexander-akait:
- [x] This is a **bugfix**
- [ ] This is a **feature**

it is just refactor, we don't need assets info here

### Motivation / Use-Case

fixes #3905

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -36,13 +36,13 @@ class Server {
     this.webSocketProxies = [];
     this.sockets = [];
     this.compiler = compiler;
+    this.currentHash = null;
   }
 
   static get DEFAULT_STATS() {
     return {
       all: false,
       hash: true,
-      assets: true,
       warnings: true,
       errors: true,
       errorDetails: false,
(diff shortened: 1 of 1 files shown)


lib/Server.js
```diff
@@ -2011,15 +2011,15 @@ class Server {
       stats &&
       (!stats.errors || stats.errors.length === 0) &&
       (!stats.warnings || stats.warnings.length === 0) &&
-      stats.assets &&
-      stats.assets.every((asset) => !asset.emitted);
+      this.currentHash === stats.hash;
```
alexander-akait, line 2014: @markjm I remove unnecessary `??` check, because it is redundant, when client connected to dev server we always send stats (so we write currentHash), on initial run we send stats too and write hash (client can connected, but compilation can be still not finished, so after it will be finished we should always send stats, it will be never `still-ok`)

alexander-akait: Also don't forget about disconnected, developer can run dev server again and compilation can take time (i.e. client connected, but no stats right now) and we should send stats always too

snitin315 (approved):

markjm:
Looks good, thanks! Also confirming after patching in these changes he's to our large project we continue to see perf gains. 80th percentile recompile is around 10s for us, and we are really trying to squeeze out more perf wherever we can! Great to see small improvements like this acknowledged and acted upon!

PR #3912, opened 2021-10-04
feat: add `client.reconnect` option

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

New option `client.reconnect: boolean | number` : allows to enable/disable/count of auto connect when client is disconnected

### Breaking Changes
None

### Additional Info
No


lib/Server.js
```diff
@@ -474,6 +478,13 @@ class Server {
         };
       }
 
+      if (
+        typeof options.client.reconnect === "undefined" ||
+        options.client.reconnect === true
```
alexander-akait, line 483: `true` should not have count of re connection, i.e.
- true - unlimited reconnections
- false - no reconnections
- number - count of reconnections

lib/Server.js
```diff
@@ -1627,6 +1638,10 @@ class Server {
         this.sendMessage([client], "progress", this.options.client.progress);
       }
 
+      if (this.options.client && this.options.client.reconnect) {
+        this.sendMessage([client], "reconnect", this.options.client.reconnect);
+      }
```
alexander-akait, line 1643: We need handler in our client for this like for `progress`/etc, so developers can change value using web socket API

lib/Server.js
```diff
@@ -474,6 +478,14 @@ class Server {
         };
       }
 
+      if (typeof options.client.reconnect === "undefined") {
+        options.client.reconnect = 10;
+      }
+
+      if (options.client.reconnect === true) {
```
alexander-akait, line 485: `else if`

alexander-akait (commented):
Looks good, one note above, and let's add tests

alexander-akait:
Thanks

snitin315:
@alexander-akait sounds good, I will implement 👍

PR #3913, opened 2021-10-04
fix: schema for web socket server type

alexander-akait:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

`false` is allowed only on for whole option, not for type

### Breaking Changes

No, because it was broken before

### Additional Info

No


snitin315 (commented):
Good catch. Let's fix CI.

snitin315 (approved):

Issue #3794, opened 2021-09-04
reload browser when hash is changed but no modules was updated


alexander-akait:
@alexander-akait Thanks for the feedback!

I have completely removed the `staticOptions` and setup the `directory` for my static files. I also removed `wds` + `src` watch content base (old) option. After that, all edited twig files triggers a new compilation, but still no browser reload. With `hot: false`, it works, weird :thinking:

Then I have set the `watchFiles` option to `['src/**/*.twig']` with `hot: true`: it works!
Every file change trigger a compilation and browser reload.

Is that configuration looks fine to you or should I share my private repo with you for further check?
Thanks a lot mate :wink:

_Originally posted by @xavierfoucrier in https://github.com/webpack/webpack-dev-server/issues/2758#issuecomment-902289933_

alexander-akait:
@xavierfoucrier but here some limitations, if you have `copy-webpack-plugin` or plugins which added files to compilation, hash was changed too, but not each file require reload browser...

xavierfoucrier:
Sure. In my case I will probably drop the `copy-webpack-plugin`, for now I will stick with the `watchFiles` option and wait to test this feature to be implemented.

liby:
@alexander-akait 
I'd like to ask a digression. If you just want to enable Live Reload, how to configure it? Like this?
```js
module.exports = {
  //...
  devServer: {
    static: {
      directory: path.join(__dirname, "dist"),,
      // watch: true,
    },
    watchFiles: ["src/**/*"],
    hot: false,
  },
};
```
The above configuration works, but I still have a few questions:
1. Do we need to set `watch: true`? Does it conflict with `watchFiles`? After all, if the file changes, the project will be rebuilt and the "dist" file will naturally change as well.
2. Similar to the first point, in my case, is it enough to set only one of `devServer.static.watch` and `devServer.watchFiles`?
3. The reason for setting `hot: false` is that without this line, the console would output the following:
    ```
    [HMR] Waiting for update signal from WDS... index.js:48 
    [WDS] Hot Module Replacement enabled.
    ```
    I don't have the HMR plugin installed, and I don't know why the default value for `hot` is not `false`, so would it be better for me to set it explicitly?

alexander-akait:
As was described in migration guide, you should use `watchFiles`, webpack don't know is this modules with hot or just copied file

alexander-akait:
[...]

PR #3972, opened 2021-10-23
fix: fix live reload for child compiler scenario

### LINKED ISSUE #3971: live reload does not works when the change is happened on entry of child compiler
### Bug report

I have a plugin which creates a child compiler and use a different entry for the compilation. It happens at parent's `compilation.hooks.processAssets` at `Compilation.PROCESS_ASSETS_STAGE_SUMMARIZE` stage . When I modify the child entry, it does not have a live reload but webpack did re-compile.

### Actual Behavior
webpack did re-compile but webpack-dev-server does not have a live reload when the change is done in the child entry 

### Expected Behavior

webpack-dev-server should have reload when child compilation has change

### How Do We Reproduce?

I have a fix ready. See my PR

wood1986:
fix #3971

- [X] This is a **bugfix**
- [ ] This is a **feature**

Currently, webpack-dev-server only cares `stats.hash` but this hash does not count child compilation when you run the childCompiler after parent's `seal` stage.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -1184,7 +1184,7 @@ class Server {
       result.hash = `${result.hash || ""}|${statsJson.hash}`;
       result.errors = [...(result.errors || []), ...statsJson.errors];
       result.warnings = [...(result.warnings || []), ...statsJson.warnings];
-      statsJson.children.forEach((child) => {
+      (statsJson.children || []).forEach((child) => {
         reduceFn(child, result);
       });
       return result;
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -42,6 +42,7 @@ class Server {
   static get DEFAULT_STATS() {
     return {
       all: false,
+      children: true,
       hash: true,
       warnings: true,
       errors: true,
@@ -1179,7 +1180,17 @@ class Server {
       stats.warningsFilter = compilerOptions.stats.warningsFilter;
     }
 
-    return statsObj.toJson(stats);
+    const reduceFn = (statsJson, result = {}) => {
+      result.hash = `${result.hash || ""}|${statsJson.hash}`;
+      result.errors = [...(result.errors || []), ...statsJson.errors];
+      result.warnings = [...(result.warnings || []), ...statsJson.warnings];
+      statsJson.children.forEach((child) => {
+        reduceFn(child, result);
+      });
+      return result;
+    };
+
+    return reduceFn(statsObj.toJson(stats));
   }
 
   setupHooks() {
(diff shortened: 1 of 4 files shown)


wood1986:
Hey @alexander-akait, may I know if you are still looking at this issue?

alexander-akait:
@wood1986 Yes, sorry for delay, tomorrow I will look, some busy

wood1986:
Thanks for letting me know!! FYI, I can confirm that it was working before webpack-dev-server 4.0.0

alexander-akait:
@wood1986 Can you provide example for v3?

wood1986:
Here is https://github.com/wood1986/demo/tree/wds3

I have just found. It always has auto reload when it is a new change. However when you undo the previous saved change, it will not have auto reload. I cannot reproduce in own repository but I can find it in this repository

```
`->` means "Save"

1st change ->                         reload
1st change -> 2nd change ->           reload
1st change -> undo 1st change ->      no reload
```

alexander-akait:
@wood1986

> Here is https://github.com/wood1986/demo/tree/wds3

> I have just found. It always has auto reload when it is a new change. However when you undo the previous saved change, it will not have auto reload. I cannot reproduce in own repository but I can reproduce it in this repository


I see, this related to https://github.com/webpack/webpack-dev-server/issues/3794, due `hot: true` by default (I think we really need to fix it), i.e. try to hot, if no modules, check hash, is hash was changed, do live reload

Generally we have two problems here:
1. `hot` is `true` by default and we need fix a bug described above
2. need to update hash of compilation (WIP on solution for you)

alexander-akait:
Oh, I see running `childCompiler.runAsChild` inside `compilation.hooks.processAssets` is too late, we already calculate `hash` of `compilation` (main) here, so we need update it manually, just do it:
```
compilation.hash = `${compilation.hash}${childCompilation.hash}`;
```

Should fix the problem, feel free to feedback

wood1986:
I did mention this fix

See
```js
parentCompilation.fullHash = parentCompilation.fullHash + childCompilation.fullHash
parentCompilation.hash = parentCompilation.fullHash
 ```

but it is not a native webpack hash. And do you think if other 3rd parties will have some logic related to this field `compilation.hash`?

alexander-akait:
@wood1986 It is valid solution, for multi compiler mode we do the same, just concat two hashes in string, we have this logic https://github.com/webpack/webpack-dev-server/blob/master/client-src/utils/reloadApp.js#L12, i.e. `indexOf`, so invalidation works fine

> but it is not a native webpack hash. And do you think if other 3rd parties will have some logic related to this field compilation.hash?

Hash is just string and can have any length (types), so even a plugin will have logic for `hash` it should not broken

wood1986:
It works. Thanks!!! I am curious the multi compiler mode implementation. Is it possible to show me the which files have hash concat?

alexander-akait:
@wood1986 All used modules, you can track it in `creatHash` and look at used values

alexander-akait:
@wood1986 Can we close?

PR #3998, opened 2021-11-01
fix: invalid host message is missing on clientwith https (#3997)

### LINKED ISSUE #3997: HTTPS clients do not receive Invalid Host/Origin header errors, they just get disconnected with no data
### Bug report

The `Invalid Host/Origin header` error frame does not get sent to a client that is connecting using HTTPS.

This is due to `client.terminate()` call after sending the message, which I think does not allow sufficient time for the frame containing the error to get sent, presumedly slowed down by the TLS layer, so the client never receives it. (Terminate calls `socket.destroy` under the hood so assumes all data has been sent already.)

Changing this line from `client.terminate()` to `client.close()` resolves the issue: https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L1636

I would raise a PR but I'm unsure if you would want timeout handling here too to ensure it does not linger around for a bad client - but since this is a development tool that perhaps is fine. Also I think there's likely no timeout handling anywhere else so perhaps this one line change is fine?

### Actual Behavior

With a non-HTTPS connection, and deliberately creating an incorrect Host/Origin (by loading index.html from a different server that then loads bundles from http://localhost:8080 as an example) - you receive the message:

    [webpack-dev-server] Invalid Host/Origin header
    [webpack-dev-server] Disconnected!

With a HTTPS connection, using same configuration as above, you receive only:

    [webpack-dev-server] Disconnected!

### Expected Behavior

Same result on both - specifically the Invalid Host/Origin header message should appear in the console.

### How Do We Reproduce?

(Will look at getting a basic example together if needed - but I could reproduce this with two different large configurations so I believe would apply to any configuration, caveat maybe needing similar environment to me, as long as you make the error occur and run HTTPS.)

Google Chrome Version 95.0.4638.54

  System:
    OS: macOS 12.0.1
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 41.94 MB / 16.00 GB
  Binaries:
    Node: 14.18.1 - ~/.nvm/versions/node/v14.18.1/bin/node
    npm: 6.14.15 - ~/.nvm/versions/node/v14.18.1/bin/npm
  Browsers:
    Chrome: 95.0.4638.69
    Firefox: 89.0.2
    Safari: 15.1
    Safari Technology Preview: 15.4
  Packages:
    babel-loader: ^8.1.0 => 8.2.2
    clean-webpack-plugin: ^4.0.0 => 4.0.0
    css-loader: ^6.4.0 => 6.4.0
    css-minimizer-webpack-plugin: ^3.1.1 => 3.1.1
    file-loader: ^6.1.1 => 6.2.0
    html-webpack-exclude-assets-plugin: 0.0.7 => 0.0.7
    html-webpack-plugin: ^5.4.0 => 5.4.0
    imports-loader: ^3.0.0 => 3.0.0
    postcss-loader: ^6.2.0 => 6.2.0
    sass-loader: ^12.2.0 => 12.2.0
    script-loader: ^0.7.2 => 0.7.2
    style-loader: ^3.3.0 => 3.3.0
    terser-webpack-plugin: ^5.2.4 => 5.2.4
    webpack: ^5.61.0 => 5.61.0
    webpack-bundle-analyzer: ^4.5.0 => 4.5.0
    webpack-cli: ^4.9.1 => 4.9.1
    webpack-dev-server: ^4.4.0 => 4.4.0
    webpack-merge: ^5.2.0 => 5.8.0

driskell:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Added a new test that successfully fails with `client.terminate`

### Motivation / Use-Case

See #3997. When host/origin headers do not match, `client.terminate` was being called. However, when HTTPS is enabled, this message is encrypted asynchronously by NodeJS before it gets written to the socket. Therefore, when `client.terminate` destroys the underlying socket the previous `sendMessage` had not yet been written to the socket.

Only applies when HTTPS is enabled.

Using `client.close` ensures that the socket is closed after all data is sent.

### Breaking Changes

None

### Additional Info

Granted, with a bad client, this could leak forever as there is no timeout. However, I do not see timeouts elsewhere so they could be added as part of adding them elsewhere. It's also a development tool that is unlikely to be very long lived, and will be extremely easy to restart upon any issue (always attended), so it's likely timeouts would unnecessarily overcomplicate the implementation with little benefit.

I also fixed loads of warnings I was getting running tests about EventEmitter leaks due to the signal listener that is registered when creating an instance. I can remove this code from this commit if it's a problem - it seemed very minor and fine to add as part of same commit - and either omit it entirely or raise separately for you.

Fixes #3997

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -1633,7 +1633,9 @@ class Server {
       ) {
         this.sendMessage([client], "error", "Invalid Host/Origin header");
 
-        client.terminate();
+        // With https enabled, the sendMessage above is encrypted asynchronously so not yet sent
+        // Terminate would prevent it sending, so use close to allow it to be sent
+        client.close();
 
         return;
       }
--- a/test/e2e/__snapshots__/allowed-hosts.test.js.snap.webpack4
+++ b/test/e2e/__snapshots__/allowed-hosts.test.js.snap.webpack4
@@ -278,6 +278,30 @@ Array [
 
 exports[`allowed hosts should disconnect web socket client using custom hostname from web socket server with the "auto" value based on the "host" header ("ws"): page errors 1`] = `Array []`;
 
+exports[`allowed hosts should disconnect web socket client using custom hostname from web socket server with the "auto" value based on the "host" header when "https" is enabled ("sockjs"): console messages 1`] = `
+Array [
+  "[HMR] Waiting for update signal from WDS...",
+  "Hey.",
+  "[webpack-dev-server] Invalid Host/Origin header",
+  "[webpack-dev-server] Disconnected!",
+  "[webpack-dev-server] Trying to reconnect...",
+]
+`;
+
+exports[`allowed hosts should disconnect web socket client using custom hostname from web socket server with the "auto" value based on the "host" header when "https" is enabled ("sockjs"): page errors 1`] = `Array []`;
+
+exports[`allowed hosts should disconnect web socket client using custom hostname from web socket server with the "auto" value based on the "host" header when "https" is enabled ("ws"): console messages 1`] = `
+Array [
+  "[HMR] Waiting for update signal from WDS...",
+  "Hey.",
+  "[webpack-dev-server] Invalid Host/Origin header",
+  "[webpack-dev-server] Disconnected!",
+  "[webpack-dev-server] Trying to reconnect...",
+]
+`;
+
+exports[`allowed hosts should disconnect web socket client using custom hostname from web socket server with the "auto" value based on the "host" header when "https" is enabled ("ws"): page errors 1`] = `Array []`;
+
 exports[`allowed hosts should disconnect web socket client using custom hostname from web socket server with the "auto" value based on the "origin" header ("sockjs"): console messages 1`] = `
 Array [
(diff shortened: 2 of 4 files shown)


snitin315 (dismissed):
Thanks

rishabh3112 (approved):

setupTest.js
```diff
@@ -3,3 +3,10 @@
 process.env.CHOKIDAR_USEPOLLING = true;
 
 jest.setTimeout(140000);
+
+global.afterEach(() => {
+  // We add listeners to signals when creating a new Server instance, but have no cleanup method
+  // So ensure they are removed after each run to prevent EventEmitter memory leak warnings
+  process.removeAllListeners("SIGINT");
+  process.removeAllListeners("SIGTERM");
+});
```
alexander-akait, line 12: Please remove it, it was memory leak on our side, it will be fixed in other PR

driskell: This is removed now.

alexander-akait (approved):

snitin315 (approved):

PR #4000, opened 2021-11-02
fix: schema for `server` option

snitin315:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

`server` can not be any string because we use it to require a node module. https://github.com/webpack/webpack-dev-server/blob/6d060ed68992875d82a0a77a2a5bf9d67d46817e/lib/Server.js#L1591

### Breaking Changes
No

### Additional Info
No

### DIFF
--- a/test/e2e/__snapshots__/server.test.js.snap.webpack4
+++ b/test/e2e/__snapshots__/server.test.js.snap.webpack4
@@ -624,6 +624,17 @@ exports[`server option as object spdy server with options should handle GET requ
 "
 `;
 
+exports[`server option as string custom-http should handle GET request to index route (/): console messages 1`] = `Array []`;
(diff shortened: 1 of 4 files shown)


lib/options.json
```diff
@@ -655,13 +652,6 @@
         "exclude": true
       }
     },
-    "ServerString": {
-      "type": "string",
-      "minLength": 1,
-      "cli": {
-        "exclude": true
-      }
-    },
```
alexander-akait, line 664: In theory developer can use `type: "own-server-package"` which compatibility with connect, so let's add test for this case (just reexport express for tests)

alexander-akait:
Thanks

PR #4006, opened 2021-11-04
fix: infinity refresh on warnings

### LINKED ISSUE #4005: Compilation failure causes infinite refresh
### Bug report

[Minimal reproducible repository](https://github.com/jeiea/webpack_report).

### Steps to reproduce

```
git clone https://github.com/jeiea/webpack_report.git
cd webpack_report
npm install
npm run start
```
- Open http://localhost:[port appearing on terminal]
- Add `#` at the end of `index.js` and save.

### Expected behavior

Error overlay appears, only once at the time of modification.

### Actual behavior

Error overlay appears, it refreshes very rapidly.

```

  System:
    OS: macOS 11.6
    CPU: (6) x64 Intel(R) Core(TM) i5-8500 CPU @ 3.00GHz
    Memory: 18.21 MB / 24.00 GB
  Binaries:
    Node: 14.15.1 - ~/.asdf/installs/nodejs/14.15.1/bin/node
    npm: 7.16.0 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 95.0.4638.69
    Firefox: 92.0
    Safari: 15.0
  Packages:
    babel-loader: ^8.2.3 => 8.2.3 
    eslint-webpack-plugin: ^3.1.0 => 3.1.0 
    html-webpack-plugin: ^5.5.0 => 5.5.0 
    webpack: ^5.61.0 => 5.61.0 
    webpack-cli: ^4.9.1 => 4.9.1 
    webpack-dev-server: ^4.4.0 => 4.4.0 
```

alexander-akait:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes, WIP

### Motivation / Use-Case

fixes #4005

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/client-src/index.js
+++ b/client-src/index.js
@@ -183,8 +183,6 @@ const onSocketMessage = {
     if (needShowOverlayForWarnings) {
       show("warning", warnings);
     }
-
-    reloadApp(options, status);
   },
   errors(errors) {
     log.error("Errors while compiling. Reload prevented.");
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,6 +54,8 @@
         "body-parser": "^1.19.0",
         "core-js": "^3.12.1",
         "css-loader": "^5.2.4",
+        "css-tree": "^1.1.3",
+        "csstree": "^0.0.3",
         "eslint": "^8.0.1",
         "eslint-config-prettier": "^8.3.0",
         "eslint-config-webpack": "^1.2.5",
@@ -5282,6 +5284,28 @@
         "url": "https://github.com/sponsors/fb55"
       }
     },
+    "node_modules/css-tree": {
+      "version": "1.1.3",
+      "resolved": "https://registry.npmjs.org/css-tree/-/css-tree-1.1.3.tgz",
+      "integrity": "sha512-tRpdppF7TRazZrjJ6v3stzv93qxRcSsFmW6cX0Zm2NVKpxE1WV1HblnghVv9TreireHkqI/VDEsfolRF1p6y7Q==",
+      "dev": true,
+      "dependencies": {
+        "mdn-data": "2.0.14",
+        "source-map": "^0.6.1"
+      },
+      "engines": {
+        "node": ">=8.0.0"
+      }
+    },
+    "node_modules/css-tree/node_modules/source-map": {
+      "version": "0.6.1",
+      "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.6.1.tgz",
+      "integrity": "sha512-UjgapumWlbMhkBgzT7Ykc5YXUT46F0iKu8SGXq0bcwP5dz/h0Plj6enJqjz1Zbq2l5WaqYnrVbwWOWMyF3F47g==",
+      "dev": true,
+      "engines": {
+        "node": ">=0.10.0"
+      }
+    },
     "node_modules/css-what": {
       "version": "5.1.0",
       "resolved": "https://registry.npmjs.org/css-what/-/css-what-5.1.0.tgz",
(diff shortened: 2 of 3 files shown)


obsius:
@alexander-akait Were the ramifications of this change intentional?  I upgraded an old deployment recently and this change was counter to the old behavior and also my expectation that warnings would not prevent an HMR update or full page reload.

For example, the log message for an error states:
`Errors while compiling. Reload prevented.`
this is not the case for warnings:
`Warnings while compiling.`

Was the intent to only prevent this reload when the overlay is used?

```js
    if (needShowOverlayForWarnings) {
      show("warning", _warnings);
    } else {
      reloadApp(options, status);
    }
```

alexander-akait:
Yes, expected, we should not do refresh on warnings, please provide reproducible example and I will show how to fix

obsius:
@alexander-akait Ok, but what I'm saying is that this behavior (calling `reloadApp` on warnings)  has been around for a long time, at least 3 years:

image

I think this change should be communicated somehow, or perhaps a configuration setting to allow the old behavior made accessible.

alexander-akait:
@obsius Can you provide example of the problem? We should not add new options on every problem, if it is bug, we should fix it, there are a lot of changes over the past three years, as you can see keeping do infinity reload, and we can't just revert it

obsius:
@alexander-akait This PR does not contain a bug, but does present a significant change to how the webpack-dev-server client works.  Anyone expecting HMR or page reloading after a compilation with warnings (the old behavior) may think their configuration is wrong or that some other plugin is in error.

I haven't looked at this repo before today, but from the commit history it seems that the old behavior was stable for a very long time, so it would make sense to let developers know of this sudden change.

For example, the log message for an error states:
`Errors while compiling. Reload prevented.`

Now that a compilation with warnings does not reload, the warning log message could be updated to this:
`Warnings while compiling. Reload prevented.`

This new behavior could also be mentioned in the release notes.

**An example of this issue would be:**

*Developer has turned off the overlay for warnings.*

Before this change:
- `webpack-dev-server` sends message `{ type: "warnings", ... }`
- `webpack-dev-server` client prints warnings to console
- `webpack-dev-server` client calls `reloadApp`
- developer sees the page page update

After this change:
- `webpack-dev-server` sends message `{ type: "warnings", ... }`
- `webpack-dev-server` client prints warnings to console
- `webpack-dev-server` client does ***not***  call `reloadApp`
- developer wonders why the page is not updating

alexander-akait:
@obsius Again, please create reproducible test repo, sending warnings should not prevent reloading, if hash was changed, we should do reload

alexander-akait:
If you send warnings without hash changes, it means nothing was changed in compilation, so no reasons to reload page

obsius:
@alexander-akait Since you agree that a reload is desired after compilation with warnings, perhaps you can help me understand how a hash change propagates into a reload in the client?

The listener in the client for `hash` does not seem to call `reloadApp`:
```js
  hash(hash) {
    status.previousHash = status.currentHash;
    status.currentHash = hash;
  },
```

The only place where `reloadApp` is called in version 4.5.0 is in the `ok` handler, but the `ok` message type is not sent from the server when a compilation produces warnings.  In the previous version, `reloadApp` was called from both `ok` and `warnings`.

alexander-akait:
dev server connected with compilation and send ws message from server to client with hash

obsius:
@alexander-akait Here is the server log for a compilation that produces warnings:

image

(1) `invalid` is sent when compilation begins
(2) `hash` is sent after compilation (something has changed)
(3) `warnings` is sent after hash

What I am trying to explain to you, but for some reason is being lost in communication, is that the client is not updating after `warnings` anymore. I don't see any mechanism in the client code to respond solely to the `hash` message, so what I am asking you to help me understand is how `reloadApp` gets called in the client after a compilation that produced warnings.  It no longer seems possible with this latest change.

alexander-akait:
Please create reproducible example, if you need help ASAP, it is the best choice

obsius:
@alexander-akait I don't need help ASAP, I'm trying to explain that this PR prevents HMR and page reloads after compilations with `warnings`. You're not taking my comments seriously.  I'm asking a very straightforward question to you and you are being obstinate.  I cannot make this issue any clearer to you at this point.  You removed the only place besides the `ok` handler where `reloadApp` is called.  The server does not send an `ok` message after a compilation with warnings, so HMR and page reloads no longer happen in this scenario.

alexander-akait:
@obsius Open an original issue and look at infinity loop, it was fixed here, we send hash on each compilation, no hash changes = no reloading, no blockers on client.

And stop blaming me, I'm trying to figure out what the problem is and request additional information, but you ignore my request. I am a developer like you, and I spend my free time on this. I believe that we may have a problem, I do not argue and do not claim that you are wrong, all I ask is to show an example.

If you have idea how to fix - PR welcome

obsius:
@alexander-akait Don't mean to blame you, and appreciate the work you are doing on this. I'm unfamiliar with this project and don't know what changes are desired, which is why I thought this was an undocumented change at first. I cross referenced an issue that may be related and has a repo linked.

PR #4013, opened 2021-11-08
fix: remove process listeners after stopping the server

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

fix memory leak warnings in e2e tests.

Screenshot 2021-11-07 at 5 41 57 PM

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -2258,6 +2258,11 @@ class Server {
         this.middleware = null;
       }
     }
+
+    // We add listeners to signals when creating a new Server instance
+    // So ensure they are removed to prevent EventEmitter memory leak warnings
+    process.removeAllListeners("SIGINT");
+    process.removeAllListeners("SIGTERM");
   }
 
   stopCallback(callback) {
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -32,6 +32,7 @@ class Server {
 
     this.options = options;
     this.staticWatchers = [];
+    this.listeners = [];
     // Keep track of websocket proxies for external websocket upgrade.
     this.webSocketProxies = [];
     this.sockets = [];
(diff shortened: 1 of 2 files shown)


setupTest.js
```diff
@@ -3,3 +3,12 @@
...
 jest.setTimeout(140000);
+
+if (expect.getState().testPath.includes("test/e2e")) {
+  afterEach(() => {
+    // We add listeners to signals when creating a new Server instance, but have no cleanup method
+    // So ensure they are removed after each run to prevent EventEmitter memory leak warnings
+    process.removeAllListeners("SIGINT");
+    process.removeAllListeners("SIGTERM");
+  });
+}
```
alexander-akait, line 14: hm, it is because we attach to `process`? Maybe we should do it in server in close

snitin315: Yes, we should move it to the `stop` method.

alexander-akait: Let's do it

lib/Server.js
```diff
@@ -2258,6 +2258,11 @@ class Server {
         this.middleware = null;
       }
     }
+
+    // We add listeners to signals when creating a new Server instance
+    // So ensure they are removed to prevent EventEmitter memory leak warnings
+    process.removeAllListeners("SIGINT");
+    process.removeAllListeners("SIGTERM");
```
alexander-akait, line 2265: We should not remove all of them just only our, move our close function above the file and use `removeListener`

snitin315: above where?

alexander-akait: I.e. just move it in function above the Server class and remove this listener in close

PR #4025, opened 2021-11-14
feat: allow to pass all options from `chokidar`

snitin315:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Updated snapshots.

### Motivation / Use-Case

 - Better defaults 
 - allow to pass all options from `chokidar` for `static.watch` and `watchFiles.options`
 

### Breaking Changes
None

### Additional Info
No


lib/Server.js
```diff
@@ -2143,17 +2167,16 @@ class Server {
...
-      ignoreInitial: true,
-      persistent: true,
-      followSymlinks: false,
-      atomic: false,
-      alwaysStat: true,
-      ignorePermissionErrors: true,
+      ...watchOptions,
       ignored: watchOptions.ignored,
       usePolling,
       interval,
```
alexander-akait, line 2173: Please add tests there we override `usePolling` and `interval`

lib/Server.js
```diff
@@ -445,13 +445,22 @@ class Server {
...
     const watchOptions = compilerOptions.watchOptions || {};
+    const defaultWatchOptions = {
+      ignoreInitial: true,
+      persistent: true,
+      followSymlinks: false,
+      atomic: false,
+      alwaysStat: true,
+      ignorePermissionErrors: true,
+      ...watchOptions,
+    };
```
alexander-akait, line 456: Let's move normalization from https://github.com/webpack/webpack-dev-server/blob/b4a0cfd326cfa73a4c932254244df44a3deec1a1/lib/Server.js#L2156 to here, wecan calculate default `usePolling` and `interval` value here

lib/Server.js
```diff
@@ -444,16 +444,49 @@ class Server {
...
+        ...compilerWatchOptions,
+        ...watchOptions,
+        ignored: watchOptions.ignored,
+        usePolling,
+        interval,
+      };
+
+      if (finalWatchOptions.poll) {
+        delete finalWatchOptions.poll;
+      }
```
alexander-akait, line 483: Avoid it, you already know about `poll` is not valid for chokidar, only pass valid options

snitin315: `poll` can also come from `compilerWatchOptions`.

alexander-akait: Just not put `poll` in your default options

lib/Server.js
```diff
@@ -464,24 +468,23 @@ class Server {
           : // eslint-disable-next-line no-undefined
             undefined;
 
+      if (watchOptions.poll) {
+        delete watchOptions.poll;
+      }
```
alexander-akait, line 473: Why not get only valid options from `watchOptions`?

snitin315: Because we respect `poll` option for normalizing `usePolling` and `interval` option.

alexander-akait: Using this will modify webpack compiler options, you should not touch them

alexander-akait: You need `Default Options` + `Options from watchOptions` + `DeveloperOptions`, so developer can override even `usePolling` and `interval` (but by default we get them from `watchOptions`) and other

snitin315: Here we are not modifying `compiler.options.watchOptions`, only `devServer.static.watch` and `devServer.watchFiles.options`

snitin315: got it.

snitin315: Previously we were also respecting `devServer.static.watch.poll` which is not valid.

alexander-akait: We should respect them too, i.e.:
```
{
  ...defaultOptions,
  ...watchOptions
  ...developerOptionsFroChokidar
}
```

snitin315: This will pass `poll` to `chokidar` when developer defines `devServer.static.watch.poll`.

PR #4038, opened 2021-11-18
chore: regex free isAbsoluteURL

### LINKED ISSUE #4037: Alternative isAbsoluteURL solution
https://github.com/webpack/webpack-dev-server/blob/497e615f24d5a16266f0faa8db1063f1b0ec2b82/lib/Server.js#L53-L62
```js
 static isAbsoluteURL(url) { 
   try { return new URL(url), true } catch(e) {}
   return false
 } 
```
then you can use both a URL class and a strings and weird domain names that are very special

jimmywarting:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

using existing tests

### Motivation / Use-Case

Domain names can be tricky with weird names... like punycode

### Breaking Changes

It parse url a bit differently now. It can throw when the url is invalid...

### Additional Info

closes #4037

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -52,7 +52,7 @@ class Server {
 
   static isAbsoluteURL(uri) {
     try {
-      new URL(uri); // eslint-disable-line
+      new URL(uri); // eslint-disable-line no-new
     } catch (err) {
       return false;
     }
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -50,15 +50,13 @@ class Server {
     };
   }
 
-  static isAbsoluteURL(URL) {
-    // Don't match Windows paths `c:\`
-    if (/^[a-zA-Z]:\\/.test(URL)) {
+  static isAbsoluteURL(uri) {
+    try {
+      new URL(uri); // eslint-disable-line
+    } catch (err) {
       return false;
     }
-
-    // Scheme: https://tools.ietf.org/html/rfc3986#section-3.1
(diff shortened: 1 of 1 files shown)


lib/Server.js
```diff
@@ -50,15 +50,13 @@ class Server {
     };
   }
 
-  static isAbsoluteURL(URL) {
-    // Don't match Windows paths `c:\`
-    if (/^[a-zA-Z]:\\/.test(URL)) {
+  static isAbsoluteURL(uri) {
+    try {
+      new URL(uri); // eslint-disable-line
```
snitin315, line 55: Let's only disable the specific rule.

snitin315:
@jimmywarting can you fix CI?

jimmywarting:
I'm not really sure what needs to be fixed...

snitin315:
The following two cases are failing.

https://github.com/webpack/webpack-dev-server/blob/a27b55389108a1b212fab5a39fe01b083fb27c94/test/e2e/static-directory.test.js#L521-L526

https://github.com/webpack/webpack-dev-server/blob/a27b55389108a1b212fab5a39fe01b083fb27c94/test/e2e/static-directory.test.js#L535-L540

alexander-akait:
Yes, I see, looks like we need check on `isWindows`

alexander-akait:
We decide do not change logic, because no benefit, please read https://github.com/webpack/webpack-dev-server/pull/4074, anyway thanks for PR

PR #4068, opened 2021-12-02
feat: add `setupMiddlewares` option

snitin315:
- [ ] This is a **bugfix**
- [X] This is a **feature**

Yes.

### Motivation / Use-Case

```js
setupMiddleware: (arrayOfMiddlewares, devServer) => {

  arrayOfMiddlewares.push('something');

  return arrayOfMiddlewares;
}
```

**TODO:** 

- [x] Deprecate `onAfterSetupMiddleware` and `onBeforeSetupMiddleware`.
- [x] E2E tests
- [x] Add examples in docs

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/examples/setup-middlewares/README.md
+++ b/examples/setup-middlewares/README.md
@@ -1,0 +1,40 @@
+# setupMiddlewares
+
+Provides the ability to execute a custom function and apply custom middleware(s).
+
+**webpack.config.js**
+
+```js
+module.exports = {
+  // ...
+  devServer: {
+    setupMiddlewares: (middlewares, devServer) => {
+      if (!devServer) {
+        throw new Error("webpack-dev-server is not defined");
+      }
+
+      const sendResponses = () => {
+        devServer.app.get("/setup-middleware/some/path", (_, response) => {
+          response.send("setup-middlewares option GET");
+        });
+      };
+
+      middlewares.push(sendResponses());
+
+      return middlewares;
+    },
+  },
+};
+```
+
+To run this example use the following command:
+
+```console
+npx webpack serve --open
+```
+
+## What Should Happen
+
+1. The script should open `http://localhost:8080/` in your default browser.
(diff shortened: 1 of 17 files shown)


alexander-akait:
Refactored and updated examples + improve perf for `setHeaders`, we don't need to attach middleware if we don't have headers

alexander-akait:
Also refactor some places, so we now try to use `connect` [insta](https://github.com/senchalabs/connect) instead express and maybe even other servers, so in future we can migrate on any application

alexander-akait:
@snitin315 Ideally we need add new options `server: { type: "connect", options: ...options }`, `type` can be "string" (i.e. package name or `require.resolve("connect")`) or `function`, and ideally we should tests all our functions on compatibility with connect - I think we can create new CI pipeline and setup tests to using `connect` instead `express` using jest aliases

snitin315:
> @snitin315 Ideally we need add new options `server: { type: "connect", options: ...options }`, `type` can be "string" (i.e. package name or `require.resolve("connect")`) or `function`, and ideally we should tests all our functions on compatibility with connect - I think we can create new CI pipeline and setup tests to using `connect` instead `express` using jest aliases

Yes, I think we can create a separate pipeline for `connect`. I will try adding one.

alexander-akait:
@snitin315 I was wrong, we need `application: { type: "connect", options: ...options }` (default value is `express`) in long term we should migrate on `connect` or maybe something more simple, express is not get updated a long time, but we need to support `http2`/`https` without any problems and workarounds, I think now we can setup our tests and fix code for `connect` and them add this options, so we can have two-three PRs

Airkro:
The API is not easy to understand like before, what's the advantage of `setupMiddlewares`?

alexander-akait:
What is misleading? It was implemented to allow developers use own middlewares and even replaces built-in

PR #4074, opened 2021-12-03
refactor: update logic for `isAbsoluteURL` method

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

Fix https://github.com/webpack/webpack-dev-server/issues/4037
Fix https://github.com/webpack/webpack-dev-server/pull/4038

### Breaking Changes

None

### Additional Info
No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -50,15 +50,18 @@ class Server {
     };
   }
 
-  static isAbsoluteURL(URL) {
-    // Don't match Windows paths `c:\`
-    if (/^[a-zA-Z]:\\/.test(URL)) {
+  static isAbsoluteURL(uri) {
+    try {
+      // eslint-disable-next-line no-new
+      new URL(uri);
+
+      // Don't match Windows paths `c:\`
+      const isWindowsPath = /^[a-zA-Z]:\\/.test(uri);
+
+      return !isWindowsPath;
+    } catch (error) {
       return false;
     }
-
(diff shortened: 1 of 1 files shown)


lib/Server.js
```diff
@@ -50,15 +50,18 @@ class Server {
...
+      const isWindowsPath = /^[a-zA-Z]:\\/.test(uri);
+
+      return !isWindowsPath;
+    } catch (error) {
       return false;
     }
-
-    // Scheme: https://tools.ietf.org/html/rfc3986#section-3.1
-    // Absolute URL: https://tools.ietf.org/html/rfc3986#section-4.3
-    return /^[a-zA-Z][a-zA-Z\d+\-.]*:/.test(URL);
```
alexander-akait, line 61: To be honestly I don't think it is faster :+1: Before we have one simple regexp, now we have `new URL(...)` and regexp...

snitin315: Hmm, So, how should we proceed here? Do you have any alternative in mind? Or should we close the related issue?

alexander-akait: I think we can keep code as is and do not modify it, it is not bottleneck for us

snitin315: Let's close it then 👍

alexander-akait: Yep, maybe we revisit it in future, but I don't see real problems with our check right now, we don't need to parse whole URL, we just need to check string is absolute URL

PR #4098, opened 2021-12-11
fix: update `--web-socket-server` description

snitin315:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

 update `--web-socket-server` description.

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/lib/options.json
+++ b/lib/options.json
@@ -1031,7 +1031,7 @@
           "$ref": "#/definitions/WebSocketServerType"
         }
       ],
-      "description": "Deprecated: please use 'webSocketServer.type'/'--web-socket-server-type' option."
+      "description": "Deprecated: please use '--web-socket-server-type' option."
     },
     "WebSocketServerFunction": {
       "instanceof": "Function"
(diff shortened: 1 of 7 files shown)


lib/options.json
```diff
@@ -1031,7 +1031,7 @@
           "$ref": "#/definitions/WebSocketServerType"
         }
       ],
-      "description": "Deprecated: please use 'webSocketServer.type'/'--web-socket-server-type' option."
+      "description": "Deprecated: please use '--web-socket-server-type' option."
```
anshumanv, line 1034: not sure if displaying the flag makes sense when the user is using the feature from config

snitin315: Ideally, we should have a `cliDescription` property which we should respect if available. https://github.com/webpack/webpack/blob/c854e30df7f879d9989a7ab7aa83d1f0d1fa6862/lib/cli.js#L97

alexander-akait (approved):

PR #4110, opened 2021-12-14
fix: allow to pass options for custom server

snitin315:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

We already allow `server: "own-server-package"`, so update schema to allow` server: { type: "own-server-package" }`

### Breaking Changes
None

### Additional Info
No

### DIFF
--- a/lib/options.json
+++ b/lib/options.json
@@ -667,7 +667,14 @@
       "type": "object",
       "properties": {
         "type": {
-          "$ref": "#/definitions/ServerType"
+          "anyOf": [
+            {
+              "$ref": "#/definitions/ServerType"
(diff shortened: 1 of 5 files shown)


alexander-akait:
Will be great if we will try to integrate `fastily` and pure `http` module there, just in tests and if it will be ok, provide examples, because we should move out from `express`

snitin315:
> Will be great if we will try to integrate `fastily` and pure `http` module there, just in tests and if it will be ok, provide examples, because we should move out from `express`

Yes, putting it in my TODO.

PR #4116, opened 2021-12-16
ci: update strategy matrix to run tests

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

Reduce CI time.
 
- Run all node versions for `ubuntu-latest`.
- Run `windows-latest` and `macos-latest` with node 16 only.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -63,8 +63,10 @@ jobs:
         include:
           - os: windows-latest
             node: "16.x"
+            webpack-version: latest
           - os: macos-latest
             node: "16.x"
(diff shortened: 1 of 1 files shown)


alexander-akait:
Better to test on all versions and platforms, from my experience problems sometimes happen on different Node.js versions and different patforms

snitin315:
Random failures on CI have increased recently, need to investigate.

alexander-akait:
There is no random, we wait pages on changes and collects logs, but in some places we have wrong logic (you can try to debug locally), so if CI is slow we have more or less logs

PR #4141, opened 2021-12-29
fix:Compatible with onAfterSetupMiddleware

### LINKED ISSUE #4140: Compatible with onAfterSetupMiddleware
Note that `app.get(...)` in `onAfterSetupMiddleware` is not the same as `app.get(...)` in `setupMiddlewares`.

It seems `app.get()` in `setupMiddlewares` are always applied **before** middlewares are applied, if we have a config like this:

```js
{
    onAfterSetupMiddleware: server => {
        server.app.get('/__index__.html', (req, res) => res.send('foo'));
    },
    config.historyApiFallback = {
        index: '/__index__.html',
        disableDotRule: true,
    };
}
```

The route `/__index__.html` will take place for history api fallback, however if we changed it to this:

```js
{
    setupMiddlewares: (middlewares, server) => {
        server.app.get('/__index__.html', (req, res) => res.send('foo'));
        return middlewares;
    },
    config.historyApiFallback = {
        index: '/__index__.html',
        disableDotRule: true,
    };
}
```

History api fallback are broken, we have to transform it into a middleware like:

```js
middlewares.unshift({
    path: '/__index__.html',
    handler: (req, res, next) => {
        if (req.method === 'GET') {
            res.send('foo');
        }
        else {
            next();
        }
    }
});
```

In this way we have to add a `request.method` check in middleware implementation.

_Originally posted by @otakustay in https://github.com/webpack/webpack-dev-server/issues/4129#issuecomment-1000897903_

godky:
- [x] This is a **bugfix**
- [ ] This is a **feature**

### Additional Info
close #4140

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -2375,10 +2375,6 @@ class Server {
       });
     }
 
-    if (typeof this.options.onAfterSetupMiddleware === "function") {
-      this.options.onAfterSetupMiddleware(this);
-    }
-
     if (typeof this.options.setupMiddlewares === "function") {
       middlewares = this.options.setupMiddlewares(middlewares, this);
     }
@@ -2395,6 +2391,10 @@ class Server {
         (this.app).use(middleware.middleware);
       }
     });
+    
+    if (typeof this.options.onAfterSetupMiddleware === "function") {
+      this.options.onAfterSetupMiddleware(this);
+    }
   }
 
   /**


rishabh3112 (changes requested):
Looks good, but need to make CI is failing.
A test for this would be great as well.

alexander-akait (approved):

alexander-akait:
We have failed CI not due this commit, I want to merge and restore CI in the other PR

Issue #4156, opened 2022-01-05
Client overlay/progress options not configurable via resource URL


jnu:
### Actual Behavior

When including the client manually via an entry point (instead of using the webpack devServer config object), the `overlay` and `progress` options are not configurable (`logging`, `hot`, and `liveReload` _are_ configurable).

### Expected Behavior

Expect an entry point configuration like:
```
{
  app: "./index.js",
  client: "webpack-dev-server/client/index.js?hot=false&live-reload=true&progress=true&overlay=true"
}
```
to set the `progress` and `overlay` client options to true, as it does for`liveReload`.

### How Do We Reproduce?

Use the above entrypoint config for Webpack, with the `devServer` config set as:
```
devServer: {
  client: false,
  hot: false,
}
```
See e.g. these docs for more info: https://webpack.js.org/guides/hot-module-replacement/#via-the-nodejs-api

### Please paste the results of `npx webpack-cli info` here, and mention other relevant information

FWIW I am configuring the `client` manually since I am exporting two Webpack configurations (one for the main app and one for the service worker). Webpack uses the first `devServer` config for *all* configs, but injecting the hot/client assets into the service worker causes issues. This is probably a separate bug to take up with webpack, but in the meantime this manual setup works around the issue, I'd just also still like to have the error overlay.

```
> $ npx webpack-cli                                                                                                                                                                                                  ⬡ 16.13.0 [±fixreload ●]
NudgeWeb:
  assets by path *.js 6.69 MiB
    asset nudge.js 6.42 MiB [emitted] (name: nudge) 1 related asset
    asset client.js 215 KiB [emitted] (name: client) 1 related asset
    asset hot.js 58.5 KiB [emitted] (name: hot) 1 related asset
  asset assets/favicon.png 86 KiB [emitted]
  asset nudge.png 86 KiB [emitted] [from: src/img/nudge.png] [copied]
  asset index.html 348 bytes [emitted]
  orphan modules 3.82 MiB [orphan] 2236 modules
  runtime modules 77.5 KiB 32 modules
  cacheable modules 5.63 MiB
    modules by path ./node_modules/ 5.52 MiB 658 modules
    modules by path ./src/ 111 KiB
      modules by path ./src/net/ 29.4 KiB 29 modules
      modules by path ./src/*.tsx 40.2 KiB 18 modules
      modules by path ./src/state/*.ts 14.9 KiB 5 modules
[...]

PR #4157, opened 2022-01-05
fix: allow more client options to be configured via resource url (#4156)

jnu:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes, added tests for parsing resource query options

### Motivation / Use-Case

See #4156

### Breaking Changes

No breaking changes.


client-src/index.js
```diff
@@ -58,6 +58,18 @@ if (parsedResourceQuery["live-reload"] === "true") {
...
+  options.progress = true;
+
+  log.info("Progress reporting enabled.");
+}
+
+if (parsedResourceQuery.overlay === "true") {
+  options.overlay = true;
+
+  log.info("Error overlay enabled.");
+}
```
alexander-akait, line 71: Overlay can be have more options, ideally we should support them too

alexander-akait: I think we can use `JSON.stringify`/`JSON.parse` to support more options

jnu:
ok sure, I will update this afternoon. thanks!

client-src/index.js
jnu: implemented this and added tests. tests are a little messy because the resource query parsing is a side-effect of the module import, but let me know if you have suggestions for cleaning them up.

client-src/index.js
```diff
@@ -58,6 +58,30 @@ if (parsedResourceQuery["live-reload"] === "true") {
...
+if (parsedResourceQuery.overlay) {
+  try {
+    options.overlay = JSON.parse(parsedResourceQuery.overlay);
+    log.info("Error overlay enabled.");
+  } catch (e) {
+    log.error("Error parsing overlay options from resource query:", e);
+  }
+
+  // Fill in default "true" params for partially-specified objects.
+  if (typeof options.overlay === "object") {
```
jnu, line 76: according to the [docs](https://webpack.js.org/configuration/dev-server/#overlay) the defaults should be true (though i don't see where this is implemented when using the overlay object options in the webpack config)

client-src/index.js
```diff
@@ -58,6 +58,30 @@ if (parsedResourceQuery["live-reload"] === "true") {
...
+if (parsedResourceQuery.progress === "true") {
+  options.progress = true;
+
+  log.info("Progress reporting enabled.");
+}
+
+if (parsedResourceQuery.overlay) {
+  try {
+    options.overlay = JSON.parse(parsedResourceQuery.overlay);
+    log.info("Error overlay enabled.");
```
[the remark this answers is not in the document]
jnu: done

snitin315 (commented):
I believe we also need to set these options automatically if provided in configuration, as we do for other client options like `logging`. 

https://github.com/webpack/webpack-dev-server/blob/e99c0bc4de371f9c3c2c14c0006d52ea104ad306/lib/Server.js#L596

PR #4163, opened 2022-01-11
fix: `ProxyConfigArray` in type def is not an `Array`

### LINKED ISSUE #4161: ProxyConfigArray in type def (types/lib/Server.d.ts) is not an Array
### Bug report

In the bundled type delcaration file "types/lib/Server.d.ts", type `ProxyConfigArray` is defined to not an array but what its item should be.

### Actual Behavior

```ts
type ProxyConfigArray = {
  path?: string | string[] | undefined;
  context?: string | string[] | HttpProxyMiddlewareOptionsFilter | undefined;
} & HttpProxyMiddlewareOptions &
  ByPass;
```

### Expected Behavior

Below codes are modified base on `@types/webpack-dev-server` codes:
```ts
type ProxyConfigArrayItem = {
    path?: string | string[] | undefined;
    context?: string | string[] | HttpProxyMiddlewareOptionsFilter | undefined;
} & HttpProxyMiddlewareOptions &
  ByPass;

type ProxyConfigArray = ProxyConfigArrayItem[];
```

### How Do We Reproduce?

```ts
import type { Configuration } from 'webpack-dev-server'
const apiPrefixes: string[] = ['/api/']
const devServerOptions: Configuration = {
  proxy: [
    {
      // TS2322: Type '{ context: string[]; target: string; }' is not assignable to type 'Options'.
      // Object literal may only specify known properties, and 'context' does not exist in type 'Options'.
      context: apiPrefixes,
      target : '',
    },
  ],
}
```

Not using webpack-cli.
using webpack-dev-server@4.7.2
using typescript@4.5.4

whitetrefoil:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No, it's a pure type declaration fix, no actual code.

### Motivation / Use-Case

Fixes #4161

### Breaking Changes

No breaking change.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -121,11 +121,15 @@ const schema = require("./options.json");
  * @callback ByPass
  * @param {Request} req
  * @param {Response} res
- * @param {ProxyConfigArray} proxyConfig
+ * @param {ProxyConfigArrayItem} proxyConfig
  */
 
 /**
- * @typedef {{ path?: string | string[] | undefined, context?: string | string[] | HttpProxyMiddlewareOptionsFilter | undefined } & HttpProxyMiddlewareOptions & ByPass} ProxyConfigArray
+ * @typedef {{ path?: string | string[] | undefined, context?: string | string[] | HttpProxyMiddlewareOptionsFilter | undefined } & HttpProxyMiddlewareOptions & ByPass} ProxyConfigArrayItem
+ */
+
+/**
+ * @typedef {ProxyConfigArrayItem[]} ProxyConfigArray
  */
 
 /**
@@ -194,7 +198,7 @@ const schema = require("./options.json");
  * @property {boolean} [http2]
  * @property {"http" | "https" | "spdy" | string | ServerConfiguration} [server]
  * @property {boolean | "sockjs" | "ws" | string | WebSocketServerConfiguration} [webSocketServer]
- * @property {ProxyConfigMap | ProxyConfigArray | ProxyArray} [proxy]
+ * @property {ProxyConfigMap | ProxyConfigArray | ProxyConfigArrayItem | ProxyArray} [proxy]
  * @property {boolean | string | Open | Array<string | Open>} [open]
  * @property {boolean} [setupExitSignals]
  * @property {boolean | ClientConfiguration} [client]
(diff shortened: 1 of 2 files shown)


alexander-akait (approved):

lib/Server.js
```diff
@@ -121,11 +121,15 @@ const schema = require("./options.json");
  * @callback ByPass
  * @param {Request} req
  * @param {Response} res
- * @param {ProxyConfigArray} proxyConfig
+ * @param {ProxyConfigArrayItem} proxyConfig
```
alexander-akait, line 124: I think something wrong here `ByPass` can't have `ByPass` inside

whitetrefoil: I haven't actually run the code yet, but looked at line 2222 it seemed to just pass the whole `proxyConfig` as-is:
```js
const isByPassFuncDefined =
  typeof proxyConfig.bypass === "function";
const bypassUrl = isByPassFuncDefined
  ? await proxyConfig.bypass(req, res, proxyConfig)
  : null;
```

alexander-akait: Can you provide configuration where we have a problem with types?

whitetrefoil: Sure~ I've put a minimial repro. in the original issue #4161:
```ts
import type { Configuration } from 'webpack-dev-server'
const devServerOptions: Configuration = {
  proxy: [
    {
      // TS2322: Type '{ context: never[]; target: string; }' is not assignable to type 'Options'.
      // Object literal may only specify known properties, and 'context' does not exist in type 'Options'.
      context: [],
      target : '',
    },
  ],
}
```
That code works when using the old `@types/webpack-dev-server`, but throws TS error when using `webpack-dev-server`'s new bundle types.  (And it actually works if I just ignores the TS type error)

alexander-akait: We have more errors - try https://github.com/webpack/webpack-dev-server/pull/4173

whitetrefoil: #4173 works for my case, should I close this one?

alexander-akait: Yep, let's close, merged https://github.com/webpack/webpack-dev-server/pull/4173, anyway thanks for PR

Issue #4180, opened 2022-01-14
Didn't handle preflight request correctly


TrickyPi:
### Bug report

Didn't handle preflight request correctly, if browser send a CORS non simple request to get a static resource.
### Actual Behavior

Occur a CORS error.
The specific info: 
Response to preflight request doesn't pass access control check: It does not have HTTP ok status.

### Expected Behavior
Request successful

### How Do We Reproduce?

1. type `npx ll-create create react-js bug-reproduce` command in terminal
2. cd bug-reproduce
3. yarn && yarn start
4. open google.com in browser, type `fetch("http://localhost:{port}/",{headers:{"another-header":"1"}})` in devtool

### Please paste the results of `npx webpack-cli info` here, and mention other relevant information
System:
   - OS: macOS 11.5.2
   - CPU: (8) arm64 Apple M1
   - Memory: 131.83 MB / 8.00 GB

Binaries:
   - Node: 16.13.1 - /usr/local/bin/node
   - Yarn: 1.22.15 - /usr/local/bin/yarn
   - npm: 8.1.2 - /usr/local/bin/npm

Browsers:
   - Chrome: 97.0.4692.71
   - Safari: 14.1.2

Other relevant information
   - webpack version: "5.66.0"
   - webpack-dev-sever version:"4.7.3"

alexander-akait:
PR welome

nickdeakin:
Alans change fixes this issue

PR #4185, opened 2022-01-16
fix: response correctly when receive an OPTIONS request

TrickyPi:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes

### Motivation / Use-Case

fix: #4180 

### Breaking Changes

No
### Additional Info

No


TrickyPi:
I try to run the test on my local machine. But it always hangs whenever it goes to the following steps.

RUNS  test/cli/basic.test.js
RUNS  test/cli/server-option.test.js
RUNS  test/cli/https-option.test.js

snitin315:
> I try to run the test on my local machine. But it always hangs whenever it goes to the following steps.
> 
> RUNS test/cli/basic.test.js RUNS test/cli/server-option.test.js RUNS test/cli/https-option.test.js

Make sure you have linked `webpack-dev-server` -

`sudo npm link && npm link webpack-dev-server`

TrickyPi:
> > I try to run the test on my local machine. But it always hangs whenever it goes to the following steps.
> > RUNS test/cli/basic.test.js RUNS test/cli/server-option.test.js RUNS test/cli/https-option.test.js
> 
> Make sure you have linked `webpack-dev-server` -
> 
> `sudo npm link && npm link webpack-dev-server`

Oh, i didn't run this command. I'm sorry about i didn't read CONTRIBUTING.md carefully.

lib/Server.js
```diff
@@ -2110,6 +2110,31 @@ class Server {
...
+          return;
+        }
+        next();
+      };
+
+      middlewares.push({
+        name: "options-request-response",
+        path: "*",
+        middleware: optionsRequestResponseMiddleware,
+      });
```
alexander-akait, line 2136: It can be problem, because we should not send response for any request...

TrickyPi: In my personal opinion. I thought it would be ok if we responded to OPTIONS requests immediately in the dev server...

alexander-akait: Do you want it for static assets or webpack assets?

TrickyPi: Both

alexander-akait: okay, implementation it for webpack assets is not a problem, with express static it can be problem due https://github.com/expressjs/serve-static/blob/master/index.js#L73... anyway I understand you problem, current solution is not good, so let's move it under an option, I think it will be better

TrickyPi: I forgot to mention that webpack-dev-server@4.6.0 process OPTIONS request correctly. During the time i didn't reply your comment(i'm sorry for that), I found the difference between 4.6.0 and latest, the change of serveMagicHtml(the last middleware/route) cause the different result, the commit hash is c13aa56. In the 4.6.0, it is implemented by using app.get, but in the latest version, it is implemented by using app.use. 
if we use app.use to implement it, this code will not be executed,
https://github.com/expressjs/express/blob/master/lib/router/index.js#L248
because `route` is undefined
https://github.com/expressjs/express/blob/master/lib/router/index.js#L232

Finally, back to express static, we can avoid this problem by moving the options-request-response middleware to the bottom of the middleware list. Anyway i agree with you that under an option is a better solution, but i think it's not necessary (just my stupid opinion).

alexander-akait: Yes, we change logic in magicHtml, because we want to remove it in the next major release, that is why we improve middlewares logic and now you can inject own middleware, I think your solution is good too, ideally - yes, we need handle it only for existing files, but it can be problem with express static, so currently I think we can move it under the option and improve it in future

TrickyPi: Uh,"under the options",is it mean that just move the code to the bottom of the options? I always thought the solution need with a devServer config, please forgive my poor english.

TrickyPi: Understood

TrickyPi: Finished

alan-agius4:
@alexander-akait, any way we can get this in?

alexander-akait:
@alan-agius4 Do you want this as built-in feature?

alan-agius4:
Ideally, the dev-server can handle preflight requests out-of-the-box as it did in version 3.

alexander-akait:
@alan-agius 4 interesting, because previously we don't handle it (i.e. we don't have code for this), it means one from middlewares handle this, I am not agait this merge, just want to check...

alexander-akait:
@alan-agius4 Do you have time to check it (just a little busy right now), thank you

TrickyPi:
Please let me know if some code needs to be changed.

TrickyPi:
a lot of differences in `types/lib/Server.d.ts`, is this normal?

alan-agius4:
The OPTIONS request/response is handled by the express router https://github.com/expressjs/express/blob/master/lib/router/index.js#L164-L169

In version 4 whoever [this check](https://github.com/expressjs/express/blob/2c47827053233e707536019a15499ccf5496dc9d/lib/router/index.js#L254) is always falsey, which causes the options not be handled properly.

In version 3 this worked because of the additional express Layer that caused the options to be handled.

```ts
setupMagicHtmlFeature() {
   this.app.get('*', this.serveMagicHtml.bind(this));
}
```

This is because the above cause a creation of an additional layer where `Boolean(this.methods[name])` is.
https://github.com/expressjs/express/blob/7ec5dd2b3c5e7379f68086dae72859f5573c8b9b/lib/router/route.js#L58-L70 is false.


In fact adding the below in `setupBuiltInRoutes` in v4, the OPTIONS are also handled correctly. 

```ts
app.get('*', (req, res, next) => { next(); });    
```

alexander-akait:
@alan-agius4 thank for this, I think we can enable it by default, should be safe for dev env

@TrickyPi Can you enable it by default?

TrickyPi:
Ok

alexander-akait (approved):

PR #4209, opened 2022-01-20
chore: update deps and types

snitin315:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No

### Motivation / Use-Case

update deps and types

### Breaking Changes

None

### Additional Info

No


lib/Server.js
```diff
@@ -754,7 +755,7 @@ class Server {
     // TODO remove `{}` after drop webpack v4 support
     const compilerWatchOptions = compilerOptions.watchOptions || {};
     /**
-     * @param {WatchOptions & WebpackConfiguration["watchOptions"]} watchOptions
+     * @param {WatchOptions & { aggregateTimeout?: number, ignored?: AnymatchMatcher | string | RegExp | string[], poll?: number | boolean }} watchOptions
```
alexander-akait, line 758: Ideally we should extend type here, i.e. `WebpackConfiguration["watchOptions"] & { ignored?: AnymatchMatcher | string | RegExp | string[] }`

snitin315: I did that at first, but it was throwing lint errors with this.

alexander-akait (approved):

PR #4258, opened 2022-02-07
chore: get rid of glob dependencies by switching del with rimraf

ludofischer:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

Replacing del gets rid of about 10 extra dependencies,
most are related to globbing, like globby, fast-glob and dir-glob, but also others like fastq, queue-microtask, reusify.

### Breaking Changes

No breaking changes.

### Additional Info

del wraps rimraf in promisify and adds
1. more sophisticated globbing
2. deletion of multiple files at once by passing an array
3. a check to prevent deleting the current working directory and outside

Since the code base does not use any of the extra features, we can just wrap `rimraf` with `promisify()`.


lib/Server.js
```diff
@@ -1133,13 +1133,15 @@ class Server {
 
           // cert is more than 30 days old, kill it with fire
           if ((now - Number(certificateStat.ctime)) / certificateTtl > 30) {
```
ludofischer, line 1135: This is the only production file that uses `del`, the rest of the usages are all in tests.

lib/Server.js
```diff
@@ -1133,13 +1133,15 @@ class Server {
...
-            const del = require("del");
+            const { promisify } = require("util");
+            const rimraf = require("rimraf");
+            const del = promisify(rimraf);
 
             this.logger.info(
               "SSL certificate is more than 30 days old. Removing..."
             );
 
-            await del([certificatePath], { force: true });
```
ludofischer, line 1142: `certificatePath` is not a glob, it's the result of calling `const certificatePath = path.join(certificateDir, "server.pem")` on line 1119. 
Also line 1123 uses `fs.stat` on `certificatePath` to check for existence. 
`{force: true}` bypasses a check added by `del`, so it's not needed for rimraf.

ludofischer (commented):
You can see how `del` wraps rimraf in the [del source](https://github.com/sindresorhus/del/blob/7c756b393804ad7df5cb534cecb80677565add59/index.js#L69).

alexander-akait (approved):

snitin315 (approved):
LGTM. Thank you.

alexander-akait:
@snitin315 Feel free to merge when CI will be green

PR #4304, opened 2022-02-20
feat: export initialized socket client

### LINKED ISSUE #4261: Expose Socket Client for external integrations
### Feature Proposal

Allow external integrations to tap into the socket client of WDS in order to consume compilation events, such as build logs, errors, progress, etc.

Right now, the client is done as a module level singleton:

https://github.com/webpack/webpack-dev-server/blob/87401b4927c75f1c5114508216101fd1a3ec9e67/client-src/socket.js#L19

Perhaps we can assign it to some special global variable on `window`, or export this variable, so external parties can just consume the client.

### Feature Use Case

This is mainly for custom error overlays that include wider coverage outside of what WDS provides (`react-error-overlay`, overlay from `@pmmmwh/react-refresh-webpack-plugin`), but it would also enable client-only tooling to sniff on HMR lifecycle, build status and more.

It is not impossible to achieve this functionality right now since the `__webpack_dev_server_client__` global is defined, and we can use that as a starting point, but we would have to do two things that is not preferrable:

1. Duplicate socket URL logic - which tightly couples the code to 1 version of WDS (and we have to keep up with all the special cases handled here)
2. Duplicate socket connection - since we have to create the client from scratch, the connection would happen multiple times, which could be confusing to users (#4224, https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/279, https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/481)

### Inspirations

`webpack-hot-middleware`

https://github.com/webpack-contrib/webpack-hot-middleware/blob/e140693fa3f648c078a47a678d66b53579f28ee2/client.js#L148-L159

pmmmwh:
- [ ] This is a **bugfix**
- [x] This is a **feature**

~No new tests are added as I believe this is a simple change and no additional functionality is added, it just exposes a bit more of the private parts.~

Added a simple test for `null` value before init, and a "valid" client instance after init.

### Motivation / Use-Case

Fixes #4261
Fixes #4324

### Additional Info

I chose the simplest way to expose the initialized client, so consumers can just import the file and use it - but I'm also ok with setting it to some globals (on `globalThis`/`window`) if you think that's better (I'm unsure if mutable exports are implemented properly when compiled?).

### DIFF
--- a/client-src/socket.js
+++ b/client-src/socket.js
@@ -18,6 +18,8 @@ let retries = 0;
 let maxRetries = 10;
 
 // Initialized client is exported so external consumers can utilize the same instance
+// It is mutable to enforce singleton
+// eslint-disable-next-line no-mutable-exports
 export let client = null;
 
 /**
--- a/client-src/socket.js
+++ b/client-src/socket.js
@@ -19,7 +19,7 @@ let maxRetries = 10;
 
 // Initialized client is exported so external consumers can utilize the same instance
 // It is mutable to enforce singleton
-// eslint-disable-next-line no-mutable-exports
+// eslint-disable-next-line import/no-mutable-exports
 export let client = null;
 
 /**
--- a/client-src/socket.js
+++ b/client-src/socket.js
@@ -16,7 +16,9 @@ const Client =
 
 let retries = 0;
 let maxRetries = 10;
-let client = null;
+
+// Initialized client is exported so external consumers can utilize the same instance
+export let client = null;
 
 /**
  * @param {string} url
--- a/test/client/socket-helper.test.js
+++ b/test/client/socket-helper.test.js
@@ -5,7 +5,7 @@
 "use strict";
 
 describe("socket", () => {
-  afterEach(() => {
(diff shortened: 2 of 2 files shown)


snitin315 (changes requested):
Let's fix lint.

pmmmwh:
I'm not sure if I should just do eslint ignore here, or should I do something more

alexander-akait:
Just add `eslint-plugin-ignore` and small comment(s) why we need to it

alexander-akait:
@pmmmwh Can we add small test cases (where we try to get client) to avoid future regressions?

pmmmwh:
Added test!

pmmmwh:
Completely rebased onto latest HEAD + fixed commitlint issue 😄

alexander-akait:
Thanks

alexander-akait:
@pmmmwh Published https://github.com/webpack/webpack-dev-server/releases/tag/v4.8.0

PR #4357, opened 2022-03-27
feat: make allowedHosts accept localhost subdomains by default

jdufresne:
Use Case:

Some web applications will serve multiple domains and render different
content depending on the incoming header. In a development environment,
these applications may use localhost subdomains to simulate these
alternative domains.

To reduce overall configuration, make the webpack-dev-server allow
"localhost" as well as its subdomains. As "localhost" is already
considered safe and convenient, the same can be said of its subdomains.

Available Workaround:

Without this feature, developing a multi-domain application requires
configuring webpack-dev-server with:

    allowedHosts: ".localhost"

- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes.

### Motivation / Use-Case

Some web applications will serve multiple domains and render different
content depending on the incoming header. In a development environment,
these applications may use localhost subdomains to simulate these
alternative domains.

To reduce overall configuration, make the webpack-dev-server allow
"localhost" as well as its subdomains. As "localhost" is already
considered safe and convenient, the same can be said of its subdomains.

### Breaking Changes

No.

### Additional Info

Available Workaround:

Without this feature, developing a multi-domain application requires
configuring webpack-dev-server with:

    allowedHosts: ".localhost"


alexander-akait:
I think it is expected, because it is not standard, localhost is built-in loop, but *.localhost are not

jdufresne:
> localhost is built-in loop, but *.localhost are not

On Fedora Linux, all subdomains resolve to localhost out of the box:

```
$ gethostip localhost
localhost 127.0.0.1 7F000001
$ gethostip app1.localhost
localhost 127.0.0.1 7F000001
$ gethostip app2.localhost
localhost 127.0.0.1 7F000001
```

It is this fact that originally drew me using localhost subdomains as a viable workflow.

ylemkimon:
RFC 2606 states and reserves:

> The ".localhost" TLD has traditionally been statically defined in host DNS implementations as having an A record pointing to the loop back IP address and is reserved for such use.  Any other use would conflict with widely deployed code which assumes this use.

So it's unlikely other applications use `.localhost` or it has security implications.

alexander-akait:
@ylemkimon Yeah, I see, let's merge it

PR #4384, opened 2022-04-13
chore: replace portfinder with custom implementation

### LINKED ISSUE #4383: node-portfinder not supported anymore
### Bug report

Please consider replacing node-portfinder with some other module. node-portfinder is not supported anymore, meaning webpack-dev-server users are struggling now how to update async ( [CVE-2021-43138](https://nvd.nist.gov/vuln/detail/CVE-2021-43138), https://github.com/http-party/node-portfinder/issues/126).

ludofischer:
Fix #4383 
- Get rid of security warnings related to async dependencies
- Drop lodash from prod.
- The implementation is based mostly on the get-port package, but searches for ports incrementing by 1 like portfinder and I;ve added TypeScript annotations.

- [ ] This is a **bugfix**
- [ ] This is a **feature**

I've added new unit tests for the custom port finder implementation.

### Motivation / Use-Case
Remove dependency on vulnerable package async, also remove dependency on lodash.

### Breaking Changes

There should be no breaking changes.
### Additional Info
I am in doubt whether to go with this or the get-port package as in https://github.com/ludofischer/webpack-dev-server/tree/switch-port-finder
The code is a bit complex because it needs to take into account differences across operating systems, so it might be useful to rely on the expertise of the `get-port` maintainters. On the other hand there are more test failures with the `get-port` package and we would be stuck on the previous get-port major version, since it switched to ESM-only.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -393,9 +393,8 @@ class Server {
     }
 
     const pRetry = require("p-retry");
-    const portfinder = require("portfinder");
-
-    portfinder.basePort =
+    const getPort = require("./getPort");
+    const basePort =
       typeof process.env.WEBPACK_DEV_SERVER_BASE_PORT !== "undefined"
         ? parseInt(process.env.WEBPACK_DEV_SERVER_BASE_PORT, 10)
         : 8080;
@@ -407,7 +406,7 @@ class Server {
         ? parseInt(process.env.WEBPACK_DEV_SERVER_PORT_RETRY, 10)
         : 3;
 
-    return pRetry(() => portfinder.getPortPromise(), {
+    return pRetry(() => getPort(basePort), {
       retries: defaultPortRetry,
     });
   }
--- a/lib/getPort.js
+++ b/lib/getPort.js
@@ -1,0 +1,122 @@
+"use strict";
+
+/*
+ * Based on the packages get-port https://www.npmjs.com/package/get-port
+ * and portfinder https://www.npmjs.com/package/portfinder
+ * The code structure is similar to get-port, but it searches
+ * ports deterministically like portfinder
+ */
+const net = require("net");
+const os = require("os");
+
+const minPort = 1024;
+const maxPort = 65_535;
+
+/**
+ * @return {Set<string|undefined>}
+ */
+const getLocalHosts = () => {
+  const interfaces = os.networkInterfaces();
+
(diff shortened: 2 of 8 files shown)


alexander-akait:
> The code is a bit complex because it needs to take into account differences across operating systems, so it might be useful to rely on the expertise of the get-port maintainters.

Can you clarify?

ludofischer:
> Can you clarify?
For example you need to check the port on every local interface, else the port might look free even though it's already taken: 
https://github.com/webpack/webpack-dev-server/blob/983a0aa0a9dfd1df841eab6e726587cd88ab4194/lib/getPort.js#L18
Or the error codes you need to check for. This sort of thing someone who hasn't been maintaining a package like this for some time years will not understand.

ludofischer:
I also think there is a mistake in the current webpack-dev-server implementation. When you pass a custom host, the custom host is not passed to portfinder, so portfinder might check the ports on the wrong host.
See this line https://github.com/webpack/webpack-dev-server/blob/c9b643360f09d7ecb9106a82f5e08c1f5a6af49a/lib/Server.js#L3200 and the implementation of `Server.getFreePort()`.

alexander-akait:
> I also think there is a mistake in the current webpack-dev-server implementation. When you pass a custom host, the custom host is not passed to portfinder, so portfinder might check the ports on the wrong host.

Yes, looks like a bug, let fix it too, do you want to fix it here or in separate PR?

ludofischer:
> Yes, looks like a bug, let fix it too, do you want to fix it? If yes - here or in separate PR?

SInce it's a longstanding bug I would fix it in a different PR. I think it's better to release the current PR on its own.

alexander-akait (approved):

ludofischer:
> Let's fix a problem with host and I will do patch release

Do you mean fix the problem mentioned in https://github.com/webpack/webpack-dev-server/pull/4384#issuecomment-1098617180?

alexander-akait:
@ludofischer Yep :+1:

ludofischer:
@alexander-akait See https://github.com/webpack/webpack-dev-server/pull/4385

ins429:
Thank you ❤️

cjwilsontech:
> Let's fix a problem with host and I will do patch release

@alexander-akait, any update on the patch release?

PR #4385, opened 2022-04-14
fix: use the host in options to check if port is available

ludofischer:
- [x] This is a **bugfix**
- [ ] This is a **feature**

Yes, but it is hard to test for non-localhost hosts.

### Motivation / Use-Case
The server should check if the port is available on the host specified by the user.

### Breaking Changes
No breaking changes.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -385,9 +385,10 @@ class Server {
 
   /**
    * @param {Port} port
+   * @param {string} host
    * @returns {Promise<number | string>}
    */
-  static async getFreePort(port) {
+  static async getFreePort(port, host) {
     if (typeof port !== "undefined" && port !== null && port !== "auto") {
       return port;
     }
@@ -406,7 +407,7 @@ class Server {
         ? parseInt(process.env.WEBPACK_DEV_SERVER_PORT_RETRY, 10)
         : 3;
 
-    return pRetry(() => getPort(basePort), {
+    return pRetry(() => getPort(basePort, host), {
       retries: defaultPortRetry,
     });
   }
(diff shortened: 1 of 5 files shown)


types/lib/Server.d.ts
```diff
@@ -716,9 +717,9 @@ declare class Server {
           description: string;
           multiple: boolean;
           path: string;
-          type: string;
+          type: string /** @type {WebSocketURL} */;
```
ludofischer, line 720: Regenerating the Server types introduces a lot unrelated comments in the type definitions, but running `build:types` on master does not create any diff.

ludofischer (commented):
These changes should result in the same code executing as before if the host is localhost. I think the case where someone wants the dev server to listen on non-localhost must be fairly rare.

types/lib/Server.d.ts
snitin315: Hmm, types changed here since you added [` * @param {string} host`](https://github.com/webpack/webpack-dev-server/pull/4385/files#diff-b706bbadb3d2315d99678f05e1f4bfd78a4bb4a7a6c65f3e81f1386fee3c883a)

snitin315 (approved):
LGTM, thanks!

lib/Server.js
```diff
@@ -406,7 +407,7 @@ class Server {
         ? parseInt(process.env.WEBPACK_DEV_SERVER_PORT_RETRY, 10)
         : 3;
 
-    return pRetry(() => getPort(basePort), {
```
ludofischer, line 409: I am not sure `retry` is going to help a lot as the `getPort()` (and portfinder) before is going to try every port up to the maximum port number anyway, but it was introduced on purpose in [this commit](https://github.com/webpack/webpack-dev-server/pull/1993) back in 2019, so there might be a reason.

types/lib/Server.d.ts
alexander-akait: It is bug in ts, reported about it, but still not fixed

alexander-akait (approved):

alexander-akait:
Thanks

Issue #4400, opened 2022-04-23
Trusted Types support


Siegrift:
### Feature Proposal

Hello, 

I propose adding [Trusted Types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/trusted-types) support which are already [supported by Webpack 5](https://webpack.js.org/guides/csp/#trusted-types). From my research, the only place that would need to be changed is [showing the error overlay](https://github.com/webpack/webpack-dev-server/blob/7c1d680008c986ed4721095c4babfa802ff31445/client-src/overlay.js#L196). The HTML markup of the error is added to the DOM via `innerHTML` which causes a Trusted Types violation. For applications with Trusted Types enforcement enabled this causes a browser error and the markup is not added to the DOM (so the error is not shown to the user).

The solution is to pass the value assigned to the `innerHTML` through a Trusted Types policy. This could be done in backwards compatible way by checking if Trusted Types are available in a current browser. This could work roughly like this:

1. Check if Trusted Types API is available in the current browser. 
2. If not, assign the value through `innerHTML` property same as of now.
3. If yes, get (or create) a Trusted Types policy - We want to create a policy one once, because when applications enforce Trusted Types via CSP, creating a policy with same name throws an error unless the `'allow-duplicates'` is enabled. This policy would only allow `createHTML` operation.
4. Pass the value to `createHTML` function of the policy and assign the return value (TrustedHTML) to the `innerHTML`.

There are no observable changes for applications which do not enforce Trusted Types via a CSP header because in that case the Trusted Types is "API only" and any policy can be created. Applications which enforce Trusted Types and restrict which policies can be created need to explicitly allow this policy in the CSP header. The name of the policy could be configurable by creating a new property in the [overlay configuration options](https://webpack.js.org/configuration/dev-server/#overlay). 

### Feature Use Case

The purpose of this feature is to enable development of applications which use Webpack and webpack-dev-server, for example applications created by [create-react-app](https://github.com/facebook/create-react-app) and many others.

### Please paste the results of `npx webpack-cli info` here, and mention other relevant information

I have already experimented with this change and created a fixed name policy called `webpack-dev-server#overlay`. I've then created a simple React application via CRA which use this as a dependency. See:
- webpack-dev-server changes = https://github.com/Siegrift/webpack-dev-server/commit/045d02a7d43b0ded181d531a4066f8d5c07645e0
[...]

PR #4404, opened 2022-04-24
feat: support Trusted Types for client overlay

Siegrift:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes. LMK if more are needed.

### Motivation / Use-Case

Addresses https://github.com/webpack/webpack-dev-server/issues/4400

### Breaking Changes

No breaking changes. 

This change is noop for applications which do not enforce Trusted Types. Applications which do enforce Trusted Types and use webpack-dev-server can now whitelist `webpack-dev-server#overlay` policy to make error overlay work. Optionally, users can choose a different name via `policyName` attribute (but maybe this could change to more verbose `trustedTypesPolicyName`?). 

### Additional Info

See https://github.com/webpack/webpack-dev-server/issues/4400


client-src/index.js
```diff
@@ -15,7 +15,7 @@ import createSocketURL from "./utils/createSocketURL.js";
  * @property {boolean} hot
  * @property {boolean} liveReload
  * @property {boolean} progress
- * @property {boolean | { warnings?: boolean, errors?: boolean }} overlay
+ * @property {boolean | { warnings?: boolean, errors?: boolean, policyName?: string }} overlay
```
Siegrift, line 18: I am wondering whether the `policyName` should be renamed to more verbose `trustedTypesPolicyName`. WDYT?

client-src/overlay.js
```diff
@@ -23,10 +23,26 @@ let iframeContainerElement;
 let containerElement;
 /** @type {Array<(element: HTMLDivElement) => void>} */
 let onLoadQueue = [];
+/** @type {any} */
```
Siegrift, line 26: There are TS type definitions located in https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/trusted-types

However, I am not sure if it is worth typing so I've added `any` for now.

alexander-akait:
Thanks for PR, can you accept CLA?

client-src/overlay.js
```diff
@@ -23,10 +23,26 @@ let iframeContainerElement;
...
 
 ansiHTML.setColors(colors);
 
-function createContainer() {
+/**
+ * @param {string | null} policyName
+ */
+function createContainer(policyName) {
+  // Enable Trusted Types if they are available in the current browser.
+  if (window.trustedTypes) {
```
Siegrift, line 37: We could optionally check whether the `policyName` is not `null`. However, I think using trustedTypes whenever they are available is better because:

1. It is still noop and backwards compatible change for applications which do not enforce Trusted Types
2. This enables Trusted Types integration without the need to change anything in webpack configuration (which can be hidden, such as in [CRA](https://github.com/facebook/create-react-app)). This means that the only change the application needs to do is whitelist the default policy name.

types/lib/Server.d.ts
```diff
@@ -1610,6 +1610,10 @@ declare class Server {
                     negatedDescription: string;
                   };
                 };
+                policyName: {
```
Siegrift, line 1613: I am not sure why there are so many unrelated changes in this autogenerated `d.ts` file. Especially, since I've added just one field.

Please, LMK if I have something wrong in my setup.

Siegrift:
I also see that there needs to be a documentation update in https://webpack.js.org/configuration/dev-server/#overlay, specifically this file https://github.com/1ilsang/webpack.js.org/blob/master/src/content/configuration/dev-server.mdx#overlay

I guess that is only updated after there is a new release?

Also, please LMK if there are is anything in this PR I've missed :)

client-src/index.js
alexander-akait: I don't think we will have the `policyName` option here, but I think make sense to rename it to `trustedTypesPolicyName`, to avoid misleading

client-src/overlay.js
alexander-akait: I think we need install them in dev depedencies, because when types is not found it is `any`

types/lib/Server.d.ts
alexander-akait: It is bug in `ts`, ignore it

alexander-akait:
> I also see that there needs to be a documentation update in https://webpack.js.org/configuration/dev-server/#overlay, specifically this file https://github.com/1ilsang/webpack.js.org/blob/master/src/content/configuration/dev-server.mdx#overlay

We will open issue in our docs after merge, dont worry

> I guess that is only updated after there is a new release?

Yes, I will do release

examples/client/trusted-types-overlay/README.md
```diff
@@ -0,0 +1,36 @@
...
+```js
+module.exports = {
+  // ...
+  output: {
+    trustedTypes: { policyName: "webpack" },
+  },
+  devServer: {
+    client: {
+      overlay: {
+        policyName: "overlay-policy",
```
koto, line 14: Consider using `webpack#dev-overlay` or something like this to feature the recommended naming scheme.

koto (dismissed):

alexander-akait:
@Siegrift frinedly ping

Siegrift:
Sorry, I got a bit busy with work stuff this week. I'll hopefully find some time tomorrow or over the weekend. 

Thanks for the ping :)

alexander-akait:
Bih thanks, feel free to ping me

client-src/index.js
Siegrift: Fixed in https://github.com/webpack/webpack-dev-server/pull/4404/commits/c63489b3372539de2b51fffa738249435f5ff6bc

client-src/overlay.js
Siegrift: Fixed https://github.com/webpack/webpack-dev-server/pull/4404/commits/68d39742f2c13dc3beb68913eecee0170f7a69af

examples/client/trusted-types-overlay/README.md
Siegrift: Fixed in https://github.com/webpack/webpack-dev-server/pull/4404/commits/8ef1a488fff474ff39097fc1394f6486b52b97b1

client-src/overlay.js
```diff
@@ -23,10 +23,25 @@ let iframeContainerElement;
...
 ansiHTML.setColors(colors);
 
-function createContainer() {
+/**
+ * @param {string | null} trustedTypesPolicyName
+ */
+function createContainer(trustedTypesPolicyName) {
+  // Enable Trusted Types if they are available in the current browser.
+  if (window.trustedTypes) {
+    overlayTrustedTypesPolicy = window.trustedTypes.createPolicy(
```
Siegrift, line 37: In case there is an error creating a Trusted Types policy the overlay is not shown at all. 

A different approach would be to capture the exception outside this function (same place we define `overlayTrustedTypesPolicy`) and show an overlay without using `innerHTML` showing that the policy creation failed. 

However I think having an error in the console is sufficient, since this is a development only feature so we can assume the console is opened anyways.

Siegrift:
I've fixed the comments and rebased on current `master` and left one more comment.

alexander-akait:
@Siegrift Hello, sorry for delay, can you update snapshots?

Siegrift:
Sorry for not running the tests locally first :/ 
However, there seem to be some changed snapshots which are unrelated to my feature. I am not sure about that.

snitin315:
@Siegrift help output changed because you added `client-overlay-trusted-types-policy-name` CLI flag, hence need to update the snapshot.

Siegrift:
Thanks, I am having problems passing the tests locally. When I run `yarn test:only --updateSnapshot` most of the tests pass, but there are sometimes a few unrelated changes (I've reverted those manually). When I try to run the tests now, they seem to remove the `basic should output help should generate correct cli flags 1` in webpack5 snapshot which is wrong.

Is there something I am doing wrong?

snitin315:
Are you using macOS?

Siegrift:
Yes, the m1 chip.

snitin315:
Yeah, help output snapshots are not quite well on macOS, known bug in nodejs it strips the actual output on stdout.

Siegrift:
I'll try to fix the tests on my Linux machine later today. I don't think I'll get anywhere by testing this on macOS :D

alexander-akait:
I want to merge it and fix tests in the next PR, because we need to do release today due security problems in portfinder, big thanks for PR

Siegrift:
No problem, sorry for taking it too long :/ 

Good luck with the release.

PR #4424, opened 2022-05-03
fix: improve schema for watch options

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**


lib/options.json
```diff
@@ -1076,6 +1071,64 @@
...
+            {
+              "type": "boolean"
+            },
+            {
+              "type": "number"
+            }
+          ]
+        }
+      }
+    },
```
alexander-akait, line 1131: hm, so chokidar doesn't validate own options?

snitin315: @alexander-akait No, it doesn't.


https://user-images.githubusercontent.com/46647141/166394402-7898c98b-c3b7-4181-bc8c-4d0ec2b907d6.mov

snitin315: hmm, I think we allow some extra options for watching, but `chokidar` will only understand its API options.

https://github.com/webpack/webpack-dev-server/blob/c2fee3b585307017c66b322afd78220e6b6c835a/lib/Server.js#L766-L772

https://github.com/webpack/webpack-dev-server/blob/c2fee3b585307017c66b322afd78220e6b6c835a/lib/Server.js#L805

snitin315: Yeah, ideally chokidar should have its own validation. Maybe we can add an option in scema-utils `additionalProperties: "warn"` which will warn if you use another property instead of throwing an error and exiting the process.

alexander-akait: Sounds better, we will still needed update them when new options were added, but it happens rarely

Issue #2716, opened 2020-08-20
behavior of static and historyApiFallback


Loonride:
### Expected Behavior

Static serving middleware for the dev server should only need to be applied to the express server once via `this.app.use(...)`

### Actual Behavior

static middleware is applied to the express server multiple times in order to make `historyApiFallback` work as expected. This is explained here: https://github.com/webpack/webpack-dev-server/pull/2670#discussion_r464946541

A similar thing seems to have been added for the `middleware` feature as seen here, as it is added multiple times:

https://github.com/webpack/webpack-dev-server/blob/4ab1f21bc85cc1695255c739160ad00dc14375f1/lib/Server.js#L548-L559

### Solution

We need to find a way to either apply all of this middleware once without having any breaking changes, or have intentional slight breaking changes while only applying the middleware once. We should also look into if applying the same middleware many times causes performance losses, or if it is an acceptable thing to do.

csr632:
Yes, if  the `proxy` feature or the `historyApiFallback` feature is enabled, `http-proxy-middleware` will be applied to the express server multiple times.

alexander-akait:
`historyApiFallback` also break HEAD request in some cases, we need fix it, more information here historyApiFallback

alexander-akait:
Partially fixed https://github.com/webpack/webpack-dev-server/pull/4501

BRMatt:
I ran into this issue earlier and spent a few hours trying to debug it. In my case it turned out that configuring the webpack-dev-server `proxy` option with `context` and `router` options, but not `target` causes webpack-dev-server to ignore the `context` option when configuring the proxy:

https://github.com/webpack/webpack-dev-server/blob/49efdf87341d4e40106baae60670a4aabf3285fb/lib/Server.js#L2174-L2189

My understanding is that if you're using `router` you don't need to specify `target`, but if I don't then webpack-dev-server configures the proxy to intercept _every_ request, rather than just the requests that match the context.

In my case I'm only proxying a single endpoint, so I've configured `context` and `target` to the same value, and let the router reconfigure the target on each request. This appears to fix the conflict with `historyApiFallback`.

alexander-akait:
I think we can improve/fix:
```
return createProxyMiddleware(proxyConfig); 
```

So we will always pass options even you don't have router, what do you think?

PR #4501, opened 2022-06-27
chore: bump connect-history-api-fallback from 1.6.0 to 2.0.0 (#2716)

awwit:
- [x] This is a **bugfix**
- [x] This is a **feature**

Do I really need to add a test for this case? tell me how to do this and how to run the tests correctly, otherwise I have problems with this (some tests seem to fall due to a timeout).

### Motivation / Use-Case

Link to issue: https://github.com/webpack/webpack-dev-server/issues/2716

I faced the same problem, I need to solve it =)

### Breaking Changes

The `history-api-fallback` now supports `HEAD` requests and handles them the same as `GET`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "chokidar": "^3.5.3",
         "colorette": "^2.0.10",
         "compression": "^1.7.4",
-        "connect-history-api-fallback": "^1.6.0",
+        "connect-history-api-fallback": "^2.0.0",
         "default-gateway": "^6.0.3",
         "express": "^4.17.3",
         "graceful-fs": "^4.2.6",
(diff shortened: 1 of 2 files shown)


alexander-akait (commented):
Let's add small test case where we send HEAD request when `connect-history-api-fallback` is enabled, looks at our tests, it should be easy, thank you for the PR

snitin315:
You also need to create a snapshot for webpack 4

awwit:
@snitin315 It looks like `webpack-dev-server / Test - macos-latest - Node v12.x, Webpack latest (pull_request)` task is stuck. I think it needs to be restarted if possible.

It's been running for almost 3 hours now.

alexander-akait (approved):

PR #4559, opened 2022-09-06
fix: auto reply to OPTIONS requests only when unhandled

### LINKED ISSUE #4551: Automatically replying to OPTIONS requests is potentially problematic
### Bug report

`OPTIONS` requests are automatically handled by webpack-dev-server since https://github.com/webpack/webpack-dev-server/issues/4180 was implemented by https://github.com/webpack/webpack-dev-server/pull/4185, bypassing the application's backend.

### Actual Behavior

webpack-dev-server automatically replies to `OPTIONS` requests regardless of how the backend application would handle it.

### Expected Behavior

Much like any other kind of requests, like `GET` and `POST`, `OPTIONS` requests should by default be forwarded to the application. Applications can reply with a failure depending on the request, and can include a [variety of response headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS#preflighted_requests_in_cors) which can be important for the application.

### How Do We Reproduce?

Use webpack-dev-server 4.10.0, which contains https://github.com/webpack/webpack-dev-server/pull/4185, and any application that handles `OPTIONS` requests in the backend to see that the backend application isn't even contacted.

  System:
    OS: Linux 5.14 Fedora 33 (Workstation Edition) 33 (Workstation Edition)
    CPU: (8) x64 Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
    Memory: 7.60 GB / 15.38 GB
  Binaries:
    Node: 14.19.3 - /nix/store/cnma41dzr06jp8j6vigfp4lmy1dqfg3x-nodejs-14.19.3/bin/node
    npm: 6.14.17 - /nix/store/cnma41dzr06jp8j6vigfp4lmy1dqfg3x-nodejs-14.19.3/bin/npm

alan-agius4:
Prior to this change the internal options middleware always responded to such requests. This is because the middleware was registered too early and was not being used as a fallback. With this change we register this middleware as the last middleware to ensure that this is only used as a fallback when OPTIONS requests are not handled.

Closes #4551

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -2131,32 +2131,6 @@ class Server {
       });
     }
 
-    {
-      /**
-       *
-       * @param {Request} req
-       * @param {Response} res
-       * @param {NextFunction} next
-       * @returns {void}
-       *
-       */
-      const optionsRequestResponseMiddleware = (req, res, next) => {
-        if (req.method === "OPTIONS") {
-          res.statusCode = 204;
-          res.setHeader("Content-Length", "0");
-          res.end();
-          return;
-        }
-        next();
-      };
-
-      middlewares.push({
-        name: "options-middleware",
-        path: "*",
-        middleware: optionsRequestResponseMiddleware,
-      });
-    }
-
     middlewares.push({
       name: "webpack-dev-middleware",
       middleware:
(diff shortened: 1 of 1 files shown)


snitin315 (approved):
LGTM, Thanks for the PR ⭐

PR #4575, opened 2022-09-19
test: fix examples for open target not working

malcolm-kee:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No new test. Old tests are fixed.

Before

before

After

after


malcolm-kee:
Code coverage dropped but I not sure how could that happen as I just fix the example config.

PR #4581, opened 2022-09-24
fix: respect `NODE_PATH` env variable

snitin315:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No.

### Motivation / Use-Case

Porting changes from https://github.com/webpack/webpack-cli/pull/3411

Reference https://github.com/webpack/webpack-cli/issues/3355

### Breaking Changes
No.

### Additional Info
No.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -43,6 +43,18 @@ const isInstalled = (packageName) => {
   const path = require("path");
   const fs = require("graceful-fs");
 
+  // Respect NODE_PATH environment variable
+  try {
+    if (
+      process.env.NODE_PATH &&
+      fs.statSync(path.join(process.env.NODE_PATH, packageName)).isDirectory()
+    ) {
+      return true;
+    }
+  } catch (_error) {
+    // Nothing
+  }
+
   let dir = __dirname;
 
   do {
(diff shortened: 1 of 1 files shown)


alexander-akait (changes requested):
Le't simprove this using:
```js
if (typeof process.env.NODE_PATH !== undefined) {
 // Logic
}
```

I think we need the same change in `webpack-cli` and `webpack` itself, can you fix it? thank you

alexander-akait:
hm, from docs:

> NODE_PATH is still supported, but is less necessary now that the Node.js ecosystem has settled on a convention for locating dependent modules. Sometimes deployments that rely on NODE_PATH show surprising behavior when people are unaware that NODE_PATH must be set. Sometimes a module's dependencies change, causing a different version (or even a different module) to be loaded as the NODE_PATH is searched.

And

> Additionally, Node.js will search in the following list of GLOBAL_FOLDERS:

> 1: $HOME/.node_modules
> 2: $HOME/.node_libraries
> 3: $PREFIX/lib/node

Also paths on windows and linux are platform specific (and windows can has multiple paths separated by `;`), so fix is not simple as I thought.

So I see two solutions here:
1. Don't support it, but it can be paintful for some developers
2. Fix our logic and handle all cases (not sure how we can handle `$HOME` and `$PREFIX`)

I am working on `2.`

alexander-akait:
@snitin315 Found easy solution, we should use it in `webpack-cli`/`webpack`

alexander-akait:
Also it should be after `node_modules` check

snitin315:
Nice solution, I'll update in webpack/webpack-cli 👍🏻

alexander-akait (approved):

Issue #4348, opened 2022-03-19
hardcoded sockjs iframe url


Zekfad:
Why this is hardcoded? 
https://github.com/webpack/webpack-dev-server/blob/95e26fea168440c0f694c16478aae672091d06a3/lib/servers/SockJSServer.js#L46

In our project this is failing because we have a different from `__webpack_dev_server__` path.

snitin315:
Hmm, we can add an option for this.

Zekfad:
> Hmm, we can add an option for this.

That would be nice!

alexander-akait:
Why you need to change it? It is potential security problem

Zekfad:
Because we have only a few restricted paths lead to a webpack server backend in development.
Also its a security problem *not* to change it, as it can be a really easy for scanners and so on.

alexander-akait:
Okay PR welcome

PR #4586, opened 2022-09-27
feat: allow to set  the `sockjs_url` option (only `sockjs`)

alexander-akait:
- [ ] This is a **bugfix**
- [x] This is a **feature**

To test it we should disable websocket supporting in headless chrome (but it is impossible), as an alternative solution we can allow to set all options for sockjs server, but sockjs is old and only for old IE and Opera, I don't think many developers support it now and we should spend time on it, really, so I tested it manually and all works fine, in the next major release we will remove sockjs

### Motivation / Use-Case

fixes https://github.com/webpack/webpack-dev-server/issues/4348

### Breaking Changes

No

### Additional Info

Should we document it?

### DIFF
--- a/lib/servers/SockJSServer.js
+++ b/lib/servers/SockJSServer.js
@@ -41,9 +41,16 @@ module.exports = class SockJSServer extends BaseServer {
   constructor(server) {
     super(server);
 
+    const sockjsUrl =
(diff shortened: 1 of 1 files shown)


snitin315 (approved):
LGTM. We may need to document this somewhere.

alexander-akait:
Yeah feel fre to update our docs :+1: And thank you

Issue #3689, opened 2021-08-19
improve our overlay


alexander-akait:
- [ ] This is a **bug**
- [x] This is a **modification** request

### Code

improve our default overlay:
1. Allow to catch `runtime` errors
2. Improve errors/warnings reporting (ideally we should have )
3. Improve styles
4. More?

CRA has `react-error-overlay`, I think we can union with them and create the one great overlay

/cc @raix  

### Expected Behavior

Better DX

### Actual Behavior

Our overlay is very simple and in some places have bad DX

Bosch-Eli-Black:
Not sure if it's possible, but it would be awesome if compilation errors were shown after refreshing the page! 🙂  Currently if the page is refreshed when there's a compilation error, we just get a whitescreen and no overlay is shown.

SuneelFreimuth:
I'd be happy to try implementing these improvements.

alexander-akait:
> Not sure if it's possible, but it would be awesome if compilation errors were shown after refreshing the page! slightly_smiling_face Currently if the page is refreshed when there's a compilation error, we just get a whitescreen and no overlay is shown.

Can you provide reproducible steps, I think it is bug

ylemkimon:
`react-error-overlay` seems to be easily integrable, adding it to the client works out of the box. Two notes:
- it doesn't support build warnings, i.e., build errors are not dismissable
- it also displays runtime errors, which is not the existing overlay did. If the dev server decides to support this, source code fetch logic should be also updated

alexander-akait:
@ylemkimon yep, feel free to start it, for runtime errors we need a new options for overlay, i.e. `overlay.runtimeErrors`

sibelius:
how can I customize overlay behavior?

I've tried this

````jsx
devServer: {
   overlay: {
      warnings: false,
      errors: true
    }
}
```

but it did not worked

alexander-akait:
Should work, what is the problem?

ylemkimon:
@sibelius If you're using dev server v4, you should put `overlay` in `client` (https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md):
```js
devServer: {
   client: {
      overlay: {
         warnings: false,
         errors: true
       }
   }
}
```

malcolm-kee:
I think all the 3 listed can be implemented independently.

I would like to share my current thought on number 3 (improve styles). 

[...]

PR #4587, opened 2022-09-27
feat: open editor when clicking error on overlay

malcolm-kee:
- [ ] This is a **bugfix**
- [x] This is a **feature**

### Motivation / Use-Case

This PR make webpack-dev-server open editor when user clicks on the error message on overlay.

Part of #3689.

### Additional Info

- New endpoint `"/webpack-dev-server/open-editor"` is added on the dev server which the client would calls using `fetch`.
- Right now I assume `moduleIdentifier` from the `stats.errors` is the file name, but I not sure if that assumption is correct.
- Use [`launch-editor`](https://github.com/yyx990803/launch-editor) to open editor (it is based on `react-dev-utils` of Create React App, and is being used by Vite as well).
- For webpack 4 the logic does not work because there is no `moduleIdentifier`, but I assume that is not a dealbreaker.

### DIFF
--- a/client-src/overlay.js
+++ b/client-src/overlay.js
@@ -184,7 +184,7 @@ function formatProblem(type, item) {
 // Compilation with errors (e.g. syntax error or missing modules).
 /**
  * @param {string} type
- * @param {Array<string  | { file?: string, moduleName?: string, loc?: string, message?: string }>} messages
+ * @param {Array<string  | { moduleIdentifier?: string, moduleName?: string, loc?: string, message?: string }>} messages
  * @param {string | null} trustedTypesPolicyName
  */
 function show(type, messages, trustedTypesPolicyName) {
@@ -202,6 +202,16 @@ function show(type, messages, trustedTypesPolicyName) {
 
       typeElement.innerText = header;
       applyStyle(typeElement, msgTypeStyle);
+
+      if (message.moduleIdentifier) {
+        applyStyle(typeElement, { cursor: "pointer" });
+        typeElement.dataset.canOpen = true;
+        typeElement.addEventListener("click", () => {
+          fetch(
+            `/webpack-dev-server/open-editor?fileName=${message.moduleIdentifier}`
+          );
+        });
+      }
 
       // Make it look similar to our terminal.
       const text = ansiHTML(encode(body));
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -2034,6 +2034,19 @@ class Server {
     );
 
     /** @type {import("express").Application} */
+    (app).get("/webpack-dev-server/open-editor", (req, res) => {
+      const fileName = req.query.fileName;
+
+      if (typeof fileName === "string") {
+        // @ts-ignore
+        const launchEditor = require("launch-editor");
+        launchEditor(fileName);
+      }
+
+      res.end();
+    });
+
+    /** @type {import("express").Application} */
     (app).get(
       "/webpack-dev-server",
       /**
--- a/test/helpers/run-browser.js
+++ b/test/helpers/run-browser.js
@@ -3,6 +3,16 @@
 const puppeteer = require("puppeteer");
 const { puppeteerArgs } = require("./puppeteer-constants");
 
+/**
+ * @typedef {Object} RunBrowserResult
+ * @property {import('puppeteer').Page} page
+ * @property {import('puppeteer').Browser} browser
+ */
+
+/**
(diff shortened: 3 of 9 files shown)


lib/launchEditor.js
```diff
@@ -0,0 +1,307 @@
...
+      }
+    });
+
+    _childProcess.on("error", (error) => {
+      printInstructions(logger, fileName, error.message);
+    });
+  }
+}
+
+module.exports = launchEditor;
```
alexander-akait, line 307: Maybe there is a package for this? Also can we add unit tests for this?

lib/Server.js
```diff
@@ -2033,6 +2034,25 @@ class Server {
       }
     );
 
+    /** @type {import("express").Application} */
+    (app).get(
```
alexander-akait, line 2038: Let's add e2e test for this

alexander-akait (commented):
Looks good, we need to test it to avoid problems, also maybe we have a package for launching (just question)?

lib/launchEditor.js
malcolm-kee: I've installed [`launch-editor`](https://github.com/yyx990803/launch-editor) package, which apparently is extracted from `react-dev-utils`. It's also used by Vite as well.

lib/Server.js
malcolm-kee: Yup, it's added!

test/helpers/run-browser.js
```diff
@@ -3,6 +3,16 @@
...
 
+/**
+ * @typedef {Object} RunBrowserResult
+ * @property {import('puppeteer').Page} page
+ * @property {import('puppeteer').Browser} browser
+ */
+
+/**
+ * @param {Parameters<import('puppeteer').Page['emulate']>[0]} config
+ * @returns {Promise<RunBrowserResult>}
```
malcolm-kee, line 14: Additional typing to improve intellisense while writing tests.

lib/Server.js
```diff
@@ -2033,6 +2033,19 @@ class Server {
       }
     );
 
+    /** @type {import("express").Application} */
+    (app).get("/webpack-dev-server/open-editor", (req, res) => {
+      const fileName = req.query.fileName;
+
+      if (typeof fileName === "string") {
+        // @ts-ignore
+        const launchEditor = require("launch-editor");
```
malcolm-kee, line 2042: require lazily to make mock possible.

client-src/overlay.js
```diff
@@ -203,6 +203,16 @@ function show(type, messages, trustedTypesPolicyName) {
       typeElement.innerText = header;
       applyStyle(typeElement, msgTypeStyle);
 
+      if (message.moduleIdentifier) {
+        applyStyle(typeElement, { cursor: "pointer" });
+        typeElement.dataset.canOpen = true;
```
malcolm-kee, line 208: this is hook for e2e test. If there is other preferred convention, let me know!

alexander-akait (approved):
Sorry for delay, it looks good, @snitin315 Can you review this too?

snitin315 (approved):
Looks great. Thanks ⭐

alexander-akait:
@malcolm-kee Thank you, do we will have more PRs before release?

malcolm-kee:
@alexander-akait I plan to have one more enhancement for reporting runtime error, but I foreseen that will probably take time as we probably want to allow configuration (e.g. `react-error-overlay` has logic specific to React).

alexander-akait:
Thanks for answer, no rush, I will wait, thank you

malcolm-kee:
@alexander-akait Right now I'm a bit stuck with how to differentiate between error thrown by webpack vs application code, do you have any suggestion?

Let me know if this should be discussed in another channel!

Meanwhile I'll continue try to understand how it was done in `react-error-overlay`.

alexander-akait:
@malcolm-kee Feel free to send a PR how you see to improve it and we will continue dicussion in the PR, thank you

nanianlisao:
@malcolm-kee This function does not seem to work, how should I verify this function

malcolm-kee:
@nanianlisao can you raise an issue with a reproduction? We have test for this so it's apparently working for us.

nanianlisao:
@malcolm-kee Sorry about that, my bad. I thought we could do without the error-overlay-webpack-plugin plugin, but it looks like we're not there yet. We can only catch errors that are passed through the WebSocket for now.

PR #4605, opened 2022-10-16
feat: catch runtime error

malcolm-kee:
- [ ] This is a **bugfix**
- [x] This is a **feature**

### Motivation / Use-Case

1. Use finite state machine to model the overlay behavior, which can be represented with the diagram below

image

2. Currently the runtime error is caught but would need additional handling to display the source code that cause the error properly as the `filename` received in the `ErrorEvent` will be an URL prefixed with `webpack:///`. Not sure if we would be able to find the actual code path based on this.

image

Part of https://github.com/webpack/webpack-dev-server/issues/3689.

### Additional Info

Test results in IE10

compilation error

runtime error

### DIFF
--- a/client-src/overlay/state-machine.js
+++ b/client-src/overlay/state-machine.js
@@ -1,0 +1,89 @@
+import { createMachine, assign } from "@xstate/fsm";
+
+/**
+ * @typedef {Object} ShowOverlayData
+ * @property {'warning' | 'error'} level
+ * @property {Array<string  | { moduleIdentifier?: string, moduleName?: string, loc?: string, message?: string }>} messages
+ */
+
+/**
+ * @typedef {Object} CreateOverlayMachineOptions
+ * @property {(data: ShowOverlayData) => void} showOverlay
+ * @property {() => void} hideOverlay
+ */
+
+/**
+ * @param {CreateOverlayMachineOptions} options
+ */
+const createOverlayMachine = (options) => {
+  const { hideOverlay, showOverlay } = options;
+  const overlayMachine = createMachine(
+    {
+      id: "overlay",
+      initial: "hidden",
+      context: {
+        level: "error",
+        messages: [],
+      },
+      states: {
+        hidden: {
+          entry: "hideOverlay",
+          on: {
+            BUILD_ERROR: {
+              target: "displayBuildError",
+              actions: ["setMessages", "showOverlay"],
+            },
+            RUNTIME_ERROR: {
+              target: "displayRuntimeError",
+              actions: ["setMessages", "showOverlay"],
+            },
+          },
+        },
+        displayBuildError: {
+          on: {
+            DISMISS: { target: "hidden", actions: "dismissMessages" },
+            BUILD_ERROR: {
+              target: "displayBuildError",
+              actions: ["appendMessages", "showOverlay"],
+            },
+          },
+        },
+        displayRuntimeError: {
+          on: {
+            DISMISS: { target: "hidden", actions: "dismissMessages" },
+            RUNTIME_ERROR: {
+              target: "displayRuntimeError",
+              actions: ["appendMessages", "showOverlay"],
+            },
+            BUILD_ERROR: {
+              target: "displayBuildError",
+              actions: ["setMessages", "showOverlay"],
+            },
+          },
+        },
+      },
+    },
+    {
+      actions: {
+        dismissMessages: assign({
+          messages: [],
+          level: "error",
+        }),
+        appendMessages: assign({
+          messages: (context, event) => context.messages.concat(event.messages),
+          level: (context, event) => event.level || context.level,
+        }),
+        setMessages: assign({
+          messages: (_, event) => event.messages,
+          level: (context, event) => event.level || context.level,
+        }),
+        hideOverlay,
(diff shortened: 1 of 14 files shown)


malcolm-kee:
@alexander-akait the issue that I mentioned in [the previous PR](https://github.com/webpack/webpack-dev-server/pull/4587#issuecomment-1278945206) about differentiating build error vs runtime error is no longer a concern with this implementation as once any `BUILD_ERROR` event is emitted, the runtime error will not be displayed.

The next step I need to work on is to find out the source code that throws the runtime error (currently the info I get is the URL prefix with `webpack:///`), not sure if this is even possible if we do not have control on the source map.

malcolm-kee:
The modeling using finite-state machine is just a random idea that come to me today. Let me know if it's involving too much change and I will revert it. 

I'm using it to think through the desired behavior.

client-src/overlay/state-machine.js
```diff
@@ -0,0 +1,89 @@
+import { createMachine, assign } from "@xstate/fsm";
```
alexander-akait, line 1: What is size `@xstate/fsm`? Because extra size reduce build time

malcolm-kee: https://bundlejs.com/?q=%40xstate%2Ffsm

It's 3.94kb/1.82kb gzip.

alexander-akait: Looks like big... Maybe we can reimplement only necessary things or use more small alternative?

malcolm-kee: What's the acceptable size? I can shop around.

malcolm-kee: I just look at `@xstate/fsm`'s [implementation](https://github.com/statelyai/xstate/blob/main/packages/xstate-fsm/src/index.ts), which seems to be quite minimal.

I can think of two options:
1. use `@xstate/fsm` as baseline and continue code golf it to make it as small as possible (probably by removing some error check and some convenient API).
1. pre-minify it before build, but I not sure if this would help the compilation efficiency.

alexander-akait: I think the `1.` will be better, we can add a new API when we will need it

malcolm-kee: Sorry for no update for more than a month! 😅 

I've just created a custom `createMachine` while maintain enough compatibility so we can use their visualizer.

alexander-akait (approved):
Looks very good for me

@snitin315 Can you review too, also what do you think about include it in the next (or we can include it in master and get feedback faster)

snitin315 (dismissed):
Looks great ⭐ . @malcolm-kee can you also update the webpack 4 snapshots?

@alexander-akait I think we can release it in master. `next` will take time.

malcolm-kee:
I've updated the snapshot for webpack 4.

There are few things that I want to continue work on once this is merged:

1. Uncaught Promise rejection
1. Expose additional options to catch custom error for frameworks, e.g. the `react-error-overlay` has some custom code to catch React component errors
1. Ability to figure out the actual line of code that throws error. I suppose ability to do this would be dependent on the `devtool`/source map options, but I think it's still worthwhile to include that capability even it only supports certain use case only.

alexander-akait (dismissed):
Thank, you will do release tomorrow

alexander-akait:
@malcolm-kee looks like two tests for webpack v4 is failed, can you look at them?

alexander-akait (commented):
Just to ensure, can you test it using IE11-IE10 (some developer still need it), to ensure we don't use the modern browser API and babel works correctly here, just a manual test, and thank you for your hard work

malcolm-kee:
@alexander-akait is there any workflow/test case that I can refer for testing in IE? It seems like webpack code itself use arrow function, which is not supported in IE.

IE console showing syntax error of arrow function

malcolm-kee:
Also on side note, is it possible for this repo to request free account from Browserstack or Saucelab? I paid for a month subscription to do the testing above 😅

alexander-akait:
@malcolm-kee You can send a request to them :smile: I use virtual machines in such case

For testing please use `target: ['web', 'es5']`, so webpack will generate ES5 code

malcolm-kee:
Based on the fix above, overlay is displayed when there is build error. 

However the runtime error is not caught, probably something wrong with `window.addEventListener('error', cb)` is not behaving as expected in IE. Would need to check further.

alexander-akait:
Great, thank you fro this work, let's check if `window.addEventListener('error', cb)` is not fixable for IE, it is not a problem, because we I think no one use IE for development

alexander-akait:
@malcolm-kee friendly ping :star:

malcolm-kee:
Sorry for the delayed response.
I've find out the reason runtime error catching not working and apply the fix.
The screenshots are attached in PR description.

alexander-akait (approved):
@snitin315 Can you review too, thank you

snitin315 (approved):
Let's release this and gather feedback!

snitin315:
@alexander-akait you can merge and release.

alexander-akait:
Sorry for big delay, let's merge and release, really very sorry

PR #4694, opened 2022-12-24
refactor!: allow only Array for the proxy option

snitin315:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

Yes, updated tests to use `Array` syntax only.

### Motivation / Use-Case

Resolve TODO. 

### Breaking Changes

BREAKING CHANGE: the proxy option schema now only accepts an array

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -1272,53 +1272,6 @@ class Server {
      * }
      */
     if (typeof options.proxy !== "undefined") {
-      // TODO remove in the next major release, only accept `Array`
-      if (!Array.isArray(options.proxy)) {
-        if (
-          Object.prototype.hasOwnProperty.call(options.proxy, "target") ||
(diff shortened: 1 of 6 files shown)


lib/Server.js
```diff
@@ -1272,53 +1272,6 @@ class Server {
...
-                  .../** @type {ProxyConfigMap} */ (options.proxy)[context],
-                };
-                proxyOptions.context = correctedContext;
-              }
-
-              return proxyOptions;
-            }
-          );
-        }
-      }
```
alexander-akait, line 1320: We will need to write migration guide on this, because it is still the popular syntax :disappointed:

alexander-akait (approved):

lib/Server.js
snitin315: Yes. We will create a separate migration guide for v5.

Issue #4733, opened 2023-02-01
environment variable to indicate the presence of hot-loading


AprilArcus:
It would be convenient for Webpack to set an environment variable such as `process.env.WEBPACK_HOT`, akin to `process.env.WEBPACK_SERVE` and `process.env.WEBPACK_WATCH`, which would definitively indicate the availability of hot-loading (i.e. respecting both the `--no-hot` command-line argument and the `devServer: { hot: false }` config). This could then be read inside a babel config to enable or disable transforms relevant to hot loading.

alexander-akait:
PR welcome

alexander-akait:
Hard to implement without hacks (because we know you are using hot only from your configuration), please consider to use `--env`

PR #4734, opened 2023-02-11
set process.env.WEBPACK_HOT

AprilArcus:
Resolves https://github.com/webpack/webpack-dev-server/issues/4733

- [ ] This is a **bugfix**
- [x] This is a **feature**

WIP

### Motivation / Use-Case

Allow Babel configs, etc. to deterministically detect hot loading and enable/disable features as appropriate.

### Breaking Changes

None

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -3326,6 +3326,12 @@ class Server {
     if (typeof this.options.onListening === "function") {
       this.options.onListening(this);
     }
+
+    if (this.options.hot) {
+      // TODO fix me in the next major release
+      // @ts-ignore
+      process.env.WEBPACK_HOT = 'true';
+    }
   }
 
   /**


snitin315:
@AprilArcus Can you accept CLA and point this PR to the `next` branch?

lib/Server.js
```diff
@@ -3326,6 +3326,12 @@ class Server {
     if (typeof this.options.onListening === "function") {
       this.options.onListening(this);
     }
+
+    if (this.options.hot) {
+      // TODO fix me in the next major release
+      // @ts-ignore
+      process.env.WEBPACK_HOT = 'true';
+    }
```
alexander-akait, line 3334: It can be too late here, do you test it locally?

snitin315: Hmm, I think we should move this to `normalizeOptions()`
https://github.com/webpack/webpack-dev-server/blob/8817fa88edbe11259375a889cbabc4c7d0c603ac/lib/Server.js#L799

alexander-akait: Ideally we should set it in webpack-cli because it is an enviroment variable and we have to set it before we loading webpack configuration, so let's close it here and implement this on webpack-cli level, because I think we will not remove `--hot` never we can hardcode such logic

snitin315: But we can also use webpack-dev-server via the Nodejs API that doesn't require webpack-cli. In that case this variabke will be missing.

snitin315: But we can also use webpack-dev-server via the Nodejs API that doesn't require webpack-cli. In that case this variable will be missing.

alexander-akait: But we still run webpack-cli https://github.com/webpack/webpack-dev-server/blob/master/bin/webpack-dev-server.js#L103 :smile:

alexander-akait: Yeah, but in this case it is not our problem to set env variables (because when you have custom cli runner you can set any own variables without problem because I control own code), for webpack-cli we just have couple of built-in env variables and not more - popular cases

alexander-akait:
Closed in favor https://github.com/webpack/webpack-cli/pull/3659

PR #4773, opened 2023-03-16
fix: add client.overlay.runtimeErrors options

### LINKED ISSUE #4771: Webpack-dev-server version 4.12.0 causing compile errors in application
### Bug report
Webpack-dev-server version 4.12.0 causes compile errors in application on browser viewport resize

### Actual Behavior

We run an application which uses webpack-dev-server 4.12.0 as a dependency. With this version of webpack-dev-server, whenever we resize the browser viewport in our application (e.g. by opening developer tools), an error overlay pops up stating "ResizeObserver loop limit exceeded". _This behaviour does not occur in the exact same version of the application that uses webpack-dev-server version 4.11.1_.

image

### Expected Behavior

The error should not appear when resizing the browser viewport in any way.

### How Do We Reproduce?

- Run an application that uses webpack-dev-server version 4.12.0
- Open developer tools in the browser

image

image

image

malcolm-kee:
- [x] This is a **bugfix**
- [x] This is a **feature**

Reported in #4771

fixes https://github.com/webpack/webpack-dev-server/issues/4774
fixes #4771

- Added `client.overlay.runtimeErrors` options - default to `true`.

### DIFF
--- a/client-src/index.js
+++ b/client-src/index.js
@@ -116,23 +116,17 @@ self.addEventListener("beforeunload", () => {
   status.isUnloading = true;
 });
 
-const overlay = options.overlay
-  ? createOverlay(
-      typeof options.overlay === "object"
-        ? {
-            trustedTypesPolicyName: options.overlay.trustedTypesPolicyName,
-            catchRuntimeError: options.overlay.runtimeErrors,
-          }
-        : {
-            trustedTypesPolicyName: false,
-            catchRuntimeError: options.overlay,
-          }
-    )
-  : {
-      send() {
-        // noop
-      },
-    };
+const overlay = createOverlay(
+  typeof options.overlay === "object"
+    ? {
+        trustedTypesPolicyName: options.overlay.trustedTypesPolicyName,
+        catchRuntimeError: options.overlay.runtimeErrors,
+      }
+    : {
+        trustedTypesPolicyName: false,
+        catchRuntimeError: options.overlay,
+      }
+);
 
 const onSocketMessage = {
   hot() {
--- a/client-src/index.js
+++ b/client-src/index.js
@@ -116,17 +116,20 @@ self.addEventListener("beforeunload", () => {
   status.isUnloading = true;
 });
 
-const overlay = createOverlay(
-  typeof options.overlay === "object"
-    ? {
-        trustedTypesPolicyName: options.overlay.trustedTypesPolicyName,
-        catchRuntimeError: options.overlay.runtimeErrors,
-      }
-    : {
-        trustedTypesPolicyName: false,
-        catchRuntimeError: options.overlay,
-      }
-);
+const overlay =
+  typeof window !== "undefined"
+    ? createOverlay(
+        typeof options.overlay === "object"
+          ? {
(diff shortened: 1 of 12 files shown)


malcolm-kee:
@alexander-akait I added the fix and verified it by testing it manually. 

If automated test is a must-have I'll have to do it tomorrow as I'm very tired now (it's 3am here).

client-src/index.js
```diff
@@ -118,9 +118,15 @@ self.addEventListener("beforeunload", () => {
 const trustedTypesPolicyName =
   typeof options.overlay === "object" && options.overlay.trustedTypesPolicyName;
 
-const overlay = createOverlay({
-  trustedTypesPolicyName,
-});
+const overlay = options.overlay
```
crobinson42, line 121: @malcolm-kee  this is not sufficient. You are only checking if `options.overlay` is truthy. All options need to be passed into the `createOverlay()` function. The `overlay` options are `errors: boolean` & `warnings: boolean`. 

Then, the `show()` function in `overlay.js` needs to check if the `type` is "warning" or "error" and compare the passed options.

alexander-akait: Yeah, we have before https://github.com/webpack/webpack-dev-server/blob/v4.11.1/client-src/index.js#L253

malcolm-kee: @crobinson42 good catch!

I just pushed a fix so that runtime error check the options as well.

For build error/warning, it's already handled:

- (errors) https://github.com/webpack/webpack-dev-server/blob/master/client-src/index.js#L297
- (warnings) https://github.com/webpack/webpack-dev-server/blob/master/client-src/index.js#L265

alexander-akait: Testa are failed, can you look what is wrong (no rush), I think today after rest we will fix all issues

client-src/index.js
```diff
@@ -115,12 +115,23 @@ self.addEventListener("beforeunload", () => {
...
-
-const overlay = createOverlay({
-  trustedTypesPolicyName,
-});
+const overlay = options.overlay
+  ? createOverlay(
+      typeof options.overlay === "object"
+        ? {
+            trustedTypesPolicyName: options.overlay.trustedTypesPolicyName,
+            catchRuntimeError: options.overlay.errors,
```
alexander-akait, line 123: What abour warning?

alexander-akait, line 123: Also ideally we need a new option for such stuff, i.e. `catchRuntimeError`, because as you can see some application works in very differents envs and it can be errors outside own application

client-src/index.js
```diff
@@ -80,6 +80,7 @@ if (parsedResourceQuery.overlay) {
     options.overlay = {
       errors: true,
       warnings: true,
+      runtimeErrors: true,
```
malcolm-kee, line 83: Default it to true.

malcolm-kee:
@alexander-akait I'm having problem to run the test for webpack5 to update snapshot. I tried to run it it will stuck without any reason. It's the same with `master` branch so my setup has issue. 😅

alexander-akait:
I will update tests don't worry

alexander-akait (approved):

PR #4778, opened 2023-03-17
fix: make webpack optional peer dependency

### LINKED ISSUE #4776: Add webpack as optional peer dependency
### Feature Proposal
Rspack use webpack-dev-server internally to compatible with devServer functionality but it will cause install warning because lack of webpack. see https://github.com/web-infra-dev/rspack/issues/2318
```sh
└─┬ @rspack/cli 0.1.1
  └─┬ @rspack/dev-server 0.1.1
    ├─┬ webpack-dev-server 4.11.1
    │ ├── ✕ missing peer webpack@"^4.37.0 || ^5.0.0"
    │ └─┬ webpack-dev-middleware 5.3.3
    │   └── ✕ missing peer webpack@"^4.0.0 || ^5.0.0"
    └─┬ @rspack/dev-middleware 0.1.1
      └─┬ webpack-dev-middleware 6.0.0
        └── ✕ missing peer webpack@^5.0.0
Peer dependencies that should be installed:
  webpack@">=5.0.0 <6.0.0"
```
we are wondering whether webpack-dev-server could list webpack as optional peer dependency to help avoid this warning in rspack

### Feature Use Case
support use webpack-dev-server in rspack
none

hardfist:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No, not easy to test in this repo

### Motivation / Use-Case
fixes #4776 

### Breaking Changes
No

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,6 +139,9 @@
   "peerDependenciesMeta": {
     "webpack-cli": {
       "optional": true
+    },
+    "webpack": {
+      "optional": true
     }
   }
 }


package.json
```diff
@@ -139,6 +139,9 @@
   "peerDependenciesMeta": {
     "webpack-cli": {
       "optional": true
+    },
+    "webpack": {
+      "optional": true
     }
```
snitin315, line 145: You may also want to remove it from the `peerDependencies`field  above

https://github.com/webpack/webpack-dev-server/blob/4e846078b3b9befafbc56b1cd6db7778175355c9/package.json#L137

hardfist: if we remove `webpack` from `peerDependencies` it would mean any `webpack` version would be ok for `webpack-dev-server` but `webpack-dev-server` did require specific `webpack` version according to the documentation of https://pnpm.io/next/package_json#peerdependenciesmetaoptional

alexander-akait (approved):

package.json
alexander-akait: Yeah, we still need versions, in theory this PR can break a code for some inattentive developers (which ignored warnings from package managers), if you don't have webpack in dependcies and rely on the new behaviour of npm (autoinstall peer deps), but it should be fixed on their side, because you should always list of you dependecies

PR #4785, opened 2023-03-22
Fix host ip all

Phecda:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No, but updated an old test case

### Motivation / Use-Case

`0.0.0.0` is not a good address to visit. Some developers working on Windows cannot open `0.0.0.0` in browser. 

### Additional Info
Before this fix:

screenshot

After this fix, it will open `http://localhost:8080` directly.

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -2893,7 +2893,11 @@ class Server {
       }
 
       if (/** @type {NormalizedOpen[]} */ (this.options.open).length > 0) {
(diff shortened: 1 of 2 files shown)


Phecda:
OK, I add a screenshot demonstrating how it looks before this fix.

alexander-akait:
Very funny, I always thought that it works on all platforms, but it turns out that Windows does not support confused

Issue #4777, opened 2023-03-17
[feat] allow to filter runtime errors/errors/warning in overlay


alexander-akait:
The idea:
```
client: {
 errors?: boolean | (error: Error) => boolean
 warnings?: boolean | (warn: unknown) => boolean
 runtimeErrors?: boolean | (error: Error) => boolean
}
```

But let's implement boolean firstly, so you can disable it if you need, and then we can look at serilization

_Originally posted by @alexander-akait in https://github.com/webpack/webpack-dev-server/issues/4771#issuecomment-1472913070_

malcolm-kee:
@alexander-akait I thought boolean already supported? Or the intention is to move it from `client.overlay.[messageType]` to `client.[messageType]`?

alexander-akait:
@malcolm-kee Yeah, boolean already supports, I mean we need to allow to filter them, i.e. pass function to our client (serilized) and use (deserilize) in the client code, so some errors can be filtered without filly disable functional of `errors`/`warnings`/`runtimeErrors`, for example:
```
client: {
  runtimeErrors: (error) { 
   return !/Any custom/.test(error.message);
  }
}
```

malcolm-kee:
@alexander-akait I would love to work on this but quite clueless how to do this, as all current options are JSON values. Do you have any pointer, such as plugins with similar capabilities that I can take inspiration from?

CarryOnHxy:
Very looking forward to see this new feature when i saw the constant ```ResizeObserver``` problem on the screen!!

alexander-akait:
Fixed https://github.com/webpack/webpack-dev-server/pull/4813, release will be soon

malcolm-kee:
This is now documented: https://webpack.js.org/configuration/dev-server/#overlay

PR #4813, opened 2023-04-11
feat: allow filter overlay message with function

malcolm-kee:
- [ ] This is a **bugfix**
- [x] This is a **feature**

### Motivation / Use-Case

- Allow filtering of overlay message
- Fix misleading overlay title when catching runtime error

Resolves

- #4777

### DIFF
--- a/client-src/index.js
+++ b/client-src/index.js
@@ -11,11 +11,19 @@ import reloadApp from "./utils/reloadApp.js";
 import createSocketURL from "./utils/createSocketURL.js";
 
 /**
+ * @typedef {Object} OverlayOptions
+ * @property {boolean | (error: Error) => boolean} [warnings]
+ * @property {boolean | (error: Error) => boolean} [errors]
+ * @property {boolean | (error: Error) => boolean} [runtimeErrors]
+ * @property {string} [trustedTypesPolicyName]
+ */
+
+/**
  * @typedef {Object} Options
  * @property {boolean} hot
  * @property {boolean} liveReload
  * @property {boolean} progress
- * @property {boolean | { warnings?: boolean, errors?: boolean, runtimeErrors?: boolean, trustedTypesPolicyName?: string }} overlay
+ * @property {boolean | OverlayOptions} overlay
  * @property {string} [logging]
  * @property {number} [reconnect]
  */
@@ -83,6 +91,23 @@ if (parsedResourceQuery.overlay) {
       runtimeErrors: true,
       ...options.overlay,
     };
+
+    ["errors", "warnings", "runtimeErrors"].forEach((property) => {
+      if (typeof options.overlay[property] === "string") {
+        const overlayFilterFunctionString = decodeURIComponent(
+          options.overlay[property]
+        );
+
+        // eslint-disable-next-line no-new-func
+        const overlayFilterFunction = new Function(
+          "message",
+          `var callback = ${overlayFilterFunctionString}
+        return callback(message)`
+        );
+
+        options.overlay[property] = overlayFilterFunction;
+      }
+    });
   }
   enabledFeatures.Overlay = true;
 }
@@ -266,17 +291,24 @@ const onSocketMessage = {
       log.warn(printableWarnings[i]);
     }
 
-    const needShowOverlayForWarnings =
+    const overlayWarningsSetting =
       typeof options.overlay === "boolean"
         ? options.overlay
         : options.overlay && options.overlay.warnings;
 
-    if (needShowOverlayForWarnings) {
-      overlay.send({
-        type: "BUILD_ERROR",
-        level: "warning",
-        messages: warnings,
-      });
+    if (overlayWarningsSetting) {
+      const warningsToDisplay =
+        typeof overlayWarningsSetting === "function"
+          ? warnings.filter(overlayWarningsSetting)
+          : warnings;
+
+      if (warningsToDisplay.length) {
+        overlay.send({
+          type: "BUILD_ERROR",
+          level: "warning",
+          messages: warnings,
+        });
+      }
     }
 
     if (params && params.preventReloading) {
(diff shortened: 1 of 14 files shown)


malcolm-kee:
Haven't look into test, but based on manual testing on the example it's working. In case anyone want to take over feel free to proceed to fork my implementation and update on top, my time for OSS is quite limited recently. 😢

lib/Server.js
```diff
@@ -628,12 +632,28 @@ class Server {
...
-          searchParams.set(
-            "overlay",
+          /**
+           *
+           * @param {OverlayMessageOptions} [setting]
+           * @returns
+           */
+          const encodeOverlaySettings = (setting) =>
+            typeof setting === "function"
+              ? encodeURIComponent(setting.toString())
```
malcolm-kee, line 642: Usage of encodeURIComponent here is intentional, because without it `URLSearchParams` will encode white space into `+` character, which could not be decoded correctly in `parseURL` in client.

Alternative is to use `URLSearchParams` in client side, but I think the drawback of that is cannot support newer browser.

client-src/index.js
```diff
@@ -83,6 +91,23 @@ if (parsedResourceQuery.overlay) {
       runtimeErrors: true,
       ...options.overlay,
     };
+
+    ["errors", "warnings", "runtimeErrors"].forEach((property) => {
+      if (typeof options.overlay[property] === "string") {
```
malcolm-kee, line 96: This logic could be a breaking change for those passing `string` to the options, as we will assume string = encoded function, meanwhile previously it was coerce to boolean.

malcolm-kee: Maybe not since we have schema validation.

malcolm-kee:
@alexander-akait slightly out-of-topic, any plan to migrate from puppeteer to playwright or upgrade puppeteer? I can't run test locally which I suspect is because of puppeteer 😢

alexander-akait:
@malcolm-kee Yeah, it will be great, I think it is not a problem problem, because most of our code is encupsulated also API is almost the same, so feel free to try it I will help

client-src/index.js
```diff
@@ -10,12 +10,20 @@ import sendMessage from "./utils/sendMessage.js";
 import reloadApp from "./utils/reloadApp.js";
 import createSocketURL from "./utils/createSocketURL.js";
 
+/**
+ * @typedef {Object} OverlayOptions
+ * @property {boolean | (message: unknown) => boolean} [warnings]
+ * @property {boolean | (message: unknown) => boolean} [errors]
+ * @property {boolean | (message: unknown) => boolean} [runtimeErrors]
```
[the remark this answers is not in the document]
malcolm-kee: Fix types of message to `Error`.

alexander-akait (commented):
Looks good, I see we don't have tests on it right? is it hard to test?

alexander-akait:
Yeah, thank you, feel free to ping me if you will need help

alexander-akait (approved):

alexander-akait:
Thank you good job, I think we can merge and release it

malcolm-kee:
@alexander-akait I'm adding more tests and realize that the following code will overwrite the overlay settings:

https://github.com/webpack/webpack-dev-server/blob/master/client-src/index.js#L170-L176

What's the use case of supporting that socket message from server?

alexander-akait:
@malcolm-kee some tools can't inject your code in webpack bundle process, but they need to enable/disable some features, so we provide ability to do, i.e you can use own websocket server and do it, but I am thinking no one currently use it

malcolm-kee:
Got it. For completeness let me handle that.

client-src/index.js
```diff
@@ -10,12 +10,20 @@ import sendMessage from "./utils/sendMessage.js";
 import reloadApp from "./utils/reloadApp.js";
 import createSocketURL from "./utils/createSocketURL.js";
 
+/**
+ * @typedef {Object} OverlayOptions
+ * @property {boolean | (error: Error) => boolean} [warnings]
```
malcolm-kee, line 15: @alexander-akait When adding tests I realize that the error here can be `string` in webpack4. Should I update the type to reflect that?

malcolm-kee:
I found bug. Checking...

client-src/index.js
alexander-akait: hm, is it only type problem? Because if yes, no need, I want to do this release and focus on the next major release, drop old node.js version, remove deprecation and unnecessary options and etc and drop webpack v4

malcolm-kee:
@alexander-akait by the way I also plan to let `runtimeErrors` catch unhandled promise rejection as well, but probably in next PR after this.

client-src/index.js
```diff
@@ -266,17 +301,24 @@ const onSocketMessage = {
...
+      const warningsToDisplay =
+        typeof overlayWarningsSetting === "function"
+          ? warnings.filter(overlayWarningsSetting)
+          : warnings;
+
+      if (warningsToDisplay.length) {
+        overlay.send({
+          type: "BUILD_ERROR",
+          level: "warning",
+          messages: warnings,
```
snitin315, line 319: Shouldn't this be `BUILD_WARNING`?

malcolm-kee: the event type purpose is to determine how to transition the state:

https://github.com/webpack/webpack-dev-server/blob/f7386f60da170b158d13a2c41cc668f0b83d9679/client-src/overlay/state-machine.js#L32-L38

I don't mind refactor to add event type `BUILD_WARNING` and remove `level` property, although it does introduce some duplication in the state machine.

snitin315: I think it's fine 👍🏻

snitin315 (approved):
LGTM, thank you for working on this 🚀 

We will also need to document this on https://webpack.js.org/configuration/dev-server/#overlay

alexander-akait (approved):

client-src/index.js
```diff
@@ -27,6 +35,30 @@ import createSocketURL from "./utils/createSocketURL.js";
...
+        const overlayFilterFunctionString = decodeURIComponent(
+          overlayOptions[property]
+        );
+
+        // eslint-disable-next-line no-new-func
+        const overlayFilterFunction = new Function(
+          "message",
+          `var callback = ${overlayFilterFunctionString}
+        return callback(message)`
+        );
```
wentokay, line 54: this `new Function()` breaks devserver for chrome extensions with the following error

`Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'".`

it could be fixed by adding `'unsafe-eval'` in the manifest but that isn't really an option for us so we're going to roll back for now

alexander-akait: @wentokay We can optionally inject this, can you open an issue?

PR #4818, opened 2023-04-15
fix(perf): reduced initial start time

alexander-akait:
- [x] This is a **bugfix**
- [ ] This is a **feature**

No need

### Motivation / Use-Case

Start faster

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -6,8 +6,6 @@ const url = require("url");
 const util = require("util");
 const fs = require("graceful-fs");
 const ipaddr = require("ipaddr.js");
-const defaultGateway = require("default-gateway");
-const express = require("express");
 const { validate } = require("schema-utils");
 const schema = require("./options.json");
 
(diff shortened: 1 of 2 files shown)


lib/Server.js
```diff
@@ -355,7 +381,7 @@ class Server {
    */
   static internalIPSync(family) {
     try {
-      const { gateway } = defaultGateway[family].sync();
+      const { gateway } = require("default-gateway")[family].sync();
```
alexander-akait, line 384: We use only one method or async or sync, so it is fine to use `require`

alexander-akait: Also we don't it when a developer provide IP, so avoid it for some cases

lib/Server.js
```diff
@@ -210,6 +208,34 @@ if (!process.env.WEBPACK_SERVE) {
...
+    // Allow to clean up memory for fn
+    // and all dependent resources
+    // eslint-disable-next-line no-undefined
+    fn = undefined;
+
+    return result;
+  };
+};
+
+const getExpress = memoize(() => require("express"));
```
alexander-akait, line 237: Yeah, we always run `express`, but let's do it lazy

snitin315 (approved):
Looks good.

PR #4849, opened 2023-05-07
feat: overlay displays unhandled promise rejection

malcolm-kee:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

`overlay.runtimeErrors` should capture unhandled Promise rejection, as it's still runtime error, just async in nature.

### Breaking Changes

Not breaking change.

### Additional Info

Part of #3689


alexander-akait (approved):

yoyo837:
How do I make unhandled promise rejection not overlay displays, so that it behaves the same as before.

yoyo837:
> How do I make unhandled promise rejection not overlay displays, so that it behaves the same as before.

```diff
overlay: {
++runtimeErrors: false,
  errors: true,
  warnings: false,
},
```

This is the same behavior as before.

keremcanb:
This feature gave us so much headache that we had to downgrade to the previous version.

alexander-akait:
@keremcanb just disable it using options

keremcanb:
We are running on CRA, which is not ejected. Also we don't want to disable
all runtime error messages.

alexander-akait:
You can filter them, sorry we will not remove this feature

tomasztomys:
It should be break change!

PR #4856, opened 2023-05-09
fix: replace :: with localhost before openBrowser()

### LINKED ISSUE #4855: Using --host :: with --open results in ERR_ADDRESS_INVALID on Windows (and less-than-ideal behavior on GitHub Codespaces)
### Bug report

Currently, using `webpack-dev-server` with `--host ::` and `--open` options results in an `ERR_ADDRESS_INVALID` error in browsers (Chrome, Edge, etc.) on Windows.

The reason is that we are sending `http://[::]` to the browser and the IPv6 unspecified address `::` is considered an invalid host address on Windows.

Note that on macOS and Linux, `::` is automatically mapped to `::1` and `0.0.0.0` is automatically mapped to `127.0.0.1`. So we have this problem only on Windows.

Interestingly, although `0.0.0.0` is also considered invalid on Windows, using --host `0.0.0.0` with `--open` is fine because we are replacing the IPv4 unspecified address `0.0.0.0` with `localhost` before opening the browser.

So, to fix the problem, we can do the same thing for the IPv6 address: replace `::` with `localhost` when the `--open` option is used.

Besides, on GitHub Codespaces, if we use `localhost`, the browser opening event will be intercepted and redirected correctly. If we use `::`, the browser will be redirected to the wrong address.

I am working on a PR to fix the problem.

### Actual Behavior

When running `webpack-dev-server --host :: --open`, `http:://[::]` is sent to the browser.

### Expected Behavior

When running `webpack-dev-server --host :: --open`, `http://localhost` is sent to the browser.

### How Do We Reproduce?

Execute an example with the options `--host ::` and `--open` **on Windows**:

```
$ cd examples/api/simple
$ webpack-dev-server --host :: --open
```

  System:
    OS: **Windows** 10 10.0.20348
    CPU: (1) x64 Intel(R) Xeon(R) CPU E5-2676 v3 @ 2.40GHz
    Memory: 414.89 MB / 1023.61 MB
  Binaries:
    Node: 20.1.0 - C:\\Program Files\\nodejs\\node.EXE
    npm: 9.6.4 - C:\\Program Files\\nodejs\\npm.CMD
  Browsers:
    Internet Explorer: 11.0.20348.380
    Google Chrome: 113.0.5672.93
    Microsoft Edge: 113.0.1774.35

Related: #1233

chousheng:
- [x] This is a **bugfix**
- [ ] This is a **feature**

[test/server/open-option.test.js](https://github.com/webpack/webpack-dev-server/compare/master...chousheng:webpack-dev-server:fix-ipv6-unspecified-addr-open-browser?expand=1#diff-9b579ac0614b984311440c72f67e512825ea9e553424a2360fb7503556d15625) is modified.

### Motivation / Use-Case

This PR resolves #4855. See #4855 for more details.

### Breaking Changes

N/A

### Additional Info

Below are the screencasts showing the fix on Windows and GitHub Codespaces.

Windows:

| Before fix     | After fix      |
| -------------- | -------------- |
| <video src="https://github.com/webpack/webpack-dev-server/assets/38355699/57d2fa7f-744b-47aa-b274-553351c27833"> | <video src="https://github.com/webpack/webpack-dev-server/assets/38355699/12cf49f9-03a9-49c2-b347-b0b40f271b97"> |

GitHub Codespaces:

| Before fix     | After fix      |
| -------------- | -------------- |
| <video src="https://github.com/webpack/webpack-dev-server/assets/38355699/f8ed93d4-65fc-4cce-a4bf-5b72289d7755"> | <video src="https://github.com/webpack/webpack-dev-server/assets/38355699/b56161c3-7f74-48b6-bd11-06a7190332cb"> |

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -2955,7 +2955,9 @@ class Server {
 
       if (/** @type {NormalizedOpen[]} */ (this.options.open).length > 0) {
         const openTarget = prettyPrintURL(
-          !this.options.host || this.options.host === "0.0.0.0"
+          !this.options.host ||
+            this.options.host === "0.0.0.0" ||
+            this.options.host === "::"
             ? "localhost"
             : this.options.host
         );
--- a/test/server/open-option.test.js
+++ b/test/server/open-option.test.js
@@ -134,7 +134,7 @@ describe('"open" option', () => {
     await server.start();
     await server.stop();
 
-    expect(open).toHaveBeenCalledWith(`http://[${host}]:${port}/`, {
+    expect(open).toHaveBeenCalledWith(`http://localhost:${port}/`, {
       wait: false,
     });
   });


lib/Server.js
```diff
@@ -2955,7 +2955,9 @@ class Server {
 
       if (/** @type {NormalizedOpen[]} */ (this.options.open).length > 0) {
         const openTarget = prettyPrintURL(
-          !this.options.host || this.options.host === "0.0.0.0"
+          !this.options.host ||
+            this.options.host === "0.0.0.0" ||
+            this.options.host === "::"
```
snitin315, line 2960: I think we can do it only for windows OS, otherwise we can keep the behavior same as before.

alexander-akait: I think it is fine for all OS

alexander-akait (approved):

lib/Server.js
chousheng: Sorry, I accidentally deleted my previous comment. (I thought canceling the review would not delete the associated comment.) I am adding the comment back here:

Hi snitin315, 

Thank you for reviewing the change.

I thought I would do that only for Windows, but I encountered a similar problem on GitHub Codespaces (as I mentioned in #4855), where `::` was not redirected correctly. So I think using `localhost` directly for all platforms is universal. That works on macOS, Linux, Windows, and GitHub Codespaces.

I just found PR #4785, where we made the change to replace the IPv4 unspecified address `0.0.0.0` with `localhost` to support Windows. However, we did not add an additional check to do that only for Windows. I believe doing the same thing for the IPv6 unspecified address `::` would make the behavior and the code more consistent (for both the IPv4 and IPv6 unspecified addresses, `0.0.0.0` and `::`).

How do you feel?

I am looping in alexander-akait, who merged #4785.

chousheng:
@alexander-akait, @snitin315. Thank you for reviewing!

Please let me know if I need to rebase the commit or do anything else before getting the PR merged.

snitin315 (approved):

snitin315:
We will merge this once CI is green.

PR #4861, opened 2023-05-13
refactor: refactor overlay styling to use css class

malcolm-kee:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

N/A

### Motivation / Use-Case

This PR refactor styling for overlay from inline styles to CSS class, which provides the following benefits:

- unlocks more CSS features, such as pseudo-selectors
- make the snapshots more readable, as the snapshots would only show the markup instead of all the CSS styles

### Breaking Changes

N/A

### Additional Info

This is prepare for better overlay styling that I show in https://github.com/webpack/webpack-dev-server/issues/3689#issuecomment-1545920152

### DIFF
--- a/client-src/overlay.js
+++ b/client-src/overlay.js
@@ -268,9 +268,7 @@ const createOverlay = (options) => {
         // Make it look similar to our terminal.
         const text = ansiHTML(encode(body));
         const messageTextNode = document.createElement("div");
-        messageTextNode.className = msgTextStyle.className;
-
-        iframeCssLoader.load(msgTextStyle.css);
+        applyCss(messageTextNode, msgTextStyle);
 
         messageTextNode.innerHTML = overlayTrustedTypesPolicy
           ? overlayTrustedTypesPolicy.createHTML(text)
--- a/client-src/overlay.js
+++ b/client-src/overlay.js
@@ -11,6 +11,7 @@ import {
 import createOverlayMachine from "./overlay/state-machine.js";
 import {
   containerStyle,
+  createCssLoader,
   dismissButtonStyle,
   headerStyle,
   iframeStyle,
@@ -93,6 +94,8 @@ const createOverlay = (options) => {
   let containerElement;
   /** @type {HTMLDivElement | null | undefined} */
   let headerElement;
+  /** @type {import('./overlay/styles.js').CssLoader} */
+  let iframeCssLoader;
   /** @type {Array<(element: HTMLDivElement) => void>} */
   let onLoadQueue = [];
   /** @type {TrustedTypePolicy | undefined} */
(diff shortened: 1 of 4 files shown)


client-src/overlay/styles.js
```diff
@@ -24,58 +16,131 @@ const iframeStyle = {
...
+ * @typedef {Object} CssLoader
+ * @property {(css: string) => void} load
+ */
+
+/**
+ *
+ * @param {Document} doc
+ * @return {CssLoader}
+ */
+const createCssLoader = (doc) => {
```
malcolm-kee, line 129: css loader could be confusing? Any suggestion is welcome.

client-src/overlay.js
```diff
@@ -129,12 +142,10 @@ const createOverlay = (options) => {
     applyStyle(iframeContainerElement, iframeStyle);
```
malcolm-kee, line 142: Intentionally keep iframe styling to use inline style because we can't scope CSS for it.

client-src/overlay/styles.js
```diff
@@ -24,58 +16,131 @@ const iframeStyle = {
...
 
 const msgTextStyle = {
-  lineHeight: "1.5",
-  fontSize: "1rem",
-  fontFamily: "Menlo, Consolas, monospace",
+  className: "webpack-msg-text",
+  css: /* css */ `.webpack-msg-text {
+      line-height: 1.5;
+      font-size: 1rem;
+      font-family: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
```
malcolm-kee, line 95: Lovingly copied from TailwindCSS `font-mono`: https://tailwindcss.com/docs/font-family

client-src/overlay/styles.js
```diff
@@ -24,58 +16,131 @@ const iframeStyle = {
...
+    load: (css) => {
+      // ignore CSS rule that has loaded before
+      if (!loadedCss.includes(css)) {
+        const style = doc.createElement("style");
+        style.innerHTML = css;
+        doc.head.appendChild(style);
+        loadedCss.push(css);
+      }
+    },
+  };
```
alexander-akait, line 143: It can create a lot of problems. Shorty - we already tried this solution and we had a lot of issues about `style` tags and its order and overlapping and etc and a lot, so I prefer do not use style and classes at all, I am fine with refactor and move styles in own objects with good names, but using classes again make more problems than benefits

malcolm-kee: Got it. Any issues that was raised before on this? Maybe worth include them in the comments

alexander-akait: They were, but they are difficult to find because they are scattered and many headings do not match the original problem, I tried to search for them, but unfortunately I can not find but I remember very well that we had problems with this :smile:

PR #5080, opened 2024-03-01
chore: refactor e2e tests to playwright

mahdikhashan:
### Final Report GSoC 2024

- Student Name: [Mahdi Khashan](https://github.com/mahdikhashan)
- Organization: [Webpack](https://github.com/webpack)
- Mentors: @snitin315 @alexander-akait @evenstensberg 
- Project: [webpack-dev-server end to end tests refactor](https://summerofcode.withgoogle.com/programs/2024/projects/vvPVCqzQ)

---

#### Project

The project has been to substitute Puppeteer framework with MS Playwright in webpack-dev-server project and refactor tests afterward.

I have set up Playwright and configured it in a way to be similar to the existing code, I have extended the default `test` object and added a custom matcher for snapshot comparison of objects and arrays. Since previously Jest was used as Puppeteer's test runner and Playwright has its own runner, I have added Sinon.js as a mock / stubbing library. 

Tests are ran over Linux, Mac and Windows using node 18, 20 and 22. The code coverage is calculated using `nyc` and `instanbul` and it's babel plugin and pushed to [CodeCov](https://app.codecov.io/gh/webpack/webpack-dev-server/pull/5080).

#### Constraints and Future Improvement

Since by using the Playwright we could benefit from its flaky test marking mechanism, some tests are labeled as flaky which we may want to refactor them later.

Additionally, some tests were failing which I decided to do a root-cause analysis and figured out that by replacing `default-gateway` npm package, it could pass them successfully. ([link to pr](https://github.com/webpack/webpack-dev-server/pull/5255))

#### Thank you

I want to sincerely thank Webpack team member @evenstensberg for taking the time to review my proposal for the project in advance. Special thanks to my mentors @snitin315 and @alexander-akait which helped me when I ran into any issues or had questions.

---

WIP: will refactors tests one by one and rename the e2e-playwright folder to e2e after its done

- [ ] This is a **bugfix**
- [ ] This is a **feature**

### Motivation / Use-Case

It will introduce playwright in the project to replace the existing puppeteer end to end framework.


alexander-akait (commented):
Ideally, we need to rewrite existing tests on `playwright` + some refactor our tests, because we have a lot of code duplication

mahdikhashan:
@alexander-akait thanks for your comment, for participating in the gsoc-2024, is this PR a valid patch requirements? if not could you introduce an issue which I can start working on? I'm a master student and a junior developer. regarding the proposal, is there anything specific to be mentioned on or do you have a template from the previous years? also, if there is a discord of slack channel which I can share the draft version of proposal and receive early feedback, I appreciate it.

snitin315:
> I think this should work for a gsoc project. (90h project)

Yes, I agree.

playwright.config.js
```diff
@@ -0,0 +1,20 @@
+"use strict";
+
+module.exports = {
+  testDir: "./test/e2e-playwright",
+  fullyParallel: true,
+  forbidOnly: !!process.env.CI,
+  retries: process.env.CI ? 2 : 0,
```
evenstensberg, line 7: make the CI env a const variable instead

mahdikhashan: done

playwright.config.js
```diff
@@ -0,0 +1,29 @@
...
+
+const CI = false;
+
+module.exports = {
+  testDir: "./test/e2e-playwright",
+  // however this can have benefits, in tests I couldn't start a server on multiple ports
+  // for tests "stats-refactored.test.js" the port is read from "ports-map.js" file somehow
+  // I should manage to handle it if I want to run the tests in parallel mode
+  fullyParallel: false,
+  forbidOnly: !!CI,
```
snitin315, line 11:
```suggestion
  forbidOnly: process.env.CI !== "true",
```

playwright.config.js
```diff
@@ -0,0 +1,29 @@
+"use strict";
+
+const CI = false;
```
snitin315, line 3: Use process.env.CI

playwright.config.js
```diff
@@ -0,0 +1,29 @@
...
+module.exports = {
+  testDir: "./test/e2e-playwright",
+  // however this can have benefits, in tests I couldn't start a server on multiple ports
+  // for tests "stats-refactored.test.js" the port is read from "ports-map.js" file somehow
+  // I should manage to handle it if I want to run the tests in parallel mode
+  fullyParallel: false,
+  forbidOnly: !!CI,
+  retries: CI ? 2 : 0,
+  workers: 1,
+  reporter: "html",
```
snitin315, line 14: We don't need HTML report when running tests on CI

mahdikhashan: you are right, while I was checking the playwright documentations, I noticed there is an option for Github Action annotations, so I have set it to report to Github Action in case its running in CI, otherwise use list format.

playwright.config.js
mahdikhashan: fixed.

playwright.config.js
mahdikhashan: done.

playwright.config.js
snitin315: It should be like - `const isCI = process.env.CI === 'true'` at the top, then directly use `isCI` everywhere else.

.github/workflows/nodejs.yml
```diff
@@ -12,6 +12,18 @@ on:
...
+    inputs:
+      update-snapshots:
+        description: "Update snapshots?"
+        type: boolean
+
+  workflow_dispatch:
+    inputs:
+      update-snapshots:
+        description: "Update snapshots?"
+        type: boolean
```
snitin315, line 25: Let's not add this. Tests should ideally fail if snapshots are not in sync with the master branch, this would let user bypass failing CI.

.github/workflows/nodejs.yml
```diff
@@ -119,6 +126,24 @@ jobs:
...
+          key: cache/${{github.repository}}/${{github.ref}}
+          restore-keys: cache/${{github.repository}}/refs/heads/master
+          path: .test/**
+
+      - name: Initialize snapshots
+        if: ${{steps.cache.outputs.cache-hit != 'true' || inputs.update-snapshots == 'true'}}
+        run: npx playwright test --update-snapshots
+
+      - name: Run E2E tests for webpack version ${{ matrix.webpack-version }}
+        run: npm run test:e2e
```
snitin315, line 145: Remove the steps at line 121 and 125, also this should be:

```suggestion
      - name: Run E2E tests for webpack version ${{ matrix.webpack-version }}
        run: npm run test:e2e -- --shard=${{ matrix.shard }}
```

playwright.config.js
mahdikhashan: done

.github/workflows/nodejs.yml
mahdikhashan: I set a specific folder for tests and now its working.

package.json
```diff
@@ -41,6 +41,7 @@
     "test:coverage": "npm run test:only -- --coverage",
     "test:watch": "npm run test:coverage --watch",
     "test": "npm run test:coverage",
+    "test:e2e": "npx playwright test",
```
mxschmitt, line 44:
```suggestion
    "test:e2e": "playwright test",
```

playwright.config.js
```diff
@@ -0,0 +1,39 @@
+"use strict";
+
+const MAX_RETRIES = 3;
+const isCI = process.env.CI === "true";
+
+/** @type { import('@playwright/test').PlaywrightTestConfig} */
+module.exports = {
```
mxschmitt, line 7: Ideally you use [defineConfig](https://github.com/microsoft/create-playwright/blob/11c9c573970ae3c9e1abbf0dcb9d106d2a9729d8/assets/playwright.config.js#L13) + '// @ts-check'

playwright.config.js
```diff
@@ -0,0 +1,39 @@
...
+  globalSetup: require.resolve("./scripts/setupPlaywright.js"),
+  testIgnore: "**/*.ignore.*",
+  testDir: "./test/e2e",
+  snapshotPathTemplate: "./test/e2e/__snapshots__/{testFilePath}/{arg}{ext}",
+  fullyParallel: false,
```
mxschmitt, line 12: Ideally you enable fullyParallel - didn't it work with having it enabled? Might be a potential follow-up.

playwright.config.js
```diff
@@ -0,0 +1,39 @@
...
+    // fixes: net::ERR_HTTP2_PROTOCOL_ERROR
+    // https://github.com/webpack/webpack-dev-server/actions/runs/10043417455/job/27756147116#step:10:297
+    ignoreHTTPSErrors: true,
+  },
+  projects: [
+    {
+      name: "chromium",
+      use: {
+        browserName: "chromium",
+        launchOptions: {
```
mxschmitt, line 33: Since you set `ignoreHTTPSErrors` in line 26, you can remove the `launchOptions` lock.

test/helpers/playwright-test.js
```diff
@@ -0,0 +1,22 @@
...
+
+      await use(done);
+
+      return donePromise;
+    },
+    { option: true },
+  ],
+});
+
+module.exports = { test: mergeTests(customTest) };
```
mxschmitt, line 22: in theory no need to merge
```suggestion
module.exports = { test: customTest };
```

playwright.config.js
mahdikhashan: yes, it didn't work at first since we need to set ports, we planed to benefit from it later on.

PR #5162, opened 2024-04-23
fix(deps): Replace `rimraf` with `rm`, move `rimraf` to a dev dep

askoufis:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

No fixes or features

### Motivation / Use-Case

Now that this package's minimum supported node version is high enough, and given the relatively simple use case that `rimraf` is used for at runtime, IMO it's worth just using node's `rm` instead (via `graceful-fs`).

There is one more `rimraf` usage inside a test, but I figured that could stay. Happy to remove that too, though `rimraf` is still used as a CLI so it has to stay at least as a dev dep.

### Breaking Changes

None

### Additional Info

None


alexander-akait (approved):

alexander-akait:
Great job :+1:

askoufis:
Should I rename the commit so commitlint will pass?

alexander-akait:
Ignore it, I will rename when will merge

alexander-akait:
Looks like github actions is down - https://www.githubstatus.com/, I will rerun it when it will work

PR #5186, opened 2024-05-28
feat: add visual progress indicators

issacgerges:
- [ ] This is a **bugfix**
- [x] This is a **feature**

### Motivation / Use-Case

Adds built-in visual progress indicators to `webpack-dev-server`

### Breaking Changes

No breaking changes


issacgerges:
Should be fixed. @alexander-akait the re-run of `build:types` seems to have stripped all comments out. let me know if theres another way to run that command that preserves them and I can do.

alexander-akait (commented):
I am fine with it, but I am afraid some developers still can use our client in old browsers, so shadow dom can't work, let's add a check and don't run this code

issacgerges:
@alexander-akait will do, sorry for moving so slow on this. 

As an alternative, we could just expose some explicit API (or tell folks to lean on the `postMessage` you're already doing with progress) and leave implementing the visual parts to people who want them. What do you think?

alexander-akait:
@issacgerges Yeah, we can, but I am fine with your solution, just let's add additional check to avoid problems with old browsers

issacgerges:
Updated
* Added a guard
* Made the options "linear" and "circular" to be more consistent
* Combined both into a single custom element and simplified it a bit

client-src/index.js
```diff
@@ -236,6 +239,18 @@ const onSocketMessage = {
       );
     }
 
+    if (isProgressSupported()) {
```
alexander-akait, line 242: Let's move `defineProgressElement` after this check to avoid extra checks and have `customElements.define("wds-progress", WebpackDevServerProgress);` only in this function

alexander-akait (commented):
Looks very good, let's do one small improvement and we can merge

alexander-akait (dismissed):

alexander-akait:
Thank you, good work :star:

PR #5308, opened 2024-09-26
refactor: use methods to get client entries

LingyuCoder:
- [ ] This is a **bugfix**
- [ ] This is a **feature**

A test case was added in `test/e2e/client.test.js` to test that the override can take effect correctly.

### Motivation / Use-Case

Use methods to get client entries, so that users can introduce custom client entries instead by overriding these methods.

### Breaking Changes

No breaking changes

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -793,9 +793,7 @@ class Server {
         webSocketURLStr = searchParams.toString();
       }
 
-      additionalEntries.push(
-        `${this.getClientEntry()}?${webSocketURLStr}`,
-      );
+      additionalEntries.push(`${this.getClientEntry()}?${webSocketURLStr}`);
     }
 
     const clientHotEntry = this.getClientHotEntry();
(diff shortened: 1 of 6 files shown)


alexander-akait (approved):

lib/Server.js
```diff
@@ -1676,6 +1675,26 @@ class Server {
     return implementation;
   }
 
+  /**
+   * @private
```
alexander-akait, line 1679: Let's remove `@private`, I think it should be public

lib/Server.js
```diff
@@ -1676,6 +1675,26 @@ class Server {
...
+   * @returns {string}
+   */
+  getClientEntry() {
+    return require.resolve("../client/index.js");
+  }
+
+  /**
+   * @private
```
alexander-akait, line 1687: The same

alexander-akait:
Thank you

PR #5374, opened 2024-12-08
chore(deps): update chokidar to v4

### LINKED ISSUE #5368: Update chokidar to v4
### Bug report

This is not a bug report but I didn't know where to put it.
I would like to propose updating chokidar to v4. This would reduce the dependency count by 11 and also improves typing.
I am opening this issue as part of the ecosystem cleanup initiative (https://e18e.dev). We are trying to modernize the ecosystem by reducing dependency bloat, package size or improve speed.

If you are open for this, I would be happy to create a PR and do the necessary changes!

Fuzzyma:
This PR aims to close https://github.com/webpack/webpack-dev-server/issues/5368

In order to replace chokidars removed glob functionality, we have to use some trickery.

- `glob-parent` is used to get the common ancenstor dir that needs to be watched.
- `picomatch` is used as ignore-filter to make sure that only files/dirs are watched that match the given glob
- `is-glob` is used to only use glob logic when needed

In total this adds 4 direct or transitive dependencies. However, the update to chokidar v4 removes 12. So its a net positive

I also added a test to ensure that creation of nested directories is detected properly.

- [ ] This is a **bugfix**
- [ ] This is a **feature**

**NOTE:** This PR needs to wait for https://github.com/paulmillr/chokidar/issues/1394. If that issue is not resolved, we have to make sure to not let undefined values slip through for watcher options.

I didn't add this change to this PR because it would add more changed lined and it might be not needed after all.
That's why this PR is still draft


lib/Server.js
```diff
@@ -3254,9 +3254,36 @@ class Server {
    * @param {string | string[]} watchPath
    * @param {WatchOptions} [watchOptions]
    */
-  watchFiles(watchPath, watchOptions) {
+  watchFiles(watchPath, watchOptions = {}) {
     const chokidar = require("chokidar");
-    const watcher = chokidar.watch(watchPath, watchOptions);
+    const isGlob = require("is-glob");
```
alexander-akait, line 3259: What do you think about passing parameters here? Before we have them in `watchOptions` for globs

lib/Server.js
```diff
@@ -3254,9 +3254,36 @@ class Server {
...
+    // No need to do all this work when no globs are used
+    if (watchPathGlobs.length > 0) {
+      const globParent = require("glob-parent");
+      const picomatch = require("picomatch");
+
+      watchPathGlobs.forEach((p) => {
+        watchPathArr[watchPathArr.indexOf(p)] = globParent(p);
+      });
+
+      const matcher = picomatch(watchPathGlobs);
```
alexander-akait, line 3273: What do you think about passing parameters here? Before we have them in watchOptions for globs

Also why do not use `tinyglobby`?

lib/Server.js
alexander-akait, line 3259: Why do not use https://github.com/webpack/webpack-dev-server/issues/5368#issuecomment-2520984933 ?

lib/Server.js
Fuzzyma: Because tinyglobby gets passed a pattern and will get you back a file list matching that pattern. It doesn't have a simple match function that matches a path against a glob. But that's what I need here.

Tinyglobby simply doesn't work if you want to have a watch functionality. Chokidar already gives me back the filename. I only need to check if they match. Different use case -> different lib.

Tinyglobby also uses picomatch unter the hood. I wouldn't even save something if I would use tinyglobby instead (which I can't because it doesn't expose a matcher)

lib/Server.js
Fuzzyma: See my other comment.

is-glob is also part of parent-glob so it's not actually pulling in more dependencies because it's in the dependency tree already

lib/Server.js
alexander-akait: @Fuzzyma Got it

What about? https://github.com/micromatch/picomatch?tab=readme-ov-file#options

> What do you think about passing parameters here? Before we have them in watchOptions for globs

Some developers may well want to watch dot files for an example (ideally, we should initially have the same values ​​that we had as `chokidar`)

Fuzzyma: I have to look into that.
I couldn't completely figure out the possible things that could be passed here - so I forgot about it 😅.
Maybe you can help me out here by giving me the possible options taht can be passed?

alexander-akait: Before the globe was removed from chokidar, for example `dot` was part of `watchOptions`, now we have two options - `picomatchOptions` and `chokidarOptions`, I would suggest trying to pass` `watchOptions` to `picomatch` (by index if it is an array)

Fuzzyma: OK I finally got what you were referring to!
I will go over the picomatch options and see if I can make it behave like chokidar v3.

Fuzzyma: Ok, I saw that dotfiles were enabled by default in v3. Besides that, I passed the cwd option to picocolors as well. I am not 100% sure if that is correct behavior but it should. If all paths are relative to the cwd, it should work out. Maybe we need to add a test for that as well.

I also added glob support back for ignored files which I forgot about before

alexander-akait:
@Fuzzyma Can you rebase?

Fuzzyma:
@alexander-akait i am still waiting for the chokidar issue to be resolved. So this isn't passing tests atm anyway.

That said, surely can rebase :)

lib/Server.js
```diff
@@ -3255,9 +3255,70 @@ class Server {
...
+        const globParent = require("glob-parent");
+        const picomatch = require("picomatch");
+
+        watchPathGlobs.forEach((p) => {
+          watchPathArr[watchPathArr.indexOf(p)] = globParent(p);
+        });
+
+        const matcher = picomatch(watchPathGlobs, {
+          cwd: watchOptions.cwd,
+          dot: true,
```
alexander-akait, line 3278: `watchOptions.dot` (by default `false`)

It might even be better to pass all `watchOptions`

lib/Server.js
```diff
@@ -3255,9 +3255,70 @@ class Server {
...
+          if (
+            watchOptions.ignored &&
+            typeof watchOptions.ignored === "string" &&
+            isGlob(watchOptions.ignored)
+          ) {
+            watchOptions.ignored = picomatch(watchOptions.ignored, {
+              dot: true,
+              cwd: watchOptions.cwd,
+            });
+          }
```
alexander-akait, line 3312: Can we simplify it because there is a lot of duplicate logic here? We can always convert `watchOptions.ignored` to array (when it is provided)

Fuzzyma: agreed! will do

lib/Server.js
Fuzzyma: reading the code from chokidar v3 it seems this defaulted to `true`. It was also not an option you could pass to `watchOptions`.

See here a constant is passed to anymatch:

https://github.com/paulmillr/chokidar/blob/7c50e25d10a497ce4409f6e52eb630f0d7647b97/index.js#L214

And the constant is defined as `{ dot: true }`

https://github.com/paulmillr/chokidar/blob/7c50e25d10a497ce4409f6e52eb630f0d7647b97/lib/constants.js#L56

The readme doesnt mention `dot` as an option: 

https://github.com/paulmillr/chokidar/tree/3.6.0?tab=readme-ov-file#api

So I dont think we should pass any other options besides that one here.

Fuzzyma:
@alexander-akait I saw that when I run the tests a lot of them fail - unrelated to my changes. What is that about?

lib/Server.js
```diff
@@ -3255,9 +3255,65 @@ class Server {
...
+          const ignoreMatcher = picomatch(ignoredGlobs, {
+            dot: true,
+            cwd: watchOptions.cwd,
+          });
+          ignoredArr.push(ignoreMatcher);
+        }
+
+        ignoredArr.push(ignoreFunc);
+
+        watchOptions.ignored = ignoredArr;
```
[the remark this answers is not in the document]
alexander-akait: I would like to finish this soon and make a release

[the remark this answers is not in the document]
Fuzzyma: Will do!

As long as chokidar is not handling the issue linked in the first post, this is blocked anyway. As alternative we have to filter out undefined options ourselves before passing them to chokidar. Wdyt?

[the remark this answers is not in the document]
Fuzzyma: @alexander-akait I tried it your way. But there is one detail that might end up being performance sensitive.
I dont want to generate multiple matchers (one for each glob) if one is sufficient for all globs. Therefore I need to filter out the glob strings twice.

I tried to simplify the code further and think the code that I have now is easy enough to understand. I made it so that everything ends up in the ignoreArr.
Also fixed a bug while I was at it. 

We need tests for the ignore case though! I am not quite sure how to do the ignore case. I need to take a look.

Also I made the snapshots break because ignore is an array now when passied to chokidar. Need to fix that as well

Fuzzyma:
@alexander-akait so there are 2 remaining issues:

- chokidar doesnt allow undefined as options anymore and crashes (even tho the types allow it... I raised an issue but they dont really wanna change it?). So we have to clean the options before we pass them to chokidar _or_ wait for chokidar to come to the conclusion that its useful to follow their own types
- I dont know how to create tests for "ignored" files. How do you detect that nothing should happen? Maybe you can help me out here

alexander-akait:
@Fuzzyma 

> chokidar doesnt allow undefined as options anymore and crashes (even tho the types allow it... I raised an issue but they dont really wanna change it?). So we have to clean the options before we pass them to chokidar or wait for chokidar to come to the conclusion that its useful to follow their own types

I.e. our logic is broken somewhere? Because I don't think someone write `option: undefined`, maybe you can provide an example?

> I dont know how to create tests for "ignored" files. How do you detect that nothing should happen? Maybe you can help me out here

There is a test - https://github.com/webpack/webpack-dev-server/blob/master/test/e2e/watch-files.test.js#L160, just add a several cases where we have `ignored` from `watchFiles` - negative glob, nested glob (i.e. like `test/**/a/**`), just static value, and just when developer set directly.

In fact, this is one of the reasons why we moved away from `chokidar` inside webpack - it is very difficult to interact and make changes. Perhaps we should consider https://github.com/webpack/watchpack as `chokidar` replacer in future... we need to improve some things there.

Another solution that I am considering is the implementation of this function fundamentally in webpack (based on `watchpack`), so it will work not only for `serve`, but for `watch` too

Fuzzyma:
@alexander-akait no your logic is not broken. It is usually totally acceptable to pass undefined options to functions because typescripts optional type is `{ foo?: string | undefined  }`. Chokidar breaks this because they pass the options through to the native node api and that api throws if you pass undefined.

However, yes, we do create undefineds when calling getInterval and no interval or poll is defined:

https://github.com/webpack/webpack-dev-server/blob/55220a800ba4e30dbde2d98785ecf4c80b32f711/lib/Server.js#L899-L911

If they dont fix it, we have to fix it on our side.

For the tests: I already added tests there for the globs. But again: how do you test if something does _not_ happen because it is ignored?

Fuzzyma:
Yeah, mine: https://github.com/paulmillr/chokidar/pull/1396

Well since chokidar doesnt support globbing anymore, everything that touches globs, we have to test on our side.
Or we just test how specific globs are correctly resolved into correct ignore functions or matchers? Maybe easier?

alexander-akait:
> Or we just test how specific globs are correctly resolved into correct ignore functions or matchers? Maybe easier?

Yeah, maybe

Let's first try to simplify everything in the code, maybe we can even use unit testing here and one or two integrations to understand that everything really works

Fuzzyma:
Then I will pull out the creation of the ignore functions into own units. The we can simply test those.

Fuzzyma:
@alexander-akait I added tests for the glob matching and also added wrapper code to support chokidars weird behavior with undefined values. It can be removed once they make up their mind. This should work now and is ready for review

talentlessguy:
Are there any blockers to this pull request?

Fuzzyma:
Not from my side

Fuzzyma:
I maybe found an issue with negated glob patterns. I need to figure out if everything is alright before merge

alexander-akait:
@Fuzzyma Sorry for long delay (there were complicated personal things), what is current status? What we need to improve or wait? Will be great to landing it

Fuzzyma:
@alexander-akait the problem i encountered was, that I was filtering out potential nested directories because i couldn't partially match glob patterns. The new idea is, to wrap chokidar and emit only the events that match the passed glob pattern.
According to @43081j, this is also how it was done internally (if i remember correctly). However, it has the disadvantage that you watch directories and files you probably don't need to watch.

Because that is such a delicate topic, I created a package [chokidar-glob](https://github.com/Fuzzyma/chokidar-glob) that re-implements glob watching. It passes all old (v3) chokidar tests for globbing except for some windows tests which I cant debug on my machine. 
However, this is achieved by utilizing the approach above (=ignoring events and overwatching).
In order to improve on this situation I would need to implement partial glob matching again and I didn't come across an easy solution for this yet.

According to james, even with this approach it should perform good enough.

The situation is not ideal. I don't feel comfortable recommending chokidar-glob yet (at minimum I have to fix the windows tests). I am open to suggestions on how to approach this topic further.

alexander-akait:
@Fuzzyma I see your problem... A possible solution is to also abandon the use of globes in favor of `ignore`, using RegExp's or `picomatch` is not difficult, yes, it looks like a small regression in usability

Fuzzyma:
The problem with `ignore` in chokidar is, that it will ignore a directory and also all subdirectories. If you would rather go with watchpak, thats obviously a solution, too. However, how far away is a new major?

alexander-akait:
>  how far away is a new major?

We don't have roadmap for it current...

> If you would rather go with watchpak, thats obviously a solution, too.

I am if we don't find solution for globing I prefer to use `watchpach` (less deps), because it does the same

Fuzzyma:
@alexander-akait does it @alexander-akait as far as I can see watchpak also doesn't support globs for watched files. Only the ignore prop supports globs (at least that's what I grasped from their docs and tests)

43081j:
Just a couple of comments:

Chokidar does have support. I joined as a maintainer last year and have been very active maintaining it, responding to issues, etc

Shipping glob support in a FS watcher doesn't make much sense. It makes more sense to be able to combine an fs watcher with a glob library (also what fdir does). We did the right thing removing glob support from chokidar but it may make sense to have a higher level library soon that combines them with it

Chokidar exists to deal with the various inconsistencies between file systems and to add a higher level interface to FS watch. It doesn't exist to provide you globs 

If you need any help, here I am. 

It's possible the pr for undefined options will get merged at some point, but you should really solve that here too. It seems odd to be passing that

43081j:
I'm not saying we won't merge the change to chokidar to deal with undefined options, just that it is worth looking on your end too at why this is a thing. From what I understood, it wasn't the user passing undefined, it was webpack. But I may have misunderstood

Removing globs was a big change. But I do believed it is the right thing to do, since chokidar should focus on being a file watcher rather than a glob library. 

I'm not sure yet if we should have a library which joins the two or we should let you pass one in. One or the other should exist to make this easier in future though I think

For now, it really shouldn't be so difficult to bring your own. So I'd like to understand what the difficulties are a bit better. That will help me with the end solution too

Fuzzyma:
@43081j passing one in sounds interesting. But then again: maybe just give me more control. e.g. an event filter to effectively ignore events but not ignore watching. We talked about this at some point that chokidars kinda needs 2 types of ignore. Then we could easily just filter the events by a glob. Right now, I need a whole chokidar wrapper for that.

I think until this is figured out, a library that combines chokidar and glob makes sense. Also useful for people that don't want to wire this up themselves.

To the undefined problem: Yes, the one undefined came from something internally. But I think webpack-dev-server allows the user to pass arbitrary watcher options. So if they pass undefined, it will break even if the one occurrence in webpack-dev-server is fixed.

43081j:
> To the undefined problem: Yes, the one undefined came from something internally. But I think webpack-dev-server allows the user to pass arbitrary watcher options. So if they pass undefined, it will break even if the one occurrence in webpack-dev-server is fixed

I understand but one doesn't justify the other. We should still fix the code here to stop passing undefined. Even if a user can forcefully do it themselves

I'll have a deeper think about how best to solve this glob stuff and get back to you soon

Fuzzyma:
> I understand but one doesn't justify the other. We should still fix the code here to stop passing undefined. Even if a user can forcefully do it themselves

Oh yeah definitely!

kevinmpowell:
Just chiming in as someone following this thread. Our org would benefit from this being merged (it would clear some security vulnerability flags we're having to ignore currently).

I see it's still a "draft" - any timeline on when this might get merged?

alexander-akait:
@kevinmpowell we can't merge it because we will lost glob support in watch files (it is very popular usage) and there are not alternative solutions/API to make it works again

43081j:
I'll bump the work up my list to have a look at this on the chokidar side 👍 

as a lot of people are currently blocked in a similar position (needing chokidar + globs)

if we can at least document a solution to that, or publish something to help with it, we can unblock everyone

I'll catch up with you on this soon @Fuzzyma and whoever else may have taken a look

alexander-akait:
@43081j thanks you, resolving it will be great and unblock updating for us too, I am fine if we will extra code or extra package(s) to get support of globs, but without such complex code with limitations as we try to resolve here

bjohansebas:
fixed by  #5657

PR #5434, opened 2025-03-16
feat: publicPath output

adubrouski:
- [ ] This is a **bugfix**
- [x] This is a **feature**

Yes

### Motivation / Use-Case

Some applications use public paths (to avoid strong differences between dev and prod builds). I set it via output.publicPath, and it seems a bit inconvenient to me that the urls in the terminal are obtained without taking publicPath into account. These can be simple applications without routing, and you have to go to / and then manually add publicPath. I suggest outputting the url with publicPath if it is set statically, if it is determined on the fly - output without it.

Now, with config like this:

```javascript
output: {
  publicPath: "/my-public-path/",
}
```

We getting this:

```
<i> [webpack-dev-server] Loopback: http://localhost:8080/, http://[::1]:8080/
<i> [webpack-dev-server] On Your Network (IPv4): http://10.252.1.16:8080/
<i> [webpack-dev-server] Content not from webpack is served from '/home/user/Project/src' directory
<i> [webpack-dev-server] 404s will fallback to '/index.html'
```

My proposal is:

```
<i> [webpack-dev-server] Loopback: http://localhost:8080/my-public-path/, http://[::1]:8080/my-public-path/
<i> [webpack-dev-server] On Your Network (IPv4): http://10.252.1.16:8080/my-public-path/
<i> [webpack-dev-server] Content not from webpack is served from '/home/user/Project/src' directory
<i> [webpack-dev-server] 404s will fallback to '/index.html'
```

### Breaking Changes

No

### Additional Info

No


lib/Server.js
```diff
@@ -2883,8 +2886,19 @@ class Server {
...
-        url.format({ protocol, hostname: newHostname, port, pathname: "/" });
+      const prettyPrintURL = (newHostname) => {
+        const publicPath = compilerOptions.output.publicPath;
+
+        return url.format({
+          protocol,
+          hostname: newHostname,
+          port,
+          pathname:
+            typeof publicPath === "function" || publicPath === "auto"
```
alexander-akait, line 2897: We need respect `function` too... But yeah, we known about it only after compilation, so maybe we can improve our message and write - something like `you can have custom public path, please navigate in your browsers  to the point that you required` (just an example, feel free to improve it)

adubrouski: @alexander-akait I think adding an info message if public path is `auto` or `function` would be a good idea, can you check this [commit](https://github.com/webpack/webpack-dev-server/pull/5434/commits/89eb799e49eeb8b218e446eca7ad87cbaa858515)?

lib/Server.js
```diff
@@ -3031,6 +3045,17 @@ class Server {
...
+    const publicPath = compilerOptions.output.publicPath;
+
+    const isOnTheFlyDefinedPublicPath =
+      typeof publicPath === "function" || publicPath === "auto";
+
+    if (isOnTheFlyDefinedPublicPath) {
+      this.logger.info(
+        `You probably have a custom public path, please navigate in your browser to the point that you required`,
+      );
+    }
```
alexander-akait, line 3058: We should output this only for `function`, `auto` is like `/`, so it is unnecessary

adubrouski: @alexander-akait 

Hmm, webpack documentation says that the "auto" value allows you to automatically infer the public path value based on the values ​​in the scripts.  Are you sure that we should keep default behavior for "auto" value?

https://webpack.js.org/guides/public-path/#automatic-publicpath

alexander-akait: Yes, you are right, but mostly it used only for `/`, so I think it can be a little bit spammy

adubrouski: Ok, it fixed in this [commit](https://github.com/webpack/webpack-dev-server/pull/5434/commits/2e68d84c567b72145aed4d1cb275486d2a6418f95). Can you check?

alexander-akait (commented):
Looks good, I will look deeply before the next release, thank you

kmk142789 (approved):

PR #5564, opened 2025-08-14
feat: bump to express 5

bjohansebas:
- [ ] This is a **bugfix**
- [x] This is a **feature**

### Additional Info

While with Express there are hardly any changes other than adding support for some encodings or changing the regex path match, so nothing in the Express code will change for now. However, it is necessary to update `http-proxy-middleware`, since in version 2 they have a resolution to use Express 4, whereas version 3 of that package already allows a different version of Express and other packages. :)


bjohansebas:
I still need to investigate why some content-types are not being changed, since JavaScript files are now "text/javascript", but somewhere that content-type is still not being updated, and also run the allowed-host tests.

bjohansebas:
Okay, after #5626 , I’ll debug this since there will be fewer tests to run.

bjohansebas:
Okay, the proxy-related tests are failing, so I’m going to look into the reason

edit: I found a small issue in the loggers, since they changed the system, so I’m working on updating it

bjohansebas:
Even though it does respect level logger, it still calls the function, so I’ll need to use snapshots instead of checking whether the function was called like the current tests do. This happens when the logger is set to none, it doesn’t show the message, probably due to webpack’s internal logic, but it still calls it.

alexander-akait:
Sound good to me

bjohansebas:
Wow, the CI passed on the first try, first time I’ve seen that happen

bjohansebas:
Okay, I’ll continue with removing spdy, I’m curious how it will behave.

PR #5627, opened 2026-01-12
fix: improved hostname validation logic in Server.js

VeerRaj-007:
This PR fixes an issue in isValidHost where setting the devServer.host option could unintentionally block otherwise valid hostnames.

Currently, when validateHost is true, the code combines IP checks, localhost checks, and this.options.host into a single expression. This makes the logic harder to reason about and causes unexpected rejections when a custom host is configured.

This change makes the behavior explicit and correct:

- Always allow:

  -   IPv4 and IPv6 addresses

  -   localhost and *.localhost

- Only allow this.options.host when validateHost is enabled

This preserves the existing security model while fixing cases where valid requests were rejected when using a custom host.

### Why this is needed

When users configure devServer.host, they expect it to be accepted consistently. The previous logic could cause valid requests to be rejected depending on validateHost, which is confusing and not intended behavior.

This change aligns the implementation with the documented behavior and avoids unexpected host blocking.

Fixes #5603

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -3235,15 +3235,18 @@ class Server {
     // For convenience, always allow localhost (hostname === 'localhost')
     // and its subdomains (hostname.endsWith(".localhost")).
     // allow hostname of listening address  (hostname === this.options.host)
-    const isValidHostname = validateHost
-      ? ipaddr.IPv4.isValid(hostname) ||
-        ipaddr.IPv6.isValid(hostname) ||
-        hostname === "localhost" ||
-        hostname.endsWith(".localhost") ||
-        hostname === this.options.host
-      : false;
-
-    return isValidHostname;
+    if (
+      ipaddr.IPv4.isValid(hostname) ||
+      ipaddr.IPv6.isValid(hostname) ||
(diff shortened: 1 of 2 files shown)


alexander-akait:
Sorry, I don't understand what you fixed here

VeerRaj-007:
The bug happens only for no-cors cross-site requests.
In that code path isValidHost() is called with validateHost = false, which causes this line to always return false:

`const isValidHostname = validateHost ? (...) : false;`


That means even localhost and 127.0.0.1 are rejected and a 403 is returned, even though the comment above says they should always be allowed.

My change makes the behavior match the comment:

- IPv4, IPv6, localhost, and *.localhost are always allowed

- this.options.host is only allowed when validateHost === true

So no-cors requests from localhost now work without requiring allowedHosts: ['localhost'], restoring the old behavior and fixing #5603.

alexander-akait:
Please add a test case

VeerRaj-007:
Hi @alexander-akait, I have added the test as requested and committed it to the repository. Please let me know if any further changes are needed

PR #5634, opened 2026-01-23
feat: remove cli-flags

bjohansebas:
**Summary**

removing these options that are deprecate


alexander-akait (commented):
Can you add this change (and replacement for these options) in migration guide to ensure we don't remove something more than expected

migration-v6.md
```diff
@@ -22,3 +51,24 @@ This document serves as a migration guide for `webpack-dev-server@6.0.0`.
   ```js
   const ip = Server.findIp("v4", true);
   ```
+
+- The following CLI options were deprecated and have now been removed. Please use the recommended alternatives:
+  - http2 -> server
+  - https -> server
```
alexander-akait: Will just `server` work? Maybe `server-type`? Just to ensure we don't make a typo here 😄

alexander-akait: oh, do you provide here not CLI options, let's write CLI options here

alexander-akait: Just write `webpack help serve --verbose` and you will see all CLI options

bjohansebas: Thanks, I was looking for a way to reproduce and see them

bjohansebas: @alexander-akait  I’m not fully understanding this. At what point is the `cli-flags.js` file executed within `webpack-cli`?

I’m asking because many of these options were removed in `webpack-dev-server` v5 — for example in this commit: https://github.com/webpack/webpack-dev-server/commit/1e86e3703627155cabca66fd969111d48f7653b0
And this is the commit that updated the README accordingly https://github.com/webpack/webpack-dev-server/commit/5c4006ac63280873bc4a7f86340edb5b0f18ca3e
So it feels like those options no longer exist starting from `webpack-dev-server` v5, and this file was simply never updated — or maybe I’m missing something?
In any case, even when running from `main`, those options still don’t show up in the CLI. From what I understand, the file that actually defines the CLI options is `options.json`, which would make `cli-flags.js` unnecessary.

alexander-akait: > I’m not fully understanding this. At what point is the cli-flags.js file executed within webpack-cli?

Yes, webpack-cli reads this file and getting all CLI flag from this file. Looks like we forgot to remove them. To avoid such things let's add a test case in CLI with `--help` and `--verbose` so we will have all CLI arguments in our tests

alexander-akait: Also let's use here the new CLI options = `--server-type=https` and etc

alexander-akait (approved):

alexander-akait:
Let's fix lint and feel free to merge

bjohansebas:
The linter issue regarding the CVE in the webpack subdependency

PR #5636, opened 2026-01-25
fix: always allow localhost in cross-origin no-cors requests

Jeston10:
Implement the intended behavior to always allow localhost host for development convenience. The isHostAllowed method had a comment stating localhost should always be allowed, but the code did not implement this check.

This fixes the issue where no-cors cross-site requests to localhost were incorrectly blocked with 403 Forbidden. Now localhost is always allowed regardless of allowedHosts configuration, matching the intended behavior for development environments.

Fixes the cross-origin-header-check middleware blocking legitimate localhost requests from blob URLs and other cross-origin contexts.

Fixes #5603

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -3145,6 +3145,10 @@ class Server {
     }
 
     // always allow localhost host, for convenience
+    if (value === "localhost" || value.endsWith(".localhost")) {
+      return true;
+    }
+
     // allow if value is in allowedHosts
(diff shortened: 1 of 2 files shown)


alexander-akait (changes requested):
That is wrong solution, attack can be done throughout localhost

Jeston10:
> That is wrong solution, attack can be done throughout localhost
Yes Sir, I'll find out a way to prevent attacks from localhost. Thank you sir for the review.

PR #5637, opened 2026-01-26
feat: use new URL instead of url.parse

bjohansebas:
**Summary**

An important change, IPv6 hostnames now return with square brackets (`[]`) (for new URL()), so they need to be formatted for `ipaddr` to correctly validate whether it’s an IPv6 URL. Additionally, the hostname also needs to be formatted in the IPv6 case, because right now, when an IPv6 hostname is passed in `host`, users are not including the square brackets. We could change this behavior, what do you think?

Also, it seems this will make retrieving the hostname faster.

imagen

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -3059,12 +3059,17 @@ class Server {
     if (!header) return null;
     try {
       // If the header does not have a scheme, prepend // so URL can parse it
-      const url = new URL(/^(.+:)?\/\//.test(header) ? header : `//${header}`);
-      let hostname = url.hostname;
+      const parseUrl = new URL(
+        /^(.+:)?\/\//.test(header) ? header : `//${header}`,
+        "http://localhost/",
+      );
+
+      let hostname = parseUrl.hostname;
       // Normalize IPv6: remove brackets if present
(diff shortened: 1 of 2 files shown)


alexander-akait (commented):
I am fine with this change, feel free to merge when CI will be green, also we have `url.parse` in our webpack-dev-middleware, make sense to improve this there too for the next major release

bjohansebas:
note for myself: always remember to add a base in case the URL doesn’t come complete

imagen

lib/Server.js
```diff
@@ -3050,6 +3050,32 @@ class Server {
...
+   * @param {string} header header value
+   * @returns {string|null} hostname or null
+   */
+  #parseHostnameFromHeader = function (header) {
+    if (!header) return null;
+    try {
+      // If the header does not have a scheme, prepend // so URL can parse it
+      const parseUrl = new URL(
+        /^(.+:)?\/\//.test(header) ? header : `//${header}`,
+        "http://localhost/",
```
bjohansebas, line 3064: This is only for a fallback.

alexander-akait (approved):

alexander-akait:
Thanks

PR #5641, opened 2026-01-31
feat: remove bypass option from proxy configuration

bjohansebas:
**Summary**

This removes the option to bypass the proxy.

### DIFF
--- a/test/server/proxy-option.test.js
+++ b/test/server/proxy-option.test.js
@@ -18,38 +18,14 @@ const proxyOptionPathsAsProperties = [
     target: `http://localhost:${port1}`,
   },
   {
-    context: "/api/proxy2",
+    path: "/api/proxy2",
     target: `http://localhost:${port2}`,
     pathRewrite: { "^/api": "" },
   },
   {
-    context: "/foo",
-    bypass(req) {
-      if (/\.html$/.test(req.path || req.url)) {
-        return "/index.html";
-      }
-
-      return null;
-    },
-  },
-  {
-    context: "proxyfalse",
-    bypass(req) {
-      if (/\/proxyfalse$/.test(req.path || req.url)) {
-        return false;
-      }
-    },
-  },
-  {
-    context: "/bypass-with-target",
+    pathFilter: "**/*.html",
     target: `http://localhost:${port1}`,
-    changeOrigin: true,
-    secure: false,
-    bypass(req) {
-      if (/\.(html)$/i.test(req.path || req.url)) {
-        return req.url;
-      }
-    },
+    pathRewrite: () => "/index.html",
   },
 ];
 
@@ -139,6 +115,9 @@ describe("proxy option", () => {
     });
     proxyApp1.get("/api", (req, res) => {
       res.send("api response from proxy1");
+    });
+    proxyApp1.get("/index.html", (req, res) => {
+      res.send("Hello");
     });
     proxyApp2.get("/proxy2", (req, res) => {
       res.send("from proxy2");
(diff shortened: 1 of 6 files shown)


test/server/proxy-option.test.js
```diff
@@ -23,48 +22,6 @@ const proxyOptionPathsAsProperties = [
...
-    context: "/bypass-with-target",
-    target: `http://localhost:${port1}`,
-    changeOrigin: true,
-    secure: false,
-    bypass(req) {
-      if (/\.(html)$/i.test(req.path || req.url)) {
-        return req.url;
-      }
-    },
-  },
```
alexander-akait, line 67: Can we rewrite these test to using new options from `http-proxy-middleware`

Also let's add in migration guide how to achive the same behaviour with new http proxy middleware

alexander-akait (commented):
Let's improve couple things and we can merge

bjohansebas:
Okay, I’ve removed those two tests because neither router nor pathRewrite allows sending responses. The ServerResponse object is not available there, unlike in bypass. In router, only the req object and the options passed as arguments can be accessed, and the same applies to pathRewrite.

I’m working on adapting the other tests, because the behavior of bypass is completely different from what router and pathRewrite do.

Hmm, I think it’s not possible to recreate those tests. The reason is that with bypass, the request never went through http-proxy-middleware, so in the first tests it was able to redirect to /index.html. With proxies, that route no longer exists, which means we would likely need to create a new route in the proxies.

test/server/proxy-option.test.js
```diff
@@ -19,51 +18,14 @@ const proxyOptionPathsAsProperties = [
...
-            res.end("proxy async response");
-            resolve(true);
-          }, 10);
-        });
-      }
-    },
-  },
-  {
-    context: "/bypass-with-target",
+    pathFilter: "**/*.html",
```
bjohansebas: I would expect that using router: "() => `http://localhost:${port3}`" should work and be somewhat similar to what `bypass` did, but it doesn’t fully work. I’m still thinking about a way to recreate `bypass`.

With `bypass`, the request never went through the proxy, which is why it could return `index.html` from the webpack-dev-server itself. Now we need to find a way for the request to go through the proxy and still reach the original server, without having to create a new route in the proxy.

test/server/proxy-option.test.js
```diff
@@ -295,13 +227,6 @@ describe("proxy option", () => {
         expect(response.status).toBe(404);
       });
 
-      it("should wait if bypass returns promise", async () => {
-        const response = await req.get("/proxy/async");
-
-        expect(response.status).toBe(200);
-        expect(response.text).toContain("proxy async response");
-      });
-
```
bjohansebas, line 304: Neither `router` nor `pathRewrite` can send data directly since they don’t have access to the `res` object. So either it gets removed, or a new route is created in the proxy — but I’m not sure what the point of that would be.

test/server/proxy-option.test.js
```diff
@@ -498,13 +423,6 @@ describe("proxy option", () => {
       expect(response.text).toContain("from proxy2");
     });
 
-    it("should allow req, res, and next", async () => {
```
bjohansebas, line 501: same https://github.com/webpack/webpack-dev-server/pull/5641/changes#r2752370675

alexander-akait (approved):
Feel free to merge

alexander-akait (approved):

test/server/proxy-option.test.js
chimurai: Hi @bjohansebas 

`res` and `options` are accessible in `router` and `pathRewrite` (Since 4.1.0)

https://github.com/chimurai/http-proxy-middleware/pull/1247
https://github.com/chimurai/http-proxy-middleware/pull/1244

Release: https://github.com/chimurai/http-proxy-middleware/releases/tag/v4.1.0

bjohansebas: hi @chimurai see #5689

PR #5647, opened 2026-02-16
feat: use compiler.platform to determine the target

bjohansebas:
**Summary**

The use of `compiler.platform` is available starting from webpack 5.92.0 ([https://github.com/webpack/webpack/commit/e11fb128a24222e2fce1ad10e55b140c2a64a2ff](https://github.com/webpack/webpack/commit/e11fb128a24222e2fce1ad10e55b140c2a64a2ff)).

The logic that added WebSocket communication when a `conditionNames` entry included `browser` has been removed. However, I’m not sure whether we also want to remove the behavior related to `externalsPresets`. Do we want to remove that as well?

If you want to know where it determines whether the web is available for a specific target, take a look here (https://github.com/webpack/webpack/blob/c835794b7fcd9f1e2f6750ae94c95aadf23a63dd/lib/config/target.js#L101)

refactor, and feat

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -560,42 +560,9 @@ class Server {
    * @returns {boolean} true when target is `web`, otherwise false
    */
   static isWebTarget(compiler) {
-    if (compiler.platform && compiler.platform.web) {
-      return compiler.platform.web;
-    }
-
-    // TODO improve for the next major version and keep only `webTargets` to fallback for old versions
-    if (
-      compiler.options.externalsPresets &&
-      compiler.options.externalsPresets.web
-    ) {
-      return true;
-    }
-
-    if (
-      compiler.options.resolve.conditionNames &&
-      compiler.options.resolve.conditionNames.includes("browser")
-    ) {
-      return true;
-    }
-
-    const webTargets = [
-      "web",
-      "webworker",
-      "electron-preload",
-      "electron-renderer",
-      "nwjs",
-      "node-webkit",
-
-      undefined,
-      null,
-    ];
-
-    if (Array.isArray(compiler.options.target)) {
-      return compiler.options.target.some((r) => webTargets.includes(r));
-    }
-
-    return webTargets.includes(/** @type {string} */ (compiler.options.target));
+    return (
+      compiler.platform?.web || compiler.options.externalsPresets?.web || false
+    );
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lib/Server.js
```diff
@@ -560,42 +560,9 @@ class Server {
...
-      null,
-    ];
-
-    if (Array.isArray(compiler.options.target)) {
-      return compiler.options.target.some((r) => webTargets.includes(r));
-    }
-
-    return webTargets.includes(/** @type {string} */ (compiler.options.target));
+    return (
+      compiler.platform?.web || compiler.options.externalsPresets?.web || false
```
alexander-akait: I think we can up minimum webpack version and don't use `compiler.options.externalsPresets?.web` all all, also we always have `compiler.platform`

bjohansebas: That sounds good to me. We’ve already bumped the minimum version to webpack 5.101, I just wasn’t sure if you also wanted to remove the externalPresets part.

alexander-akait: Yeah, we don't need `externalPresets` anymore here, only `platform`, it was literally design for such cases

lib/Server.js
```diff
@@ -560,42 +560,7 @@ class Server {
...
-      undefined,
-      null,
-    ];
-
-    if (Array.isArray(compiler.options.target)) {
-      return compiler.options.target.some((r) => webTargets.includes(r));
-    }
-
-    return webTargets.includes(/** @type {string} */ (compiler.options.target));
+    return compiler.platform?.web || false;
```
alexander-akait: Let's remove `?` here, you got it because webpack compiler can be `Compiler` or `MultiCompiler`, so you need

```
if (Array.isArray(compiler)) {
  return compiler.some((r) => compiler.platform.web);
}

return compiler.platform.web;
```

bjohansebas: In fact, the "?" is not necessary. And when this function is called, it will be a compiler, not a multi-compiler (https://github.com/webpack/webpack-dev-server/blob/a4012b6b9a27a36f20dd632306aeacc5ec8a5281/lib/Server.js#L605, https://github.com/webpack/webpack-dev-server/blob/a4012b6b9a27a36f20dd632306aeacc5ec8a5281/lib/Server.js#L1698 are already compilers and not multi-compilers). It could be considered a new feature if that’s what you want.

lib/Server.js
```diff
@@ -560,42 +560,7 @@ class Server {
...
-      undefined,
-      null,
-    ];
-
-    if (Array.isArray(compiler.options.target)) {
-      return compiler.options.target.some((r) => webTargets.includes(r));
-    }
-
-    return webTargets.includes(/** @type {string} */ (compiler.options.target));
+    return compiler.platform.web || false;
```
alexander-akait, line 563: Oh, forgot, we need to improve support of universal target too, i.e. `target: "node" | "web"`, i..e when `web === null && node === null`, but in our client code we should not run code when it is node environment and run when it is web , but let's do it in other PR

alexander-akait (approved):

alexander-akait:
Feel free to merge

PR #5650, opened 2026-02-21
feat: plugin support

bjohansebas:
**Summary**

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -2362,6 +2362,7 @@ class Server {
           // @ts-expect-error
           this.compiler,
           this.options.devMiddleware,
+          true,
         );
       }
 
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -362,6 +362,11 @@ class Server {
      */
 
     this.currentHash = undefined;
+    /**
+     * @private
+     * @type {boolean}
+     */
+    this.isPlugin = false;
   }
 
   static get schema() {
@@ -2362,7 +2367,7 @@ class Server {
           // @ts-expect-error
           this.compiler,
           this.options.devMiddleware,
-          true,
+          this.isPlugin,
         );
       }
 
@@ -3436,6 +3441,7 @@ class Server {
    */
   apply(compiler) {
     this.compiler = compiler;
+    this.isPlugin = true;
     this.logger = this.compiler.getInfrastructureLogger(pluginName);
 
     this.compiler.hooks.watchRun.tapPromise(pluginName, async () => {
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -328,11 +328,14 @@ class Server {
       baseDataPath: "options",
     });
 
-    this.compiler = compiler;
-    /**
-     * @type {ReturnType<Compiler["getInfrastructureLogger"]>}
-     */
-    this.logger = this.compiler.getInfrastructureLogger(pluginName);
+    if (compiler) {
+      this.compiler = compiler;
+
+      /**
+       * @type {ReturnType<Compiler["getInfrastructureLogger"]>}
+       */
+      this.logger = this.compiler.getInfrastructureLogger(pluginName);
+    }
     this.options = options;
     /**
      * @type {FSWatcher[]}
@@ -1589,6 +1592,9 @@ class Server {
    * @returns {void}
    */
   setupProgressPlugin() {
+    // In the case where there is no compiler and it’s not being used as a plugin.
+    if (this.compiler === undefined) return;
+
     const { ProgressPlugin } =
       /** @type {MultiCompiler} */
       (this.compiler).compilers
@@ -1633,6 +1639,7 @@ class Server {
    * @returns {Promise<void>}
    */
   async initialize() {
+    if (this.compiler === undefined) return;
     this.setupHooks();
 
     await this.setupApp();
@@ -1708,7 +1715,7 @@ class Server {
           needForceShutdown = true;
 
           this.stopCallback(() => {
-            if (typeof this.compiler.close === "function") {
+            if (typeof this.compiler?.close === "function") {
               this.compiler.close(() => {
                 // eslint-disable-next-line n/no-process-exit
                 process.exit();
@@ -1783,11 +1790,14 @@ class Server {
    * @returns {void}
    */
   setupHooks() {
+    if (this.compiler === undefined) return;
+
     this.compiler.hooks.invalid.tap("webpack-dev-server", () => {
       if (this.webSocketServer) {
         this.sendMessage(this.webSocketServer.clients, "invalid");
       }
     });
+
     this.compiler.hooks.done.tap(
       "webpack-dev-server",
       /**
@@ -1842,6 +1852,7 @@ class Server {
    * @returns {void}
    */
   setupMiddlewares() {
+    if (this.compiler === undefined) return;
     /**
      * @type {Array<Middleware>}
      */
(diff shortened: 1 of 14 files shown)


alexander-akait:
We should start it in webpack-dev-middleware, currently `webpack-dev-middleware` uses `console.log` to output stats, it means we will spam logs, we need an options or another function wrapper - `require("webpack-dev-middleware").plugin(options)`

lib/Server.js
```diff
@@ -1587,6 +1597,9 @@ class Server {
    * @returns {void}
    */
   setupProgressPlugin() {
+    // In the case where there is no compiler and it’s not being used as a plugin.
+    if (this.compiler === undefined) return;
```
alexander-akait, line 1601: Ideally we should always have `compiler` here, so we can use `/* {Compiler} */ (this.compiler)` to say typescript we have it is

lib/Server.js
```diff
@@ -3414,6 +3434,30 @@ class Server {
...
+   * @returns {void}
+   */
+  apply(compiler) {
+    this.compiler = compiler;
+    this.isPlugin = true;
+    this.logger = this.compiler.getInfrastructureLogger(pluginName);
+
+    let started = false;
+
+    this.compiler.hooks.watchRun.tapPromise(pluginName, async () => {
```
alexander-akait, line 3449: I think we should use another hook here, we should:
1. Start dev server after the first compilation happend
2. We should use hook when `CleanPlugin` will remove old files and webpack will not write new files, why? Because we are using memory fs here, old assets should be removed and new assets should be written to memory

bjohansebas: I’m going to use the afterEmit hook

alexander-akait: @bjohansebas let's also add a test case with `output.clean: true` to ensure we work as expected

test/e2e/api.test.js
```diff
@@ -4,11 +4,94 @@ const path = require("node:path");
...
+    expect(consoleMessages.map((message) => message.text())).toMatchSnapshot(
+      "console messages",
+    );
+    expect(pageErrors).toMatchSnapshot("page errors");
+
+    await browser.close();
+    await new Promise((resolve) => {
+      compiler.close(resolve);
+    });
+  });
```
alexander-akait, line 47: Let's add full plugin tests, where dev server plugin is a part of `plugins` in webpack.config.js, also let's add a test case for `clean: true`, these tests not fully test our plugin logic, also in future we will need to rewrite all our tests to the plugin support (not in this pr, but for future), so let's start is here (and if we need required helpers)

bjohansebas (commented):
These changes should work without requiring any changes to webpack-cli yet. Once this is merged, we can work on a patch for webpack-cli to use the plugin support, but it is independent of changes to webpack-cli.

bjohansebas:
cc: @alexander-akait friendly ping for feedback

alexander-akait:
Will look soon, I little bit busy, we should make the new major release the next week, I will help you to finish remains tasks

alexander-akait:
@bjohansebas let's rebase

alexander-akait:
@bjohansebas can take a look at our failed tests, a lot of PRs without green, we need to resolve it, thanks

bjohansebas:
> @bjohansebas can take a look at our failed tests, a lot of PRs without green, we need to resolve it, thanks

Those are just flaky tests, like before. You just need to rerun them and that's it.

alexander-akait:
Looks like problem with lint here

lib/Server.js
```diff
@@ -3542,6 +3578,76 @@ class Server {
...
+    const seenFirstDone = new WeakSet();
+    let firstDoneCount = 0;
+
+    // A one-shot `compiler.run()` (plain `webpack` build) is detected when no
+    // child compiler is in watch mode. In that case we skip both `setup()` and
+    // `listen()` so the build can finish and the process can exit normally —
+    // the user is not in control of the plugin lifecycle here, so we stay
+    // silent rather than logging a warning.
+    const isBuildMode = () =>
+      childCompilers.every((child) => !child.watching && !child.options.watch);
```
alexander-akait: Each compiler have watching when it is watching, no need to check on watch, at least I can see situation when watch will be true, but watching undefined

lib/Server.js
```diff
@@ -3542,6 +3578,76 @@ class Server {
...
+    /**
+     * @param {Compiler} childCompiler child compiler
+     * @returns {Promise<void>} promise
+     */
+    const onChildDone = async (childCompiler) => {
+      if (listening || isBuildMode()) return;
+      if (seenFirstDone.has(childCompiler)) return;
+      seenFirstDone.add(childCompiler);
+      firstDoneCount++;
+      if (firstDoneCount < childCompilers.length) return;
```
alexander-akait: Can you clarify when we need firstDoneCount?

lib/Server.js
```diff
@@ -3542,6 +3578,76 @@ class Server {
...
+      listening = false;
+      await this.stop();
+    };
+
+    for (const childCompiler of childCompilers) {
+      childCompiler.hooks.beforeCompile.tapPromise(pluginName, ensureSetup);
+      childCompiler.hooks.done.tapPromise(pluginName, () =>
+        onChildDone(childCompiler),
+      );
+      childCompiler.hooks.shutdown.tapPromise(pluginName, onChildShutdown);
```
alexander-akait: Ideally we should have the same hooks for multi compiler, and so don't use `for/of`, but we can postpone it and improve later

lib/Server.js
```diff
@@ -3542,6 +3578,76 @@ class Server {
...
+    const onChildShutdown = async () => {
+      if (stopped) return;
+      stopped = true;
+      setupPromise = undefined;
+      listening = false;
+      await this.stop();
+    };
+
+    for (const childCompiler of childCompilers) {
+      childCompiler.hooks.beforeCompile.tapPromise(pluginName, ensureSetup);
```
alexander-akait: Can we avoid `beforeCompile` at all, it is unsafe in some cases, ideally we should use only `done` and `shutdown`

alexander-akait: I think we can done everything on `done`, also please add a test case what we really change outputFileSystem and use it, not webpack built-in the real fs

bjohansebas:
okay tomorrow i will continue with this

lib/Server.js
bjohansebas: Doing it only through `done` is complicated because there are processes that need to be initialized much earlier. That's why `beforeCompile` is used.

bjohansebas:
### `watchRun` vs `beforeCompile`

`setup()` must run before the first compilation writes assets, because `webpack-dev-middleware` replaces `outputFileSystem` with an in-memory filesystem. Running only on `done` is too late—the first build has already written to disk.

Both `watchRun` and `beforeCompile` run before `emit`, but `watchRun` is a better fit because it is watch-only (no build-mode guard required) and, on a `MultiCompiler`, a single top-level tap covers all child compilers.

### Why not do everything on `done`?

`done` fires after `emit`, so it's too late for setup that needs to affect the compilation output. The only thing that belongs on `done` is `listen()`, since the HTTP server should start after the first successful build.

### Why use `shutdown`?

Teardown needs to work when `compiler.close()` is called. On a `MultiCompiler`, the aggregate `watchClose` does not fire for `compiler.close()`, while child `shutdown` hooks do.

`shutdown` is also async, allowing `compiler.close()` to wait for the server to fully stop and release its port before returning.

Issue #5024, opened 2023-12-17
Overlay by thrown error at initial load is closed immediately


woochanleee:
### Actual Behavior

Overlay by thrown error at initial load is closed immediately.

https://github.com/webpack/webpack-dev-server/assets/48552260/5f8184b4-60d5-4fc3-ba2c-c1cbb715cada

#### With debugger mode

https://github.com/webpack/webpack-dev-server/assets/48552260/deda2386-87e2-4965-a6ef-a5a60134b527

### Expected Behavior

Overlay by thrown error at initial load is **not** closed immediately.

### How Do We Reproduce?

Just edit like below the photo to the overlay-example code. https://github.com/webpack/webpack-dev-server/blob/v4.15.1/examples/client/overlay/app.js#L49

<img width="1231" alt="image" src="https://github.com/webpack/webpack-dev-server/assets/48552260/d2be81c7-dfc4-45c4-ab8f-f81a9f180072">

### Please paste the results of `npx webpack-cli info` here, and mention other relevant information

```
System:
  OS: macOS 13.4.1
  CPU: (10) arm64 Apple M1 Max
  Memory: 89.55 MB / 32.00 GB
Binaries:
  Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
  Yarn: 3.6.4 - ~/.nvm/versions/node/v16.14.0/bin/yarn
  npm: 8.3.1 - ~/.nvm/versions/node/v16.14.0/bin/npm
Browsers:
  Chrome: 120.0.6099.109
  Safari: 16.5.2
Global Packages:
  webpack-dev-server: 4.15.1
```

snitin315:
I'll check this

avinashii:
Hi Nitin,

I was thinking to contribute to this issue , I am new to it.

is this a good issue to try my hands on , or its just opened since long.

Anubhab2003:
Hey I am Interested in contributing this issue, How to get started ?

geanfarias:
@woochanleee did you solved this at your project?

woochanleee:
> [@woochanleee](https://github.com/woochanleee) did you solved this at your project?

No, i didn't. 😢

snitin315:
Will check this today

kehach07:
Hi @snitin315  👋

This PR #5653  fixes the issue where the error overlay closes immediately when a runtime error occurs during the initial page load (#5024).

I would really appreciate it if someone could take a look when available. Thank you!

PR #5653, opened 2026-03-10
fix(client): prevent overlay from closing immediately on runtime error at initial load

kehach07:
…r at initial load

**Summary**

**Use of AI**

Fixes an issue where the error overlay was closed immediately when a runtime error occurred during the initial page load.

Previously, the overlay was dismissed by socket events such as `invalid`, `still-ok`, and `ok`, even when a runtime error existed. This caused the runtime error overlay to disappear instantly, making debugging difficult.

This change tracks runtime errors and prevents the overlay from being dismissed until the error is resolved.

Issue: #5024


alexander-akait:
is is AI generated code? Please add a test case

test/e2e/overlay-initial-runtime-error.test.js
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  it("should keep overlay visible on runtime error during initial load", async () => {
+    await page.goto("http://localhost:8081");
+
+    const overlay = await page.$("webpack-dev-server-client-overlay");
+
+    expect(overlay).not.toBeNull();
+  });
+});
```
alexander-akait: Please use e2e overlay existing tests

kehach07: Thanks for the suggestion. I moved the test into the existing overlay e2e tests.
It reproduces the runtime error on initial load and verifies that the overlay stays visible.
Let me know if anything else should be adjusted.

ovflowd (dismissed):

client-src/index.js
```diff
@@ -84,8 +85,15 @@ const decodeOverlayOptions = (overlayOptions) => {
...
   currentHash: __webpack_hash__,
+  hasRuntimeError: false,
 };
+window.addEventListener("error", () => {
+  status.hasRuntimeError = true;
+});
 
+window.addEventListener("unhandledrejection", () => {
+  status.hasRuntimeError = true;
+});
```
alexander-akait: This code can be run in web workers and it will broken

alexander-akait (changes requested):

client-src/index.js
kehach07: Thanks for pointing this out. I’ve updated it to use self instead of window so it works correctly in web workers as well. Please take another look.

client-src/index.js
```diff
@@ -87,13 +87,15 @@ const status = {
...
+    status.hasRuntimeError = true;
+  });
 
-window.addEventListener("unhandledrejection", () => {
-  status.hasRuntimeError = true;
-});
+  window.addEventListener("unhandledrejection", () => {
+    status.hasRuntimeError = true;
+  });
+}
```
alexander-akait, line 98: What about if the error happens in web worker? This will not work

alexander-akait (changes requested):
We need to search better solution

PR #5659, opened 2026-04-20
fix: treat loopback aliases as equivalent in isSameOrigin

ThierryRakotomanana:
**Summary**

This is a follow-up to webpack/webpack-cli#4715, where Alexander pointed out after merging. That PR fixed the webpack-cli templates by removing the hardcoded `host: "localhost"` that was causing WebSocket disconnection loops. But the root cause lives deeper, in webpack-dev-server itself.

When `host: 'localhost'` is configured, the OS decides at bind time whether `localhost` resolves to `127.0.0.1` (IPv4) or `::1` (IPv6). This is OS-dependent and not under the user's control, Windows 10/11 with modern network stacks commonly prioritizes IPv6, linux `127.0.0.1`

During the WebSocket handshake, `createWebSocketServer()` calls `isSameOrigin()` which compares the parsed `origin` header against the parsed `host` header. When the OS resolves `localhost` to `::1`, the browser sends:

```
origin: http://localhost
host:   ::1
```
The final line of `isSameOrigin()` is a strict string comparison:

```js
return origin === host; // "localhost" === "::1" → false
```
This causes the server to reject the WebSocket connection and send `Invalid Host/Origin header` back to the client, triggering an infinite reconnection loop visible in the browser console:

```
[webpack-dev-server] Invalid Host/Origin header
[webpack-dev-server] Disconnected!
[webpack-dev-server] Trying to reconnect...
```

This happens even though `localhost`, `127.0.0.1`, and `::1` are all loopback addresses and already individually trusted by `isValidHost()` earlier in the same handshake check.

a fix when host is set to `localhost`

Four tests added to `test/e2e/allowed-hosts.test.js` inside the existing `describe("check host headers")` block, following the same direct `server.isSameOrigin()` call pattern as existing `isValidHost()` 

No, and this fix does not widen the trust boundary in any way. Here is the exact reasoning:

- All three values are already individually trusted. `isValidHost()`, called earlier in the same WebSocket handshake, already passes `localhost`, `127.0.0.1`, and `::1` unconditionally. This fix only resolves the inconsistency when comparing two values that are both already trusted.

- Both sides must be loopback simultaneously. The condition uses `&&`, it only passes when *both* `origin` and `host` are in the loopback set. The `host` header comes from the server's own socket binding. An external attacker cannot control the server's `host` header to be `127.0.0.1` or `::1`.

- Browsers enforce origin strictly. A page served from `https://evil.com` will always send `Origin: https://evil.com`. No browser will send `Origin: http://localhost` for a cross-origin request from an external site, making spoofing of the loopback `origin` impossible.

no need

**Use of AI**
no

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -3091,10 +3091,15 @@ class Server {
       return true;
     }
 
-    // Treat all loopback aliases as equivalent, localhost may resolve to
+    // Treat all loopback aliases as equivalent: localhost may resolve to
     // 127.0.0.1 or ::1 depending on the OS, causing a false mismatch.
+    // Only widen when allowedHosts is "auto" (default) or already permits a
+    // loopback alias, so an explicit allow-list excluding loopback is honored.
     const loopbacks = new Set(["localhost", "127.0.0.1", "::1"]);
-    if (loopbacks.has(origin) && loopbacks.has(host)) {
+    const loopbackPermitted =
+      this.options.allowedHosts === "auto" ||
+      [...loopbacks].some((alias) => this.isHostAllowed(alias));
+    if (loopbacks.has(origin) && loopbacks.has(host) && loopbackPermitted) {
       return true;
     }
 
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -3091,6 +3091,13 @@ class Server {
       return true;
     }
 
+    // Treat all loopback aliases as equivalent, localhost may resolve to
+    // 127.0.0.1 or ::1 depending on the OS, causing a false mismatch.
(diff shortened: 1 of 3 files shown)


alexander-akait (commented):
/cc @bjohansebas Can you look at this and investigate why we have such problem, thanks

/cc @ThierryRakotomanana thanks for PR

bjohansebas (approved):
LGTM, although now that I think about it, maybe I should add port validations for security reasons (I haven’t thought it through much yet, it was just an idea that came to mind)?

PR #5660, opened 2026-04-21
types: Restrict client property to false or ClientConfiguration

Mister-Hope:
`client: true` is never documented nor supported. It is confusing to support `client: true` in types.

Schema here:

https://github.com/webpack/webpack-dev-server/blob/df073c53a8cefb54210b43813fa6ee60364a554e/lib/options.json#L50-L85

Docs here:

https://webpack.js.org/configuration/dev-server/#devserverclient


alexander-akait:
Our types are auto generate, it means we have typo somewhere

PR #5665, opened 2026-04-26
refactor: migrate to ES module syntax

bjohansebas:
**Summary**

**Use of AI**


bjohansebas:
For this, we’ll first need to migrate to Node.js’s test runner so everything works properly and we avoid weird issues, since Jest only has experimental support for it (in a separate PR to make it easier to review). Then we can proceed with this one by migrating to ESM, so it stays focused on that change. After that, we continue with the http-proxy-middleware PR, which will allow us to update that dependency and complete the migration to ESM-only.

eslint.config.mjs
```diff
@@ -7,6 +7,11 @@ export default defineConfig([
   {
     extends: [config],
     ignores: ["client-src/**/*", "!client-src/webpack.config.js"],
+    languageOptions: {
+      // ES2025 needed for import attributes (`import(x, { with: ... })`).
+      // eslint-config-webpack pins ecmaVersion to 2024 for Node 22.
+      ecmaVersion: "latest",
+    },
```
alexander-akait: This should work without this, because we automatically takes ecmaVersion from `node` in `package.json`

alexander-akait:
Also feel free to use experimental support of ecma modules in jest, it should work, to avoid unnecessary work right now, we will migrate to nodejs test runner later, will be great to make a release soon

eslint.config.mjs
bjohansebas: see https://github.com/webpack/eslint-config-webpack/blob/32eaa6f7518e0fd64f305e055354320ad3d1ec2e/configs.js#L73 and the problem is that estree treats that functionality mentioned in the comment as part of ES2025, but for Node 22, we have it available

alexander-akait: hm, I think we need to fix it in our shared config, but let's do it separately, not here

bjohansebas: Okay, Friday is a public holiday in Colombia, which means I’ll be able to dedicate most of my time to finishing all of this and getting the new version launched.

bjohansebas:
Getting Jest to work with ESM is being complicated; I'm going to try the Node test runner.

eslint.config.mjs
bjohansebas: [https://github.com/webpack/eslint-config-webpack/pull/154](url)

lib/Server.js
```diff
@@ -1495,6 +1495,12 @@ class Server {
             clientImplementation = fileURLToPath(
               import.meta.resolve(clientTransport),
             );
+            if (
+              path.isAbsolute(clientTransport) &&
+              !fs.existsSync(clientImplementation)
```
alexander-akait: Ideally we should avoid fs operation, it is slow, special sync

PR #5667, opened 2026-05-02
refactor: migrate to node:test and setup jsdom environment

bjohansebas:
**Summary**

**Use of AI**


alexander-akait:
Looks like test are failed...

scripts/run-tests.mjs
bjohansebas, line 1: t may look weird to run the test runner like that, but with so many flags there was a point where the fixtures were getting executed too, so I had to filter them out. And when using test-shard, it wouldn’t recognize them because of a Node.js bug, so this was the workaround.

scripts/node-test-setup.mjs
```diff
@@ -0,0 +1,29 @@
+import path from "node:path";
+import { snapshot } from "node:test";
+import { format } from "pretty-format";
+import webpack from "webpack";
+
+process.env.CHOKIDAR_USEPOLLING = "true";
+
+// Normalize "\r\n" and "\r" to "\n" so snapshots are platform-agnostic.
+snapshot.setDefaultSnapshotSerializers([
```
bjohansebas, line 9: So the snapshots can keep working very similarly to how they worked with Jest, since there isn’t enough time to migrate to the Node.js format.

bjohansebas (commented):
Okay, once CI passes, I’m going to merge it. There really aren’t any major changes, and I made it so the snapshots wouldn’t change too much, although some of them did. But those are changes I verified myself. Jest components are still being used, but everything works normally. Otherwise, this PR would become even bigger, and I need this to unblock the http-proxy-middleware update, since Jest was a factor preventing us from updating it.

PR #5678, opened 2026-05-26
fix(proxy): match the HMR upgrade path exactly like the ws server

UlisesGascon:


lib/Server.js
```diff
@@ -1860,16 +1860,50 @@ class Server {
...
+    // percent-encoding (so `/%77%73` matches `/ws`), and strip trailing
+    // slashes. Returns null when decoding fails on malformed percent escapes.
+    /**
+     * @param {string} pathToNormalize URL path to normalize for HMR comparison.
+     * @returns {string | null} Normalized path, or null on decode failure.
+     */
+    const normalizeHmrPath = (pathToNormalize) => {
+      try {
+        return decodeURIComponent(pathToNormalize)
+          .toLowerCase()
```
bjohansebas: I think the routes should be case-sensitive, just like Express handles them by default.

test/server/proxy-option.test.js
```diff
@@ -859,6 +859,90 @@ describe("proxy option", () => {
...
+      });
+    });
+
+    // The HMR server's default path is /ws. These variants should all be
+    // recognized as the HMR socket and not forwarded through the user proxy.
+    const variants = [
+      ["trailing slash", "/ws/"],
+      ["uppercase", "/WS"],
+      ["mixed case", "/wS"],
+      ["percent-encoded path", "/%77%73"],
```
bjohansebas: All of these should go to the proxy if we implement case-sensitive routes.

lib/Server.js
```diff
@@ -1860,16 +1860,50 @@ class Server {
...
+    // Normalize a URL path for HMR-path comparison: lowercase, decode
+    // percent-encoding (so `/%77%73` matches `/ws`), and strip trailing
+    // slashes. Returns null when decoding fails on malformed percent escapes.
+    /**
+     * @param {string} pathToNormalize URL path to normalize for HMR comparison.
+     * @returns {string | null} Normalized path, or null on decode failure.
+     */
+    const normalizeHmrPath = (pathToNormalize) => {
+      try {
+        return decodeURIComponent(pathToNormalize)
```
bjohansebas: By normalizing this ourselves, how much would it affect performance? It might be worth running some benchmarks, just as a side note, since I don't remember whether decodeURIComponent is expensive performance-wise.

lib/Server.js
bjohansebas: The comment also comes from how ws handles its internal server, and partly from how I remember Express handling routes.

https://github.com/websockets/ws/blob/2120f4c8c625a76316792680a231496e1b615252/lib/websocket-server.js#L224

bjohansebas:
Okay, I removed the normalization/decoding step so that it follows the ws implementation and avoids any unexpected edge cases. I'm also going to add tests for SockJS since it's still used in this version. It can be removed in the next major release.

bjohansebas:
Interesting, Node 24 doesn't work because of Chrome-related issues.

bjohansebas (approved):

PR #5690, opened 2026-06-25
feat: support universal platform as a web target

bjohansebas:
**Summary**

We already supported this on `main`—the `[node, web]` combination—but it was removed when the `isWebTarget` method was simplified. So I'm adding it back now, along with support for the `universal` platform. If we were to raise the peer dependency version, we could simplify the function even further, but I don't think it's worth doing that yet.

**Use of AI**

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -574,10 +574,11 @@ class Server {
   /**
    * @private
    * @param {Compiler} compiler compiler
-   * @returns {boolean} true when target is `web`, otherwise false
+   * @returns {boolean} true when target is `web` or `universal`, otherwise false
    */
   static isWebTarget(compiler) {
-    return compiler.platform.web || false;
+    // `universal` is only available in webpack `>= 5.108.0`
+    return compiler.platform.web || compiler.platform?.universal || false;
   }
 
   /**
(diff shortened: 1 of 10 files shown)


test/is-web-target.test.js
bjohansebas, line 1: I'm not sure a unit test is worth it, but I'll add one. If you'd rather remove it, that's fine with me. Or if you'd prefer an E2E test, just let me know.

alexander-akait: Let's add one e2e test for this and wen can merge

bjohansebas:
We have an issue even with universal, I'm sending a patch to the core because it's still trying to use require in the browser for HMR.

bjohansebas:
Lol, it works locally

bjohansebas:
Lesson learned: always rebase

lib/Server.js
```diff
@@ -1683,6 +1693,38 @@ class Server {
...
+                  /webpack[/\\]hot|webpack-dev-server[/\\]client/.test(
+                    result.context,
+                  )
+                ) {
+                  result.request = emitter;
+                }
+              },
+            ).apply(compiler);
+          }
+        }
```
bjohansebas, line 1726: The same thing needs to be done on the v5 branch, but I'd prefer that we ship v6 first. After that, I can backport the patch to v5 to avoid having to do another rebase on next.

bjohansebas: It'll probably be a bit more involved for the v5 branch, but that's something to figure out when I put together that patch.

PR #5691, opened 2026-06-30
fix: reject cross-site requests to open-editor and invalidate endpoints

bjohansebas:
**Summary**

Basically, this only protects against external websites calling those endpoints. It doesn't protect against Node.js scripts or `curl`, since in those cases the headers can be freely modified.

In contrast, when a request is made with `fetch()` from a web page, Chrome automatically sends the `Sec-Fetch-Site` header and doesn't let the page override it, even if an attacker tries. The same applies to the `Origin` header—it can't be spoofed by JavaScript running in the browser.

**Use of AI**

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -2084,6 +2084,12 @@ class Server {
           return;
         }
 
+        if (!this.#isSameOriginRequest(req)) {
+          res.statusCode = 403;
+          res.end("Cross-Origin request blocked");
+          return;
+        }
+
         this.invalidate();
 
         res.end();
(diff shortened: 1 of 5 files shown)


lib/Server.js
```diff
@@ -3225,6 +3237,36 @@ class Server {
...
+    // `none` is a user-initiated navigation.
+    if (typeof secFetchSite === "string") {
+      return secFetchSite === "same-origin" || secFetchSite === "none";
+    }
+
+    // Without `Sec-Fetch-*` metadata (non-browser/legacy clients): a request
+    // with no `Origin` header was not initiated by another origin's script;
+    // otherwise fall back to the `Origin`/`Host` comparison.
+    if (typeof headers.origin !== "string") {
+      return true;
```
bjohansebas, line 3327: From what I've tested, the browser doesn't send an `Origin` header in this case

bjohansebas: Also, without this, we could break a simple `curl` request, which isn't a good idea.

lib/Server.js
```diff
@@ -3288,6 +3300,36 @@ class Server {
...
+    }
+
+    return this.isSameOrigin(headers);
+  }
```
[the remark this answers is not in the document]
bjohansebas: false

PR #5698, opened 2026-07-02
fix: reject cross-site requests to open-editor and invalidate endpoints

bjohansebas:
**Summary**

ref: #5691

**Use of AI**

### DIFF
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -2166,6 +2166,12 @@ class Server {
           return;
         }
 
+        if (!this.#isSameOriginRequest(req)) {
+          res.statusCode = 403;
+          res.end("Cross-Origin request blocked");
+          return;
+        }
+
         this.invalidate();
 
         res.end();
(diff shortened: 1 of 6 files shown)


lib/Server.js
```diff
@@ -3390,6 +3402,36 @@ class Server {
...
+    const secFetchSite = headers["sec-fetch-site"];
+
+    // Prefer `Sec-Fetch-Site`: browsers send it even for same-origin GET
+    // `fetch`es (which omit `Origin`), and a cross-site page cannot forge it.
+    // `none` is a user-initiated navigation.
+    if (typeof secFetchSite === "string") {
+      return secFetchSite === "same-origin" || secFetchSite === "none";
+    }
+
+    // Without `Sec-Fetch-*` metadata (non-browser/legacy clients): a request
```
bjohansebas, line 3425: Same as in the other PR: without this, these routes would no longer be usable with curl. This functionality should only protect against requests coming from external websites that the developer visits.

PR #5710, opened 2026-07-22
ci: node 26

bjohansebas:
**Summary**

We don't need to upload them on every system. We can collect them all in one place and then upload everything at the end.

**Use of AI**


scripts/node-test-setup.mjs
```diff
@@ -4,8 +4,10 @@ import { format } from "pretty-format";
 import webpack from "webpack";
 
 process.env.CHOKIDAR_USEPOLLING = "true";
+process.env.WATCHPACK_POLLING = "true";
```
bjohansebas, line 7: Hmm, something only changed in Node.js 26, but I can't figure out what it is

PR #5728, opened 2026-08-28
fix: handle incomplete network interfaces and IPv6 listen addresses

OskarEichler:
## Summary

- Skip absent network-interface entries and null CIDRs instead of crashing local/gateway lookup.
- Return the first matching family/interface directly without allocating flattened and filtered intermediate arrays.
- Remove URL brackets before passing discovered IPv6 addresses to `net.listen`, while preserving `findIp`'s existing bracketed return format.
- Correct host integration expectations so IPv6 addresses are not double-bracketed and the actual bound address is asserted.
- Use loopback addresses when testing wildcard listeners; for discovered network addresses, verify binding directly because VPN/container interfaces are not necessarily browser-routable.

A focused regression covers undefined interface groups, null entries/CIDRs, first-match selection, gateway matching, stable bracketed IPv6 output, and unbracketed listen hostnames.

## Compatibility

No public type or API removals. `local-ipv6` now binds the discovered IPv6 interface as intended instead of passing a bracketed hostname to Node. This can change the listen address on hosts with a non-internal IPv6 interface.

## Verification

- Full host integration suite: 30 passed
- Focused IP lookup suite: 1 passed
- Package build passes
- Full lint suite passes: ESLint, Prettier, TypeScript (server/client), and spelling
- Changeset validation and `git diff --check` pass

Type: fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved local network address detection when interfaces or CIDR information are missing.
  * Ensured address lookup stops at the first matching interface.
  * Corrected IPv6 hostname handling so listening sockets receive unbracketed addresses.
  * Improved support for local IPv4 and IPv6 host resolution across port configurations.

* **Tests**
  * Added coverage for incomplete network interface data and local address resolution.

### DIFF
--- a/.changeset/local-ip-lookup.md
+++ b/.changeset/local-ip-lookup.md
@@ -1,0 +1,5 @@
+---
+"webpack-dev-server": patch
+---
+
+Skip missing interfaces and CIDRs during local-address lookup, stop after the first matching interface, and pass unbracketed IPv6 addresses to the listening socket.
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -429,24 +429,21 @@ class Server {
    */
   static findIp(gatewayOrFamily, isInternal) {
     if (gatewayOrFamily === "v4" || gatewayOrFamily === "v6") {
-      let host;
-
-      const networks = Object.values(os.networkInterfaces())
-        .flatMap((networks) => networks ?? [])
-        .filter((network) => {
+      for (const networks of Object.values(os.networkInterfaces())) {
+        for (const network of networks ?? []) {
           if (!network || !network.address) {
-            return false;
+            continue;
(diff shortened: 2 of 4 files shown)


OskarEichler:
Resolved the draft blocker and found one additional null-entry edge case during the re-review. `host.test.js` now preserves `findIp`’s bracketed URL contract without double-bracketing, asserts the actual bound local IPv6 address, and uses loopback for wildcard listeners. The gateway path now also skips null interface entries, not only null CIDRs. The full host suite passes 30/30, the focused incomplete-interface/gateway regression passes, and build plus all lint/type/format/spelling checks are clean. I’ve marked the PR ready for review.

PR #5730, opened 2026-08-28
fix: repair example configuration and shared asset serving

OskarEichler:
## Fixes

- Derive configuration context/README/output from each example's URL rather than the shell working directory; let webpack resolve entry forms.
- Extract the README heading without mutating global marked settings or rendering unused HTML.
- Serve shared assets through express.static with correct query/404 handling, removing the undeclared mime dependency and manual stream handling.
- Modernize the plugin example to ESM, actually register the middleware example, migrate proxy arrays and removed hooks, and use the declared httpxy package for IPC proxying.
- Close example backend/proxy servers with the dev server. Use an available backend port to avoid local port-5000 conflicts.
- Repair CommonJS assets, dev-middleware index behavior, quote-insensitive universal output checks and affected README examples.

- Repair root-relative shared asset and font paths, contain the layout at narrow widths, add language/decorative-logo/status semantics, and raise status/link contrast with visible keyboard focus.

## Compatibility

Examples/docs only; no published runtime or package/engine changes and no changeset required. The demo proxy backend uses an ephemeral loopback port rather than assuming port 5000 is available.

## Verification

All 50 example configurations import and pass the dev-server schema. Eight runtime configurations compile and exercise asset responses; the node-target universal configuration is compilation-only for assets. The actual plugin example separately compiles, listens and serves its layout assets. The proxy route returns the expected backend response; IPC configuration serves HTTP over its socket. Browser review passes at 320px and 1280px; desktop scroll width equals viewport width. Status/link contrast ratios are 5.08:1, 5.90:1 and 5.12:1. Full lint, types, spelling, formatting and diff checks pass. No claim that every example's browser interaction or HTTPS certificate path was exercised.

## Audit scope

## Summary by CodeRabbit

* **New Features**
  * Updated examples for current development-server, middleware, proxy, static asset, and HTTP/2 configurations.
  * Added request logging, dynamic proxy routing, default index serving, and static asset hosting.
  * Improved IPC proxy lifecycle handling and Hono request processing.

* **Bug Fixes**
  * Corrected example URLs, compression expectations, proxy formats, and server content checks.
  * Improved accessibility with clearer focus states, stronger contrast, language metadata, and reduced-motion support.

* **Documentation**
  * Revised example instructions and headings to match current behavior and configuration.

### DIFF
--- a/examples/.assets/layout.html
+++ b/examples/.assets/layout.html
@@ -1,22 +1,23 @@
 <!doctype html>
-<html>
+<html lang="en">
   <head>
     <title>WDS ▻ <%= htmlWebpackPlugin.options.title %></title>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <link rel="shortcut icon" href=".assets/favicon.ico" />
+    <link rel="shortcut icon" href="/.assets/favicon.ico" />
     <link
       rel="stylesheet"
       href="https://fonts.googleapis.com/css?family=Source+Code+Pro:400,600|Source+Sans+Pro:400,400i,500,600"
     />
-    <link rel="stylesheet" href=".assets/style.css" />
+    <link rel="stylesheet" href="/.assets/style.css" />
   </head>
   <body>
     <main>
       <header>
         <h1>
           <img
-            src=".assets/icon-square.svg"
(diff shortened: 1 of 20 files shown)


alexander-akait (changes requested):
A lot of AI slop and bugs

PR #5731, opened 2026-08-28
fix: watch future glob matches without mutating options

OskarEichler:
## Fixes

- Watch each glob's static parent and filter file registrations, so future matching files/directories remain discoverable.
- Match ignored globs against future paths, preserving function, regex and matcher-object ignores.
- Keep explicit paths in mixed arrays and apply negative patterns as exclusions.
- Clone watch options instead of mutating ignored; frozen options are supported.
- Remove synchronous filesystem glob expansion from watch setup and deduplicate watched roots.

## Compatibility

No API/engine removal. Adds picomatch ^4.0.4 as a direct dependency (already present transitively in the lockfile). Relative patterns remain cwd-relative; events retain chokidar's cwd semantics. Newly created matches now trigger live reload when subsequently edited. Directory watching is necessary to discover future matches; nonmatching files are excluded.

## Verification

All 32 unchanged upstream watch-files tests pass. Ten real filesystem scenarios pass: absolute/cwd paths, negative patterns, function/regex/object ignores, mixed literal/glob paths, missing roots, escaped bracket paths and frozen options. Original source reproduces missed new files and option mutation. Build and full lint/types/formatting/spelling pass. Windows filesystem execution remains unverified.

## Audit scope

## Summary by CodeRabbit

- **New Features**
  - File watching now automatically includes files created after the development server starts when they match configured glob patterns.
  - Ignore patterns continue to apply to newly created files and paths.
  - Watch configurations remain unchanged while monitoring dynamic file matches.

- **Bug Fixes**
  - Improved handling of negated patterns and dynamically added files in file watch configurations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "launch-editor": "^2.14.1",
     "open": "^11.0.0",
     "p-retry": "^8.0.0",
+    "picomatch": "^4.0.4",
     "schema-utils": "^4.3.3",
     "selfsigned": "^5.5.0",
     "serve-index": "^1.9.2",
--- a/package.json
+++ b/package.json
@@ -77,7 +77,6 @@
     "launch-editor": "^2.14.1",
     "open": "^11.0.0",
     "p-retry": "^8.0.0",
-    "picomatch": "^4.0.4",
     "schema-utils": "^4.3.3",
     "selfsigned": "^5.5.0",
     "serve-index": "^1.9.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -77,6 +77,7 @@
     "launch-editor": "^2.14.1",
     "open": "^11.0.0",
     "p-retry": "^8.0.0",
+    "picomatch": "^4.0.4",
```
alexander-akait: We should avoid extra packages here

alexander-akait (changes requested):
We don't need multiple glob related packages

PR #5732, opened 2026-08-28
fix: export and type the custom server extension class

OskarEichler:
## Fixes

- Export the documented webpack-dev-server/lib/servers/BaseServer.js extension point for ESM, CommonJS and TypeScript consumers.
- Unwrap the CommonJS class consistently with the main Server entrypoint, retaining .default interoperability.
- Type webSocketServer.type as a constructor extending BaseServer instead of a zero-argument callable returning a configuration object.

## Compatibility

Restores the documented extension path blocked by the v6 export map. No runtime/engine/dependency removal. The declaration now accepts the class actually instantiated at runtime; invalid callable factories may require correction to constructible classes.

## Verification

Unchanged v6 export map reproduces ERR_PACKAGE_PATH_NOT_EXPORTED. Real ESM and CommonJS subclasses construct successfully; CommonJS default interop is retained. Strict NodeNext TypeScript consumer accepts a custom subclass in webSocketServer configuration. Build and full lint/types/spelling/formatting pass.

## Audit scope

## Summary by CodeRabbit

- **New Features**
  - Exported the `BaseServer` class as a public package entry point.
  - WebSocket server configuration now accepts custom server class constructors.
  - Added TypeScript declarations for the new public export.
  - Improved CommonJS imports so server classes are returned directly.

### DIFF
--- a/.changeset/export-base-server.md
+++ b/.changeset/export-base-server.md
@@ -1,0 +1,5 @@
+---
+"webpack-dev-server": patch
+---
+
+Export the documented `BaseServer` extension class and accept its constructors in the public `webSocketServer.type` declaration.
--- a/lib/Server.js
+++ b/lib/Server.js
@@ -138,7 +138,7 @@ function loadWebpackPeer() {
 
 /**
  * @typedef {object} WebSocketServerConfiguration
- * @property {("ws" | string | (() => WebSocketServerConfiguration))=} type type
+ * @property {("ws" | string | typeof import("./servers/BaseServer.js").default)=} type type
  * @property {Record<string, EXPECTED_ANY>=} options options
  */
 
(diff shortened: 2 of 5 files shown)


.changeset/export-base-server.md
bjohansebas, line 1: That would be a minor, but honestly, I’m not sure it’s necessary. What’s your use case?

OskarEichler:
The use case targeted here is a TypeScript/ESM consumer providing webSocketServer.type as a custom class that extends the documented BaseServer. Today that requires a deep import of an unexported CommonJS file, and the public type models the extension as a zero-argument factory instead of a constructor. The export and type change make the documented extension point usable without a private-path import.

PR #5741, opened 2026-09-18
ci: stop sharding the test matrix and drop Node 25

alexander-akait:
**Summary**

Two cuts to what CI spends, together taking the test matrix from **48 jobs per push to 9**.

**No more sharding.** Four shards per OS and Node version meant each job repeated checkout, `npm ci`, `npm run build` and a Puppeteer Chrome download before running about a quarter of the suite. That setup is a fixed cost per job, so most of the spend went on preparing runners rather than on testing. Nothing runs less in parallel as a result: Node's test runner already executes files concurrently at `availableParallelism() - 1` within a job, and that is unchanged — there are simply far fewer job setups. The coverage artifact name was keyed on `strategy.job-index`, which only stayed unique because of the shard axis, so it is now keyed on the matrix values that identify a job.

**Only the LTS lines.** Node 25 is odd-numbered, so it never became LTS, and it reached end of life on 2026-06-01 — testing it buys nothing. The matrix keeps 22 (maintenance LTS until 2027-04-30), 24 (active LTS until 2028-04-30) and 26 (current, LTS from 2026-10-28).

ci

n/a — no source change. Verified instead that the suite works unsharded. Locally: `npm run test:coverage` without `--test-shard` discovers all 73 test files, exits 0 and writes `coverage/lcov.info` (the file the upload step requires with `if-no-files-found: error`); a full single-process run completed at 1005 tests, 916 passing, with every failure a sandbox limitation that reproduces the same way when the suite is run in pieces. `test/ports-map.js` was checked for overlap, since suites that previously sat in different shards can now run at the same time: 76 suites, 85 distinct ports, no collisions. On CI, the first push of this branch ran the full unsharded suite green on all four ubuntu jobs in 7–9 minutes each.

No. `engines` still declares `>= 22.15.0`; this only changes which runtimes CI exercises.

n/a

**Use of AI**

Claude Code made the workflow edits and ran the verification above — the unsharded command end to end, the full single-process suite run, the port-map overlap check, and a read of the Node release schedule to confirm which lines are LTS today.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UjuMAuk9o6UazjHzcAQCTA)_

## Summary by CodeRabbit

- **Chores**
  - Simplified automated test execution by removing test sharding.
  - Updated supported CI test environments by removing Node.js 25.x.
  - Updated coverage artifact naming to identify the associated webpack version.


bjohansebas:
Hmm, should I really stop sharding? The E2E tests are pretty expensive, and if one shard fails, we'd have to rerun all of them, right?

PR #5742, opened 2026-09-18
ci: cancel superseded pull request runs

alexander-akait:
`nodejs.yml` already cancelled superseded jobs, but through per-job concurrency groups that also cancel pushes to `main`, `next` and `v4`. This replaces them with one workflow-level group that only cancels pull request runs.

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

## Why the branch pushes should not be superseded

That is the actual bug here. `main`, `next` and `v4` are built on push, and those runs feed the coverage that later comparisons are measured against. A cancelled run also hides a breakage that is already on the branch, which is exactly when you want to know.

Note that `cancel-in-progress: false` alone is **not** enough: a concurrency group holds at most one *pending* run, and GitHub evicts that pending run whenever a newer one enters the group regardless of the flag — *"any existing pending job or workflow in the same concurrency group will be canceled and the new queued job or workflow will take its place."* With one `main` run executing and a second waiting, a third push would evict the second. Hence `github.run_id` for non-pull-request events: each gets a group of one, so it can be neither cancelled nor evicted. (Thanks to the review on the sibling PR for catching this; the first version of this PR had the flaw.)

## Why one workflow-level group rather than fixing the per-job ones

Adding the condition to the existing blocks would have been a smaller diff, but the per-job arrangement has three other costs:

- **It cancels job by job.** Each old job dies only as its replacement is queued; a workflow-level group cancels the whole superseded run at once.
- **`upload-coverage` has no group at all.** Today a superseded run can still reach it and publish coverage for a commit nobody is looking at. The workflow-level group covers it.
- **Every matrix dimension has to be repeated in the group key**, and a dimension added later and forgotten would make cells cancel each other — a quiet and confusing failure. #5741 is a live example of the maintenance burden: it had to edit the group key purely because the matrix changed. Keying on the workflow and ref cannot drift.

Happy to switch to the minimal version (keep the per-job groups, just add the condition) if you would rather keep the existing shape.

## Verification

`actionlint` 1.7.7 is clean on all three workflows, and I checked that this means something by typo'ing the expression to `github.event_nam`, which it catches at exactly that line. The file is also Prettier- and cspell-clean using this repo's own configs.

`release.yml` is untouched: it already uses `concurrency: ${{ github.workflow }}-${{ github.ref }}`, whose string form leaves `cancel-in-progress` false, so releases queue instead of cancelling. `dependency-review.yml` is untouched too — one short job, nothing to save.

Rebased onto #5741 (merge commit), which stopped sharding the test matrix and dropped Node 25. It edited the same per-job group this PR removes; the removal is kept, since the workflow-level group supersedes it. An earlier version of this description quoted a job count, which that PR made wrong within the hour — the count is now gone from both the description and the workflow comment.

The same change is going to `enhanced-resolve` (webpack/enhanced-resolve#673) and `webpack-dev-middleware` (webpack/webpack-dev-middleware#2412).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016aGHrb1YaEvaGwNELGEHjF

## Summary by CodeRabbit

- **Chores**
  - Updated automated workflow run coordination.
  - Pull request workflow runs now cancel earlier in-progress runs for the same branch, prioritizing the latest changes.
  - Push-triggered workflows continue running independently without automatic cancellation.
  - Workflow concurrency is managed across the complete workflow rather than separately for each test and lint configuration.
  - This provides more predictable validation results while reducing unnecessary duplicate work.


alexander-akait:
Pushed a correction in 2a57d72 before anyone spent time reviewing the first version — the review on the sibling PR (webpack/enhanced-resolve#673) caught a real flaw that this PR had too.

`cancel-in-progress: false` protects only the run that is already executing. The concurrency group still holds at most one *pending* run, and GitHub evicts that pending run whenever a newer one enters the group, whatever the flag says — *"any existing pending job or workflow in the same concurrency group will be canceled and the new queued job or workflow will take its place."*

So keying every run on the ref did not actually deliver what this PR claims. With one `main` run executing and a second waiting, a third push would have evicted the second — exactly the "two merges in quick succession" case the change exists to protect, and it would have looked correct while failing.

The group is now:

```yaml
group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

Pull requests still group by ref and supersede as described above; every other run gets a group of one and can be neither cancelled nor evicted. `actionlint` 1.7.7, Prettier and cspell are clean on the result using this repo's own configs.

---
_Generated by [Claude Code](https://claude.ai/code)_

